PR #2, opened 2015-11-20
[SYSML-375] [Optimizer] Address Optimizer Inconsistencies Due to Casting of Doubles To Ints

dusenberrymw:
This addresses a few inconsistencies with our optimizer in which accidental truncation of double values to integer values could lead to rewrite rules being mistakenly applied.  As an example, a rewrite rule targeting the expression `sum(v^2)` could be mistakenly applied to the expression `sum(v^2.3)`.


dusenberrymw:
Test build started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/16/).

dusenberrymw:
Testing successful.

dusenberrymw:
@mboehm7 Unless you have any additional thoughts, I can go ahead and push this.

mboehm7:
LGTM

PR #6, opened 2015-12-02
Creating optional Maven profiles for the various binary distributions.

dusenberrymw:
Creating optional Maven profiles for the various binary distribution builds (source, cluster, standalone, in-memory) in order to speed up regular builds.

Now, a regular `mvn clean package` will *only* generate the SystemML JAR, and runs in ~20 seconds versus > 1 min.

In order to generate both the SystemML JAR *and* one of the other distribution builds, use one of the following commands:
* Source distribution
  * `mvn clean package -P SourceDist`
* Cluster distribution
  * `mvn clean package -P ClusterDist`
* Standalone distribution
  * `mvn clean package -P StandaloneDist`
* In-memory distribution
  * `mvn clean package -P InmemoryDist`

cc @mboehm7, @deroneriksson, @lresende, @akchinSTC

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -186,71 +186,6 @@
 			</plugin>
 
 			<plugin>
-				<artifactId>maven-assembly-plugin</artifactId>
-				<version>2.4</version><!--$NO-MVN-MAN-VER$-->
-				<executions>
-					<execution>
-						<id>create-package-assembly</id>
-						<phase>package</phase>
-						<goals>
-							<goal>single</goal>
-						</goals>
-						<configuration>
-							<skipAssembly>true</skipAssembly>
-						</configuration>
-					</execution>
-					<execution>
-						<id>create-source-distribution-assembly</id>
-						<phase>package</phase>
-						<goals>
-							<goal>single</goal>
-						</goals>
-						<configuration>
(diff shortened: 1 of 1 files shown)


dusenberrymw:
Test build started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/22/).

mboehm7:
that's very useful - thanks @dusenberrymw; I'd be fine with either approach with regard to distribution profiles.

dusenberrymw:
Great. Yeah as I mentioned on the dev list, I'm also fine with either approach since the main goal is to just speed up the regular dev builds. Based on that, I'll change it to a single profile and then merge it.

dusenberrymw:
Alright, now we can use the following commands:
* `mvn clean package` <- Creates the SystemML JAR.
* `mvn clean package -P distribution` <- Creates the SystemML JAR *and* the various distribution binaries.

dusenberrymw:
Test build started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/24/).

dusenberrymw:
Testing successful.

PR #10, opened 2015-12-04
Remove copyrights from license, add license where needed

deroneriksson:


deroneriksson:
This removes the old copyright notices from the license messages. In addition, many files were missing Apache licenses, so they have been added. Additionally, replaced OCO Source Materials notices with Apache licenses where they were found.

Tests completed successfully [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/27/).

@lresende @dusenberrymw One thing I am a bit confused about is that some files (such as AParserWrapper.java) at the end of the GitHub "Files Changed" view show up with "No changes" messages even though it is showing that the files have changed. If you click the View button for the file, you see that the file contains the correct information with the correct license. If I check my local history on my machine, I see that for a file such as AParserWrapper.java, the old license message with the copyright has been replaced with the new license message without the copyright. So I'm not sure why the "No changes" messages appear. Maybe it's an artifact of changing so many files?

An additional point to be aware of is that updating the licenses was more difficult that I originally expected because files in the project have a mix of Unix-style (LF) and Windows-style (CR+LF) linefeeds. The linefeeds in the project should probably be standardized to Unix-style to prevent such difficulties from occurring.

dusenberrymw:
@deroneriksson Yeah I think that is just a GitHub artifact.  Looks fine checked out locally.

LGTM.

deroneriksson:
Added Apache project name and copyright to NOTICE files, as described by @lresende. I amended the commit message rather than creating a new commit, so I needed to do a force push. I will rerun the tests.

Test build started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/28/).

lresende:
LGTM

PR #11, opened 2015-12-09
Update doc links to incubator on GitHub and to website

deroneriksson:
Minor updates to URLs in documentation to point to new SystemML GitHub site and to Apache SystemML Incubator website.

Does not require test suite to be run since only *.md and *.html updated.


dusenberrymw:
LGTM.

PR #15, opened 2015-12-10
Update transform function in DML Language Reference

deroneriksson:
Updated transform() documentation.

The update can be reviewed here: http://deroneriksson.github.io/incubator-systemml/dml-language-reference.html#data-pre-processing-built-in-functions


shirisht:
LGTM

PR #19, opened 2015-12-11
Update sparkDML.sh to remove some hard-coded parameters and fix a pat…

nishi-t:
Update scripts/sparkDML.sh to remove some hard-coded parameters and fix a path of SystemML configuration in spark-submit command-line options.

The sparkDML.sh overwrite $SPARK_HOME and $SYSTEMML_HOME by hard-coded parameters. If these environment variables is already defineded, using it without overwritting is helpful for users. And, some default parameters (eg. driver_memory, num_executors, ...) for spark-submit are hard-coded too. I think that these parameters setting in ($SPARK_HOME | $SPARK_CONF_DIR)/conf/spark-defaults.conf is better than hard-coded in the shell-script. This PR fixes these.

In addition, this PR change a path of SystemML-config.xml in spark-submit comand-line options. Because, SystemML-config.xml is placed under conf/ now.

### DIFF
--- a/scripts/sparkDML.sh
+++ b/scripts/sparkDML.sh
@@ -39,12 +39,13 @@ fi
 
 # Default Values
 
-master=yarn-client
-driver_memory=20G
-num_executors=5
-executor_memory=60G
-executor_cores=24
+master="--master yarn-client"
+driver_memory="--driver-memory 20G"
+num_executors="--num-executors 5"
+executor_memory="--executor-memory 60G"
+executor_cores="--executor-cores 24"
 conf="--conf spark.driver.maxResultSize=0 --conf spark.akka.frameSize=128"
+
 
 # error help print
 
@@ -83,17 +84,16 @@ EOF
   exit 1
 }
 
-
 # command line parameter processing
 
 while true ; do
   case "$1" in
     -h)                printUsageExit ; exit 1 ;;
-    --master)          master=$2 ; shift 2 ;;
-    --driver-memory)   driver_memory=$2 ; shift 2 ;;
-    --num-executors)   num_executors=$2 ; shift 2 ;;
-    --executor-memory) executor_memory=$2 ; shift 2 ;;
-    --executor-cores)  executor_cores=$2 ; shift 2 ;;
+    --master)          master="--master "$2 ; shift 2 ;;
+    --driver-memory)   driver_memory="--driver-memory "$2 ; shift 2 ;;
+    --num-executors)   num_executors="--num-executors "$2 ; shift 2 ;;
+    --executor-memory) executor_memory="--executor-memory "$2 ; shift 2 ;;
+    --executor-cores)  executor_cores="--executor-cores "$2 ; shift 2 ;;
     --conf)            conf=${conf}' --conf '$2 ; shift 2 ;;
      -f)               f=$2 ; shift 2 ;;
     --stats)           stats="-stats" ; shift 1 ;;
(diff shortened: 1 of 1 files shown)


shirisht:
Hi Tatsuya

Thanks for the PR. 
1) Agree with your change reg. SPARK_HOME and SYSTEMML_HOME. However, note that SPARK_HOME is not set by default when we install Spark. This is something the user has to do. You can keep that change, but please change DEFAULT_SPARK_HOME to something generic (e.g., */usr/bin/spark-1.4.0-SNAPSHOT-bin-hadoop2.4*) and put a comment noting that users must override it by pointing to Spark installation on their cluster.

2) Reg. Spark defaults like driver-memory, please keep them for now. This is because the values in spark-defaults.conf affects all users and cluster-wide applications. They can probably go into SystemML-config.xml but let us keep them for now.

3) Please also revert the location of *SystemML-config.xml* (i.e., remove *conf/*). It is always good to have both *SystemML.jar* and *SystemML-config.xml* at the same level. 

Shirish

deroneriksson:
Hi @nishi-t,

Thank you very much. I have a couple comments from looking at the script.

(1)
I see that the DEFAULT_SYSTEMML_HOME variable is created but isn't used. So probably need to update

```
if [ -z ${SYSTEMML_HOME} ]; then
  SYSTEMML_HOME="."
fi
```

to

```
if [ -z ${SYSTEMML_HOME} ]; then
  SYSTEMML_HOME=${DEFAULT_SYSTEMML_HOME}
fi
```

(2)
At bottom of script, certain variables have ${...} notation such as ${master} but others do not, such as $driver_memory. Do variables such as $driver_memory and $num_executors need to be updated to ${driver_memory} and ${num_executors}?

nishi-t:
Hi Shirish

Thank you for review. I understand the purpose of some hard-coded default parameters and the location of SystemML-config.xml.
As you pointed out, I have fixed the PR.

Thanks.

Tatsuya

nishi-t:
Hi Deron

Thank you for your comments. 
It's my mistake. I have taken your comments to the PR. 

Thanks.

Tatsuya

nishi-t:
@shirisht, @deroneriksson,
Addressed your comments in the commit 8604ce40efedd393b29070249f22c07f8a5f7a65
could you take a look?

nishi-t:
@shirisht 
> I actually like your previous change where you build full Spark configuration parameters in the while loop. 

OK. I reverted to it and tested the script with various combinations of possible aruguments in the commit 4a4426bf311eed273db6a52dbf44f2c3d4d6c7c5
As a result of this change, I had to change the description of setting default parameters in the script.

For example, default spark parameters is written in my latest commit:
```sh
master="--master yarn-client"
driver_memory="--driver-memory 20G"
```
Instead of (previous commit)
```sh
master=yarn-client
driver_memory=20G
```
This change make setting default parameters somewhat redundant, but I choose this simplest way for now.
Please review again.

shirisht:
@nishi-t Thanks for the update.
A last comment .. could you please remove ``--properties-file`` option that loads ``spark-defaults.conf`` from ``spark-submit`` invocation? It is the default behavior anyway.

nishi-t:
@shirisht Thank you for review. You're right. It is the default behavior. I removed it.

deroneriksson:
@shirisht Is this PR ready to merge?

shirisht:
Yes, it looks good. 

@deroneriksson Please merge this PR.

deroneriksson:
Thank you @nishi-t

PR #26, opened 2015-12-16
Initial Hadoop Batch Mode doc

deroneriksson:
Here is some documentation to help users with SystemML in Hadoop Batch Mode.


deroneriksson:
Rendered doc can be reviewed here: http://deroneriksson.github.io/incubator-systemml/hadoop-batch-mode.html

shirisht:
@deroneriksson, Thanks for the PR. I was just thinking last night that we need to reorganize documentation reg. execution modes. Since you started this thread, can we actually have **Execution Modes** as the main section under **SystemML Documentation**? Let us make that section as the one-stop for all possible ways of running SystemML -- standalone, Hadoop, Spark, JMLC, MLContext, etc. Each of this can be a subsection. We now have this information in various different places. Let us consolidate and make it flow better through this PR. Please feel free if you have any suggestions.

shirisht:
BTW, I think we should also change DML examples in page **DML and PyDML Programming Guide**. For example, we currently show examples that iterate through matrix cells and print them. First, it is not related to ML, and more importantly, it unintentionally advocates bad/less-efficient programming practice of indexing matrices at the cell level. In SystemML, we want to matricize the computations as much as possible. We can however discuss this change as a separate PR. You can create an Documentation issue if the JIRA server is up and running.

deroneriksson:
@shirisht

I think reorganizing the docs to have an Execution Modes section is an excellent idea. There is a lot of value to that idea. Perhaps I can do that as a separate PR since reorganizing seems like a separate unit of work.

With regards to the **DML and PyDML Programming Guide**, I created it for a complete beginner to start becoming familiar with the absolute basics of the language (data types, matrix indexing, control statements, etc) rather than being a guide for ML. It should probably be renamed to something like **Beginner's Guide to DML and PyDML** or **Language Basics** or something like that. I think you are right in that a **DML and PyDML Programming Guide** should detail practices such as writing efficient ML code that scales. I can put something like this together if I can get lots of input from the ML experts on the project.

shirisht:
Thanks @deroneriksson  for the detailed documentation. Here are the Initial set of comments. Will provide more as I finish reading the text.

* "We will first run SystemML on a single machine in Hadoop Standalone mode" --> "We will first run SystemML on a single machine with Hadoop running in Standalone mode"
* Please remove IBM internal hostnames such as ``bdavm015.svl.ibm.com``. Similarly in command prompts, suh as ``[hadoop@bdavm015 ~]$``. 
* Reference to Hadoop and SystemML release tar balls -- are we planning to update them whenever there is a new release (for example, system-ml-0.9..) ?
* Please explain the generated console output -- at least, for the first example.
  * INFO ``No default SystemML config file (./SystemML-config.xml) found`` (BTW, seems like SystemML-config.xml is not in your working directory)
  * WARN message ``hops.OptimizerUtils: Auto-disable multi-threaded text read for 'text' and 'csv' due to thread contention on JRE < 1.8 (java.version=1.7.0_79).``. 
  * Briefly explain ``scratch_space/``

deroneriksson:
Thanks @shirisht for your feedback. I will make those updates.

With regards to the tar ball version number question, that is a very good point to bring up. Version numbers can be extracted to a yml file for the documentation so that the version can be stored in a single place and it can be referenced from multiple locations. Ideally it would be nice if that version number in the yml file could be automatically generated based on the version number in the pom. Maven resource filtering might allow that to work. It would also be nice if the particular yml file/value being updated was gitignored so that it didn't need to be committed. I will think about this to see if I can come up with a clean solution.

deroneriksson:
I made the requested updates. I also removed the screen captures since they had internal hostnames in them. They are also very hard to keep up-to-date over time, so it's probably a lot easier to maintain the document by removing them.

deroneriksson:
Updated, rendered doc can be reviewed here: http://deroneriksson.github.io/incubator-systemml/hadoop-batch-mode.html

dusenberrymw:
LGTM.  Ideally, we also need a Spark Batch Guide as well.

deroneriksson:
Agreed regarding the Spark Batch Guide. Thank you for reviewing.

PR #27, opened 2015-12-16
[SYSML-282] Update in place for intermediate matrix objects

asurve:
This request is to get code review and following question answered.
   1. How to add matrix object variable into symbol table?
   I have left place holder in OptimizerRuleBased.java, to do this. I tried few options with no success.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/DataOp.java
+++ b/src/main/java/org/apache/sysml/hops/DataOp.java
@@ -79,6 +79,12 @@ public class DataOp extends Hop
 			setInputFormatType(FileFormatTypes.BINARY);
 	}
 
+	public DataOp(String l, DataType dt, ValueType vt, DataOpTypes dop,
+			String fname, long dim1, long dim2, long nnz, boolean updateInPlace, long rowsPerBlock, long colsPerBlock) {
+		this(l, dt, vt, dop, fname, dim1, dim2, nnz, rowsPerBlock, colsPerBlock);
+		setUpdateInPlace(updateInPlace);
+	}
+
 	/**
 	 * READ operation for Matrix
 	 * This constructor supports expressions in parameters
@@ -178,10 +184,11 @@ public class DataOp extends Hop
 		_dataop = type;
 	}
 	
-	public void setOutputParams(long dim1, long dim2, long nnz, long rowsPerBlock, long colsPerBlock) {
+	public void setOutputParams(long dim1, long dim2, long nnz, boolean updateInPlace, long rowsPerBlock, long colsPerBlock) {
 		setDim1(dim1);
 		setDim2(dim2);
 		setNnz(nnz);
+		setUpdateInPlace(updateInPlace);
 		setRowsInBlock(rowsPerBlock);
 		setColsInBlock(colsPerBlock);
 	}
@@ -229,7 +236,7 @@ public class DataOp extends Hop
 			case PERSISTENTREAD:
 				l = new Data(HopsData2Lops.get(_dataop), null, inputLops, getName(), null, 
 						getDataType(), getValueType(), false, getInputFormatType());
-				l.getOutputParameters().setDimensions(getDim1(), getDim2(), _inRowsInBlock, _inColsInBlock, getNnz());
+				l.getOutputParameters().setDimensions(getDim1(), getDim2(), _inRowsInBlock, _inColsInBlock, getNnz(), getUpdateInPlace());
 				break;
 				
 			case PERSISTENTWRITE:
--- a/src/main/java/org/apache/sysml/hops/Hop.java
+++ b/src/main/java/org/apache/sysml/hops/Hop.java
@@ -83,6 +83,7 @@ public abstract class Hop
 	protected long _rows_in_block = -1;
 	protected long _cols_in_block = -1;
 	protected long _nnz = -1;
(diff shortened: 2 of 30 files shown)


mboehm7:
@asurve , it is impossible to put these intermediates during parfor optimization into the symbol table
because they might be created multiple times in deep control structures. Please, try to add the related code to VariableCPInstruction.processInstruction (createvar). If the newly created matrix object is marked as update in-place, then allocate the matrix block similar to DataConverter.createDeepCopyExecutionContext. The parfor optimizer should only mark the hop for update in-place, the flag is then compiled into the instruction and during runtime you can create the special matrix object during createvar instructions. Note that you need to update Hop.clone() too, otherwise the update in-place flag is lost on deep copy of HOP DAGs during dynamic recompile. For debugging your change, please, double check that your update-in place flag is still existing on processing the related createvar instruction. Thanks.

shirisht:
@asurve Also, please note that parfor optimizer uses recompilation framework to generate the execution plan. Please follow the code path to see how properties from Hop are propagated to runtime instructions.

mboehm7:
Arvind, thanks for the changes (I really would like to have this feature as soon as possible) but this patch needs a major rework:

1) Algorithm correctness: There are various scenarios, where your algorithm would currently create incorrect results: (a) missing else branches, (b) missing function handling, and (c) incorrect hop consumer checks. For example in "isInLoop", you only investigate the if branch, you never jump into related functions, and you only investigate hop DAG root nodes (missing recursive DAG processing w/ memoization) - we want to check parents of transient read, not inputs to transient write. 
2) Optimizer performance: Please recompile each statement block just once if required, not for every candidate that might be in the same block.
3) Simplifications: I would also strongly suggest various simplifications to make the code easier to understand and debug. First, you don't need separate cases for parfor and for if you do equivalent things; parfor is subclass of for, so any for handling covers both. Second,  you could do almost everything in getUIPConsumerList, isInLoop and validateUIPConsumerList on OptNodes which allows for simple tree processing (because it only represents containment) - only for leaf nodes, you get the associated Hop and do the remaining checking. Third, is this UIPCandidateHop really required? Almost all the information is available at hops and optnodes.
4) Code cleanup: First, please wrap any log debug messages into LOG.isDebugEnabled, especially if you construct strings. Second, please use foreach instead of indexed access on collections  unless you need the index or remove elements from the lists. Third, please don't use "} else {" - it is unreadable.
5) Tests: First, you need to put the (1==1){} cut into the function; otherwise your function is inlined. Second, a couple of tests regarding the issues raised under 1) would be good. Third, please move your specific code out of AutomatedTestBase into your own tests. 
6) Statistics: The maintenance of update in-place counters is currently not thread-safe. Since the only place where this is used is parfor, you need to add this. Please either remove or use atomics/synchronized.
7) Explain: It would be good not to introduce unnecessary complexity. Since, update in-place markers are only set on DataOp, we should extend the output for DataOps of type tread, maybe only if it is marked for update in place.

mboehm7:
thanks @asurve this looks already much better - a couple of issues to cleanup

1) Potential side effects: You maintain the list of update-in-place variables in debug mode as a static variable in order to facilitate automated tests. However, since you always maintain that in debug mode, it might create problems in specific cases, where multiple parfor optimizations happen concurrently (e.g., parfor w/ fixed parameters w/ nested optimized parfor). Please, either use ThreadLocal state or a specific flag that is only enabled by your test.  
2) Logging: Furthermore, you currently output a lot of detailed information. Since debug is used to get a high-level idea of parfor plans and applied rewrites, please move your detailed output wrt candidates and things alike to log level TRACE.
3) Flag in parfor optimizer: Please add a flag 
APPLY_REWRITE_UPDATE_INPLACE_INTERMEDIATE to OptimizerRuleBased (enabled by default). This will allow us to quickly disable this feature if issues show up after the initial commit.
4) Statistics: First, please only show the update-in-place stats if it was a parfor script (because parfor is the only place where we apply it right now). Have a look at (parforOptCount>0). Second, currently you redundantly use atomics and synchronized, please remove the unnecessary sychronized from the increment methods.
5) Code Style: First, please update the comment "guaranteed to be a last-level block (see hop change)" - it's wrong because you recompile the entire parfor program. Second, some minor stylistic issues: size() == 0 -> isEmpty; size() > 0 -> !isEmpty(); listUIPRes.removeAll(listUIPRes) -> listUIPRes.clear(); and if( (bInLoop) || (pb instanceof WhileProgramBlock ... ) becomes if( bInLoop || pb instanceof WhileProgramBlock ). Third, please make your methods private or protected (not package visible). Fourth, please rename getUIPConsumerList -> rGetUIPConsumerList
and validateUIPConsumerList -> rValidateUIPConsumerList to make clear these are recursive methods.
6) Tests: First, it would be nice if your UpdateInPlaceTest would include a class-level comment what the individual tests are (e.g., test for functions, test for interleaved statements, etc). Second, updateinplace12.dml does not need the if(1==1) cut because it already includes control flow. Third, please remove unused code.

mboehm7:
thanks @asurve, we're almost there - please (1) use LOG.isTraceEnabled() in places where you only use LOG.trace(), (2) merge the update-in-place stats output with the existing parfor output, where update-in-place being the last two entries in this category.

mboehm7:
LGTM; thanks @asurve - there are a couple of potential improvements but those can be addressed in separate PRs.

PR #28, opened 2015-12-17
Remove Token and ParseException, refactor to DMLParseException

deroneriksson:
Deleted Token and ParseException classes. Changed all ParseException references to DMLParseException.

@mboehm7 The only real change in code functionality is in DMLParseException, where newMsg code was replaced my msg code. Please let me know if this looks OK or needs to be changed and if other code updates are needed.

Tests completed successfully [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/40/).


deroneriksson:
Sorry, I found an issue with this. The generated error message isn't properly displayed. I will investigate. Perhaps I should create a test too to test for an expected error message.

deroneriksson:
OK, problem exists because I folded ParseException and DMLParseException together but "public ParseException(String message)" and "public DMLParseException(String fname)" are not equivalent. Need to think more about this...

deroneriksson:
Upon further examination, ParseException is used for purposes besides DML parsing (for example, parsing DMLConfig, merging configs, wrapper for FileNotFoundException and IOException, etc). Some of these should not be converted to DMLParseExceptions because they don't have to do with parsing DML. Perhaps I can remove the Token class but not fold ParseException and DMLParseException together.

mboehm7:
@deroneriksson I still believe it would be useful to merge these exceptions. The instances of ParseException that are not related to parsing could be replaced with LanguageException.

PR #29, opened 2015-12-17
Upgrading Spark version to 1.4.1

oza:
Making the versions of spark-related dependencies consistent.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,7 @@
 	<properties>
 		<hadoop.version>2.4.1</hadoop.version>
 		<antlr.version>4.3</antlr.version>
+		<spark.version>1.4.1</spark.version>
 
 		<!-- OS-specific JVM arguments for running integration tests -->
 		<integrationTestExtraJVMArgs />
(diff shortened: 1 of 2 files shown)


oza:
After the upgrade, MLBlock cannot be compiled because of changes against Row in Spark. The PR includes the fix for the failure of compiling.

oza:
I can upgrade more recent version, so please ping me if it's necessary for us.

dusenberrymw:
Testing of rebased version started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/48/).

dusenberrymw:
@mboehm7, @niketanpansare Do you have any thoughts on updating the Spark version to 1.4.1, with particular emphasis on the change from 1.3.1 -> 1.4.1 for Spark SQL?

mboehm7:
From my perspective it's fine to switch to 1.4 since we ran all our experiments with 1.4/1.5 in the last five months. I would recommend to be a little bit lazy here and stick to 1.4 for now in order to prevent exploiting new features too eagerly which would loose backwards compatibility.

dusenberrymw:
@mboehm7 Agreed, thanks for the thoughts!

@oza Pending the results of the tests, I will merge this in.  Thanks!

niketanpansare:
@dusenberrymw  I am fine with upgrading Spark version. Once upgraded, there are few checks in api package which are no longer necessary (for example: https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/MLContext.java#L1104). I can take care of it once this pull request is merged. Thanks :)

dusenberrymw:
@niketanpansare Great!

dusenberrymw:
Rerunning the test build [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/54/).

dusenberrymw:
Test build 54 passed.  I'll merge this in now.

dusenberrymw:
Merged.  Thanks @oza for the contribution, and look forward to future ones!

PR #30, opened 2015-12-17
Run unit tests in functions packages without suites.

gweidner:
Updated pom.xml to run unit tests under functions without using Suite wrapper (previously refactored packages listed individually).  For consistency, also added missing test class with 4 test cases to misc package suite which I missed when refactoring that package.

Note the package gdfo was kept as suite since a failure was reported in on-demand build 41 although not observed on local windows development environment.

Tests completed successfully [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/42/) which includes latest commit.

PR #32, opened 2016-01-04
[SYSTEMML-217] Add New Unary Aggregates: Variance & Standard Deviation

dusenberrymw:
This introduces new "***variance***" and "***standard deviation***" unary operators to the DML language, including row and column variants for each: `var(X)`, `rowVars(X)`, `colVars(X)`, `sd(X)`, `rowSds(X)`, `colSds(X)`.  The implementation includes parser, HOP, LOP, runtime, and optimizer components.

The following provides an slightly more-detailed overview, and a full discussion can be found in the design document <sup>**[1]**</sup>:

* Adds `sd`, `colSds`, and `rowSds` built-in functions for standard deviation to the DML language.
* Adds `var`, `colVars`, and `rowVars` built-in functions for variance to the DML language.
* Creates a new `VAR` operation type for the existing `AggUnaryOp` HOP.
* Compiles the `sd`, `colSds`, and `rowSds` built-in DML functions to the square root of the variance via the `VAR` HOP type wrapped in a `OpOp1.SQRT` HOP type.
* Creates a new `Var` operation type for the existing `Aggregate` and `PartialAggregate` LOPs.
* Compiles the `VAR` HOP type to the `Var` LOP type.
* Creates new runtime opcodes: `"avar"`, `"uavar"`, `"uarvar"`, `"uacvar"`.
* Compiles the `Var` LOP type to new opcodes.
* Parses the new opcodes into runtime aggregate operators that use the `CM` runtime function with `Variance` type.
* Creates a new `VAR` aggregate type in the core unary aggregate library.
* Matches the runtime aggregate operators using the `CM` function of `Variance` type to the `VAR` case.
* Extends the core unary aggregate library with new variance functions for the `VAR` case.
* Adds logic for incremental aggregation of variances computed for multiple partitions. <sup>**[2]**</sup>
* Extends existing optimizer rewrites for column-wise and row-wise aggregation to cover row and column variances.
* Adds new optimizer rewrites to optimize row and column variance operations under certain conditions.

**[1]**: ~~*To be posted later once the JIRA project is set up.*~~  This has been posted to [SYSTEMML-217](https://issues.apache.org/jira/browse/SYSTEMML-217).
**[2]**: The current implementations of `incrementalAggregate(...)` in `MatrixBlock` still need to be refactored, however, I've elected to postpone that to a future JIRA in order to limit the scope of this one.

cc @mboehm7


dusenberrymw:
Testing started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/44/).

mboehm7:
This feature is a very useful addition and the pull request already looks very nice. Just a couple of cleanups required:

1) Rewrite: By adding VAR to LOOKUP_VALID_ROW_COL_AGGREGATE, it looks as if the general case rewrites for row/col aggregate do also apply, which is, however, not always the case. The reason why it works is the correct ordering of your new rewrites. However, you still want to use the case for rowVars -> var if we have just a single row. I would recommend to either (a) add an additional constraint to the general case rewrites, or (b) remove VAR from LOOKUP_VALID_ROW_COL_AGGREGATE and add the second general case to the specific VAR rewrites.
2) Tests: It would be great if we could increase the coverage of these new testcases. First, please use a dense sparsity of 1.0>x>0.4 instead of 1.0 in order to include 0s for dense as well. Second, please make both dimensions larger than 1000 (our default blocksize) in order to test aggregation in the distributed case (your column var test already has that but we should do the same for row var too). Third, since variance requires special handling of 0s, it would be good to include empty input matrices as well.
3) Runtime: There are a couple of places where you explicitly initialize newly created arrays with 0. This is unnecessary because the Java specification guarantees that int/double arrays are 0-initialized. 
4) Comments (nice-to-have): It's nice that you added comments about the correction layout - you just might want to extend that (in one place) by (a) some note why var is included as actual value, and (b) why these values are sufficient (as all other intermediates can be reconstructed from these values)
5) Commit messages (nice-to-have): As mentioned before, the key 'SYSTEMML' is just too long for commit messages - we might want to use the old SYSML for our commits. Similarly, I would leave away these compiler/optimizer/parser/runtime tags as they do not really provide any additional information.

dusenberrymw:
Thanks @mboehm7 for the great feedback.

1. I've incorporated the logic of the new rewrites into the existing general case rewrites with a specific constraint for each, and removed the new rewrites.
2. I've updated the dense sparsity to be 0.7, increased the dimensions to 1234 x 1432, and added tests for empty cases.  Since standard deviation relies on variance operators internally, I've tried to keep the standard deviation tests simpler to avoid growing our testing time unnecessarily.
3. I've removed those initializations.  However, I still see several zero-based initializations of the output `c` matrix in `LibMatrixAgg.java`, such as on [line 1960](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixAgg.java#L1960).  Is there a reason for those cases?
4. I have added additional comments to `getCorrectionLocation(...)` in `PartialAggregate.java`, as seen [here](https://github.com/apache/incubator-systemml/pull/32/files#diff-18451b8f1f6069ae271d031fe6cbc008R219).
5. I have removed the compiler, optimizer, runtime, and parser tags from the title.  While "SYSTEMML-###" is a bit long, I am concerned that changing this to "SYSML-###" could cause confusion.  For example, if someone wanted to look up the associated JIRA, searching for "SYSML-###" in a search engine would not pull up the JIRA issue, while "SYSTEMML-###" would.

**NOTE**:  This cleanup is still a work in progress.  Extending the test cases to have both dimensions > 1000, thus triggering the parallel threaded execution, exposes fairly consistent, yet somewhat random, failures, mostly in the CP cases.  The variances are somewhat close, but different enough to definitely cause concern.  The incremental aggregation uses the existing logic in `CM.java` for the aggregation.  The symptoms point towards possible numerical stability issues, off-by-one issues, or some other type of bug.  I will continue looking into it.

mboehm7:
thanks for the cleanups. 
ad 3: Yes, this is something we still need to cleanup - feel free make the change. 
ad 5: OK, let's go consistently with the full SYSTEMML for now, once it becomes too annoying and we all agree on a prefix, we can revisit this.
ad note: Yes, we need to follow up on this one. Does it only happen for column aggregates? If yes, we would need to look at the final aggregation. Does it only happen for variance? One thing to keep in mind is that the epsilon in our tests is used as an absolute not relative tolerance - this implies that we expect larger errors for variance than for standard deviation because the former operates on squared values (and your generated data has a large enough domain). I would be fine to reduce the eps of the variance test to something like eps = Math.pow(10, -7).

dusenberrymw:
@mboehm7 The failures were occurring with row, column, and full variance aggregates (and by extension also standard deviation), only in CP mode, and only for cases in which the dimensions prompted a local multi-threaded execution.  I figured out the problem -- thread contention involving the `_buff2` object in the `CM` function object during threaded variance calculations, located at [line 177](https://github.com/apache/incubator-systemml/pull/32/files#diff-d9237face00cb50c86d8ebb855c20000L177).  This `_buff2` is used as a buffer to calculate the updated `m2` values without updating the original `m2` object -- it appears to be useful for the other moment calculations, but extraneous for the variance calculation.  Thus, I have removed it, and changed it to update the `m2` object directly.  Was there an additional design decision behind this?  If not, I'd like to also clean it up at [line 285](https://github.com/apache/incubator-systemml/pull/32/files#diff-d9237face00cb50c86d8ebb855c20000L285).

The row/column/full variance and standard deviation tests are passing locally now.  I'll start some full tests soon.

mboehm7:
That's a great catch. I don't think there was a particular reason for this buffer use; variance was introduced after the moments and just reused the same infrastructure. Yes, please cleanup the variance computation in CM.

However, note that in general, we tackle this problem of shared state by not reusing function objects with internal state (e.g., see CM.getCMFnObject() compared to KahanPlus.getKahanPlusFnObject()). This is the reason why CM worked fine for multi-threaded groupedAggregate in LibMatrixAgg.

dusenberrymw:
Okay great, I'll cleanup the rest of the variance code as well.  I'll be out of town for a few days, so I'll finish this up first thing on Monday.

dusenberrymw:
Testing started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/47/).

dusenberrymw:
@mboehm7 The rest of the variance code in `CM.java` has been cleaned up.  I'm running a final full test suite, and pending any other comments, I can merge this in.

mboehm7:
LGTM - thanks Mike.

dusenberrymw:
We seem to have had a major test suite regression (>1000 test failures) since the last positive tests, and it is affecting various builds (different branches, i.e. from different pull requests).  The common theme seems to be that the failures are only occurring with those builds that are based on the current master.  Currently investigating.

Build [47](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/47/) (for this PR), and [48](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/48/) (for PR #29) both failed.

We'll need to pause the 0.9 release until this is fixed.

cc @mboehm7, @gweidner, @akchinSTC, @lresende

mboehm7:
I ran our testsuite last night successfully in my local dev environment. Could it be that there is one of the following issues (1) no clean separation of concurrent runs of our testsuite, or (2) hardware disk issues?

dusenberrymw:
(1) is a good possibility, and we checked on (2) and that doesn't seem to be the case.  Did you run the suite successfully using `mvn clean verify`?

mboehm7:
no I run the testsuite directly with junit from eclipse

dusenberrymw:
Actually, I read (1) wrong.  No, the Jenkins test builds are placed in separate base directories, so they shouldn't interfere with each other.

dusenberrymw:
Okay, yeah the `mvn clean verify` runs parts of the test suite in parallel, versus the synchronous testing in Eclipse (as far as I know for the Eclipse behavior).  Jenkins uses the former approach.  I don't see any issues just running the specific tests individually in an IDE.

gweidner:
I've attached mvn shell prompt console results when I ran the full suite on my local Windows environment which included the change made to pom.xml (PR #30).
[mvn_4282_pom_success_2h_50m.txt](https://github.com/apache/incubator-systemml/files/86645/mvn_4282_pom_success_2h_50m.txt)

mboehm7:
@dusenberrymw could you please try to revert commit 96019cf5e2d62410a631f754c7654e27de842bf and rerun the tests? Furthermore, there was some side effect between tests, hiding an issue - I will fix the test functions.aggregate.ColSumTest.

mboehm7:
ok, meanwhile these issues seem to be resolved - I'll deliver the fix tomorrow morning together with same other fixes.

dusenberrymw:
@mboehm7 Great!  I've started build [52](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/52/) against the current master (with the hotfix), and build [53](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/53/) is this PR branch rebased over the current master.

As a follow-up for the previous two builds, build 50 (previous current master) failed, and build 51 (rebuild of last positive build) passed.

dusenberrymw:
Great, test builds 52 & 53 have passed.  I'll merge this in now.

PR #33, opened 2016-01-08
Updated dml script class path from ibm to apache

gweidner:
Only systemml.bat referenced previous ibm path instead of new apache path which resulted in 'Could not find DMLScript.class. You must build the project.' error message.  I left java memory options as-is although I did have to lower to get to run on my local system.


deroneriksson:
Good catch @gweidner. LGTM.

dusenberrymw:
Merged.  Thanks, @gweidner!

PR #36, opened 2016-01-13
Update Quick Start and handle release links

deroneriksson:
Remove or hide links to releases until available.
Quick Start updates, such as fix build command,
fix scripts paths, miscellaneous text updates,
and remove trailing whitespace.

### DIFF
--- a/docs/hadoop-batch-mode.md
+++ b/docs/hadoop-batch-mode.md
@@ -134,10 +134,23 @@ To verify that Java and Hadoop were on the path, I used the `java -version` and 
 	From source with checksum f9ebb94bf5bf9bec892825ede28baca
 	This command was run using /home/hadoop/hadoop-2.6.2/share/hadoop/common/hadoop-common-2.6.2.jar
 
+<!--
 Next, I downloaded a SystemML binary release and unpacked it.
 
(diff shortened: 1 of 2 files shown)


deroneriksson:
Quick Start Guide updates can be seen [here](http://deroneriksson.github.io/incubator-systemml/quick-start-guide.html).
Hadoop Batch Mode updates can be seen [here](http://deroneriksson.github.io/incubator-systemml/hadoop-batch-mode.html).

dusenberrymw:
This looks good to me for now, although we really should replace the `runStandaloneSystemML` scripts with the new `bin/systemml` script soon.  That will require updating the distribution scripts.

deroneriksson:
@dusenberrymw
I agree about the runStandaloneSystemML vs. bin/systemml being an issue. Let's bring up the issue on the dev mailing list. I think the systemml/systemml.bat scripts might also need some minor updating/testing. Right now `systemml` is using `SYSTEM_ML_JAR=\"${BUILD_DIR}/classes\"` so you need to do a 'mvn package' for it to work. However, there is also a `#SYSTEM_ML_JAR=$( find $PROJECT_ROOT_DIR/target/system-ml-*-SNAPSHOT.jar )` line commented out, so it looks like it's ready to use a jar too. Maybe it can be set up to do both?

dusenberrymw:
Yeah we should probably just switch it to always use a JAR.  That way it is ready to go for both custom-built scenarios, and official binary releases.

PR #39, opened 2016-01-14
Update DML ref for rev, sign, cholesky, and aggregate

deroneriksson:
Also fix a table alignment needed after changing page fonts.


deroneriksson:
Updates can be viewed [here](http://deroneriksson.github.io/incubator-systemml/dml-language-reference.html).

mboehm7:
LGTM w/ a nice to have cleanup: instead of "in which case X is a regular matrix", we could be more specific "in which case X is a regular (n x m) matrix"

deroneriksson:
@mboehm7 Updated doc with "(n x m)".

mboehm7:
great thanks.

PR #40, opened 2016-01-15
Adding a custom Java options environment variable to bin/systemml.

dusenberrymw:
Adding a custom Java options environment variable to `bin/systemml`, as well as a new `conf/systemml-env.sh.template` template.  Now the user can provide additional Java options either directly at the command line or via a `conf/systemml-env.sh` file, both using the new environment variable `SYSTEMML_JAVA_OPTS`.  Also, edited the `bin/systemml` script to use the user's preferred bash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ buildNumber.properties
 # User configuration files
 conf/SystemML-config.xml
 conf/log4j.properties
+conf/systemml-env.sh
 
 # Documentation artifacts
 docs/_site
--- a/bin/systemml
+++ b/bin/systemml
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #-------------------------------------------------------------
 #
 # Licensed to the Apache Software Foundation (ASF) under one
@@ -29,7 +29,7 @@ Usage: $0 <dml-filename> [arguments] [-help]
 EOF
   exit 1
 }
-#    Script internally invokes 'java -Xmx4g -Xms4g -Xmn400m -jar StandaloneSystemML.jar -f <dml-filename> -exec singlenode -config=SystemML-config.xml [Optional-Arguments]'
+#    Script internally invokes 'java -Xmx4g -Xms4g -Xmn400m [Custom-Java-Options] -jar StandaloneSystemML.jar -f <dml-filename> -exec singlenode -config=SystemML-config.xml [Optional-Arguments]'
 
 while getopts "h:" options; do
   case $options in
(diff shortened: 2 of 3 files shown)


mboehm7:
@dusenberrymw could you please elaborate a little bit on the reasoning behind this change - any additional indirection via additional files and variables might create potential issues. Why is it not sufficient to have this config in the header of standalone.sh?

dusenberrymw:
The additional `conf/systemml-env.sh.template` is just a convenience should a user continuously want to add more configuration, and only lives as a template until a user chooses to make use of it.  We need to consolidate all user-editable configuration to a single location, such as the `conf/` directory, and keep the rest of the code clean.

For example, this change came about because I wanted to add a remote dubugging Java option while running standalone scripts with `bin/systemml`, and it was annoying to have to change the `bin/systemml` script itself, which is a Git-tracked file, just for my local configuration.

deroneriksson:
Hi @dusenberrymw @mboehm7
I like the idea of this. Configuration data is being abstracted to conf and there is clean separation from the execution in bin so scripts don't require modification. It follows the Spark project layout pattern, where conf contains things such as spark-env.sh.template, spark-defaults.conf.template, etc. So, someone coming from Spark to SystemML would be familiar with this pattern which is nice.

deroneriksson:
@dusenberrymw @mboehm7 
Do you happen to know if bin/systemml.bat could be adapted to handle this update in a clean fashion?

dusenberrymw:
cc @ckadner Any thoughts?

mboehm7:
Ok, I got your point. However, there is one major difference - for spark, these configuration affect the actual cluster execution, while for us it only affects standalone and users might get the wrong impression that they need this, which is not the case. For example, with hadoop, you have to set hadoop_client_opts and with spark, you set the driver memory budget. 

Anyway, please go ahead for now. However, after the release we should setup some tasks to cleanup all unnecessary scripts and configurations. I really like to keep our simple deployment model: take a SystemML.jar (with packed jars for standalone) and run it wherever you like from command line without need for anything else.

dusenberrymw:
Great, it would definitely be nice to have the end-user execution made as simple as possible.

deroneriksson:
@mboehm7 
You make a very good point. For example, I have looked at Hadoop configuration (HDFS/YARN) before, and it is possible to overwhelm users since so many things can be configured in so many places.

I am in favor of whatever makes the project as easy and understandable to end users as possible.

mboehm7:
great - thanks guys; then we're all in agreement.

deroneriksson:
Sounds like @mboehm7 approves for now. LGTM too. Maybe we should create a JIRA for bin/systemml.bat for consistency.

PR #41, opened 2016-01-16
Cleaning up JAR artifacts in prep for 0.9 release.

dusenberrymw:
Removing the `original-system-ml*.jar` that was created after shading, as it was simply causing confusion.  Also, added a task to copy the final jar to a new `SystemML.jar`, since most of the documentation deals with the later name.

cc @deroneriksson

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -368,6 +368,51 @@
 				</configuration>
 			</plugin>
 
+			<plugin>
+				<artifactId>maven-clean-plugin</artifactId>
+				<version>3.0.0</version>
+				<executions>
+					<execution>
+						<id>clean-original-jar</id>
+						<phase>package</phase>
+						<goals>
+							<goal>clean</goal>
+						</goals>
+						<configuration>
+							<excludeDefaultDirectories>true</excludeDefaultDirectories>
+							<filesets>
+								<fileset>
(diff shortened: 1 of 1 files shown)


deroneriksson:
The original*.jar has confused me when looking at the target directory following a build, so removing it is a good idea.

Eyeballing runStandaloneSystemML.sh, this 'should' be good since:

```
CLASSPATH=""
for f in ${CURRENT_PATH}/lib/*.jar; do
  CLASSPATH=${CLASSPATH}:$f;
done
```

Did you try it out with runStandaloneSystemML.sh and bin/systemml?

deroneriksson:
Oops, nevermind bin/systemml since it's using target/classes.

dusenberrymw:
I didn't run it with `runStandaloneSystemML.sh`, but that script just relies on having a `SystemML.jar` file in the same directory, which is generated with the standalone distribution build from the `system-ml-0.9.0-SNAPSHOT.jar` file.  The `original*.jar` file is just a by-product of shading the JAR to include ANTLR stuff.

dusenberrymw:
Sanity check: `runStandaloneSystemML.sh` does indeed still work.

deroneriksson:
Thank you for checking runStandaloneSystemML.sh. Then LGTM.

PR #42, opened 2016-01-16
Fixing parsing bug with sparkDML.sh in prep for 0.9 release.

dusenberrymw:
Fixing parsing bug with `sparkDML.sh` in which `sparkDML.sh -f test.dml` would be parsed as incorrect usage.


deroneriksson:
Eyeballing it, LGTM.

PR #45, opened 2016-01-19
[Systemml-475] [WIP] Add Implicit Conversion Between Scalar Values and 1x1 Matrices

dusenberrymw:
**Work in progress**

---

Currently, any DML functions that require scalars will not accept 1x1 matrices.  Matrix indexing always returns a matrix, so a statement such as `A[1,1]` from the following simple example will return a single-valued matrix, even though it logically represents a scalar.  Thus, in order to use this value as a scalar, the `as.scalar(...)` function is required.

```r
A = matrix(7, rows=10, cols=10)                                                                                                                
print( "A[1,1]: " + as.scalar(A[1,1]))  // 7
```

In this scenario, the required usage of `as.scalar(...)` can cause confusion, and can be viewed as a violation of SystemML's goal of hiding the underlying data representation from the user.

Additionally, many of the built-in functions, such as `rowVars(...)`, accept matrices without other requirements of dimensionality, and thus will accept a 1x1 matrix, yet will not accept a scalar value.  We should allow any such functions that accept 1x1 matrices to also accept scalar values.

```r
a = 4
print("Row variances of scalar are: " + as.scalar(rowVars(as.matrix(a))))  // 0
```

In order to achieve this, we can add implicit casting between scalar values and 1x1 matrices within the parser during the validation stage as needed.

With this PR, the following can be done:

```r
A = matrix(7, rows=10, cols=10)
print( "A[1,1]: " + A[1,1])  // 7
```

```r
a = 4
print("Row variances of scalar are: " + rowVars(a))  // 0
```

---

**Note:  This is still a work in progress**, as there are likely additional fixes/improvements/changes that can be made, such as determining side-effects on other operations, or HOP rewrites such as removing unnecessary casting.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/BinaryExpression.java
+++ b/src/main/java/org/apache/sysml/parser/BinaryExpression.java
@@ -122,8 +122,39 @@ public class BinaryExpression extends Expression
 			_left = constVars.get(((DataIdentifier) _left).getName());
 		if( _right instanceof DataIdentifier && constVars.containsKey(((DataIdentifier) _right).getName()) )
 			_right = constVars.get(((DataIdentifier) _right).getName());
-		
-		
+
+
+		// Perform any 1x1 matrix -> scalar casts if necessary.
+		// Note, we use the same line/column data as the
+		// current built-in function expression since this
+		// cast is implicitly tied to the function, rather
+		// than the function argument.
+		Identifier lout = _left.getOutput();
+		Identifier rout = _right.getOutput();
+		if (lout.getDataType() == DataType.SCALAR && rout.getDataType() == DataType.MATRIX
+				&& rout.getDim1() == 1 && rout.getDim2() == 1 && rout.getValueType() != ValueType.STRING) {
+			Expression[] castArgs = {_right};
+			BuiltinFunctionExpression cast = new BuiltinFunctionExpression(BuiltinFunctionOp.CAST_AS_SCALAR,
+					castArgs, this.getFilename(), this.getBeginLine(), this.getBeginColumn(), this.getEndLine(),
+					this.getEndColumn());
+			cast.setDimensions(0, 0);
+			cast.setDataType(DataType.MATRIX);
+			cast.setValueType(_right.getOutput().getValueType());
+			_right = cast;
+			_right.validateExpression(ids, constVars, conditional);
+		} else if (rout.getDataType() == DataType.SCALAR && lout.getDataType() == DataType.MATRIX
+				&& lout.getDim1() == 1 && lout.getDim2() == 1 && lout.getValueType() != ValueType.STRING) {
+			Expression[] castArgs = {_right};
+			BuiltinFunctionExpression cast = new BuiltinFunctionExpression(BuiltinFunctionOp.CAST_AS_SCALAR,
+					castArgs, this.getFilename(), this.getBeginLine(), this.getBeginColumn(), this.getEndLine(),
+					this.getEndColumn());
+			cast.setDimensions(0, 0);
+			cast.setDataType(DataType.MATRIX);
+			cast.setValueType(_left.getOutput().getValueType());
+			_left = cast;
+			_left.validateExpression(ids, constVars, conditional);
+		}
+
 		String outputName = getTempName();
 		DataIdentifier output = new DataIdentifier(outputName);
 		output.setAllPositions(this.getFilename(), this.getBeginLine(), this.getBeginColumn(), this.getEndLine(), this.getEndColumn());		
--- a/src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
+++ b/src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
@@ -20,6 +20,7 @@
 package org.apache.sysml.parser;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 
 import org.apache.sysml.parser.LanguageException.LanguageErrorCodes;
@@ -29,6 +30,7 @@ public class BuiltinFunctionExpression extends DataIdentifier
 	
 	protected Expression[] 	  _args = null;
 	private BuiltinFunctionOp _opcode;
+	private Boolean revalidationNeeded = false;
 
 	public BuiltinFunctionExpression(BuiltinFunctionOp bifop, ArrayList<ParameterExpression> args, String fname, int blp, int bcp, int elp, int ecp) {
 		_kind = Kind.BuiltinFunctionOp;
@@ -201,6 +203,12 @@ public class BuiltinFunctionExpression extends DataIdentifier
 		
 		default: //always unconditional
 			raiseValidateError("Unknown Builtin Function opcode: " + _opcode, false);
+		}
+
+		// Added casting expressions will trigger a re-validation.
+		if (revalidationNeeded) {
+			this.revalidationNeeded = false;
+			this.validateExpression(ids, constVars, conditional);
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


dusenberrymw:
cc @mboehm7, @deroneriksson

deroneriksson:
I'm highly in favor of implicit casting from 1x1 matrix to scalar in the case of displaying a matrix element value in a print statement. I think that is a very big win for usability.

I also think `rowVars(a)` is much easier on the eyes than `as.scalar(rowVars(as.matrix(a)))`, since it focuses on the essence rather than on the conversions, although I don't know if this implicit conversion brings up other issues. I don't have a deep enough knowledge of the project to make a more informed comment as to the ramifications of such conversions. Great question for @mboehm7

mboehm7:
Well, I also think this would be useful but it's hard to get right in the general case. We already did something similar for leftindexing w/ scalars, e.g, A[b,c] = s that originally required a right-hand-side matrix. The problem of adding this after right indexing is that we don't necessarily see (especially during validate) where the output is used. This causes many problems due to changed signatures of statement blocks, e.g., in case of conditional control flow, fixed function signatures, or operations that only operator over matrices, etc. I'm not in favor of modifying operations that users would only use in case of matrices (e.g., unary aggregates, matrix mult, etc). Yes, it might happen that during runtime, we have a 1x1 matrix but that is usually not known during validate in these cases. Please also keep in mind that this has nothing to do with the underlying data representation but language semantics. 

I would be fine though to add this for the restricted cases of (1) string concatenations (+ or append), and (2) print statements. These cases would be simple to realize and would probably already solve 95% of related usability concerns.

dusenberrymw:
@mboehm7 Great, thanks for the detailed feedback and guidance as always.  The continued exploration of the engine is quite interesting.  Adding this specifically for string concatenation and print statements would definitely help for usability.

---

Just in the interests of continuing to **learn** more about the engine, it would be very interesting to extend this conversation further.  If we take the `rowVars(a)` example above (yes, agreed that it's not a likely use-case, but nonetheless, let's use it as a concrete example), where `a` is a scalar, and apply the concept of injected matrix casting in order to make this statement correct, it seems like there would be two options.  We could tie the casting to the `a` variable, as in a replacement of `a` with `as.matrix(a)`, or tie the casting to the `rowVars(...)` expression, as in a replacement of `rowVars(...)` with `rowVars(as.matrix(...))`.   As you confirmed (if I understand correctly), the former case should cause issues as it changes the datatype of the `a` variable from a scalar to a matrix, without giving attention to where else in the script `a` is being used.  My thought with this experiment was that the later case of associating the casting with the expression that requires it (which would only be the case when the input is a scalar datatype), as in `rowVars(as.matrix(...))`, would be a safer option as it would not interfere with the `a` variable and any other uses of it in the script.  Thus, when validating, I was attempting to add these casting statements directly as inputs of the expressions needing them due to being supplied a scalar input while requiring a matrix input, such as `rowVars(a)`.  Once I added the matrix casting, I would mark the current expression (current being the `rowVars()` expression, in this case) as needing validation again, and then re-run the validating step on that expression.

If you have the time, @mboehm7, it would very interesting to gain any further thoughts/insights regarding this concept and it's interactions with the engine.  Just for the sake of learning, as I mentioned.

mboehm7:
@dusenberrymw, yes you're absolutely right. Whenever you tie this to the consumer, you can easily ensure correctness for scalar->matrix. All you would need is a lookup table of builtin function arguments that require matrix inputs. Without this awareness, we would affect the result data type of builtin functions like round(x) that also work on scalars. We do similar things for (1) scalar-matrix data type conversions for left indexing, but also (2) value-type conversions for function inputs (e.g., a function with integer signature but called with a double). For matrix->scalar (even if tied to the consumer), it's already more subtle because it's size dependent (only in case of 1x1 matrices) - otherwise it might be very confusing for a user to get errors in case of larger matrices from the injected cast instead of the print. As already described, binding it to the source expression A[1,1] creates many issues because we compile different instructions for different data type combinations. 

I think we all agree that it would be useful to do implicit matrix->scalar casts for print and string concatenation. So let's get this in. Thanks for initiating this discussion guys.

Wenpei:
Really appreciate for this feature to reduce redundancy DML code.
But it is a little bit narrow down to print, string concatenation only, can we make rightIndex(not sure how to call this, rangereindex?) to result scalar value when access a cell of matrix?
e.g.
a_double = b_matrix[1,1]   

a_double is matrix currently, but in this case, I guess user want cell double value.

Great work. @dusenberrymw

dusenberrymw:
@mboehm7 Thanks for the extended discussion!

dusenberrymw:
@Wenpei Yes, I agree that would make more sense from a user's perspective.  R, Python (NumPy), and Matlab/Octave all use the same idea of treating a full right-index result as a scalar value, rather than a 1x1 matrix.  Furthermore, Matlab/Octave has no distinction between the two at the user level.  

@mboehm7 What are your thoughts on this?  While this seems like a good idea at the language level, I'm sure there may have been a good reason/goal behind the original decision.

mboehm7:
ok wrt right indexing, let's try not to cycle the discussion (please see above, let me know if more clarification is needed). In comparison to R (which is an interpreter), it is more difficult in SystemML because we compile different instructions for different data type combinations.

dusenberrymw:
Oh, yeah I understand from our discussion that adding *casts* to full right indexing (i.e. `X[1,4]`) could cause problems.  Perhaps though this could be interpreted as a bigger design discussion -- what if full right indexing natively returned a scalar value?  We wouldn't be changing it post-processing such as with the casting, but rather the instruction would natively result in a scalar value.

mboehm7:
Well the problem is consistency: how do you treat X[a:b, c:d] or X[a,] without knowing a, b, c, d and the dimensions of the input matrix (especially during validate) - then you would end up with some indexing expressions that result in matrices and some others in scalars for the same 1x1 matrix. Also, we want to have the distinction between scalars and matrices because we compile completely different operations in the distributed case.

dusenberrymw:
Yeah, I agree consistency could be an issue.  However, what if we **only** assume a scalar output for the case `X[a,b]`?  In that case, it doesn't matter what `a` and `b` are -- the result is still going to be one single value.  If the user were to instead use `X[a:b, c:d]`, we just return a matrix.  Even if that evaluates to `X[1:1, 1:1]`, the user was still using range indexing, so a matrix was desired and/or expected.  This is the design that Python (NumPy) uses as well.  If we follow this approach, we could define very clear distinctions for the user, and perhaps this would also allow us to maintain clear distinctions internally as well.

mboehm7:
@dusenberrymw please don't get me wrong, of course we could define some specific cases but I'm afraid that this would cause more problems than it solves. User could easily run into issues where in one branch they do point indexing, in the other range indexing, and we "automatically" change the data type in one, leading to compile/runtime errors after the branch.

dusenberrymw:
@mboehm7 Okay, we don't have to pursue this, although if we made the decision that a `X[a,b]` right index is an operation that returns a scalar (natively, rather than in a post-processing step), then validation of conditional control flow should catch any such issues.  At that point, it would be up to the user to understand that `X[1,1]` and `X[1:1,1:1]` are different indexing operations, which already seems natural.

mboehm7:
just to clarify, we already do these checks for data type mismatches of variables from conditional control flow, which would give a compile error if this situation happen to be in an unconditionally executed block; otherwise if it's unclear if we execute the block, it only gives a warning - unfortunately, users typically completely ignore warnings which will lead to runtime errors.

dusenberrymw:
Note: I'm going to close this for now, but we should consider revisiting these ideas in the future with regards to quick fixes for print statements and string appends, as well as broader design decisions for native scalar outputs for full right-indexing (i.e. `X[1,3]`).

deroneriksson:
For user consumability, this would be a great addition.

PR #46, opened 2016-01-19
Update copyright years in NOTICE files

deroneriksson:


deroneriksson:
@lresende 
On general@incubator list in **[VOTE] Release Apache Zeppelin (incubating) 0.5.6-incubating (RC2)** thread I noticed that they needed to update copyright dates in NOTICE files. So did we, so I updated.

PR #48, opened 2016-01-19
[SYSTEMML-477] Updating the format of the metadata file JSON object

dusenberrymw:
Updating the format of the metadata file JSON object to have the commas placed at the end of the lines, rather than at the beginning of the lines.

Previously, the metadata files written would be formatted with the commas placed at the beginning of the lines, which was somewhat strange looking:

```json
{ 
    "data_type": "matrix"
    ,"value_type": "double"
    ,"rows": 1
    ,"cols": 2
    ,"nnz": 2
    ,"format": "text"
    ,"description": { "author": "SystemML" } 
}
```

This change places the commas at the end of the lines, as follows:

```json
{ 
    "data_type": "matrix",
    "value_type": "double",
    "rows": 1,
    "cols": 2,
    "nnz": 2,
    "format": "text",
    "description": { "author": "SystemML" } 
}
```

This applies to matrix and scalar metadata.  Since this is still valid (and perhaps better formatted) JSON, there shouldn't be any issues, and local testing supports this for writes and reads.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DataExpression.java
+++ b/src/main/java/org/apache/sysml/parser/DataExpression.java
@@ -78,8 +78,9 @@ public class DataExpression extends DataIdentifier
 	public static final String COLUMNBLOCKCOUNTPARAM = "cols_in_block";
 	public static final String DATATYPEPARAM = "data_type";
 	public static final String VALUETYPEPARAM = "value_type";
-	public static final String DESCRIPTIONPARAM = "description"; 
-	
+	public static final String DESCRIPTIONPARAM = "description";
+	public static final String AUTHORPARAM = "author";
+
 	// Parameter names relevant to reading/writing delimited/csv files
 	public static final String DELIM_DELIMITER = "sep";
 	public static final String DELIM_HAS_HEADER_ROW = "header";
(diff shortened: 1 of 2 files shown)


mboehm7:
@dusenberrymw I like the idea, but could please also do a test with the wink library that we use for reading the meta data file? Maybe we could get rid of this custom string concatenation code?

shirisht:
I agree with @mboehm7 that we should replace that ugly string concatenation code with appropriate JSON library API calls.

dusenberrymw:
@mboehm7, @shirisht I explored using the Wink JSON library for writing, and I've replaced the string concatenation code.  I used an `OrderedJSONObject` so that the metadata files will have a consistent ordering for the user's benefit.

dusenberrymw:
[Test build 67](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/67/) started.

mboehm7:
great thanks @dusenberrymw - could you please post an example of the output from that library?

dusenberrymw:
Definitely!

Script:
```r
A = matrix("1 2 3 4 5 6", rows=2, cols=3)
write(A, "A")
```

Old `A.mtd`:
```json
{ 
    "data_type": "matrix"
    ,"value_type": "double"
    ,"rows": 2
    ,"cols": 3
    ,"nnz": 6
    ,"format": "text"
    ,"description": { "author": "SystemML" } 
}
```

New `A.mtd`:
```json
{
    "data_type": "matrix",
    "value_type": "double",
    "rows": 2,
    "cols": 3,
    "nnz": 6,
    "format": "text",
    "description": {
        "author": "SystemML"
    }
}
```

deroneriksson:
@dusenberrymw The output is easier to read and so is the source code. Thank you.

dusenberrymw:
Test build 67 was successful.

deroneriksson:
Looks good to me.

PR #49, opened 2016-01-20
[SYSTEMML-482] [SYSTEMML-480] Force Correct Normalization Of All Line Endings to Correct Unix-style Characters

dusenberrymw:
This adds a `.gitattributes` file with a setting that will correctly normalize all files (other than those that are Windows-specific, such as .bat, .exe, etc.) upon checkin to the repo to have the correct Unix-style new line characters, rather than Windows-style carriage returns, while allowing for correct OS-specific new line characters while working locally.  Additionally, this setting will cause Git to correctly maintain the normalization for future commits.  As [seen during review](https://www.mail-archive.com/dev@systemml.incubator.apache.org/msg00254.html) of the first 0.9 release candidate, the lack of this normalization can cause major issues, and the issue is [confirmed in the official Git manual](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Formatting-and-Whitespace).


mboehm7:
@dusenberrymw could you do something similar to force the encoding utf-8? We had issues with that in the past.

dusenberrymw:
Checked into the Git literature, and it looks like there is not a setting to enforce a file encoding.

deroneriksson:
In Eclipse you can set project-level encoding at Project Properties -> Resource ->Text file encoding (to UTF-8), but that setting is stored in .settings/org.eclipse.core.resources.prefs, which can be shared but probably isn't a very good thing to check into git. Not sure how else to share file encoding with others.

dusenberrymw:
[Test build 112](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/112/) running.

mboehm7:
yes, most of us use unix delimiters / utf-8 at workspace level for a long time but it would be good to enforce.

deroneriksson:
@dusenberrymw Test build 112 at https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/112/ seems to be linking to PR 49 of the old repo https://github.com/SparkTC/systemml/pull/49 . Any idea what's going on?

dusenberrymw:
It's building the correct PR, so that just must be link configuration setting.

cc @akchinSTC

dusenberrymw:
FYI, due to this [open Eclipse issue](https://bugs.eclipse.org/bugs/show_bug.cgi?id=342372), this will not yet be enforced for those using the built-in Git included with Eclipse.

PR #50, opened 2016-01-21
[SYSTEMML-158] Update deprecated Hadoop properties

deroneriksson:
Removes the console warnings from deprecated Hadoop properties.
MRConfigurationNames includes "constants" for both v1 and v2 of these properties.

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -76,7 +76,7 @@ import org.apache.sysml.runtime.controlprogram.parfor.ProgramConverter;
 import org.apache.sysml.runtime.controlprogram.parfor.stat.InfrastructureAnalyzer;
 import org.apache.sysml.runtime.controlprogram.parfor.util.IDHandler;
 import org.apache.sysml.runtime.matrix.CleanupMR;
-import org.apache.sysml.runtime.matrix.mapred.MRProp;
+import org.apache.sysml.runtime.matrix.mapred.MRConfigurationNames;
 import org.apache.sysml.runtime.matrix.mapred.MRJobConfiguration;
 import org.apache.sysml.runtime.util.LocalFileUtils;
 import org.apache.sysml.runtime.util.MapReduceTool;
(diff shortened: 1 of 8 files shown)


mboehm7:
1) Based on which rule of thumb did you decide to centralize configurations in MRConfigurationName vs leaving them as they are in the code? Only configurations with differences in v1/v2?  
2) Could you also please simplify (shorten) the constant names ("MR" prefix instead of "MAPREDUCE" etc)?

deroneriksson:
1)
When executing SystemML, there were 5 log warnings (to the console) each time it was run about deprecated hadoop configuration properties (io.sort.mb, mapred.job.tracker, mapred.task.tracker.task-controller, mapred.local.dir, and mapred.system.dir). These were old properties that needed to be updated. So this PR is basically concerned with only these properties so that users don't see a bunch of deprecation warnings when they run SystemML.

Moving them to MRConfigurationName allows for conditionally setting these properties based on Hadoop v1 vs v2, thus eliminating the deprecation warnings for v2 and still supporting v1.

From a code maintenance standpoint, I believe that ALL Hadoop configuration properties should eventually be moved to constants in MRConfigurationName. This has several benefits:
1) centralizes all properties in a single place so you can go to one file and know which properties are being used in the project.
2) if a developer comments-out a constant in MRConfigurationName, the IDE immediately reports to the developer all of the locations where the constant is being used (or alternatively, developer can look up all the references to the constant in the project).
3) also eliminates the multiple string instances (which can be error-prone) with constants so that there is no chance of typos in the multiple string instances when referring to the constants.

---

2)
If MAPREDUCE is changed to MR, then there is less of a direct correspondence between the constant names and the properties that they represent (such as MAPREDUCE_TASK_IO_SORT_MB representing "mapreduce.task.io.sort.mb"). Although MR would be assumed to represent MAPREDUCE when dealing with Hadoop, so that's probably a good idea to reduce the size of the constants to make them easier to work with. It's nice if a developer can do code completion of the constants and immediately know which property is actually being set based on the name of the constant. (I based all the constant names on the v2 versions of the properties.)

I'll go ahead and change MAPREDUCE to MR.

deroneriksson:
@mboehm7 
Updated MAPREDUCE to MR.

Might also be nice to do something like rename MRConfigurationNames to MRProps. Also, might be nice to remove the INVALID settings of the public statics so that they can be made public static finals since they are being set in the static initialization block.

deroneriksson:
Renamed MRConfigurationNames to MRProp. Also changed properties to be public static finals rather than public static.

Tests started [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/130/).

mboehm7:
Deron, I should have commented earlier on this issue - please change it back to MRConfigurationNames (or MRConfigNames in which case you would also need to change MRConfiguration to MRConfig)

deroneriksson:
@mboehm7 
Renamed MRProp back to MRConfigurationNames. (I had changed the name to MRProp to reduce the size of the references to the string constants.)

deroneriksson:
@mboehm7 Does this PR look good to you?

mboehm7:
LGTM - thanks Deron.

PR #54, opened 2016-01-22
[SYSTEMML-480] [SYSTEMML-463] Fix Release Packaging

dusenberrymw:
This cleans up and adds missing files to the release packaging, removes outdated documentation from the releases, cleans up an old readme file, and adds the ability to create a standalone JAR.  With regards to the latter item, we can also explore replacing the `lib` folder of dependency JARs with this new, single standalone JAR file at some point in the near future.

### DIFF
--- a/docs/Language Reference/README_HADOOP_CONFIG.txt
+++ b/docs/Language Reference/README_HADOOP_CONFIG.txt
@@ -1,7 +1,6 @@
 Usage
 -----
-The machine learning algorithms described in 
-$BIGINSIGHTS_HOME/machine-learning/docs/SystemML_Algorithms_Reference.pdf can be invoked
+The machine learning algorithms described in SystemML_Algorithms_Reference.pdf can be invoked
 from the hadoop command line using the described, algorithm-specific parameters. 
 
 Generic command line arguments arguments are provided by the help command below.
@@ -13,17 +12,14 @@ Recommended configurations
 --------------------------
 1) JVM Heap Sizes: 
 We recommend an equal-sized JVM configuration for clients, mappers, and reducers. For the client
-process this can be done via 
+process this can be done via
 
    export HADOOP_CLIENT_OPTS="-Xmx2048m -Xms2048m -Xmn256m" 
    
 where Xmx specifies the maximum heap size, Xms the initial heap size, and Xmn is size of the young 
 generation. For Xmn values of equal or less than 15% of the max heap size, we guarantee the memory budget.
 
-The above option may also be set through BigR setting the "ml.jvm" option, e.g.
-   bigr.set.server.option("jaql.fence.jvm.parameters", "-Xmx2g -Xms2g -Xmn256m")
-
-For mapper or reducer JVM configurations, the following properties can be specified in mapred-site.xml, 
+For mapper or reducer JVM configurations, the following properties can be specified in mapred-site.xml,
 where 'child' refers to both mapper and reducer. If map and reduce are specified individually, they take 
 precedence over the generic property.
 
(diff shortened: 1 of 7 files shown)


shirisht:
Please include the word "incubating" in the Standalone JAR name, similar to Luciano's recent commit.  Also, the version number?

dusenberrymw:
@shirisht The word "incubating" and the version number are already taken care of by the `version` tag at [line 28](https://github.com/apache/incubator-systemml/blob/master/pom.xml#L28) in the `pom.xml` that Luciano added.  That acts as a blanket identifier for all the generated artifacts.  In the case of the new standalone JAR file, it will be created as `system-ml-0.10.0-incubating-SNAPSHOT-standalone.jar` for a local dev build.

dusenberrymw:
I'd also like to gather feedback for an update the 'distributed' release of SystemML to have the suffix 'cluster' attached, in order to make it easy for an end user to distinguish from the 'standalone' package.  

**Previously**, we generated the following (omitting the tar files):

`system-ml-0.10.0-incubating-SNAPSHOT.zip`
`system-ml-0.10.0-incubating-SNAPSHOT-standalone.zip`
`system-ml-0.10.0-incubating-SNAPSHOT-src.zip`

**Now**, we would generate the following releases (omitting the tar files):

`system-ml-0.10.0-incubating-SNAPSHOT-cluster.zip`
`system-ml-0.10.0-incubating-SNAPSHOT-standalone.zip`
`system-ml-0.10.0-incubating-SNAPSHOT-src.zip`

The goal here would be to reduce confusion for the end user as to which package to download.


@mboehm7, @shirisht, @deroneriksson, @lresende Thoughts?

deroneriksson:
Personally I like the idea of -cluster so I know the purpose of that release package. It reduces confusion.

mboehm7:
I don't have a too strong opinion on that but I would prefer the existing naming scheme (except for renaming system-ml to systemml). In any case, please make sure that system-ml-0.9.0-incubating.zip is identified as the "default" download. From my perspective, giving the "cluster" prefix can be confusing because cluster/standalone would be seen as alternatives, which they are not: systemml-standalone is a specialization of systemml (because it just includes additional libaries).

deroneriksson:
I definitely agree with @mboehm7 that I prefer the name systemml to system-ml.

WRT other release package naming, I was rather confused 2 days ago when I was trying to validate the releases since there were so many release packages to go through (unzip this jar, untargz this targz, unzip this zip, unzip this src zip and build, etc). I was particularly confused by "system-ml-0.9.0-incubating.zip". I had to open it to say to myself "This contains the jar and the algorithms, and there are no standalone scripts, so this must be what I would use to run an algorithm on a cluster".

I do see @mboehm7 's point that the standalone is a specialization of the other. I don't have a strong opinion about "-cluster". I think my confusion arises because there are so many release packages and it's hard to visually distinguish among the release packages what I should download.

If system-ml-0.9.0-incubating-standalone.zip has all the functionality of system-ml-0.9.0-incubating.zip plus extras, is the presence of system-ml-0.9.0-incubating.zip itself in some ways redundant so that it should be removed?

lresende:
Foks, as these discussions are not being archived on the mailing list, we should not be having them as part of the PR and on the dev list.

Anyway, I am +0.

dusenberrymw:
Thanks for the thoughts.  In general, I think we need to get down to one single release.  @mboehm7 What would it take internally to be able to generate one JAR that could be used for both single node Java execution, as well as any combination of execution modes on Hadoop and Spark?  As I understand it, the main issue currently is that we rely on multiple Hadoop libraries, as well as others, but we can't always package it or we would have conflicts when running on Hadoop or Spark.

mboehm7:
yes, there are advantages to either one approach - could we do a quick poll on the mailing list? This will also bring the discussion there. 

Having a single release would be great but this would require a general abstraction of core data structures. For example classes like MatrixIndexes and MatrixBlock are Hadoop writables - we would need to introduce additional abstractions in order to never touch the wrapper writables in CP. Anyway, the major problems are related to IO and cluster status - this means we would need to write, for example, sequence files by hand (according to the documented format) and emulate any cluster interactions. The bottom line is, it's possible but a ton of work that might right now outweigh its benefit.

dusenberrymw:
Alright, for now I have removed the "cluster" suffix for the distributed release, and updated the naming from `system-ml-*` to `systemml-*`.  Everything looks good on my end, so I will merge this now.

PR #58, opened 2016-01-30
[SYSTEMML-495] Simplify SystemML Configuration Loading

dusenberrymw:
Currently, SystemML always looks for a configuration file in a default location (in the current directory), and will always emit a warning if it is not found.  Additionally, SystemML allows a user to supply a configuration file location.  In this latter case, SystemML will still attempt to read the default configuration file, emitting a warning if it is not present (which causes confusion), as well as the user-supplied file.  If both are loaded, the user-supplied file will be merged into the default configuration with an overwrite strategy.

Always looking for a default file, even when a file is explicitly supplied, causes confusion, and may create unintended side effects.  If a user explicitly supplies a configuration file, we should simply use that file, rather than attempting to also read an additional default file at the same time.  In order to provide backwards compatibility, if the user does not supply a file, then we should fall back to the default file location.  If that file is also not found, we should then fall back to the internal defaults.  Note that this means that the configuration settings of a user-supplied file will *not* be merged with those of a default file, assuming both are found.

Additionally, in order to supply an additional configuration file, the `-config=path/to/file` flag has to be used, and it only works with the equals sign `=` present, which is unlike all other flags that use spaces.  Instead, we should support `-config path/to/file`.

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -295,8 +295,10 @@ public class DMLScript
 					if( rtplatform==null ) 
 						return ret;
 				}
-				else if (args[i].startsWith("-config="))
-					fnameOptConfig = args[i].substring(8).replaceAll("\"", ""); 
+				else if (args[i].startsWith("-config=")) // legacy
+					fnameOptConfig = args[i].substring(8).replaceAll("\"", "");
+				else if (args[i].equalsIgnoreCase("-config"))
+					fnameOptConfig = args[++i];
 				else if( args[i].equalsIgnoreCase("-debug") ) {					
 					ENABLE_DEBUG_MODE = true;
 				}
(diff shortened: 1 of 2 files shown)


deroneriksson:
Note: I spoke with a user today who expressed confusion with regards to the config file warning. If need be, we might even want to drop the message log level to INFO if we receive additional such feedback.

deroneriksson:
Also, I agree about the `-config` vs `-config=` issue. The use of the equals sign for only this flag is confusing and inconsistent with the other flags.

mboehm7:
1) I'm favor of the cleanup regarding the -config flag (just for completeness, there is one more configuration -Dsystemml.logging=DEBUG w/ similar syntax, but this is due to the JVM -D option)
2) Regarding the warning: let's modify the message if needed but keep it in WARN for now - otherwise people don't think about it which can be disastrous if /tmp is mounted to some small disk (potentially leading to "no space left on device"). 
3) Further, before we can remove the merge logic, let's please discuss on the JIRA. The use case is the following: Assume SystemML and it's config file is put into some system directory on each node in a shared cluster. This means all users of that cluster share the common setup. However, individual users can overwrite any subset of configurations. With the change, a user would be forced to suddenly specify ALL configurations. I'm fine with this for release 1.0 but you need to give people some time to raise concerns.

dusenberrymw:
@deroneriksson, @mboehm7 Great, thanks for the feedback.  I've written up more information regarding the decision to remove the merging logic on the associated [JIRA issue](https://issues.apache.org/jira/browse/SYSTEMML-495), and have invited further discussion there.

deroneriksson:
WRT item 2 (the warning message), right now in the project, the warning message about the default config file appears even if a config file is explicitly specified (and found) via -config(=).

Maybe the logic/wording of the messages can be changed:
1) Maybe an INFO if a config file is found saying that the config file was found (at ...). It might be nice for the user to know that the config file has indeed been found.
2) Maybe a WARN but with more information if no config file is found, such as "No SystemML-config.xml found at ... (and ...). For optimal SystemML performance, it is highly recommended that you reference a SystemML-config.xml file."

dusenberrymw:
@mboehm7 Do you have any more thoughts on this minor pr?

dusenberrymw:
[Test build 261](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/261) successful.

PR #60, opened 2016-02-01
Remove user mailing list ref

deroneriksson:
Users and developers use dev list for all questions currently. Remove "User" reference to avoid confusion since no user list yet.


shirisht:
LGTM

PR #61, opened 2016-02-02
[SYSTEMML-456] parser code deduplication

nakul02:
This is a first pass on the deduplication and cleaup effort between the visitors/listeners for the DML & PyDML generated by ANTLR4.
I've changed the grammars ever so slightly. A more thorough cleanup can follow, but will require significant changes to the Listeners that ANTLR4 generates; which i will leave for a future JIRA. 
There is a also a large chunk of code in the PydmlSyntacticValidator.java file that converts python built-in functions to it's DML equivalent. I or someone else can deal with this later.
There is a Jenkins [build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/84/) in progress.


mboehm7:
@nakul02 could you please do me a favor and squash these commits into a single commit to ease the review. Thanks.

nakul02:
Fixed and set another [Jenkins build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/85/).

nakul02:
@mboehm7 could you please take a look?

mboehm7:
@nakul02 in general, the patch looks already good - just a couple of cleanups (some of these issues exist for a while - we should use the opportunity to resolve them):

1) Redundancy: There still seems to be some unnecessary redundancy. Examples include (a) the handling of control structures, (b) function parameter handling, (c) parameter expression handling. Please try to consolidate or comment why you want to leave it - if there's a good reason, that's fine too.
2) Commented code: Could you please also delete the unused commented code fragments (it's unnecessary to keep that since it can be restored from the history and only confuses people).
3) Error messages: There are a couple of error messages that are completely useless. An example is "incorrect lvalue ... strange" - even if we think a certain code path will never be taken, it will happen and we don't want these kinds of messages showing up in error logs.
4) Package qualification: Please also remove any unnecessary package specification of classes - the only acceptable reason would be that you want to use two classes from different packages in the same file. In such a case, please use the minimal package extension to differentiate them, not the full package name.
5) Unnecessary files: Please remove any unnecessary files like "package-info.java". If the only reason is a global comment, add it to the main entry class of that package. In this particular case, we could just remove it. Also, please don't put JIRA numbers into any comments.
6) Unnecessary class nesting: Could you please flatten SyntacticErrorListener ? Why would we need a class that has only another class (CustomErrorListener ) in it? 
7)  Unknown data type parsing: Further, the current version of getConstIdFromString seems a bit like to hack. Could you please think about a cleaner version, maybe based on valid characters (please also have a look at existing commons libs like StringUtils).  
8) Code style: Finally, it would be good to fix the remaining code style issues as well. This includes (a) typos (e.g., constoubleIdExpressionHelper), (b) the unnecessary use of "compareTo()==0" ("equals" would do just fine), and (c) conditional boolean assignments (e.g., val=false; if( A.equals(B) ) val=true; should become val = A.equals(B).

nakul02:
@mboehm7 For the methods I did not refactor out (control structures and below), additional work would need to be done in DmlSyntacticValidator & PydmlSyntacticValidator and the gains (if any) did not seem worth it. Case if point - exitExternalFunctionDefExpression; A for loop over a list of StrictParameterizedKeyValueStringContext (which a Parser specific type) would need to be done in each SyntacticValidator before any common code could be refactored out.

I could have overlooked this, but I could not find the case you've mentioned in 8(c).

mboehm7:
wrt 8c: there is an example in CommonSyntacticValidator.getConstIdFromString, boolean case

nakul02:
That function has changed now as part of addressing the comments.

nakul02:
Submitted a Jenkins [build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/88/)

nakul02:
[Build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/91/) Passed

nakul02:
@mboehm7 Could you please take a look?

mboehm7:
thanks for the updates @nakul02. There are just some minor cleanups remaining. @deroneriksson could you please merge this in once Nakul did the cleanups?

1) History: Please, double check that the history of CustomErrorListener still exists. Currently it looks like SyntacticErrorListener w/ nested subclass CustomErrorListener has been deleted and re-created as CustomErrorListener. If the history is not present, please do separate commits for the rename and the code change, and squash commits.
2) Cleanup: Get rid of all unused imports (e.g., com.fasterxml.jackson.databind.ser.std.NumberSerializers.IntegerSerializer)
3) Code style: Finally, it would be good to condense unnecessary branches (e.g., exitMatrixDataTypeCheck, instead of "if(pre){}else{exception}" please do if(!pred){exception})

nakul02:
@mboehm7 
wrt 1, squashing the commits for the rename and code change loses the history for the file.
If the two commits are separate, the history is retained. 
What do you suggest?

mboehm7:
then let's keep them separate for now and bring them together when merging.

nakul02:
@mboehm7 To preserve the history, I now have 3 commits.
1) Work to refactor
2) commit that uses "git mv" to rename SyntacticErrorListener.java to CustomErrorListener.java
3) commit that addresses other comments on this PR

Squashing commit 2 (even with the rename-threshold option) gets rid of the history for SyntacticErrorListener.java

Commit 2) does not compile.
A downside to this approach is that we will have a commit in the history that does not compile.

Submitted a Jenkins [build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/95/)

nakul02:
Jenkins [build](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/95/) passed.
@deroneriksson, can you please take a look?

deroneriksson:
@mboehm7 Is this OK to merge this in as 3 commits (with commit 2 not compiling) to preserve history?

mboehm7:
well, I still think it should be possible to merge these commits into one but cannot look into it right now. @lresende any thoughts?

deroneriksson:
@mboehm7 I'll take a look (WRT merging the 3 commits)

deroneriksson:
@mboehm7 Spent a couple hours trying to squash the 3 commits to 1 commit while retaining history for the CustomErrorListener (which was renamed), but without success. I would think that rename-threshold should help do this but I was unsuccessful.

lresende:
In the past, this is what we had to do, to avoid git removing history because the changes on a file was too disrupting for git to identify as rename. There should be no reason for not committing these as different commits, but maybe you want to "cleanup" the commit messages with the JIRA info, etc.

dusenberrymw:
Yeah, let's just commit this as multiple commits, each with the proper JIRA title.

PR #62, opened 2016-02-03
Add Apache license to @header sections of g4 files

deroneriksson:


mboehm7:
LGTM; @deroneriksson could you, however, please change the title - initially I thought you're about to put generated files into our repo (what we don't want).

deroneriksson:
@mboehm7 Thank you for reviewing. I updated the title to hopefully avoid confusion.

PR #64, opened 2016-02-04
[SYSTEMML-503] Update license for CSS and JS

deroneriksson:
Add license info for bootstrap.min.js, bootstrap.min.css, anchor.min.js, and jquery-1.12.0.min.js.
Delete modernizr js file.


deroneriksson:
@lresende 
Please review. Is this the proper way to include these licenses in the project?

Added license info to src/assembly/standalone/LICENSE since this is where the other library license info is included.

deroneriksson:
@lresende Does all the license info in the standalone LICENSE need to be included in the project root-level LICENSE?

deroneriksson:
@lresende Also, do we need two LICENSE files, or can we reduce it to one?

dusenberrymw:
@lresende ping.

lresende:
@deroneriksson The changes here looks good, but as we are also making some of these html/css files under doc folder available on the source distro, we need to include another LICENSE for the source distro that only include the licenses related to doc. When we have the LICENSE for source distro, we should update the source assembly to use those when producing the source distro. Other then that LGTM.

deroneriksson:
@lresende I will go ahead and merge this PR then. The location of another LICENSE file may lead to some discussion, so I'll do that in a separate PR.

PR #65, opened 2016-02-05
Cleanup assembly for release packages

deroneriksson:
Alphabetize includes (scripts, jars, etc).
Apply same formatting to all assembly files.
Fix typos.
Remove redundancies.
Add missing util script in one assembly.


dusenberrymw:
LGTM.

PR #67, opened 2016-02-12
[SYSTEMML-494] Remove warnings from GLM (DML and PyDML) execution

deroneriksson:
Set default termination_code to int.
Initialize w outside of if-else and while statements.

### DIFF
--- a/scripts/algorithms/GLM.dml
+++ b/scripts/algorithms/GLM.dml
@@ -158,7 +158,7 @@ eps = as.double (eps);
 
 # Default values for output statistics:
 
-termination_code     = 0.0;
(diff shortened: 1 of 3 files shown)


mboehm7:
LGTM, but why only for GLM?

deroneriksson:
@mboehm7 Ethan Xu asked about these warnings when he first ran GLM , so I wanted to be responsive and address his concerns in a prompt manner since he's providing us valuable feedback. I'll create another JIRA to validate and fix any similar issues in other algorithms.

mboehm7:
thanks @deroneriksson that would be very useful

PR #69, opened 2016-02-12
Using SystemML from Zeppelin in an end to end data analytics process

sourav-mazumder:
This Notebook covers -

    Ingestion of multiple Datasets from HDFS.
    Exploration of the Data using Spark SQL
    Merging of the data from various data sources for preparing data for building Model
    Building Model using GLM.dml of SystemML
    Using GLM-predict.dml for prediction using larger population
    Relating the prediction back to original dataset
    Visualization of Prediction using R libraries using SparkR.

Please note that this notebook uses a R interpreter for Zeppelin (https://github.com/apache/incubator-zeppelin/pull/208/commits) which is not part of main branch. So the SparkR paragraphs will not work if someone is using Zeppelin main branch. Alternatively one can use the branch related to this PR (PR#208). Also one has to have R and other relevant R packages installed separately on the same machine where Zeppelin process is running. the other R packages needed for the plots are googleVis, ggplot2, maptools, htmltools, knitr, repr (from http://irkernel.github.io/).


akchinSTC:

deroneriksson:
Welcome @sourav-mazumder ! Thank you for creating this example Zeppelin notebook.

Unfortunately, the contribution did not come over properly. The 2BCHR4T1Q directory same over as an empty file rather than a directory, so the notebook JSON file isn't present. I recently created a document about contributing to SystemML that may help, at http://apache.github.io/incubator-systemml/contributing-to-systemml.html#systemml-on-github .

Please let me know if you have any questions.

rightwaitforyou:
How to use the java code to  save/load  your trained model?????

PR #70, opened 2016-02-16
[Scala Pipeline API] Add scala logisticRegression api for spark pipeline

Wenpei:
I wrote a scala ml pipeline wrapper for LogisticRegression Model as a example for scala user.  
I thought those api (java, python, scala) should be put to separated jar, but just follow what it is this time. 

Regards.
Wenpei.

### DIFF
--- a/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
+++ b/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
@@ -105,7 +105,8 @@ object LogisticRegressionModel{
 }
 class LogisticRegressionModel(
     override val uid: String)(
-        val mloutput: MLOutput) extends Model[LogisticRegressionModel] {
+        val mloutput: MLOutput) extends Model[LogisticRegressionModel]  with HasIcpt
+    with HasRegParam with HasTol with HasMaxOuterIter with HasMaxInnerIter {
   override def copy(extra: ParamMap): LogisticRegressionModel = defaultCopy(extra)
   override def transformSchema(schema: StructType): StructType = schema
   override def transform(df: DataFrame): DataFrame = {
--- a/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
+++ b/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
@@ -1,19 +1,20 @@
-package org.apache.spark.ml.systemml
+package org.apache.sysml.api.ml.scala
 
 import org.apache.spark.{ SparkContext }
-
-import org.apache.spark.sql.{ DataFrame, Row }
+import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.types.StructType
-
 import org.apache.spark.ml.{ Model, Estimator }
 import org.apache.spark.ml.classification._
-import org.apache.spark.ml.param.{ Params, Param, ParamMap }
+import org.apache.spark.ml.param.{ Params, Param, ParamMap,DoubleParam }
 import org.apache.spark.ml.param.shared._
-import org.apache.spark.mllib.linalg.Vector
-
 import org.apache.sysml.api.{ MLContext, MLOutput }
 import org.apache.sysml.runtime.matrix.MatrixCharacteristics
 import org.apache.sysml.runtime.instructions.spark.utils.{ RDDConverterUtilsExt => RDDConverterUtils }
+import org.apache.spark.SparkConf
+import org.apache.spark.mllib.linalg.Vectors
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.sysml.runtime.instructions.spark.utils.{RDDConverterUtilsExt => RDDConverterUtils}
+
 
 trait HasIcpt extends Params {
   final val icpt: Param[Int] = new Param[Int](this, "icpt", "Intercept presence, shifting and rescaling X columns")
@@ -30,14 +31,22 @@ trait HasMaxInnerIter extends Params {
   setDefault(maxInnerIter, 0)
   final def getMaxInnerIter: Int = $(maxInnerIter)
 }
+trait HasTol extends Params {
+  final val tol: DoubleParam = new DoubleParam(this, "tol", "the convergence tolerance for iterative algorithms")
+  setDefault(tol,0.000001)
+  final def getTol: Double = $(tol)
+}
+trait HasRegParam extends Params {
+  final val regParam: DoubleParam = new DoubleParam(this, "tol", "the convergence tolerance for iterative algorithms")
+  setDefault(regParam,0.000001)
+  final def getRegParam: Double = $(regParam)
+}
 object LogisticRegression{
   final val scriptPath = "MultiLogReg.dml"
 }
 class LogisticRegression(override val uid: String,val sc:SparkContext) extends Estimator[LogisticRegressionModel] with HasIcpt
     with HasRegParam with HasTol with HasMaxOuterIter with HasMaxInnerIter {
   
-  setDefault(regParam, 0.0001)
-  setDefault(tol, 0.0001)
   def setIcpt(value: Int) = set(icpt, value)
   def setMaxOuterIter(value: Int) = set(maxOuterIter, value)
   def setMaxInnerIter(value: Int) = set(maxInnerIter, value)
@@ -107,14 +116,15 @@ class LogisticRegressionModel(
 object LogisticRegressionTest {
   import org.apache.spark.{ SparkConf, SparkContext }
   import org.apache.spark.sql.types._
-  import org.apache.spark.mllib.linalg.Vectors
-  import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.mllib.linalg.Vectors
+import org.apache.spark.mllib.regression.LabeledPoint
   
   def main(args: Array[String]) = {
     val sparkConf: SparkConf = new SparkConf();
     val sc: SparkContext = new SparkContext("local", "TestLocal", sparkConf);
     val sqlContext = new org.apache.spark.sql.SQLContext(sc);
-    import sqlContext.implicits._
+
+import sqlContext.implicits._
     ScriptsUtils.setSystemmlHome("C:\\aWorkFolder\\git\\systemml\\incubator-systemml\\scripts")
     val training = sc.parallelize(Seq(
       LabeledPoint(1.0, Vectors.dense(1.0, 0.0, 3.0)),
--- a/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
+++ b/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
@@ -1,5 +1,10 @@
 package org.apache.sysml.api.ml.scala
 
+import org.apache.sysml.api.{MLContext, MLOutput}
+import org.apache.sysml.runtime.matrix.MatrixCharacteristics
+import org.apache.sysml.runtime.instructions.spark.utils.{ RDDConverterUtilsExt => RDDConverterUtils }
+import org.apache.sysml.runtime.instructions.spark.utils.{RDDConverterUtilsExt => RDDConverterUtils}
+
 import org.apache.spark.{ SparkContext }
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.types.StructType
@@ -7,14 +12,9 @@ import org.apache.spark.ml.{ Model, Estimator }
 import org.apache.spark.ml.classification._
 import org.apache.spark.ml.param.{ Params, Param, ParamMap,DoubleParam }
 import org.apache.spark.ml.param.shared._
-import org.apache.sysml.api.{ MLContext, MLOutput }
-import org.apache.sysml.runtime.matrix.MatrixCharacteristics
-import org.apache.sysml.runtime.instructions.spark.utils.{ RDDConverterUtilsExt => RDDConverterUtils }
 import org.apache.spark.SparkConf
 import org.apache.spark.mllib.linalg.Vectors
 import org.apache.spark.mllib.regression.LabeledPoint
-import org.apache.sysml.runtime.instructions.spark.utils.{RDDConverterUtilsExt => RDDConverterUtils}
-
 
 trait HasIcpt extends Params {
   final val icpt: Param[Int] = new Param[Int](this, "icpt", "Intercept presence, shifting and rescaling X columns")
--- a/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
+++ b/src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 package org.apache.sysml.api.ml.scala
 
 import org.apache.sysml.api.{MLContext, MLOutput}
@@ -113,7 +132,7 @@ class LogisticRegressionModel(
   }
 }
 
-object LogisticRegressionTest {
+object LogisticRegressionExample {
   import org.apache.spark.{ SparkConf, SparkContext }
   import org.apache.spark.sql.types._
 import org.apache.spark.mllib.linalg.Vectors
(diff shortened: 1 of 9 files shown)


akchinSTC:

mboehm7:
thanks @Wenpei; @niketanpansare could you please take a look.

niketanpansare:
Hi Wenpei, 

Thanks for your contribution. But, I believe your code does exactly the same thing as https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/ml/LogisticRegression.java. This class is fully compatible with Scala API and won't require separate jar. May be I am missing something. Is there any particular reason why you think we should create wrappers in Scala instead ?

Thanks,

Niketan.

Wenpei:
@niketanpansare 
Personally, I thought this wrapper( https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/ml/LogisticRegression.java ), more like a example but not a real api expose to end user.  

So I propose a scala version example since some weakness for java version.
1. It's not naturally to extend scala class in java code. We need know function style after compile, like
	@Override
	public void org$apache$spark$ml$param$shared$HasElasticNetParam$_setter_$elasticNetParam_$eq(DoubleParam arg0) {
		
	}
I assume it's set function, but do nothing here

2. Hard to follow ml parameter style, but define parameter like below
	private IntParam icpt = new IntParam(this, "icpt", "Value of intercept");
	private DoubleParam reg = new DoubleParam(this, "reg", "Value of regularization parameter");

Wenpei.

niketanpansare:
@Wenpei 
Fair point, Java API is little cumbersome for developers of the API as one needs to understand how Java developers can use certain Scala features. However, it does not affect usability with respect to the end user. The main advantage of having Java API is "easy integration with current dev setup".

Btw, I am OK with having a Scala API into incubator-systemml if following conditions are met:
1. "mvn verify" creates a jar with the Scala APIs in it and also runs the unit tests that verifies the Scala API. This will ensure that the future commits do not break the Scala API. 
2. Common dev setup (preferably eclipse or Scala IDE) for both the Scala API and the runtime. That is, scala files are compilation units inside the eclipse editor. This will encourage other committers to work contribute to the Scala API.

Thanks,

Niketan.

dusenberrymw:
I would be in favor of replacing the existing Java APIs ([LogisticRegression](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/ml/LogisticRegression.java), [MLContext](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/MLContext.java), etc.) with Scala versions in order to alleviate the types of issues raised that are inherent with user-facing APIs.  We could simply place these Scala APIs in `src/main/scala` and keep the same package structure as before.

So, currently our `LogisticRegression` API is is part of the `org.apache.sysml.api.ml` package, and therefore the actual [LogisticRegression.java](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/ml/LogisticRegression.java) file is located at `src/main/java/org/apache/sysml/api/ml/LogisticRegression.java`.  We could instead place the Scala file at `src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala`, and use the same `org.apache.sysml.api.ml` package identifier.

To alleviate build issues, we can just integrate Scala compilation into Maven.  A quick search shows that the [scala-maven-plugin](https://github.com/davidb/scala-maven-plugin/) can do just this by simply adding a few lines to the [pom.xml](https://github.com/apache/incubator-systemml/blob/master/pom.xml), as shown in [1].  Eclipse integration appears to also be easy by simply adding a few more lines to the [pom.xml](https://github.com/apache/incubator-systemml/blob/master/pom.xml), as show at the bottom of [2].  Then, a simply `mvn package` or Eclipse (or IntelliJ) build work work as usual.  No additional JARs would be necessary in the final packaged `SystemML.jar` file either.

This route would also be great in preparation for [SYSTEMML-451](https://issues.apache.org/jira/browse/SYSTEMML-451), which aims add DML as an embedded Scala DSL.

* [1]: [http://davidb.github.io/scala-maven-plugin/example_java.html](http://davidb.github.io/scala-maven-plugin/example_java.html)
* [2]: [http://davidb.github.io/scala-maven-plugin/faq.html](http://davidb.github.io/scala-maven-plugin/faq.html)

niketanpansare:
@dusenberrymw +1 on creating Scala API for LogisticRegression, but am not sure whether we need to convert MLContext to Scala. 

scala-maven-plugin solutions looks good to resolve condition 1. The flag "cTestGoals" should ideally allow us to run junit tests [1]. 

Wrt condition 2, I had bad experience with Scala IDE during initial prototyping of Spark backend, especially for mixed scala/java projects (eg: code completion/refactoring/references). This was an year and half ago and may be it has got better. Before we make the switch, let's double check the IDE support on both windows/mac. Then @Wenpei can push this commit to src/main/scala.

[1]: http://davidb.github.io/scala-maven-plugin/example_cctest.html

dusenberrymw:
@niketanpansare Great!  The goal for replacing [MLContext](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/MLContext.java) as well would relate to the issue of default parameters.  There are currently 18 variations of `registerInput(...)`, and 23 variations of `execute(...)`/`executeScript(...)` and it just keeps getting messier with each new parameter or parameter type.

Also, I can't speak for Eclipse, but IntelliJ works great for Spark (which is now a mixed Scala/Java project) as well as SystemML.

Wenpei:
@niketanpansare @dusenberrymw 
I work out a init pom.xml for scala support, which works for `mvn clean package` to produce a `SystemML.jar`.
But I face two issues here:
1. It's very large for `SystemML.jar` , about 30M in my compute.  Not very sure but my be reasonable.
2. I don't know how to change compile order to make compile java at first, which cause class not found issues in eclipse. 

I am not very good at maven config / pom, so anyone can help to review this.

beside this, systemml has folder like "/systemml/src/test_suites/java", we may need change it to "/systemml/src/test/java", and add "systemml/src/test/scala" to follow plugin rule.

Regards.
Yu Wenpei.

niketanpansare:
@Wenpei Will look into this today.

dusenberrymw:
@Wenpei Thanks for updating the POM file.  
1. I think the issue with regards to size is due to the inclusion of the Scala library dependencies into to final JAR.  Since this specific log reg API will only ever be used from Spark, we can assume that those Scala libraries will be available at runtime, and thus we don't need to include them in the JAR.  Can you set the "scope" on all the Scala dependencies you added to "provided", indicating that they should not be included in the JAR?

niketanpansare:
@Wenpei Please do following things before making the switch:
1. Change all the scala dependency to provided. For example:
 <dependency>
        <groupId>org.scala-lang</groupId>
        <artifactId>scala-compiler</artifactId>
        <version>${scala.version}</version>
		<scope>provided</scope>
      </dependency>
      <dependency>
        <groupId>org.scala-lang</groupId>
        <artifactId>scala-reflect</artifactId>
        <version>${scala.version}</version>
		<scope>provided</scope>
      </dependency> ....
This should bring the jar down to 6M. 

2. Please make sure the jar works on a cluster without Spark/Scala (as well as with Spark/Scala) on all the execution modes (i.e. standalone, hybrid, ....). I have done sanity test on our internal cluster which has both Spark/Scala (on the above 6M jar), but I would highly recommend using our performance test suite to ensure that there are no unintentional side-effects. Also please make sure you run the integration test suite  (i.e. mvn verify).

3. Please try the new setup on eclipse (and the IDE of your choice (IntelliJ)) on Mac/Windows and see if you are getting compilation errors. 

4. Please provide instructions to switch to new setup (for example: download certain version of Scala, certain version of IDE, etc ...) to other committers.

mboehm7:
Well, I don't have a strong opinion on the wrapper code. The only hard requirement is that it has to work nicely with eclipse - no matter what.

niketanpansare:
@Wenpei  I was not able to reproduce the second issue.
Scala IDE

Wenpei:
@niketanpansare I change pom.xml to provider, I can get 6M `SystemML.jar` now.
And second issues gone after recompiled.

I will work on instruction those days. And do some test to make sure `SystemML.jar` works fine.

niketanpansare:
Thanks @Wenpei, I really appreciate the effort :)

Wenpei:
@niketanpansare I finish instruction, test suite.
But my compute report `less memory` error after run a hour for `mvn verify`, so it would be great if some one can launch a jenkins test for this. Or @bertholdreinwald can assign a account for me? :-)

For Instruction, I just leave it there. But may @deroneriksson want add it to http://apache.github.io/incubator-systemml/ as a part of instruction for contributor, can address this in another pr.

Regards.
Wenpei.

niketanpansare:
Jenkins test this.

akchinSTC:
test this please.

niketanpansare:
@Wenpei Please check https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/185/console and fix following errors:

1. "Multiple versions of scala libraries detected" warning can be eliminated by modifying the pom file.
2. "... error: not found" is most likely because the code might be behind git. For example: src/main/java/org/apache/sysml/parser errors could be related to changes introduced in https://github.com/apache/incubator-systemml/pull/61

Wenpei:
I will rebase this pr will new.
For multi scala version, it caused by jar dependence ,  will take a look but may common issues?

niketanpansare:
Thanks :)

Wenpei:
@niketanpansare I rebase pr with new. Please launch a test again.
For warning message, we didn't directly dependent to com.twitter. It should be caused by other jar(spark?).

niketanpansare:
@akchinSTC test this please.

niketanpansare:
test this please

niketanpansare:
test this please.

akchinSTC:
test this please.

Wenpei:
@niketanpansare I found build failed in scala doc generate phase (build and test pass). I will do some search today that why automatic doc generated after add scala-maven-plugin:3.2.2 in pom. 
But actually, there are some issues in old code when generate java/scala doc.

I will try to disable automatic doc generate.
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/192/console

Wenpei:
@akchinSTC I disable doc generated now, please test again.

akchinSTC:
test this please.

Wenpei:
@niketanpansare https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/196/console
Build pass.
I think this is  ready to merge.

niketanpansare:
LGTM

deroneriksson:
@niketanpansare @mboehm7 @gweidner 
Before merging, can someone verify that Scala is currently integrating smoothly in Eclipse with a Maven project? I experienced some issues with this (changes to classes not appearing in target/classes, standard debugging) when I tried last Friday, but it sounded like @gweidner was making some progress.

niketanpansare:
Thanks @gweidner for looking into it :)

gweidner:
I've been able to build the PR successfully within Eclipse when using Juno-based development environment.  
build_juno

Build failures within Luna-based environment under investigation.

Wenpei:
I have add eclipse version to dev/instruction.md to specify Juna version.
I didn't test eclipse luna, and spark community prefer Juna also.
https://cwiki.apache.org/confluence/display/SPARK/Useful+Developer+Tools

gweidner:
Thank you @Wenpei for the information.  When using the latest scala ide sdk (4.3.0 downloaded from http://scala-ide.org/download/current.html), I've observed two sets of errors:
lr_scala_errors
test_suite_errors

Wenpei:
@gweidner I do some test today, for scala IDE and eclipse luna
scala IDE:
1. delete project from eclipse
2. run command `mvn eclipse:clean`
3. import mvn project from wizard
4. If face `incompatation issues` , change IDE scala version 
`project->propertiest->scala compiler -> scala installation`
	to Fixed scala Installation: 2.10.5
5. If face `Not found type *` issues. 
  Run command `mvn package`
  And do `project -> refresh`

For eclipse luna:
Except scala IDE pulgin install, please make sure get update from "http://alchim31.free.fr/m2e-scala/update-site" if get `maketplace not found` issues

gweidner:
Thank you WenPei for your time and tips.  That did work except for Project-Clean... build.

The configuration that worked best for me was Luna + Scala IDE 4.0 plugin + maven scala connector.  That combination had no Eclipse build type issues shown in previous captures and full Project-Clean build was successful.   Note my testing was on Windows only.

Wenpei:
@gweidner @deroneriksson @niketanpansare 
I update my `project import.md` file. Please let me know if any concern for this pr.

niketanpansare:
@gweidner @deroneriksson Do you have any more concerns for this PR ?

@Wenpei Can you please send an email with the tips on dev mailing list with a link to this PR ?

mboehm7:
Would it work with Luna too? It would not be a good idea to force people to switch their dev environment right now.

gweidner:
Yes - I'm preparing a documentation PR with specific Eclipse version
information and tips for cases where developers prefer not to include Scala
in their IDE.

Wenpei:
@niketanpansare I send a mail to dev mailing list with PR and tips link.

src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
```diff
@@ -0,0 +1,169 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sysml.api.ml.scala
```
dusenberrymw: Ideally, this should just be in the `org.apache.sysml.api.ml` package, as it shouldn't matter to the end-user if the source was written in Scala.

src/main/scala/org/apache/sysml/api/ml/scala/ScriptsUtils.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.sysml.api.ml.scala
+
+import java.io.File
+
+object ScriptsUtils {
+  var systemmlHome = System.getenv("SYSTEMML_HOME")
+  def resolvePath(filename:String):String = {
+    import java.io.File
+    ScriptsUtils.systemmlHome + File.separator + "algorithms" + File.separator + filename
```
dusenberrymw: Could you look into the possibility of packing our algorithm scripts into the JAR during build time as perhaps a Maven "resource" that would be available to the Java process without needing to look into the user's filesystem?  One issue I see with the current approach is if a user wishes to attach the SystemML JAR to a cloud notebook (such as Databricks Cloud) in which an environment variable may not be able to be set.

dusenberrymw:
@Wenpei Awesome work here!  I left a couple of comments, the later of which would be particularly interesting to explore.  Past that, this should definitely be able to be merged.

src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
Wenpei: It cause a build error since we have both java and scala Logisiticregression in same package.
So i add scala version to seperated folder.

src/main/scala/org/apache/sysml/api/ml/scala/ScriptsUtils.scala
Wenpei: I will create a Jira to trace this.

src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
dusenberrymw: Okay, it might be better to move the Java version to a `org.apache.sysml.api.ml.java` folder (or possibly remove it), and move this file to `org.apache.sysml.api.ml` in order to encourage other Scala contributions to the API.

cc @niketanpansare  Thoughts?

src/main/scala/org/apache/sysml/api/ml/scala/ScriptsUtils.scala
dusenberrymw: Great!

src/main/scala/org/apache/sysml/api/ml/scala/LogisticRegression.scala
deroneriksson: It might be good to go ahead and merge this and then work out any additional updates in separate JIRAs/PRs since the conversation count is already 50+.

dusenberrymw: Yeah, that would be fine as well.  @Wenpei Can you go ahead and make a JIRA for this as well?

src/main/scala/org/apache/sysml/api/ml/scala/ScriptsUtils.scala
dusenberrymw: One other issue to track here is that there is currently a disconnect with the directory structure between the Git repo and our distribution tar/zip packages.  In the former case, if `SYSTEMML_HOME` is set to the base of the Git repo (which is the common pattern), then this line needs to change to `ScriptsUtils.systemmlHome + File.separator + "scripts" + File.separator + "algorithms" + File.separator + filename` to reflect that `algorithms` is nested under the `scripts` folder.  Otherwise, the associated test would fail.  However, in the latter case, our tar/zip packages have the `algorithms` folder at the base (not nested), so this line would be correct for the API to be able to be used.

dusenberrymw:
Also, just to track, I just pulled this into IntelliJ, and there are no issues encountered thus far.

Wenpei:
Jenkins, test this please.

Wenpei:
Build passed https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/210/
This PR is ready for merge.

dusenberrymw:
Great, LGTM.  I'll merge this today.

dusenberrymw:
I've opened [SYSTEMML-580](https://issues.apache.org/jira/browse/SYSTEMML-580) to reference this PR, and I've set the Reporter and Assignee to you, @Wenpei.

dusenberrymw:
PR merged.  Thanks, @Wenpei!

dusenberrymw:
Also, to anyone interested, there are several follow-up JIRA issues attached to [SYSTEMML-580](https://issues.apache.org/jira/browse/SYSTEMML-580).

PR #74, opened 2016-02-19
[SYSTEMML-534] Add optional console output of stats to Univar-Stats.dml

deroneriksson:
Add optional input parameter to print stats to console (default is off).
Update INPUT/OUTPUT/INVOKE comments at top of script.

### DIFF
--- a/scripts/algorithms/Univar-Stats.dml
+++ b/scripts/algorithms/Univar-Stats.dml
@@ -20,26 +20,26 @@
 #-------------------------------------------------------------
 
 #
-# DML Script to compute univariate statistics for all attributes 
-# in a given data set
+# DML Script to compute univariate statistics for all attributes in a given data set
 #
-# Three inputs:
-#     $1) X - input data
-#     $2) TYPES - row matrix that denotes the "kind"/"type" of all attributes
-#             kind=1 for scale, 
-#             kind=2 for nominal,
-#             kind=3 for ordinal
+# INPUT PARAMETERS:
+# -------------------------------------------------------------------------------------------------
+# NAME           TYPE     DEFAULT  MEANING
+# -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


deroneriksson:
Sample (optional) console output:

```
-------------------------------------------------
Feature [1]: Scale
 (01) Minimum             | 30.0
 (02) Maximum             | 83.0
 (03) Range               | 53.0
 (04) Mean                | 52.45751633986928
 (05) Variance            | 116.71458266366658
 (06) Std deviation       | 10.803452349303281
 (07) Std err of mean     | 0.6175922641866753
 (08) Coeff of variation  | 0.20594669940735139
 (09) Skewness            | 0.1450718616532357
 (10) Kurtosis            | -0.6150152487211726
 (11) Std err of skewness | 0.13934809593495995
 (12) Std err of kurtosis | 0.277810485320835
 (13) Median              | 52.0
 (14) Interquartile mean  | 52.16013071895425
-------------------------------------------------
Feature [2]: Scale
 (01) Minimum             | 58.0
 (02) Maximum             | 69.0
 (03) Range               | 11.0
...
```

Wenpei:
@deroneriksson Should be better to add OPTIONAL column for input parameter.

deroneriksson:
Hi @Wenpei . Thank you for reviewing. I wrapped the input parameters on line 39 to a new line for readability. The OPTIONAL column is a nice idea, but the DEFAULT column includes this information, since if a default value is present, the input parameter is optional. If we receive more feedback that this is unclear, maybe we can add an OPTIONAL column to all scripts in the future.

shirisht:
I agree with @deroneriksson ... the fact that a parameter has default value means it is optional.

shirisht:
LGTM

Wenpei:
Thanks, i agree

PR #75, opened 2016-02-20
[SYSTEMML-524] Doc turning off parfor parallelization in DML Lang Ref

deroneriksson:
Add description of how to turn off parfor parallelization to DML Language Reference.


deroneriksson:
Preview [here](http://deroneriksson.github.io/incubator-systemml/dml-language-reference.html#parfor-statement).

mboehm7:
LGTM - thanks @deroneriksson; 

Could you please also create a JIRA for an improvement of our parfor dependency analysis? The example of a false positive we show in the docs is actually something we should be able to infer automatically (with minor extensions).

deroneriksson:
@mboehm7 Thank you for reviewing. Also, I created SYSTEMML-553 for improving the parfor dependency analysis.

PR #77, opened 2016-02-21
[SYSTEMML-510] Generalized wdivmm w/ eps all patterns

gweidner:
This patch with related tests extends the existing wdivmm operator by two patterns for four operands:
t(t(U)%*%(W*(U%*%t(V)+x))), and (W*(U%*%t(V)+x)%*%V, where x is a scalar (e.g., epsilon).

Supporting these patterns allows a DML script to contain epsilon similar to an equivalent R script that included epsilon to prevent divide-by-zero.

### DIFF
--- a/docs/devdocs/MatrixMultiplicationOperators.txt
+++ b/docs/devdocs/MatrixMultiplicationOperators.txt
@@ -1,6 +1,6 @@
 #####################################################################
 # TITLE: An Overview of Matrix Multiplication Operators in SystemML #
-# DATE MODIFIED: 11/21/2015                                         #
+# DATE MODIFIED: 02/20/2016                                         #
 #####################################################################
 
 In the following, we give an overview of backend-specific physical matrix multiplication operators in SystemML as well as their internally used matrix multiplication block operations.
@@ -115,7 +115,8 @@ C) CORE MATRIX MULT PRIMITIVES LibMatrixMult (incl related script patterns)
                (c) t(t(U)%*%(W*(U%*%t(V)))), (d) (W*(U%*%t(V)))%*%V, 
                (e) W*(U%*%t(V)), (f) t(t(U)%*%((X!=0)*(U%*%t(V)-X))),
                (g) (X!=0)*(U%*%t(V)-X)%*%V, (h) t(t(U)%*%(W*(U%*%t(V)-X))),  
-               (i) (W*(U%*%t(V)-X)%*%V
+               (i) (W*(U%*%t(V)-X)%*%V,
+               (j) t(t(U)%*%(W/(U%*%t(V)+x))), (k) (W/(U%*%t(V)+x))%*%V
   - sequential / multi-threaded (same block ops, par over rows in X)                 
   - all dense, sparse-dense factors, sparse/dense-* x 9 patterns
 
--- a/src/main/java/org/apache/sysml/hops/QuaternaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/QuaternaryOp.java
@@ -819,10 +819,10 @@ public class QuaternaryOp extends Hop implements MultiThreadedHop
 		//MR operator selection, part1
 		double m1Size = OptimizerUtils.estimateSize(U.getDim1(), U.getDim2()); //size U
 		double m2Size = OptimizerUtils.estimateSize(V.getDim1(), V.getDim2()); //size V
-		boolean isMapWsloss = (!wtype.hasFourInputs() &&
+		boolean isMapWdivmm = (!(wtype.hasFourInputs() && !wtype.hasScalar()) &&
 				m1Size+m2Size < OptimizerUtils.getRemoteMemBudgetMap(true)); 
 		
-		if( !FORCE_REPLICATION && isMapWsloss ) //broadcast
+		if( !FORCE_REPLICATION && isMapWdivmm ) //broadcast
 		{
 			//partitioning of U
 			boolean needPartU = !U.dimsKnown() || U.getDim1() * U.getDim2() > DistributedCacheInput.PARTITION_SIZE;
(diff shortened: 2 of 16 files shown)


gweidner:
Verified all tests passed prior to submitting PR:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/103/

Also ran overnight on local Windows system which reported no failures:
Tests run: 4799, Failures: 0, Errors: 0, Skipped: 0

mboehm7:
thanks @gweidner - couple of comments:

1) Tests: For testing purposes it would be good to use a larger eps value (e.g., 0.1 instead of 1e-17); otherwise we're always in danger that it is ignored given the test tolerance of 1e-6.
2) Rewrites: Currently, you apply this rewrite for both / and *. Since the runtime only supports /, please change it to / only when checking for VALID WDIVMM binary operations. 
3) Instruction Generation: Could you please use the same instruction layout for CP and Spark and only use "instruction patching" for MR where it is required? This applies to the lops WeightedDivMM and WeightedDivMMR.
4) Spark Quaternary Instruction: According to 3, you would need to change the instruction parsing. Also, please think about a solution where eps is not put into the operator but treated as input data.  
5) LibMatrixMult: In matrixMultWDivMMDense, could the case "scalar and x==null" happen at all; if not, please simplify. In matrixMultWDivMMSparseDense, where do you actually pass in the scalar value? In all three methods, could you please unpack the scalar value once and reuse it over all iterations?

Once, this pr is in, could you please also have a look into step 2 of [SYSTEMML-510]? I'll take care of step 3.

gweidner:
Thank you Matthias for the feedback.  I've pushed commit for item 1 to use larger eps value and will push additional commits separately to incorporate the other comments.

gweidner:
Successful tests results for latest commit available at [build 174|https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/174/].

gweidner:
Updated RewriteAlgebraicSimplificationDynamic.java for 2 new patterns to '/' only.  Verified 84 tests in WeightedDivMatrixMultTest all passed in local test environment.  Full automated test build in progress [build 175|https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/175/].

gweidner:
All tests passed in [PR build 177](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/177/).
Previous PR build 175 reported single test failure in functions.recompile.ReblockRecompileTest.testReblockGroupedAggregate most likely due to randomly generated seed value in runReblockTest.

Pushed additional commit for comment (5) to update scalar handling in matrixMultWDivMMSparseDense, matrixMultWDivMMDense, matrixMultWDivMMGeneric of LibMatrixMult.

mboehm7:
These changes look good @gweidner - just some final cleanups: 
ad 3) Please use the same approach for WeightedDivMM as you did for WeightedDivMMR (both apply to the MR backend, where "R" just stands for "Reduce"-side w/ shuffle)
ad 5) It would be good to consistently use 'eps' instead of x[0].
6) Looking over (3) made me realize that this change currently always compiles a WeightedDivMMR for MR (even if both factors fit into the broadcast budget) and always a WeightedDivMM for Spark (even if the factors do not fit in). Could you please change QuaternaryOp for both MR and Spark. For Spark, please change the condition of isMapWsloss (currently it is ignoring forced op types and memory budgets). For MR, add similar conditions as you already did for Spark. Thanks.

gweidner:
Thank you Matthias for all the pointers!
I've pushed updated WeightedDivMM (MR exec type with prepScalarInput) and LibMatrixMult (consistent eps referencing/unpacking in matrixMultWDivMMDense).
Item 6 is in progress and item 4 under investigation.

gweidner:
All tests passed in [PR build 181](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/181/) and on-demand [build 104](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/104/).

gweidner:
@mboehm7  For comment 6, I've updated the logic for isMapWdivmm and added four new tests cases which force replication.  With these changes both WeightedDivMM and WeightedDivMMR can be generated as can be seen by the resulting instructions.  For example:

testWeightedDivMMLeftEpsDenseMRRep
----------  other inst     = MR redwdivmm 0.MATRIX.DOUBLE 3.MATRIX.DOUBLE 5.MATRIX.DOUBLE 0.1.SCALAR.DOUBLE.true 6.MATRIX.DOUBLE DIV_LEFT_EPS false false

testWeightedDivMMLeftEpsDenseMR
----------  mapper inst    = MR mapwdivmm 0.MATRIX.DOUBLE 1.MATRIX.DOUBLE 2.MATRIX.DOUBLE 0.1.SCALAR.DOUBLE.true 3.MATRIX.DOUBLE DIV_LEFT_EPS 

testWeightedDivMMLeftEpsDenseSP
------SPARK mapwdivmm _mVar4.MATRIX.DOUBLE _mVar5.MATRIX.DOUBLE _mVar6.MATRIX.DOUBLE 0.1.SCALAR.DOUBLE.true _mVar7.MATRIX.DOUBLE DIV_LEFT_EPS

testWeightedDivMMLeftDenseEpsSPRep
------SPARK redwdivmm _mVar4.MATRIX.DOUBLE _mVar5.MATRIX.DOUBLE _mVar6.MATRIX.DOUBLE 0.1.SCALAR.DOUBLE.true _mVar7.MATRIX.DOUBLE DIV_LEFT_EPS false false

gweidner:
All tests passed [PR build 184](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/184/).

mboehm7:
looks good - just two style comments:
1) could you please simplfy the double negation during lops construction to (!wtype.hasFourInputs() || wtype.hasScalar())
2) also, let's rename qop.hasScalar() to qop.getScalar as it returns a double not a boolean what the name suggests.

gweidner:
Thank you for the review!  I've updated QuaternaryOp, QuaternaryOperator and MatrixBlock.  All tests passed in [PR build 194](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/194/).

mboehm7:
LGTM

PR #79, opened 2016-02-23
Optionally set genRandData4ALS output format

deroneriksson:
Allow output format to be optionally specified rather than hardcoded.


mboehm7:
LGTM; however keep in mind, the only reason why this works (despite the fact that we don't support variable formats) is constant propagation. However, this is fine because most data generators do the same thing to simply usage.

PR #80, opened 2016-02-24
[SYSTEMML-240] Added instrumentation for spark collect, broadcast & parallelize

nakul02:
This is the first step towards SYSTEMML-240 (Stats util extension for lazy evaluation).
Changes Highlighted in Red

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/context/SparkExecutionContext.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/context/SparkExecutionContext.java
@@ -353,7 +353,9 @@ public class SparkExecutionContext extends ExecutionContext
 	@SuppressWarnings("unchecked")
 	public PartitionedBroadcastMatrix getBroadcastForVariable( String varname ) 
 		throws DMLRuntimeException, DMLUnsupportedOperationException
-	{
+	{		
+		long t0 = DMLScript.STATISTICS ? System.nanoTime() : 0;
+
 		MatrixObject mo = getMatrixObject(varname);
 		
 		PartitionedBroadcastMatrix bret = null;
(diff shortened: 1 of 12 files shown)


mboehm7:
nice @nakul02 - this is a great feature. Just a couple of comments to streamline this:

1) Layout: Could you please create a more compact output? Something like (a) Spark transfer counts (par, bc, col), (b) Spark transfer times (par, bc, col)
2) Measurement: You should only include the actual buffer pool interactions into this measurements. Please exclude instructions and specific jobs that are anyway measured separately. Also for collect and similar functions please include everything that corresponds to a given matrix as one measurement.
3) Statistic maintenance: Furthermore, please also wrap statistic maintenance into the DmlScript.STATISTIC flag - for very small usecases, stats maintenance itself becomes the bottleneck so we only measure if explicitly requested via -stats.
4) Thread-safeness: Keep in mind that broadcast/collect/parallelize might be called in a multi-threaded parfor context. So please use atomics to prevent anomalies like lost updates.
5) Consistency: For internal consistency, it would be good to remove the Spark class from Statistics and try to integrate your additions similar to existing measurements. Thanks.

nakul02:
@mboehm7 - couple of questions 

wrt 1) Layout:
Does this look good?
Changes Highlighted in Red

wrt 2) Measurement - 
a) do you mean, I should exclude times counted in by the "Heavy hitter timers"?
b) "please include everything that corresponds to a given matrix as one measurement." - does this also mean all the maps and other higher order functions besides the collect?

wrt 5)Consistency - I think it would make sense to move all spark specific stats into the Spark class, all MR specific stats into the MR class and so on. Right now everything is in this huge monolithic class. thoughts?

mboehm7:
ad 1) It goes into the right direction. Try to give it a meaningful name though, something like "Spark transfer times (...)" or "Spark trans times (...)". Also, for consistency, the key separator is ',' not '/'.
ad 2) I mean please exclude, for example, measurements of custom collects in instructions, utils or parfor related features - only focus on the main data transfer - otherwise it would rather confuse than help (e.g., measuring the parallelize of a single header line). Regarding the second aspect, if we collect, for example, an RDD into a single matrix than the entire process including matrix construction needs to be measured and counted as one collect.
ad 5) Well unless there is a good reason for additional complexity, we'll always stick to the simpler solution. So far, I don't see any advantages of this additional nesting since functionality-wise they are equivalent.

nakul02:
@mboehm7 - I've made some changes. 
My initial understanding of measuring time spent in collect(), broadcast() & parallelize() was to measure any invocations.
If I now understand this correctly, then the target is to measure invocations related to operands (matrix blocks). Is this correct?

mboehm7:
well, in general, we want to capture the collect/broadcast/parallelize calls in terms of data transfer between CP and Spark. For now we're only interested in the common case of buffer pool interactions, which means purely the calls in SparkExecutionContext. It would be good to do this in a coarse-grained manner, i.e., measure once for the entire toMatrixBlock including all related overheads, not for individual branches. In more detail:

1) Please fix the statistics output, there is an unnecessary space in the second introduced line, which would be better placed before the colon.
2) Remove the measurements from any parfor-related classes, spark utils, etc and purely focus on the SparkContext
3) Simplify the measurements in SparkExecutionContext, e.g., toMatrixBlock, and  getBroadcastForVar, to have one measurement surrounding all related overheads.
4) Broaden the measurement scope of getJavaRDDHandle to include the block creation not just the actual parallelize.

nakul02:
@mboehm7 - does this look good?
Submitted [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/109/) build

mboehm7:
LGTM - thanks @nakul02

deroneriksson:
@nakul02 @mboehm7 
Just wanted to verify (before merging into master) with you that when squashed and merged, there are only two files that are being updated by this PR? See https://github.com/apache/incubator-systemml/pull/80/files

mboehm7:
yes that's fine @deroneriksson

PR #81, opened 2016-02-25
[SYSTEMML-486] Java opts to standalone scripts

deroneriksson:
Allow Java options to be set via SYSTEMML_JAVA_OPTS env var.

### DIFF
--- a/bin/systemml.bat
+++ b/bin/systemml.bat
@@ -60,7 +60,7 @@ IF "%USER_DIR%" == "%PROJECT_ROOT_DIR%\bin" (
 )
 
 
-:: if the SystemML-config.xml does not exis, create it from the template
+:: if the SystemML-config.xml does not exist, create it from the template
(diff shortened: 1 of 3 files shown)


shirisht:
Is this the best way to provide java opts to standalone execution? It may misled users into thinking that they can pass  such java opts for driver/CP process in distributed execution as well -- something similar to `HADOOP_CLIENT_OPTS` in Hadoop.

Two options:
1. Either change the env var to make it explicit that it is applicable only for standalone execution
2. Instead of env var, have such property in the config file .. just like the way we had jvm properties for distributed execution. I prefer this but does not have a strong opinion.

deroneriksson:
@shirisht WRT option 1, I renamed the env var to SYSTEMML_STANDALONE_OPTS to attempt to clarify the standalone/singlenode execution. Does that seem clearer? Thank you for reviewing.

deroneriksson:
@shirisht @mboehm7 Are any other modifications required, or is this ready for merging? Thank you.

shirisht:
LGTM.

PR #83, opened 2016-02-26
[SYSTEMML-536] KNN Algorithm

Wenpei:
Hi

We would like to contribute KNN algorithm that we implement last year. 
@fmakari , please help to take a look.

Regards.
Yu Wenpei.


akchinSTC:

deroneriksson:
@Wenpei Thank you for this contribution. At the top of the script, could you update the license to the Apache license?

@frreiss @mboehm7
What is the procedure to accept new algorithms? After @fmakari validates the script, can the algorithm be merged in? I am asking because some other potential steps are 1) corresponding data generation script, 2) test cases added to test suite, 3) update algorithms reference, and maybe 4) PyDML equivalent? However, requirement of any of these additional points might discourage contributions. Would scripts/staging be the place to put an algorithm contribution that is validated but does not have the additional points 1) through 3)?

mboehm7:
Let's put it under ./scripts/staging for now, once we ran sufficient experiments and added required performance features we can put it under algorithms and add necessary documentation. 

Subsequently we can add tests too but we cannot require it given that only a subset of our existing algorithms is included in our test suite. Also, I don't think it's a good idea to replicate all algorithms in Pydml, excepts for selected testcases.

deroneriksson:
@mboehm7 WRT the existing algorithms, I would think that we probably should have tests for all the primary algorithms already included in the project in the ./scripts directory to validate that they are performing as expected.

mboehm7:
If you mean ./scripts/algorithms, then yes I absolutely agree. However, I don't think we should replicate all of them in pydml as it creates maintenance issues. Also, down the road it will become infeasible to have tests for all scripts ever written - so let's be selective here. 

As a higher-level comment, these kinds of tests are also usually not seen as the responsibility of the algorithm writer as the behavior of scripts is not supposed to change. So we cannot really require it but need to add these tests in a best effort manner ourself.

frreiss:
With regard to replicating scripts in PyDML: I personally think that we should replicate all of our first-tier algorithms in both dialects of the language. There is a substantial community of data scientists that greatly prefers the Python syntax to R syntax. If we do not ship our small set of fully-supported and fully-tested algorithms in both dialects, it sends the wrong message to the Python community.

The larger group of second-tier algorithms with "as-is" support can stay in one dialect.

deroneriksson:
@mboehm7 Yes, I meant the ./scripts/algorithms directory. Ideally it would be to nice to have a particular set of DML scripts (the ones in ./scripts/algorithms) that are all validated/tested/documented that are the gold standard. A user can be certain of the integrity of a script in ./scripts/algorithms. It 'works', it is high performance, it scales, it's easy to generate sample data, it's tested, and it's well documented.

At the same time, there is great value in encouraging algorithm writers to contribute additional algorithms without burdening them with other requirements (an R specialist shouldn't need to write any Java). Perhaps a new script can undergo an "incubation", which can involve contributions from the community (add test cases, validate performance, etc) to assist in its move to being a main algorithm.

We should make sure that our process of accepting new DML scripts: (1) maintains the integrity of existing first-tier algorithms and (2) encourages algorithm writers to contribute new scripts. This 'process' could be as simple as starting a new script in the staging directory and moving it to ./scripts/algorithms when it graduates once it meets the requirements of a first-tier algorithm. We just need to define what those requirements are so that we maintain the integrity of the first-tier algorithms.

mboehm7:
@deroneriksson please move these kinds of discussions over to the dev list (let's not start them on PRs) and concentrate on the KNN here.

Wenpei:
It is just dml script only, i would like to write datagenerate script ,test case and algorithm doc after review script.
But i don't think we will write pydml in future.

Wenpei

deroneriksson:
Thank you @Wenpei . That sounds great.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+getErr_k = function (  matrix[double] in_m_neighbor_index, 
+        matrix[double] in_m_cl, 
+        integer in_i_cl_type , 
+        integer in_i_k_min  ) 
+    return (  matrix[double] out_m_err  ){
+    i_col = ncol( in_m_neighbor_index  );
+    i_row  = nrow( in_m_neighbor_index  );  
+    m_neighbor_value = matrix( 0,i_row,i_col  );
+    
+    #parfor
```
fmakari: These nested for loops should be matricized.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+# prepareKNNData:
+#   Do data prepare - dummy for category variable and [-1,1] transform for continues variable 
+# Argument:
+# in_m_data                     input matrix as features
+# in_m_data_type                input matrix with 1 row denoting whether columns in X are 
+#                               continuous ( =1 ) or categorical ( =2 )
+# in_i_trans_continues          Option flag for continuous feature transformed:
+#                               0 = do not transform continuous variable;
+#                               1 = transform continuous variable;
+prepareKNNData = function (
```
fmakari: The current data preparation can be  expensive specially for categorical varibaes. Would it be possible to handle categorical variables directly without dummy coding? In this case we ca decide within different methods how to handle categorical variables, e.g. in distance computation and so on.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+    for( i in 1:ncol( in_m_data ) ){
+        #categorical
+        if( as.scalar( in_m_data_type[1,i] ) == 2 ){
+            tmp_i_max = max( in_m_data[,i] );
+            for( i_c in 1:tmp_i_max ){
+                out_m_data[,i_outmatrix_iter + i_c - 1] = ppred( in_m_data[,i],i_c,"==" );
+            }
+            #iter
+            i_outmatrix_iter = i_outmatrix_iter + as.integer( max( in_m_data[,i] ) );
+        }else{ #continuous
```
fmakari: rescaling continuous variables can be done one the whole input matrix without going through each column in the for loop.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+#                                           search
+# X_T                   String   ---   Y    Location to read the feature type of matrix X an T
+#                                           whether columns in X are continuous ( =1 ) or
+#                                           categorical ( =2 )
+# Y                     String   ---    Y   Location to read the input matrix as target
+# Y_T                   String   ---    Y   Location to read the target type of matrix Y whether
+#                                           columns in Y are continuous ( =1 ) or
+#                                           categorical ( =2 )
+# K                     String   ---    Y   Location to write the result of selected k value
+# PR                    String   ---    Y   Location to write the predicted k-d tree search result,
```
fmakari: All parameters and comments related to k-d tree or hashing method should be removed.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+getErr = function (  matrix[double] in_m_neighbor_index, 
+        matrix[double] in_m_cl, 
+        integer in_i_cl_type ) 
+    return (  matrix[double] out_m_err ){
+    i_col = ncol( in_m_neighbor_index );
+    i_row  = nrow( in_m_neighbor_index );   
+    m_neighbor_value = matrix( 0,i_row,i_col );
+    
+    #parfor
+    for( i in 1:i_col,check=0 ){
```
fmakari: These nested for loops should be matricized.

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+    matrix[double] out_m_norm_feature_importance
+     )
+{
+    n_feature = ncol(in_m_data)
+    n_record = nrow(in_m_data)
+    if(n_feature <= 1){
+        stop("can't estimate feature importance when ncol = 1")
+    }
+    m_err = matrix( 0,n_record,n_feature);
+    for(i_feature in 1:n_feature){
```
fmakari: parfor instead of for

scripts/algorithms/knn.dml
```diff
@@ -0,0 +1,745 @@
...
+    
+    out_m_err = matrix( 0,i_row,i_col - in_i_k_min + 1  );
+    if( in_i_cl_type == 2  ){
+        #category
+        m_correct = ppred( m_neighbor_value,in_m_cl[1:i_row,],"!="  );
+    }else{
+        #continues
+        m_correct = m_neighbor_value - in_m_cl[1:i_row,];#ppred( m_neighbor_value,in_m_cl,"-"  );
+    }
+    for( i in 1:i_col-in_i_k_min+1  ){
```
fmakari: parfor instead of for

akchinSTC:
add to whitelist.

akchinSTC:
ok to test.

dusenberrymw:
@Wenpei Can you move this to `./scripts/staging`?  Then the next steps will be to merge this, and then build some large-scale tests to stress it.

Wenpei:
@dusenberrymw I move this script to staging. It's ready to merge now.

We will submit another PR for unit test, document.

dusenberrymw:
Merged.  Thanks, @Wenpei!

PR #84, opened 2016-02-27
Reference full Java class names in g4 files

deroneriksson:
Remove header imports for ExpressionInfo and StatementInfo.
This removes "The import ... is never used" warnings in Eclipse.


mboehm7:
LGTM

PR #85, opened 2016-02-27
[SYSTEMML-410] Fix UpdateInPlace intermediate candidate

asurve:


mboehm7:
1) Thread-safeness: Please use Recompiler.recompileProgramBlockHierarchy instead of introducing a new recompileProgramBlocks method into the recompiler. This way you get a thread-safe version of the same functionality (without the danger of concurrent size modifications between DAG recompilations, which could render the control flow reconciliation logic invalid) and you don't need to handle the RecompileStatus.
2) Explain: Could you please enable the explain output for all hops that are flagged as update-in-place, not just DataGen ops - this would give us more debugging opportunities if things go wrong. 
3) Code style: Also, try to avoid unnecessary casts, e.g., write 0L instead of (long)0. It's just easier to read.

mboehm7:
LGTM

mboehm7:
Could you please rename the patch to SYSTEMML-410 as we closed SYSTEMML-282 when we created the epic SYSTEMML-409. Thanks.

PR #89, opened 2016-03-11
[SYSTEMML-545] Document Scala build support in Eclipse

gweidner:
Added document with screen captures to help developers prepare their Eclipse development environment to handle Scala in the SystemML code base.  Note developers can choose to skip the Scala components such that the JDT will continue to be used when working inside Eclipse.  Specific Eclipse and Scala plugin versions are also recommended.  See https://github.com/apache/incubator-systemml/pull/70 for detailed discussion including additional documentation plus the actual updated pom.xml which enables building Scala.


Wenpei:
@gweidner Great share.  LGTM

gweidner:
@Wenpei Thank you!  You really did all the heavy lifting in updating the pom.xml for Scala plus your new Scala code looks good.  Thanks again for all your work.

dusenberrymw:
@gweidner, @Wenpei  Great, thanks for getting the Eclipse integration worked out in prep for the Scala addition!

dusenberrymw:
Merged.  Thanks @gweidner!

PR #92, opened 2016-03-17
[SYSTEMML-573] Frame IO interface

asurve:


mboehm7:
thanks @asurve - I would just recommend a couple of minor cleanups:
1) Remove the special cases for write (empty frames, replication) and read (ismmfile, localfs) for now
2) Remove schema and names from the write api since the frameblock already carries this information
3) Could you provide overloaded reads for the cases (1) w/ schema and colnames, (2) w/ schema, (3) w/o schema/colnames (if not specified we would assume the schema as string)

mboehm7:
great - just one thing: please remove "getSchema( String fname )" from all classes - let's construct a schema of ValueType.STRING according to the number of columns if it is not specified.

mboehm7:
LGTM

PR #94, opened 2016-03-21
[SYSTEMML-586] Fix in-memory distribution for JMLC

deroneriksson:
Update pom.xml to build in-memory distribution.
Update inmemory.xml to include necessary dependencies for JMLC.
Update JMLC example in jmlc.md.

### DIFF
--- a/docs/jmlc.md
+++ b/docs/jmlc.md
@@ -94,10 +94,8 @@ In the Java below, we initialize SystemML by obtaining a `Connection` object. Ne
 (`"scoring-example.dml"`) as a `String`. We precompile this script by calling the `prepareScript` method on the
 `Connection` object with the names of the inputs (`"W"` and `"X"`) and outputs (`"predicted_y"`) to register.
 
-Following this, we read in the model (`"sentiment_model.mtx"`) and convert the model to a 47x46 matrix, where the
-last row of the matrix is for the `b` values. We set this matrix as the `"W"` input. Next, we create a random 46x46 matrix
-of doubles for test data with a sparsity of 0.7 and set this matrix as the `"X"` input. We then execute the script and
-read the `"predicted_y"` result matrix.
+Following this, we set matrix `"W"` and we set a matrix of input data `"X"`. We execute the script and read
+the resulting `"predicted_y"` matrix. We repeat this process. When done, we close the SystemML `Connection`.
 
 
 #### Java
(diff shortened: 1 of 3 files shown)


deroneriksson:
@bertholdreinwald I updated the inmemory.xml file to include several libraries that were required to run the JMLC example. Please let me know if this is how the in-memory jar should be packaged.

bertholdreinwald:
Deron, thanks for the fix. Couple of comments:
the inmemory JAR file is 15 MB now, but it used to be 9 MB. 6MB more than 
it used to be. Same thing happened to standalone jar which is 31 MB which 
is 6 MB more than it used to be. Wondering what changed.
JMLCExample:   using readScript() to read the mtx data file looks a little 
odd ... Want to just create an in-memory toy array to have some example 
data, which would also simulate the in-memory hand-over of the application 
data to SystemML?
an exhanced example, or additonal example should show that you can invoke 
.executeScript() several times with different data, and without preparing 
the script again.
up to you whether you want to show the compile/run for the example to 
highlight the use of the inmemory jar file:
COMPILE: javac -cp systemml-0.10.0-incubating-SNAPSHOT-inmemory.jar -d bin 
-s src src/JMLCExample.java
RUN: java -cp systemml-0.10.0-incubating-SNAPSHOT-inmemory.jar:bin 
JMLCExample


Regards,
Berthold Reinwald
IBM Almaden Research Center
office: (408) 927 2208; T/L: 457 2208
e-mail: reinwald@us.ibm.com

deroneriksson:
@bertholdreinwald Thank you for reviewing.

To determine the libraries to add to the in-memory jar, I added libraries one at a time and rebuilt each time as each 'class not found' exception occurred when I ran the JMLCExample in a separate project using the built in-memory jar. I repeated this process until the example ran with no exceptions, so I think the list of included libraries is about as minimal as possible given the current set-up. Would you like me to investigate the 6M increase in size to try to determine where the increase is coming from?

I agree that readScript() for the mtx data is a little strange. I'll update the examples with your suggestions.

deroneriksson:
@bertholdreinwald 
I compared the new in-memory jar with an old version on my machine from last summer (system-ml-5.0-SNAPSHOT-inmemory.jar). The old version was 9M but is missing classes needed in order to run. For example, if I run the JMLCExample with the old jar (and update the import packages to the old packages), I get "java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration". 

BTW, the size discrepancies were mainly from the following directories (unpacked):
org/apache/commons (3.2M new vs 0M old)
org/apache/hadoop (16.1M new vs 6.3M old)
org/apache/log4j (706K new vs 0 old)
org/apache/wink (160K new vs 0 old)
org/slf4j (73K new vs 0 old)
META-INF/maven (~190K new vs ~30K old)
SystemML files (8.08M new vs 7.37M old)
root directory files (110K new vs 0 old)

bertholdreinwald:
thanks, Deron!


Regards,
Berthold Reinwald
IBM Almaden Research Center
office: (408) 927 2208; T/L: 457 2208
e-mail: reinwald@us.ibm.com

mboehm7:
I am not so sure about the usefulness of this inmemory jar any longer. Originally (back in the days, still with our ant build) it was meant to package - in a very fine-grained manner - only files that are actually needed when running in a pure in-memory scoring setting (e.g., all hadoop file system implementations and a lot of SystemML internal stuff like MR or Spark instructions are just unnecessary there but people are concerned with the footprint of their UDF jars which package and call SystemML). Furthermore, the plan was to introduce a runtime mode which would prevent the instantiation of JobConf and alike to reduce the required classes to very few APIs. Let's keep this jar for now but once we create the separate build for SystemML's core block runtime we might want to reconsider alternatives.

deroneriksson:
@mboehm7 Alternatives sound like a good idea.

@bertholdreinwald I updated the JMLC example. It can be previewed [here](http://deroneriksson.github.io/incubator-systemml/jmlc.html). Let me know if further updates are needed or if this is sufficient.

bertholdreinwald:
LGTM


Regards,
Berthold Reinwald
IBM Almaden Research Center
office: (408) 927 2208; T/L: 457 2208
e-mail: reinwald@us.ibm.com

PR #95, opened 2016-03-21
[SYSTEMML-573] Frame IO Read

asurve:
Added reading functionality for frame in the text, cell, and binary format.
May need to look closely as Binary format, when test code gets written.


mboehm7:
Thanks @arvind. It's too late now but I would have started with the writers as this allows to add separate component tests right away. The current patch has a couple of issues:

1) FrameIndexes/IndexedFrameBlock: please remove these additional data structures as they are plain copies of MatrixIndexes and IndexedMatrixValue and they are not needed for frames.
2) Unnecessary binary read methods: Also, please delete the binary read methods readBinaryBlockBlocksFromHDFS() and readIndexedFrameBlocks() as they are unnecessary.
3) Binary block serialization: Don't use the binary block serialization framework - it will reject frameblocks anyway as they are not supported.
4) ValueType handling: Please remove the custom value type handling from all text readers and use UtilFunctions.stringToObject() instead.
5) FrameBlock copy: It would be good to add some testcases to org.apache.sysml.test.integration.functions.frame for these copy operations. These tests would have shown that the copy cannot work like that: (a) the copy constructor only does a shallow copy of column references, (b) the copy of partial matrix blocks will run into arrayoutofbounds on the column references and also never copies the actual data.

mboehm7:
LGTM - thanks @asurve Just out of curiosity: Where do you intend to use the UtilFunctions.compareTo? In order to support sparse cells we would need to extend this with appropriate null checks as well.

asurve:
Thanks, its used in test program FrameCopyTest.java, which was not checked in, but its now. Please look into it. thanks

PR #96, opened 2016-03-24
[SYSTEMML-573] Frame write

asurve:
Its showing changes from earlier commit though those are push but not reflected in git repo yet.

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/io/FrameWriter.java
+++ b/src/main/java/org/apache/sysml/runtime/io/FrameWriter.java
@@ -28,7 +28,7 @@ import org.apache.sysml.runtime.DMLUnsupportedOperationException;
 import org.apache.sysml.runtime.matrix.data.FrameBlock;
 
 /**
- * Base class for all format-specific matrix writers. Every writer is required to implement the basic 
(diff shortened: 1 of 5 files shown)


mboehm7:
thanks @asurve; this is a good start - couple of comments though:

1) Binary Block Writer: The method writePartitionedBinaryBlockFrameToHDFS is not required and can be removed. Also please put a comment in there that this write uses fixed size blocks with block-encoded keys. In the future, we might change that, especially the keys (e.g., starting offset).
2) CSV File Merge: the methods  mergeCSVPartFiles and addHeaderToCSV work on bytes. Hence there is no need to replicate them here for frames.
3) Null Checks: Keep in mind that frames might contain sparse cells (w/ null objects). You need to handle this when trying to convert them to strings. Also, for textcell, sparse cells should not be materialized into the output (it's a sparse format).
4) String Conversions: The cast to string for CSV and textcell would crash for any schema other than ValueType.STRING. Please, use UtilFunctions.objectToString() here.

mboehm7:
I'm sorry for the wrong direction, but instead of UtilFunctions.objectToString, we should just call toString for objects !=null. Thanks.

mboehm7:
LGTM

PR #97, opened 2016-03-24
SYSTEMML-594 Add Jupyter Example

dusenberrymw:
This adds a PySpark Jupyter Example notebook, and associated
documentation of the PySpark MLContext API.

Documentation preview available [here](https://dusenberrymw.github.io/incubator-systemml).


dusenberrymw:
cc @deroneriksson, @lresende

deroneriksson:
Thank you @dusenberrymw. This is really great!

Couple things:
1) On line 1050 of spark-mlcontext-programming-guide.md, should "prod_i" be "prod_j"? (max_prod_j = X_train.select(F.max("prod_i")).first()[0])
2) On line 1001, the download link is empty "...can be downloaded \[here\]()". Should we mention the path in the project to the file rather than having a download link, or should we include this example notebook file under docs/ so that it can be easily downloaded from the online documentation? I guess I would lean towards the second approach since that is most convenient for the end user, and this example is a special notebook example since it is included in the official documentation. What are your thoughts?

dusenberrymw:
@deroneriksson Great catches.  I've fixed the first one, and for the second, I'd like to have the link point to the raw file in the GitHub repo.  So, in that case it will still download as easily, but we don't have to maintain two separate files.

dusenberrymw:
Also, I added two links: one to the (future) location on GitHub, which will automatically render the notebook as if it was opened locally, and one link to the raw download.

deroneriksson:
@dusenberrymw Great! Linking to the raw notebook file in the GitHub repo sounds good. At some point before our 1.0 release, we might want to revisit the issue since if it's packaged under docs/, we can have in-sync snapshots of the entire documentation site for each release version.

Great job. LGTM.

dusenberrymw:
Yeah, definitely.  Thanks!

PR #99, opened 2016-03-29
[SYSTEMML-573] Frame write cleanup

asurve:


mboehm7:
thanks @asurve  It would be good to reuse the internal arrays as well - otherwise there would be no point of such a reset functionality. You can do that by simplify setting the size of each column to 0 (the implementation already differentiated between logical size and physical length in order to allow exponential resizing for log resizing steps).

asurve:
Please check latest changes.

mboehm7:
1) Please, remove any IOExceptions from the FrameBlock api as the block itself will never do IO and hence this is misleading.
2) The reset call needs to cleanup the schema and column names too
3) Also, please don't mix the code for reset/allocation with the copy of schema because it should be possible to call reset on an empty block as well.
4) Why are the copy parameters rlen, clen, and rlDest required - isn't this redundant to rl, ru, cl, cu, and src?

asurve:
Addressed changes from point 1 (residue from other change), 2, and 3. Partially from 4.
Wrt comment 4, I keept rlen, and clen to be consistent with other copy functions, but removed now.
   I need rldest, as need to know which position data to be copied, information is not available in any other source other than this parameter.

mboehm7:
LGTM - I'm still not 100% happy with the copy signature but I'll clean this up whenever I touch it the next time

PR #100, opened 2016-03-31
[SYSTEMML-604] Undefined function error message line, col, filename

deroneriksson:
Previously, the following DML:
```
print("howdy");
a=foo();
```
produced error message with null filename and no line number or column number set:
```
Caused by: org.apache.sysml.parser.LanguageException: ERROR: null -- line 0, column 0 -- function foo is undefined in namespace .defaultNS
	at org.apache.sysml.parser.StatementBlock.isMergeableFunctionCallBlock(StatementBlock.java:201)
```

This update populates the filename, line number, and column number:
```
Caused by: org.apache.sysml.parser.LanguageException: ERROR: example.dml -- line 2, column 0 -- function foo is undefined in namespace .defaultNS
	at org.apache.sysml.parser.StatementBlock.isMergeableFunctionCallBlock(StatementBlock.java:208)
```

### DIFF
--- a/src/main/java/org/apache/sysml/parser/StatementBlock.java
+++ b/src/main/java/org/apache/sysml/parser/StatementBlock.java
@@ -195,11 +195,6 @@ public class StatementBlock extends LiveVariableAnalysis
 			//function calls (only mergable if inlined dml-bodied function)
 			if (sourceExpr instanceof FunctionCallIdentifier){
 				
-				sourceExpr.setBeginLine(stmt.getBeginLine());
-				sourceExpr.setBeginColumn(stmt.getBeginColumn());
-				sourceExpr.setEndLine(stmt.getEndLine());
-				sourceExpr.setEndColumn(stmt.getEndColumn());
-				sourceExpr.setFilename(stmt.getFilename());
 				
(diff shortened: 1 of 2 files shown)


deroneriksson:
@mboehm7 Is this the proper way to set this information?

mboehm7:
Although it works, I would expect the setup of positions and filenames during initial parsing (e.g., DmlSyntacticValidator) for atomic expressions and the propagation for composite expressions either in the syntactic parser or during validate. Also please use the one-liner 'setAllPositions'. Thanks @deroneriksson

deroneriksson:
@mboehm7 Moved to CommonSyntacticValidator. Let me know if this is satisfactory.

mboehm7:
LGTM, but it would be great to double check if this fix also applies to external functions or if additional extensions are required.

deroneriksson:
Not sure if I entirely understand WRT external functions, but if I try to cause an issue with a function defined with externalFunction (use wrong number of args, have the java function throw an exception, etc), typically an error such as "Runtime error in program block generated from statement block between lines 7 and 7" is generated.  Let me know if you'd like me to handle a specific external function situation. I will go ahead and merge this PR.

mboehm7:
that's fine, please go ahead and merge the PR.

PR #101, opened 2016-03-31
[SYSTEMML-607] Fix error message when namespace not found

gweidner:
Minor fix to error message that gets logged with exception raised when namespace not found.  This was discovered while investigating namespace handling of imported functions.


mboehm7:
LGTM

PR #102, opened 2016-03-31
Corrected typo in documentation

ethanyxu:
Correct syntax typo in 'if statement' documentation.
Added if - else if syntax.

### DIFF
--- a/docs/dml-language-reference.md
+++ b/docs/dml-language-reference.md
@@ -289,7 +289,7 @@ The syntax for an if statement is as follows:
     } ]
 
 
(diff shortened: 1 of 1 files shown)


akchinSTC:

mboehm7:
thanks @ethanyxu - the addition of "if-else if-else" is useful but the "[ ]" indicates that the else branch is optional. Maybe you can merge both into meta level (both else-if and else branches are optional but else-if can occur many times)

ethanyxu:
Ah I see thanks for the clarification. I included both 'else if' and 'else' both in the main syntax body with '[]'. Also reworded the description to reflect the change.

mboehm7:
@ethanyxu could you please do one more cleanup: remove the sentence starting "The statement(s) in any of three bodies must be surrounded by braces ..." as this is not required and maybe even add a hint that we actually support this.

ethanyxu:
Sure. It's edited accordingly.

mboehm7:
LGTM - thanks @ethanyxu

PR #103, opened 2016-04-03
[SYSTEMML-573] Frame read-write test

asurve:


mboehm7:
thanks @asurve - could you please consolidate writeAndVerifyBinaryBlockData, writeAndVerifyTextCSVData, and writeAndVerifyTextCellData into a single method that simply goes through FrameWriterFactory?

asurve:
thanks, addressed the change

mboehm7:
it goes into the right direction but why can't we just use FrameReaderFactory and FrameWriterFactory to avoid handling input and output infos - that's exactly what they're meant for, right?

asurve:
yes, I should have. I did the change. I have added another signature FrameReaderFactory.createFrameReader(iinfo, prop) to be consistent with FrameWriterFactory and to avoid burden on caller.

mboehm7:
ok great - just two more things:
1) you don't need to specify both InputInfo and OutputInfo - you can always get the corresponding input/output with InputInfo.getMatchingOutputInfo or OutputInfo.getMatchingInputInfo
2) there is a redundant semicolon in FrameReadWriteTest, line 170

mboehm7:
LGTM - @asurve could you please double check that the tests are still working after merging. Thanks.

PR #106, opened 2016-04-06
[SYSTEMML-611] Enhance display of parsing errors for user

deroneriksson:
Output actual parse issues via DMLScript invocation to standard error.
Separate file and string invocations of parser in MLContext.
Make parse issues available higher in call stack via ParseException.
Fix null pointer exception issue in exitDataIdExpressionHelper.
Some more informative parse error messages.
Update SetWorkingDirTest to read from standard error where needed.

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -83,6 +83,7 @@ import org.apache.sysml.utils.Statistics;
 import org.apache.sysml.yarn.DMLAppMasterUtils;
 // import org.apache.sysml.utils.visualize.DotGraph;
 import org.apache.sysml.yarn.DMLYarnClientProxy;
+import org.xml.sax.SAXException;
 
 
 public class DMLScript 
@@ -188,7 +189,9 @@ public class DMLScript
 		String[] otherArgs = new GenericOptionsParser(conf, args).getRemainingArgs();
 		
 		try {
-		DMLScript.executeScript(conf, otherArgs); 
+			DMLScript.executeScript(conf, otherArgs);
+		} catch (ParseException pe) {
+			System.err.println(pe.getMessage());
 		} catch (DMLScriptException e){
 			// In case of DMLScriptException, simply print the error message.
 			System.err.println(e.getMessage());
@@ -196,7 +199,7 @@ public class DMLScript
 	} 
 
 	public static boolean executeScript( String[] args ) 
-		throws DMLException
+		throws DMLException, ParseException
 	{
 		Configuration conf = new Configuration(ConfigurationManager.getCachedJobConf());
 		return executeScript( conf, args );
@@ -210,9 +213,10 @@ public class DMLScript
 	 * @param suppress
 	 * @return
 	 * @throws DMLException
+	 * @throws ParseException
 	 */
 	public static String executeScript( Configuration conf, String[] args, boolean suppress) 
-		throws DMLException
+		throws DMLException, ParseException
 	{
 		_suppressPrint2Stdout = suppress;
 		try {
@@ -230,10 +234,11 @@ public class DMLScript
 	 * @param conf
 	 * @param args
 	 * @return
-	 * @throws LanguageException 
+	 * @throws DMLException 
+	 * @throws ParseException 
 	 */
 	public static boolean executeScript( Configuration conf, String[] args ) 
-		throws DMLException
+		throws DMLException, ParseException
 	{
 		boolean ret = false;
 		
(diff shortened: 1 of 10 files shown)


deroneriksson:
Given the following DML:

```
boo();
x=blah();
print("1", "2");
write();
```

Currently the parse errors are logged to log4j. If the user does not have log4j configured, the user gets the following output:

```
Exception in thread "main" org.apache.sysml.api.DMLException: org.apache.sysml.parser.ParseException: One or more errors found during parsing (could not construct AST for file: example.dml). Cannot proceed ahead.
	at org.apache.sysml.api.DMLScript.executeScript(DMLScript.java:344)
	at org.apache.sysml.api.DMLScript.main(DMLScript.java:191)
Caused by: org.apache.sysml.parser.ParseException: One or more errors found during parsing (could not construct AST for file: example.dml). Cannot proceed ahead.
	at org.apache.sysml.parser.dml.DMLParserWrapper.parse(DMLParserWrapper.java:98)
	at org.apache.sysml.api.DMLScript.execute(DMLScript.java:581)
	at org.apache.sysml.api.DMLScript.executeScript(DMLScript.java:332)
	... 1 more
```

A stacktrace is mostly useless for a user who is writing DML and has parse issues in the DML. In addition, the above message does not tell the user anything about the parse issues that actually occurred.

If the above DML is executed with the code in this PR, the user sees the following to standard error:

```
--------------------------------------------------------------
The following 3 parse issues were encountered:
#1 example.dml [line 1:0] [Validation error] -> boo();
   function call needs to have lvalue (Quickfix: change it to 'tmpVar = boo(...)')
#2 example.dml [line 3:0] [Validation error] -> print("1", "2");
   print() has only one parameter
#3 example.dml [line 4:0] [Validation error] -> write();
   incorrect usage of write function (at least 2 arguments required)
--------------------------------------------------------------
```

This output contains the information that the user actually needs to fix the user's DML. Additionally, the stacktrace information isn't included since this information is distracting and not useful for a user trying to fix parse errors in the DML.

This PR also makes this information avaible via ParseException so that it is available at higher levels of the call stack, which can help when handling parse situations in a notebook, for example.

This PR also contains other fixes and updates.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -188,15 +189,17 @@ public static void main(String[] args)
...
+			System.err.println(pe.getMessage());
 		} catch (DMLScriptException e){
 			// In case of DMLScriptException, simply print the error message.
 			System.err.println(e.getMessage());
 		}
 	} 
 
 	public static boolean executeScript( String[] args ) 
-		throws DMLException
+		throws DMLException, ParseException
```
frreiss: If you're going to expose ParseException here, consider either renaming it to DMLParseException or making it a subclass of DMLException. Other systems and libraries use the name "ParseException", and we don't want users to have to type enter fully-qualified class names because they ended up with code that throws two kinds of "ParseException".

src/main/java/org/apache/sysml/parser/ParseException.java
```diff
@@ -19,74 +19,231 @@
...
+		StringBuilder sb = new StringBuilder();
+		sb.append("\n--------------------------------------------------------------");
+		if (parseIssues.size() == 1) {
+			sb.append("\nThe following parse issue was encountered:\n");
+		} else {
+			sb.append("\nThe following " + parseIssues.size() + " parse issues were encountered:\n");
+		}
+		int count = 1;
+		for (ParseIssue parseIssue : parseIssues) {
+			if (parseIssues.size() > 1) {
```
frreiss: Consider putting this code into ParseIssue.

src/main/java/org/apache/sysml/parser/ParseException.java
```diff
@@ -19,74 +19,231 @@
...
+	 * This constructor takes a list of parse issues that were generated during
+	 * script parsing and the original DML/PyDML script String.
+	 * 
+	 * @param parseIssues
+	 *            List of parse issues (syntax errors, validation errors, and
+	 *            validation warnings) generated during parsing.
+	 * @param scriptString
+	 *            The DML/PyDML script String.
+	 */
+	public ParseException(List<ParseIssue> parseIssues, String scriptString) {
```
frreiss: I like how you've created separate objects for each of the parse errors. UI components for editing scripts are much easier to write when errors are exposed this way.

frreiss:
LGTM.

src/main/java/org/apache/sysml/api/DMLScript.java
deroneriksson: I will create a JIRA to rename the exception and/or subclass DMLException since Matthias is currently making modifications to the exception hierarchy.

PR #111, opened 2016-04-12
Inconsistency in descriptions

ethanyxu:
Changes are all in descriptions. Didn't touch code.
- Corrected inconsistency of description and examples in argument. 

 - In the explanation of argument 'sv', example 1 and 2 contradicted the statement: "sum(sv) must be 1".  
 - Example 2 (sv = (0.25, 0.25, 0.25, 0.25, 0.25)) seems like a typo(?), because when `i == 5` in `parfor`, `T2` (line55) would be a zero column, so `P` (line57) would be empty.  

- Corrected location of 'sample.dml' in the example: From `algorithms/utils/sample.dml` to `./scripts/utils/sample.dml`.

- Added slightly more wordy explanations, as new users may try to learn dml from sample scripts like this one.

### DIFF
--- a/scripts/utils/sample.dml
+++ b/scripts/utils/sample.dml
@@ -20,26 +20,28 @@
 #-------------------------------------------------------------
 
 #
-# Sampling of a data set into number of samples specified as percentages in column vector
+# Randomly split data into a number of disjoint subsets (by rows). The sizes of the subsets are specified in terms of fractions, stored as a 1-column vector in a separate input file (see parameter 'sv'). 
 #
 # Parameters:
 #    X       : (input)  input data set: filename of input data set
-#    sv      : (input)  sampling vector: filename of 1-column vector w/ percentages. sum(sv) must be 1.
-#                         e.g. sv = (0.2)   for a 20% sample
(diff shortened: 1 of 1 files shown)


akchinSTC:

mboehm7:
thanks @ethanyxu that's a good catch - the example of 5 partitions is indeed wrong. 

Could you please slightly change the description: (1) it would be good to relax statements like "equal sized disjoint subsets" as the the outputs are not necessarily equal sized (they are *in expectation* equal sized), (2) also, "randomly split" might create the wrong impression that sum(sv)=1 although you might very well have sum(sv)<1 and use it for sampling, (3) could you please break very long lines (e.g., line 23) and fix the indentation of lines 28-30? Thanks.

dusenberrymw:
add to whitelist.

ethanyxu:
Thank you @mboehm7 . Changes modified according to your comment.

mboehm7:
LGTM - in addition, I would just slightly modify line 29 to "less than or equal 1"

ethanyxu:
Done.

mboehm7:
great - thanks @ethanyxu

PR #112, opened 2016-04-12
[SYSTEMML-551] Enhance JMLC javadocs

deroneriksson:


mboehm7:
thanks @deron that looks really good. Just two comments: 
(1) Could you please delete package-info.java (potentially move the java doc to connection) as most people only read class/method documentation. I've seen it in other projects too but our convention in SystemML is not to have these unnecessary files. You might want to bring up a discussion but let's separate this out.
(2) Non-zero arguments: Please, add to the nnz parameter description something like "-1 to indicate unknowns".

deroneriksson:
@mboehm7 Made requested updates.

mboehm7:
LGTM - thanks @deroneriksson

PR #113, opened 2016-04-13
[SYSTEMML-573] Frame CSV Spark read/write

asurve:


mboehm7:
thanks @asurve  but this requires some work - please keep the intended distributed frame representation (https://issues.apache.org/jira/browse/SYSTEMML-560) in mind.

1) binaryBlock2CSV slice: The slicing of binary blocks to rows is not required for frames as each frame block covers all columns.
2) csv2BinaryBlock nnz: Without the CSVAnalysisFunction, the accumulator from which you try to obtain the nnz will always return 0. Furthermore, for frames we don't even maintain the nnz, so please remove the related code.
3) csv2BinaryBlock shuffle: The whole point of the variable length frameblocks is to prevent shuffle. Currently, the mergeBykey causes unnecessary shuffle and should be removed. Please, also delete the related MergeBlocksFunction and mergeByKey primitives.
4) Blocksize: Again, we want to create variable size blocks at the end of partitions, so we cannot use UtilFunctions.computeBlockSize which would always give you 1k, even at the end of partitions.
5) Tests: Please factor out your new testcases into a separate converter test class as it has nothing to do with frame read/write. Given this refactoring you could also drop the execution type as it is misleading because these are component tests not integration tests on dml script level.

asurve:
I have addressed the changes.

mboehm7:
6) Binary block keys: on csv to binary block, it seems that we output the wrong keys (block indexes instead of block row offset) which is wrong with variable-sized blocks.
7) Formatting: Finally, there are also some formatting issues, e.g.,  FrameRDDConverterUtils, line 295.

asurve:
addressed the comment.

mboehm7:
LGTM

PR #114, opened 2016-04-13
[SYSTEMML-265] Update All Performance Tests to Use More Iterations

nishi-t:
Simply update some scripts to use 20 iterations, but don't touch certain tests(e.g. k-means).


akchinSTC:

mboehm7:
thanks @nishi-t for bringing up this topic. In general, I think it would be useful to consistently change the number of max iterations to something like 10. However, the currently patch modified the wrong argument (intercept which has algorithm-specific meaning and only takes the values 0/1 or 0/1/2). For the maximum number of iterations please have a look at the arguments 'maxiter' and 'moi'/'mii'. That reminds me, it would be also useful to create consistency between the names of these arguments across all dml scripts. Furthermore, for small data, we might also want to introduce a number of runs (what you most likely were interested in modifying here, right).

dusenberrymw:
add to whitelist.

nishihs:
Thank you. @mboehm7 , @dusenberrymw 
Oops! I misunderstood the issue. I'll do it following your comments.

nishi-t:
Thank you. @mboehm7 , @dusenberrymw 
Ooops! I misunderstood the issue. I'll do it following your comments.

nishi-t:
@mboehm7 , @dusenberrymw sorry for the delay, been busy the last week. I just partially addressed your comments in the commit 49b0858, but still left the task to create consistency between the names of arguments. I want to hear your opinion for the left task. My opinion is that `maxiter` in m-svm.dml (or l2-svm.dml) is inconsistent, because other scripts(e.g. LinearRegDS, Kmeans.dml) use `maxi` as name of the same meaning argument. I think `maxiter` (or `maxi`) should be `maxi` (or `maxiter`), but `moi` should be not changed because `moi` includes meaning outer-iterations. Is it correct as understanding following comment.
> That reminds me, it would be also useful to create consistency between the names of these arguments  across all dml scripts. 

Thank you.

bertholdreinwald:
Thanks @nishi-t . Wrt changing names of arguments, please be aware that this will create an incompatibility if somebody implemented an application using the old names. Hence, this should be carefully introduced and well documented for app developers to be aware of. How many names would you have to change?

nishi-t:
@bertholdreinwald Thank you for your comments. 
> How many names would you have to change?

It is about 6 words in scripts and 10 words in the document related to the scripts. I assumed changing `maxiter` to `maxi` in the following files.

- scripts/algorithms/{l2-svm.dml | m-svm.dml}
- scripts/perftest/{runL2SVM.sh | runMSVM.sh}
- docs/algorithms-classification.md

However, I understood that I should be carefully for changing names. If these names are not needed to  change now, I'm not going to do it in this PR.

dusenberrymw:
@nishi-t Thanks for updating this!  Can you increase the number of iterations to 20 instead of 10 though?  In specific cases, such as the k-means script, we've had issues when the number of iterations is below 20.

nishi-t:
Thanks @dusenberrymw . Addressed your comment in the commit 03529f3

dusenberrymw:
This **LGTM**.  Apologies for the delay.  Thanks, @nishi-t!

PR #118, opened 2016-04-14
[SYSTEMML-632] Handle source statements in imported scripts

gweidner:
This patch expands parser wrapper import statement processing to handle cases where an imported script1 also imports and calls functions in another script2.  Previously, the custom function calls in script1 to script2 would fail due to namespace undefined in main script.  Test cases were added for both dml and pydml where an imported script sources additional script(s) that also source scripts.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DMLProgram.java
+++ b/src/main/java/org/apache/sysml/parser/DMLProgram.java
@@ -48,7 +48,6 @@ import org.apache.sysml.runtime.instructions.CPInstructionParser;
 import org.apache.sysml.runtime.instructions.Instruction;
 
 
-
 public class DMLProgram 
 {
 	
@@ -63,7 +62,6 @@ public class DMLProgram
 		_blocks = new ArrayList<StatementBlock>();
 		_functionBlocks = new HashMap<String,FunctionStatementBlock>();
 		_namespaces = new HashMap<String,DMLProgram>();
-		_namespaces.put(DMLProgram.DEFAULT_NAMESPACE,this);
 	}
 	
 	public HashMap<String,DMLProgram> getNamespaces(){
--- a/src/main/java/org/apache/sysml/parser/dml/DMLParserWrapper.java
+++ b/src/main/java/org/apache/sysml/parser/dml/DMLParserWrapper.java
@@ -184,22 +184,22 @@ public class DMLParserWrapper extends AParserWrapper
 			List<ParseIssue> parseIssues = errorListener.getParseIssues();
 			throw new ParseException(parseIssues, dmlScript);
 		}
-		dmlPgm = createDMLProgram(ast);
+		dmlPgm = createDMLProgram(ast, sourceNamespace);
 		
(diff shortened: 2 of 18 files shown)


gweidner:
All tests passed in [OnDemand build 143](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/143/).

As an additional manual test outside the unit tests, reused the deep learning scripts from SYSTEMML-618 and moved the workaround source statements to files that directly use the imported functions.  For example, modified tests.dml imported grad_check.dml and test.dml only.  Similarly, grad_check.dml and test.dml imported the files each one directly used.
Verified the revised tests.dml ran successfully.

Note [SYSTEMML-631](https://issues.apache.org/jira/browse/SYSTEMML-631) opened to address handling of potential namespace conflicts.

mboehm7:
thanks @gweidner - I would recommend to also add some tests with functions that are not inlined in order to test the runtime part as well. Right now we inline any "single-statement-block functions" (recursively).  So for example a function with control flow like loops or non-folded branches would do the trick.

dusenberrymw:
@gweidner This is awesome!  I ran some tests locally, and the intended use case of executing source statements in a sourced module so that those required modules don't have to be sourced at the top-level script works great.  Thanks!

dusenberrymw:
Also, I found another specific case that we'll want to test for as part of the follow-up [SYSTEMML-631](https://issues.apache.org/jira/browse/SYSTEMML-631): modules that (inadvertently) source themselves, or more generally, circular references.  Otherwise we will end up in an infinite loop.  In general though, if we fix the handling of namespace conflicts and duplicates, this will be fixed anyway.

gweidner:
Thank you both for the review, comments and feedback.  I've push commits to address the following:
1)  Added tests with non-inlined functions at the inner-most imported script.  The corresponding test scripts were adapted from functions.misc.FunctionInliningTest except functions foo1 and foo2 were refactored into separate script files.  Manually checked under Eclipse debugger that StatementBlock.rIsInlineableFunction returned false (at reject control flow blocks).

2) Added verification of no parser error output to console (similar to what was done in functions.misc.SetWorkingDirTest of [SYSTEMML-611](https://issues.apache.org/jira/browse/SYSTEMML-611).  This ensures test fails when parse issues displayed to console.

Also investigated circular references between separate scripts and reproduced stack overflow issue [SYSTEMML-625](https://issues.apache.org/jira/browse/SYSTEMML-625).  Based on stack trace, it hangs in ParseTreeWalker loop before ParserWrapper.createDMLProgram modified in this PR.  I can next look into both issues [SYSTEMML-631], [SYSTEMML-625] as part of separate PR(s).

mboehm7:
LGTM - thanks @gweidner 

With regard to circular references, I would recommend to use a similar approach as IPA and the parfor optimizer, i.e., a simple function/namespace stack that is carried along while parsing.

dusenberrymw:
Merged.  Thanks, @gweidner!

PR #119, opened 2016-04-15
[SYSTEMML-637] Hybrid Flink Linear regression example

akunft:
This reflects the initial effort to integrate Flink as new back-end for SystemML, as an example we provide a end-to-end implementation for linear regression in hybrid flink mode.

This PR includes the necessary changes to the compiler and instructions as well as the needed functions. Apart from d5e5385, no changes were made to SystemML.


akchinSTC:

mboehm7:
thanks @akunft @FelixNeutatz @fschueler - that's impressive. I'll look into the details in the next days.

mboehm7:
Thanks again @akunft @FelixNeutatz @fschueler @carabolic for contributing this initial Flink backend - this is a really good start. After setting up a recent Flink 1.0.1, I gave it a try on LinearRegDS.dml and a 80GB scenario (10M x 1k, dense w/ binary inputs) in hybrid_flink execution mode. Unfortunately, it was failing during lops construction due to the missing integration of matrix multiply operator selection (in AggBinaryOp). I would recommend to focus on the common distributed operations first: in hybrid execution mode and with moderately large data, we would only run tsmm and mapmm (both compiled from AggBinaryOp) as Flink operations, while all other operations are compiled as pure in-memory CP operations. You have already added all required instructions - that's great, so it would just take some additional wiring in AggBinaryOp and the related lops. 


Furthermore, here are also some additional comments on the actual implementation:

1) APIs and interfaces: Unfortunately, you've put already effort into API extensions. I would recommend to remove the modifications of MLOutput and MLContext because MLOutput is not maintained and MLContext is about to get a major refactoring. Let's use the command line interface for now and extend the APIs once the backend is ready.

2) Isolation of Flink dependencies: We need to be very careful about introducing Flink API dependencies into shared classes in order to ensure that Spark and Flink backends can be used without dependency on the other. For example, there are (1) references to Flink in runtime/controlprogram/Program, and (2) references to the SparkExecutionContext in Flink execution modes. Could you please systematically go over shared classes and double check for isolation?

3) Bufferpool integration: I agree that the duplication of RDDObject and DataSetObject creates unnecessary redundancy in a core place like our bufferpool (CacheableData and ExecutionContext). Let's abstract this to something like DistDataObject with two implementations RDDObject and DataSetObject. Just let me know, I'd be happy to prepare this for you guys.

4) Broadcast memory budget: Currently, the broadcast memory budget is obtained from the SparkExecutionContext. It would be good to automatically obtain this size (e.g., size for user objects) to remove the dependency and because it is really important for operator selection.

5) Checkpointing: I see a CheckpointFLInstruction which always does a pass-through. Let's remove this instruction (until Flink supports distributed caching) and simply do not apply the rewrites for injecting these checkpoints (RewriteInjectSparkLoopCheckpointing, RewriteInjectSparkPReadCheckpointing).  

6) Instructions w/ broadcasts: There seem to be some inconsistencies wrt when and how we broadcast data (e.g., join vs join w/ hint vs withBroadcastSet). I understand that Flink offers optimization capabilities in that regard. However, from a SystemML perspective I would recommend to stick to physical operator semantics (e.g., MAPMM with broadcast, CPMM/RMM with join). If experiments show that Flink actually does better than our operator selection, we could simply compile earlier (for smaller data sizes) CPMM/RMM. Btw, is there a way to pin the broadcast in distributed memory and reuse it across multiple invocations? 

7) Advanced operations: You already added compiler support for advanced operations like weighted squared loss and cumulative aggregates. Let's put these aside for now and get a narrow prototype running well first before we broaden the scope.

8) Tests: It would be good if we could add a few tests for flink mapmm and tsmm to ensure they are producing correct results as this is the precondition for any experiments. Please have a look into FullMatrixMultiplicationTransposeSelfTest and FullDistributedMatrixMultiplicationTest.

9) Exception handling: There are a couple of instances where exceptions are simply put to stderr. Could you please double check that all caught exceptions are rethrown in order to simplify debugging. Also, please replace the assertions for dimensions mismatch checks with exceptions as these checks need to happen even if assertions are disabled (by default).

10) Formatting: Well, we don't have a style guide defined yet but the project-wide convention is to use tabs instead of spaces (e.g., issues in DMLScript). Furthermore, you've also done some code cleanups of formatting existing code. Could we separate these changes out of this PR to simplify the review? Feel free to submit them as an individual PR though.  

 
Finally, as a next step after this PR I would recommend to extend the scenario to LinearRegCG.dml as an iterative algorithm with similar core operations. I would find an investigation of Flink's iteration features very interesting. One direction could be to automatically compile this iteration construct for while loops via rewrites.

fschueler:
Hi Matthias,

thanks a lot for the extensive review! This is a great help for us and we'll get going on the suggestions right away.

shirisht:
This is a great start .. thanks @FelixNeutatz @fschueler @akunft

akunft:
Hi,
we addressed your comments and updated the PR. In addition there are some changes one has to consider in order to run SysML with Flink:

- set HADOOP_HOME environment variable
- set `akka.framesize` to a higher value (otherwise the submitted jobs exceed this limit) e.g. 300000000b in `flink-conf.yaml`
- to submit a job to flink:
	- add yarn dependencies with `hadoop-mapreduce-client-jobclient-2.7.1.jar`
	- explicitly set the main class with `-c org.apache.sysml.api.DMLScript`

We try to fix the problem with the `framesize`, so that we can set it directly in the job and will let you know when this is fixed in Flink.

Looking forward to your comments.

mboehm7:
great - thanks for the updated version and hints. I'll give it a try early next week.

mboehm7:
I gave it a try w/ LinregDS and LinregCG, on a 1+6 node cluster and the 80GB scenario (10M x 1k, dense w/ binary inputs) and it worked fine. Awesome work - thanks @akunft @fschueler @FelixNeutatz @carabolic . Couple of comments:

1) Setup: Even with HADOOP_HOME properly configured there were issues so I disabled the respective code in InfrastructureAnalyzer. This is something, we should probably do anyway. Thanks for pointing this out. Furthermore, I had to reconfigure akka.framesize to 512MB to make these scenarios work.

2) Performance: My biggest concern right now is performance but I'm sure there are simple fixes to that or I simply misconfigured Flink. Let us have a look at the comparison to the MR and Spark backends, where I exclude the time for Spark context creation because Flink runs against an initialized cluster too:

- MR: LinregDS: 147s (2 jobs); LinregCG w/ 6 iterations: 361s (8 jobs) w/ mmchain; 628s (14 jobs) w/o mmchain
- Spark: LinregDS: 71s (3 jobs); LinregCG w/ 6 iterations: 41s (8 jobs) w/ mmchain; 48s (14 jobs) w/o mmchain
- Flink: LinregDS: 212s (3 jobs); LinregCG w/ 6 iterations: 1,047s (14 jobs) w/o mmchain

3) MMchain instruction: Maybe you could add the mmchain instruction for the special case of t(X)%*%(X%*%v) to allow a better understanding what is going on here. This would remove any potential performance influence of broadcasts as the intermediate vector of 80MB does not need to be collected and redistributed via broadcast and it also reduces the number of scans per iteration from 2 to 1. You already have the necessary function implemented. 

4) JDK7 Compatibility: It would be great to ensure backwards compatibility with JDK7, i.e., replace JDK8 methods with their JDK7 counterparts as we try to simultaneously support older deployments and dev environments.

5) FlinkExecutionContext Execute: You already commented wrt a clean handling of this call to pending dataset operations. Couldn't we remove this and simply call execute on write instructions - all other jobs are triggered when the buffer pool collects matrices from producing operations. Right?

akunft:
First of all thank you very much for your detailed comments and sorry for my late reply, but we wanted to clarify some things before answering.

To your comments:

1. We are working on a fix for the akka framesize. The problem here seems to be the collect(), which uses akka to retrieve the data. For large data, this should be done via the data transport instead.

2. As we see it, the performance problems are due to 2 drawback of the current Flink version:

a) Flink currently broadcasts data by sending the data to each slot of the task managers separately instead of sharing it among the slots of a task manager. Therefore, the amount of data sent over the network is increased with increasing number of slots per task managers. 

b) As Flink does not support caching of data in the moment, we have to re-read from HDFS every time data sources are accessed.

3. We will try this, as it would at least omit the problem stated in 2 a. Thank you for pointing that out.

4. Good point, we fixed that.

5. Yes, we think this is possible and implement that.

We try to solve the problems mentioned in 1 and 2, but especially 2b. will take some time and needs to be accepted by the Flink community. What do you think is the best way to go in the meanwhile with regards to the integration in SystemML?

mboehm7:
Thanks for sharing these thoughts - that sounds good to me. In general, I'd say that, even without caching, the Flink integration would still be a valuable (experimental) feature, because it enables the SystemML community to play around with it and explore advanced features like iteration constructs. So, if we manage to bring the performance to a level that is consistently better than the MR backend (because there is no reason why it should not), I'll endorse it. If later Flink also adds support for distributed caching, even better.

Regarding the broadcast issue, I was actually thinking that the lookup map of matrix blocks might implicitly replicate blocks, if (and only if) broadcasts are internally shared in serialized form. In that case, we could have addressed it with a lazy broadcast partition read (as we do in the MR backend, because a single task seldom reads all broadcast blocks). However, if broadcast are indeed sent to each slot, it is probably a good idea to fix this for the general case.  

So let's have a look into the mmchain instruction and re-evaluate these scenarios once we have it. This should help us understand these characteristics in detail but it will also be useful in general because this operation is the core of major algorithms like LinregCG, GLM, Logreg, and HITS.

akunft:
Ok, sounds good. We are on it

dusenberrymw:
Can we add a "WIP" tag to the title of this PR?

FelixNeutatz:
Yes, we are still working on this. At the moment we are working on the Flink side to make it more performant.

deroneriksson:
@akunft @mboehm7 @fschueler @FelixNeutatz @dusenberrymw 
Hi, I just wanted to check in and see if this PR is still under active development.

akchinSTC:

akunft:
@deroneriksson: We are currently stuck due to some limitations in Flink that prevent us from reaching acceptable runtimes. Thus, I would suggest to close this PR in the meanwhile. Once we have reasonable support for caching and broadcast variables in Flink, we can think about the integration again.

What do you think?

mboehm7:
I'd prefer to keep this PR open as a reminder.

PR #122, opened 2016-04-20
[SYSTEMML]Fix DMLAppMasterStatusReporter.java thread cancel bug.

GrapeBaBa:
# Summary
This PR created for fixing multiple thread bug in DMLAppMasterStatusReporter.java.
1.Use a volatile variable for flag.
2.Invoke interrupt method to avoid block method such as Thread.sleep invocation.


mboehm7:
thanks @GrapeBaBa - since you explicitly interrupt the thread, could you please also add the related  handling of InterruptedException (e.g., something like LOG.warn("Status reporter interrupted with stop="+_stop)

GrapeBaBa:
@mboehm7 Done.

mboehm7:
LGTM

deroneriksson:
Thank you @GrapeBaBa . I will merge.

PR #123, opened 2016-04-21
[SYSTEMML-560] Frame converter between TextCell and Binary block

asurve:


mboehm7:
that's a good start @asurve - couple of comments:

1) Block merge: It would be good to implement the frame block merge via copy, not cell access for performance (intrinsics) and robustness (with regards to different schemas).

2) Convert FrameBlockToIJVLines: You don't need a lazy iterator here as the input is bounded to one block. Please simplify by simply adding all cells to to an ArrayList - this might even solve you occasional serialization issues.

3) FrameConverterUtils: Please order the actual converter methods to the top (and internal functions to the bottom) and make sure that overloaded methods call each other for consitency (e.g., textCellToBinaryBlock and textCellToBinaryBlockLongIndex).

4) FrameBinaryBlockToTextCellConverter: This converter does not implement setBlockSize - either implement or remove it.

5) Tests: Given 3, you don't need separate tests for TXTCELL2BIN and TXTCELL2BIN_LONGIDX - please remove one of them.

6) Code style: Please use x++ instead of ++x for generic increments where you do not rely on a pre-increment. Also please simplify boolean assignments e.g., 'bNextCellExists = (nextRow != _block.getNumRows())'. Finally, remove unnecessary casts (e.g., in fromIndexedFrameBlock).

mboehm7:
LGTM - Just one minor fix before you push: please remove the unnecessary list-arrary-list conversion in ConvertFrameBlockToIJVLines (Arrays.asList(cells.toArray())). Also for the future, please make sure not to use scala.actors.threadpool.Arrays but java.util.Arrays (I know it gets auto-resolved to the scala version since SystemML has been made a mixed scala/java project).

PR #124, opened 2016-04-21
Update parse method parameters from HashMap to Map

deroneriksson:
Referencing Map interface rather than HashMap concrete class allows
the methods to not be tied to a specific implementation of Map.

Updated for AParserWrapper, CommonSyntacticValidator, DMLParserWrapper, DmlSyntacticValidator, PyDMLParserWrapper, and PydmlSyntacticValidator, relating to the parse calls, so that a client calling the parser code can pass a Map reference rather than a HashMap reference. Also updated DMLScript, MLContext, MLOutput, and jmlc Connection to use Map references rather than HashMaps.


mboehm7:
LGTM - thanks @deroneriksson

PR #127, opened 2016-04-22
[SYSTEMML-645] Change Naive-bayes accuracy result to scalar

Wenpei:
I change this to scalar since it means nothing to export a string to output file.
Please help to review


shirisht:
I assume you want to export only the accuracy number so that the exported file can be easily consumed by other softwares. 

LGTM.

deroneriksson:
@Wenpei Thank you. I will merge.

PR #128, opened 2016-04-22
Troubleshoot document for java heap size error

ethanyxu:
Added a troubleshoot section about OOM errors in reduce phase.

### DIFF
--- a/docs/troubleshooting-guide.md
+++ b/docs/troubleshooting-guide.md
@@ -53,7 +53,14 @@ Maven (`mvn clean package -P distribution`).
 ## OutOfMemoryError in Hadoop Reduce Phase 
 In Hadoop MapReduce, outputs from mapper nodes are copied to reducer nodes and then sorted (known as the *shuffle* phase) before being consumed by reducers. The shuffle phase utilizes several buffers that share memory space with other MapReduce tasks, which will throw an `OutOfMemoryError` if the shuffle buffers take too much space: 
 
-    java.lang.OutOfMemoryError: Java heap space
+    Error: java.lang.OutOfMemoryError: Java heap space
+        at org.apache.hadoop.mapred.IFile$Reader.readNextBlock(IFile.java:357)
+        at org.apache.hadoop.mapred.IFile$Reader.next(IFile.java:419)
+        at org.apache.hadoop.mapred.Merger$Segment.next(Merger.java:238)
+        at org.apache.hadoop.mapred.Merger$MergeQueue.adjustPriorityQueue(Merger.java:348)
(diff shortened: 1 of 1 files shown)


shirisht:
Ethan, thank you for preparing this section under Troubleshooting guide.
I am, however, not sure about the need for this -- especially because these are Hadoop settings that are relevant for any application running on Hadoop. We are also not providing any recommended settings specific to SystemML -- for example, we recommend three times the HDFS block size for `mapreduce.task.io.sort.mb` based on the way SystemML operates. In contrast, we are not providing any SystemML-specific recommended settings here.

I would like to see what others think about it.

deroneriksson:
I am in favor of accepting this PR. @ethanyxu is running SystemML on a large dataset on a large Hadoop cluster. Another user in a similar situation may find this information helpful.

deroneriksson:
If accepted, my one request would be to update the absolute link to `http://apache.github.io/incubator-systemml/hadoop-batch-mode.html` to a relative link `[Invoking SystemML in Hadoop Batch Mode](hadoop-batch-mode.html)`

dusenberrymw:
I too am in favor of adding this, as it represents a simple solution to a particular Hadoop issue that has presented itself in the course of running SystemML on a real problem.

@ethanyxu Thanks for continuing to bring to our attention the various issues and scenarios that you have encountered.

shirisht:
Couple of comments:

1. Can you include at least first few lines from the stack trace insteadof just `java.lang.OutOfMemoryError`, so that the cause of such an error is more explicit.
2. Change the wording `per-task basis` -- the word 'task' can be misled to mean individual map reduce tasks.
3. How do the Hadoop configuration parameters specified in `SystemML-config.xml` are propagated to individual MR jobs that are created?

ethanyxu:
Oops I missed the discussion. Thanks all for the suggestions and comments. I understand these are general hadoop settings, but in my use case changing the global configuration was not an option as the cluster was shared with other people. Modifying the configurations via `SystemML-config.xml` (as suggested by @mboehm7) worked great. 

@shirisht I ran into this problem twice with different OOM errors. One traceback is attached. The wording `per-task basis` is changed to `per SystemML-task basis`. As for how the configs were propagated, it looks from the server side logs that they were just passed to all MR jobs generated by SystemML. I am not able to elaborate in more technical depth. 

@deroneriksson The link is changed to a relative link as you suggested. 

Thank you @dusenberrymw I'm glad to share my experience:)

mboehm7:
I'm also in favor of extending our trouble shooting guide as we go, even if these issues are not always related to SystemML core. Thanks @ethanyxu

deroneriksson:
LGTM. Thank you @ethanyxu

PR #129, opened 2016-04-27
[SYSTEMML-651] Fix scalar input support for JMLC

deroneriksson:


mboehm7:
thanks @deroneriksson for debugging this issue. However, I would strongly recommend not to create these additional special cases in DAG (which would lead to different conditions firing for normal transient reads and transient reads introduced by jmlc/mlcontext). Couldn't we instead modify RewriteRemovePersistentReadWrite and clear the filename in case of scalars?

deroneriksson:
@mboehm7 In RewriteRemovePersistentReadWrite, I've tried things such as the following without success:
```
case PERSISTENTREAD:
	if( _inputs.contains(dop.getName()) ) {
		dop.setDataOpType(DataOpTypes.TRANSIENTREAD);
		if (hop.getDataType() == DataType.SCALAR) {
			dop.setFileName(null);
			dop.getInput().set(dop.getParameterIndex("iofilename"), new LiteralOp(null));
		}
	}
```

Also, such things as
```dop.getInput().set(dop.getParameterIndex("iofilename"), new LiteralOp(Double.NaN));```
and
```dop.getInput().set(dop.getParameterIndex("iofilename"), new LiteralOp(""));```

Do you have any further advice?

**Comment by mboehm7** [comment]
how about removing it (i.e., not overwriting) from both _paramIndexMap and _inputs?

**Comment by deroneriksson** [comment]
@mboehm7 Let me know if the current implementation looks OK. Updated to:
```
if (hop.getDataType() == DataType.SCALAR) {
	int iofilenameIndex = dop.getParameterIndex("iofilename");
	dop.getInput().remove(iofilenameIndex);
	HashMap<String, Integer> parameterIndexMap = dop.getParameterIndexMap();
	parameterIndexMap.remove("iofilename");
	parameterIndexMap.put("format", iofilenameIndex);
}
```

mboehm7:
thanks, in general yes, although not as clean as I hoped for when I made the initial comment. 

I leave it up to you to go either with your original version or this one right here. If we use the latest version, I would recommend to move the map modifications into DataOp (together with a slight generalization for "re-indexing" arbitrary parameters, not just format). Also, you might want to change one of the existing jmlc tests to pass a scalar through this codepath (instead of the currently used input arguments).

deroneriksson:
@mboehm7 Created a removeInput method in DataOp which is called from RewriteRemovePersistentReadWrite. Also created some basic scalar input tests for JMLC.

deroneriksson:
All tests passed [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/301/).

mboehm7:
LGTM - thanks @deroneriksson

deroneriksson:
@mboehm7 Thank you for reviewing and the help.

PR #130, opened 2016-04-27
[SYSTEMML-648] Deprecate castAsScalar, update parse warning handling

deroneriksson:
Add castAsScalar deprecation warnings.
Add Expression javadocs, rename cast enum values.
Remove unused enums from Expression.
Display parse warnings if they occur without parse errors.
Don't chain ParseException in JMLC Connection.

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -32,6 +32,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Scanner;
 
@@ -62,6 +63,8 @@ import org.apache.sysml.parser.DMLProgram;
 import org.apache.sysml.parser.DMLTranslator;
 import org.apache.sysml.parser.LanguageException;
 import org.apache.sysml.parser.ParseException;
+import org.apache.sysml.parser.common.CustomErrorListener;
+import org.apache.sysml.parser.common.CustomErrorListener.ParseIssue;
 import org.apache.sysml.runtime.DMLRuntimeException;
 import org.apache.sysml.runtime.DMLScriptException;
 import org.apache.sysml.runtime.controlprogram.Program;
@@ -585,6 +588,11 @@ public class DMLScript
 		Statistics.startCompileTimer();
 		AParserWrapper parser = AParserWrapper.createParser(parsePyDML);
 		DMLProgram prog = parser.parse(DML_FILE_PATH_ANTLR_PARSER, dmlScriptStr, argVals);
+		
+		if (parser.isAtLeastOneWarning()) {
+			List<ParseIssue> parseIssues = parser.getParseIssues();
+			System.out.println(CustomErrorListener.generateParseIssuesMessage(dmlScriptStr, parseIssues));
+		}
 		
 		//Step 4: construct HOP DAGs (incl LVA and validate)
 		DMLTranslator dmlt = new DMLTranslator(prog);
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -32,7 +32,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Scanner;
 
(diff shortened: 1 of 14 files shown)


dusenberrymw:
In a follow-up PR (or this one) we should also replace all uses of `castAsScalar(...)` with `as.scalar(...)` in any of our DML/PyDML files.

deroneriksson:
Sounds good. Already created https://issues.apache.org/jira/browse/SYSTEMML-647 to address this. Should just be a big find/replace.

mboehm7:
thanks @deroneriksson - couple of comments

1) Please move the warnings after the parse call into the parse method (potentially make parse a wrapper that calls some internalParse of the subclasses). It would be very usefull to keep
the parser invocation a single call that handles everything internally as it is called multiple times.
2) Always use logging for these warnings. The only write to stdout in SystemML should come from print. In the past we had severe issues where SystemML (running in a fenced process)
eventually blocked because the stdout queue filled up. Btw, that's why we have a flag to disable any prints (_suppressPrint2Stdout).
3) What is the rational behind renaming the casting operations?
4) Conditional validate: could we please keep the conditional flag for raiseValidateError to make clear
when warnings instead of errors are raised (if validate of conditional control flow where we don't know if this is ever gonna be executed, which simplifies for example the handling of conditional inputs; otherwise we would need to always provide valid filenames there).

deroneriksson:
1) I will go ahead and remove the parse warnings to standard output in DMLScript, the current MLContext, and the JMLC Connection if that's how you think it should be done. However, note that if log4j is not set up, the user will fail to see these important messages. In the new MLContext, this information shall potentially be handled differently so that Notebook users can see this information on the Notebook page rather than in a console, since it's quite possible a Notebook user will never see anything going to a console (such as log warnings).
2) I don't think 1 potential output to standard output will clog the stdout queue. From a usability standpoint, I would be sad if there was a warning regarding my DML script and SystemML didn't tell me about it in a clear manner. This information is still being logged as log.warn()
3) Since castAsScalar was a mistake, I assumed the CAST_AS_* was a mistake. I will change the AS_* names back to the legacy names.
4) I'm sorry, I don't understand. For raiseValidateError, the flag is still there but renamed from 'conditional' to 'doWarning', since without looking at the implementation of the method, I previously had no idea what 'conditional' means. 'doWarning' is clearer WRT what this flag is doing, and the new version has an accompanying javadoc description (@param doWarning if {@code true}, display log warning message. Otherwise, the message (and optional error code) will be thrown as a LanguageException)

dusenberrymw:
I think updating the internal casting operations to `AS_*` would provide a cleaner mapping from the language representation (`as.*`) to the internal compiler representation (`AS_*`).

mboehm7:
ad 3) well, the issue with castAsScalar was that we did not follow R's convention for these kinds of casts - internally, however, the cast prefix was very appropriate. Anyway, I don't have a strong opinion here - if the majority of people preferes the AS prefix that's fine too.
ad 4) I understand why you made that change - the only reason why I would prefer to call the flag 'conditional' is that it represents a slightly different contract: with a flag like 'doWarning' the implementation has no choice but to raise a warning instead of a exception. With 'conditional' callers just report the context and it's up to the implementation to raise an exception or warning.

deroneriksson:
3) I'll change back to CAST_AS_*, but if more developers like @dusenberrymw and myself would prefer a closer match to the DML keywords, we can do a JIRA to change at all levels (hops (Hop.OpOp1, etc), lops (such as UnaryCP.OperationTypes, etc), ...).
4) I'll change back to 'conditional' but keep the javadoc comment so new developers don't have to look at the method implementation to see what's going on

mboehm7:
sounds good - I appreciate it @deroneriksson. 

Btw, with regard to logging, we should think about creating a lean wrapper where all logging goes through which would allow us to quickly switch between logging and stdout (potentially depending on the api from where SystemML is called.)

deroneriksson:
@mboehm7 Made the updates.

All tests passed [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/302/).

mboehm7:
sorry I didn't see this before, but could you please move the warning from the parsers (PydmlSyntacticValidator/DmlSyntacticValidator) into the language level (BuiltinFunctionExpression) in order to consolidate all validate-related code as the parsers are meant for pure syntactic translation and validation, not semantic like functions names and things like that. Thanks.

deroneriksson:
Is there a flag available in BuiltinFunctionExpression that distinguishes between DML and PYDML, so that I can output a language-specific deprecation warning (switch to as.scalar() for DML, switch to scalar() for PYDML)?

mboehm7:
That's a good point. I missed the different syntax - please leave it for now as we want to keep the language level independent of the parsers.

LGTM - thanks @deroneriksson

deroneriksson:
@mboehm7 Thank you. I will merge.

PR #131, opened 2016-04-29
[SYSTEMML-540] [SYSTEMML-445] Initial implementation of conv2d/maxpooling builtin functions and GPU backend

niketanpansare:
2. Added initial version of GPU backend.
3. The documentation is available in docs/devdocs folder.

@mboehm7 @shirisht @dusenberrymw @bertholdreinwald Please review the PR.

### DIFF
--- a/scripts/perftest/runAllBinomial.sh
+++ b/scripts/perftest/runAllBinomial.sh
@@ -39,11 +39,12 @@ echo $2"-- Generating binomial data: " >> times.txt;
 ./genBinomialData.sh $1 $2 &>> logs/genBinomialData.out
 
 # run all classifiers with binomial labels on all datasets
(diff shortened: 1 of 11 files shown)


mboehm7:
thanks @niketanpansare - you certainly put a lot of effort in here. Could we please separate those PRs though? It would be great to get the new dml builtin functions in first, with tests but without new dependencies and without reuse of allocated dense arrays and other custom extensions (that do not generalize to all scenarios yet). We could discuss the other extensions as soon as we're comfortable with the baseline builtin functions.

dusenberrymw:
Agreed, splitting this up would definitely facilitate easier review and discussion. :)

niketanpansare:
Sure, will do.

PR #133, opened 2016-05-03
[SYSTEMML-631] Improve handling namespace conflicts

gweidner:
This patch with new tests changes the way namespaces are tracked internally (i.e., uses qualified script file path) to prevent collisions. Previously, functions would be lost if the same namespace was specified when importing different scripts which resulted in error being reported as function undefined in namespace.  With this change, 'source(file1) as namespace1' in one script does not interfere with 'source(file2) as namespace1' in another script.  Note an error is reported at source statement if same namespace specified for different scripts within the same file only.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/common/CommonSyntacticValidator.java
+++ b/src/main/java/org/apache/sysml/parser/common/CommonSyntacticValidator.java
@@ -69,6 +69,8 @@ public abstract class CommonSyntacticValidator {
 	protected static ThreadLocal<HashMap<String, String>> _scripts = new ThreadLocal<HashMap<String, String>>() {
 		@Override protected HashMap<String, String> initialValue() { return new HashMap<String, String>(); }
 	};
+	// mapping of namespaces to full paths as defined only from source statements in this script (i.e., currentFile)
+	protected HashMap<String, String> sources;
 	
 	public static void init() {
 		_scripts.get().clear();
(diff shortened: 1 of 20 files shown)


gweidner:
For testFunctionIndirectConflict (Functions9.dml), observed revised namespace value in runtime.instructions.cp.FunctionCallCPInstruction:

Instruction: CP°extfunct°.\./src/test/scripts/functions/misc/FunctionsJ3.dml°foo3°1°1°_mVar1·MATRIX·DOUBLE°V
Instruction: CP°extfunct°.\./src/test/scripts/functions/misc/FunctionsJ2.dml°foo2°1°1°B·MATRIX·DOUBLE°V
Instruction: CP°extfunct°.\./src/test/scripts/functions/misc/FunctionsJ1.dml°foo1°1°1°X·MATRIX·DOUBLE°Y

Also considered generating unique namespace via appending runtime.controlprogram.parfor.util.IDSequence.

All tests passed on local Windows development machine and at [OnDemand build 154](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/154/).

dusenberrymw:
Awesome, Glenn!  I'l grab this locally and try it out in a little bit.

dusenberrymw:
@gweidner So sorry I didn't get to this -- sidetracked on another issue.  I will definitely review and merge tomorrow.  Really excited to get this in, as I hit the exact issue today, and still managed to take a moment before realizing what was going on.

dusenberrymw:
Alright, I tested this locally with some real use-cases I had run into, and it works great!  I like the implementation as well -- it's simple and the mapping of namespaces to direct file paths removes confusion.  Also, I noticed that this seems to have also fixed another issue in which the namespaces imported within a file would "leak" out and be available to any other files.  Now, a script only has access to the namespaces that it itself imported, which is the correct behavior.  Awesome to see that patched as well, as I think that could have been some bad technical debt.

Awesome work, @gweidner! 👍

dusenberrymw:
**LGTM**

PR #134, opened 2016-05-03
[SYSTEMML-540] Initial implementation of conv2d/maxpooling builtin functions

niketanpansare:
2. The documentation is available in docs/devdocs folder.

To test this built-in functions, I tested the results against Stanford's Python scripts for Deep Learning and also compared the results with CuDNN. To make the test suite more generic, we have following options:
1. Compare the above built-in functions against existing R package. I could not find one that fit our need. Please suggest if you know any R package that allows one to compute the above functions.
2. Add Python dependency for our dev environment and compare against existing DL packages. For example: TensorFlow/Stanford's Python scripts.
3. Write the logic of conv2d into R and compare against it. For this approach, I tried to use https://github.com/dusenberrymw/systemml-nn/blob/master/nn/layers/conv.dml but ran into some errors which will get ironed out soon. The benefit of this approach is that we donot add any additional dependency on our test suite, but the drawback is that we are comparing the semantics against our alternate implementation and not against external packages.


niketanpansare:
@dusenberrymw @mboehm7 @bertholdreinwald Can you please comment on (and then review) this PR ?

src/test/scripts/functions/tensor/Conv2DTest.R
```diff
@@ -0,0 +1,99 @@
...
+w=matrix(seq(1, numFilters*numChannels*filterSize*filterSize), numFilters, numChannels*filterSize*filterSize)
+
+
+pad_image <- function(img, Hin, Win, padh, padw){
+  C = nrow(img)
+  img_padded = matrix(0, C, (Hin+2*padh)*(Win+2*padw))  # zeros
+  for (c in 1:C) {
+    img_slice = matrix(img[c,], Hin, Win)  # depth slice C reshaped
+    img_padded_slice = matrix(0, Hin+2*padh, Win+2*padw)
+    img_padded_slice[padh+1:padh+Hin, padw+1:padw+Win] = img_slice
```
dusenberrymw: It looks like R can't handle the math within this slice statement without the use of parentheses.  Change this line to `img_padded_slice[(padh+1):(padh+Hin), (padw+1):(padw+Win)] = img_slice` and it should be fine.

src/test/scripts/functions/tensor/Conv2DTest.R
```diff
@@ -0,0 +1,99 @@
...
+  img_cols = matrix(0, C*Hf*Wf, Hout*Wout)  # zeros
+  for (hout in 1:Hout) {  # all output rows
+    hin = (hout-1) * strideh + 1
+    for (wout in 1:Wout) {  # all output columns
+      win = (wout-1) * stridew + 1
+      # Extract a local patch of the input image corresponding spatially to the filter sizes.
+      img_patch = matrix(0, C, Hf*Wf)  # zeros
+      for (c in 1:C) {  # all channels
+        img_slice = matrix(img[c,], Hin, Win)  # reshape
+        img_patch[c,] = matrix(img_slice[hin:hin+Hf-1, win:win+Wf-1], 1, Hf*Wf)
```
dusenberrymw: Interestingly, R doesn't seem to have a problem with the math in this slice statement though, although in this case it is only present in the upper bound.  Might be safer to wrap the upper bounds in parentheses anyway.

dusenberrymw: Update: The parentheses are required here as well for R.  Otherwise, the indices will just all be sifted by `Hf-1`.

```
> x = matrix(seq(1, 10), 2, 5, byrow=TRUE)
>
> x
     [,1] [,2] [,3] [,4] [,5]
[1,]    1    2    3    4    5
[2,]    6    7    8    9   10
> x[1,1:4]
[1] 1 2 3 4
> x[1,1:4+1]
[1] 2 3 4 5
> x[1,1:(4+1)]
[1] 1 2 3 4 5
```

Note, this is *not* the case with DML.

mboehm7: just to clarify that - (x:y) is just a shorthand for seq(x,y) and accordingly the second example produces seq(1,4)+1. In systemml, we never materialize these expressions and accordingly evaluate the expressions differently - we should at least throw a warning here

dusenberrymw: Interesting.  If we do throw a warning, let's only do it for DML (not PyDML), as the current behavior here mimics that in Python.

niketanpansare:
@mboehm7 @dusenberrymw @bertholdreinwald We have following options here:
1. Merge this PR with existing test-cases and figure out why the results of R code is not matching with CuDNN's result. Then update the corresponding test-cases with R checks accordingly.
2. Merge this PR with existing test-cases and ask for vote in the mailing list for adding Python dependency in our build environment. Once the vote is approved, then update the corresponding test-cases with Python checks accordingly.
3. Wait until we debug our R scripts (which we will have to write for conv2d_backward*, maxpooling, maxpooling_backward*), add R testcases and then merge this PR.
4. Wait until the vote for Python packages is approved and then add Python testcases into this PR and merge this PR.

I think the second option is the best because it allows us to keep making progress on performance improvement. Also, comparing our code with an external package is the right way to test our built-in functions.

mboehm7:
before we discuss merging this PR let's get the code and tests in shape first - I'll provide comments soon.

mboehm7:
thanks @niketanpansare for the initial prototype - here are a couple of comments to help improve the implementation:

1) Hop: Why do we need a new hop for these kinds of operations - isn't the semantic of various operations covered by ReorgOp and ParamBuiltinOp?

2) Pooling parameters: It would be good to support the general case of variables here instead of assuming that every parameter is given as a constant.

3) Size propagation: The current size propagation needs a rework. First, it always assumes entire 
dense  outputs which incorrectly propagates into subsequent operations. Second, there is no overwrite with unknown values which renders it incorrect in special cases of dynamic recompilation.

4) Validation: Please move any semantic validation out of the parser (DmlSyntacticValidator) and into the actual validate. 

5) Buffer pool: CacheableData is completely generic and should not have any references to MatrixBlock or FrameBlock, please move the reuse handling into MatrixObject or ExecutionContext.

6) Reuse allocated arrays: The current reuse of allocated arrays has two major problems: (1) it is not thread-safe, and (2) it creates incorrect results for operations that assume
zero-initialized outputs and accordingly don't overwrite each cell. Please add TODOs to fix that. 

7) Multi-threading - integration: Please integrate the decision on multi-threading into hops, lops, and runtime instructions. Note that the Hop should implement MultithreadedHop and requires treatment in the parfor optimizer - otherwise  the instruction would not be aware of the surrounding parfor parallelism.

8) Multi-threading - error handling: Currently, there is no error handling on multi-threaded operations, which means issues would lead to silent errors and hence potentially incorrect results. Please, explicitly call a get on each future. 

9) Block operations: Why are the individual operations directly implemented in the instructions? Please move to MatrixBlock and potentially create a LibMatrixDNN to groups these operations together.

10) Sparse operations: Currently, sparse operations would suffer an unnecessarily bad asympotic behavior as we access each cell, and each cell requires a binary search to find the value. It would be great to support both dense and sparse if these operations might be used by other usecases than pure dense inputs.

11) Distributed operations: how hard would it be to add the distributed operations for Spark/MR too? If infeasible, please add least add some comments into the reprective Hop as to why
these operations are not added.

12) Statistics: Please only maintain statistics whenever you -stats is enabled as the System.nanoTime and atomic counters can be an actual bottleneck. Also please, remove them from
MatrixBlock to ensure they are not unnecessarily called on distributed operations. Also the new counters need to be reset on Statistics.reset.

13) Tests: I would really appreciate to have the R scripts for tests as the unrolled expected results are really had to maintain and extend. In the worst case, let use at least provide these fixed expected results as CSV files. 

14) Codestyle: First, please move all class members to the top of the class. Second, please try to merge predicates to reduce levels of indentation (e.g., DataExpression.getMatrixParametersForTensorFn). Third, please remove the redundant opcode checks from the CP instruction. Fourth, remove commented stdout debug output. Fifth, don't use all capital names for non-constants (NON_ZEROED_DOUBLE_ARR, which is a map).

dusenberrymw:
I'll look into the issues with the R port, and I'm going to plug the new forward and backward functions into the SystemML-NN library to test them with the gradient-check framework there.

niketanpansare:
Thanks @mboehm7  for your comments :)

1. I think having a separate Op for class of tensor-related functions/instructions (such as im2col, reshape_col, rotate_180, maxpooling, maxpooling_backward, direct_conv2d, ...) is extremely useful. I don't think all of these operations qualify as Reorg as some of them do more than simple reorg, for example: col2im. Also, these operations are fairly specific to tensor to be grouped as seperate Op rather than have them under ParameterizedOp. Also, just like AggregateBinaryOp, this Op can qualify for fusing/meta-optimization. For example: http://benanne.github.io/2014/12/09/theano-metaopt.html
2. Pooling parameters: Can you please point me to the part where I assume every parameter is a constant ? If so, it is definitely a bug and will correct it.
3. Yes, the output of ConvolutionCP instructions is always in dense format. We can revisit this assumption later. I can work on fixing overwriting of unknown values.
4. There is no semantic validation in DmlSyntacticValidator. Since SystemML doesnot support list, all we do in DmlSyntacticValidator is unfold the list.
5. Done.
6. (1) Converted  non_zeroed_double_arr to ConcurrentHashMap for thread-safety of put and remove. (2) If the caller calls the method "public void allocateDenseBlock(boolean clearNNZ, boolean zeroOut)" incorrectly (i.e. allocateDenseBlock(..., false) when the operation requires allocateDenseBlock(..., true)) then and then only it will produce incorrect results. 
7/8. Will add the multi-threading in the next commit.
9. Will create LibMatrixDNN in the next commit. 
10/11. I think we should optimize "Sparse operations" and "Distributed operations" in subsequent PR once we are happy with the performance of single node Lenet script on dense MNIST dataset. I have added respective TODO comments. 
12. Done.
13. I can create a fixed expected results as CSV files in this PR and Mike can resolve issues with R port in a separate PR.
14. Done.

mboehm7:
ad 2) ConvolutionOp->parseInput->extractValue throws an exception if its not a LiteralOp. Could we simply return -1 instead to make it robust enough to tolerate variables during compilation?

ad 3) The issue is that these "estimates" are propagated to dim1/dim2/nnz during refreshSizeInformation, which will be used by the parent operations. Note that for example a sum(X!=0) is just rewritten to take the nnz. Similarly the dims need to be correct, otherwise certain rewrites would be invalid too. If you cannot exactly infer these stats, please use the MemoTable and worst-case estimates.

ad 4) It would be great if we could move the handling of default parameters, and valid parameter names (in exitFunctionCallAssignmentStatement) into our actual validate.

niketanpansare:
(ad 2/3) Done.

(ad 4) All the parameters are required in conv2d/pooling functions. So, we do not any validation except for checking matrix inputs. In exitFunctionCallAssignmentStatement, we actually create new expressions from expression list. I am not sure whether they should be pushed into validate as we would skip LV analysis of these new expressions.

dusenberrymw:
Update: The issue with the R tests is indeed due to the port from DML -> R involving different R semantics.  

By default, R works with matrices in a column-wise manner, i.e., shaping a vector into a matrix will do so in a column wise manner:
```r
> x = matrix(1:8, 2, 4)
> x
     [,1] [,2] [,3] [,4]
[1,]    1    3    5    7
[2,]    2    4    6    8
```
There is a `byrow` boolean flag to the `matrix(...)` that can be set to `TRUE` to enable a row-wise treatment:
```r
> x = matrix(1:8, 2, 4, byrow=TRUE)
> x
     [,1] [,2] [,3] [,4]
[1,]    1    2    3    4
[2,]    5    6    7    8
```

**However**, this row-wise treatment is *not* applied when unrolling a matrix back into a vector, regardless of the flag:
```r
> matrix(x, 1, 8)
     [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8]
[1,]    1    5    2    6    3    7    4    8
> matrix(x, 1, 8, byrow=TRUE)
     [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8]
[1,]    1    5    2    6    3    7    4    8
```

Therefore, in R, we have to transpose the matrix first, and the unroll into a vector:
```r
> matrix(t(x), 1, 8)
     [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8]
```

I've fixed the R script locally, and now it passes the tests and correctly outputs the same results as both SystemML-NN and the new built-in functions.  I'm going to do a few more checks on it, and then I'll either attach it here or make a PR.

niketanpansare:
Thanks @dusenberrymw ... I am fine with either (make a new PR or attach it to this PR).

niketanpansare:
@mboehm7 All the issues except following have been resolved:
1. Tests: I can add expected results in CSV or wait for Mike to add R script in a subsequent PR.
2. ad 4: Addressed in my comment 3 hours back.

Please let me know if there is any other suggestions. Thanks :)

dusenberrymw:
@niketanpansare I've [opened a PR](https://github.com/niketanpansare/incubator-systemml/pull/1) against your `deep_learning` branch to fix the ported R script.  Once you merge into `deep_learning`, this PR will automatically update as well.

src/test/scripts/functions/jmlc/scalar-input-string.dml
```diff
@@ -0,0 +1,24 @@
+#-------------------------------------------------------------
```
dusenberrymw: I believe this file has been pulled in from a previous commit by @deroneriksson due to merging.

src/test/scripts/functions/jmlc/scalar-input.dml
```diff
@@ -0,0 +1,25 @@
+#-------------------------------------------------------------
```
dusenberrymw: I believe this file has been pulled in from a previous commit by @deroneriksson due to merging as well.

samples/jupyter-notebooks/Flight_Delay_Demo.ipynb
```diff
@@ -1 +1 @@
...rid = new ParamGridBuilder().addGrid(lr.regParam, Array(0.1, 1e-3, 1e-6)).build()\ncrossval.setEstimatorParamMaps(paramGrid)\ncrossval.setNumFolds(2) // Setting k = 2\nval cvmodel = crossval.fit(trainDataset)", "outputs": [], "metadata": {"collapsed": true, "trusted": true}}, {"source": "### Evaluate the cross-validated model", "cell_type": "markdown", "metadata": {}}, {"execution_count": null, "cell_type": "code", "source": "val cvpredictions = cvmodel.transform(testDataset.withColumnRenamed(\"label\", \"OriginalLabel\"))\ncvpredictions.registerTempTable(\"cvpredictions\")\nsqlContext.sql(\"SELECT sqrt(avg(pow(OriginalLabel - label, 2.0))) FROM cvpredictions\").show", "outputs": [], "metadata": {"collapsed": true, "trusted": true}}, {"source": "## Homework ;)\n\nRead http://apache.github.io/incubator-systemml/algorithms-classification.html#multinomial-logistic-regression and perform cross validation on other hyperparameters: for example: icpt, tol, maxOuterIter, maxInnerIter", "cell_type": "markdown", "metadata": {}}], "nbformat": 4, "metadata": {"kernelspec": {"display_name": "Scala 2.10.4 (Spark 1.5.2)", "name": "spark", "language": "scala"}, "language_info": {"name": "scala"}}}
```
dusenberrymw: I believe this file has been pulled in from a previous commit due to merging as well.

docs/troubleshooting-guide.md
```diff
@@ -50,3 +50,47 @@ from `provided` to `compile`.
 SystemML can then be rebuilt with the `commons-math3` dependency using
 Maven (`mvn clean package -P distribution`).
 
+## OutOfMemoryError in Hadoop Reduce Phase
```
dusenberrymw: I believe this file has been pulled in from a previous commit due to merging as well.

src/main/java/org/apache/sysml/hops/ConvolutionOp.java
```diff
@@ -0,0 +1,478 @@
...
+			{	
+				if( et == ExecType.CP )
+				{
+					setLops(constructConvolutionLops(et, inputs));
+					break;
+				}
+				else {
+					// TODO: Add support for SPARK/MR backends once we are happy with the performance of
+					// single node Lenet script. 
+					throw new HopsException("Unimplemented ConvolutionOp for execution type: " + et.name());
```
dusenberrymw: How much work will it be to add a very light support for Spark so that these functions can be called from, say, `MLContext` in a PySpark notebook?  Particularly, it would just be nice for the code path to be active (most likely to just be optimized to run in CP mode), just to enable quick testing and iterations.

src/main/java/org/apache/sysml/hops/rewrite/RewriteRemovePersistentReadWrite.java
```diff
@@ -27,8 +27,9 @@
 import org.apache.sysml.hops.DataOp;
 import org.apache.sysml.hops.Hop;
 import org.apache.sysml.hops.Hop.DataOpTypes;
-import org.apache.sysml.hops.HopsException;
```
dusenberrymw: I believe this file has been pulled in from a previous commit due to merging as well.

src/main/java/org/apache/sysml/parser/AParserWrapper.java
```diff
@@ -23,13 +23,15 @@
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.List;
```
dusenberrymw: I believe this file has been pulled in from a previous commit due to merging as well.

src/main/java/org/apache/sysml/hops/ConvolutionOp.java
niketanpansare: To add Spark instructions (non-optimized version) for these functions, it might take a week or two.

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -1000,6 +1000,37 @@ else if(in.length == 3)
...
+		case CONV2D:
+		case CONV2D_BACKWARD_FILTER:
+		case CONV2D_BACKWARD_DATA:
+		case MAX_POOL:
+		case AVG_POOL:
+		case MAX_POOL_BACKWARD:
+		{
+			// At DML level:
+			// output = conv2d(input, filter, input_shape=[3, 2, 2], filter_shape=[3, 2, 2], 
+			// strides=[1, 1], border_mode="valid")
```
dusenberrymw: Side note... It would be really nice to have an arbitrary list construct in DML to hold various items, such as multiple matrices to be passed around.

docs/devdocs/deep-learning.md
```diff
@@ -0,0 +1,122 @@
...
+* Element-wise addition: `X + Y`
+* Element-wise subtraction: `X - Y`
+
+SystemML does not support implicit broadcast for above tensor operations, however one can write a DML-bodied function to do so.
+For example: to perform the above operations with broadcasting on second dimensions, one can use the below `rep(Z, n)` function:
+``` python
+rep = function(matrix[double] Z, int C) return (matrix[double] ret) {
+	ret = Z
+	parfor(i in 2:C) {
+		ret = cbind(ret, Z)
```
dusenberrymw: I believe this needs to be a `for` loop for the `cbind` to work.

docs/devdocs/deep-learning.md
```diff
@@ -0,0 +1,122 @@
...
+* X of shape [1, 1, H, W] and Y of shape [N, C, H, W]: `rep(X, C) + Y`
+
+TODO: Map the NumPy tensor calls to DML expressions.
+
+## Representing images in SystemML
+
+The images are assumed to be stored NCHW format, where N = batch size, C = #channels, H = height of image and W = width of image. 
+Hence, the images are internally represented as a matrix with dimension (N, C * H * W).
+
+## Convolution and Pooling built-in functions
```
dusenberrymw: Let's add a quick description of our convolution implementation that just says that we pad the image, perform `im2col` to extract patches, and perform a matrix multiplication between the patches matrix and the filters.  Also, we may want to make a small note that we follow the convention of TensorFlow [1], Caffe, and Torch by technically performing *cross-correlation* versus true convolution since we do not flip the filters (in contrast to the default in Theano).  Totally the correct move for the DL application, but may be worth noting.

(1): https://www.tensorflow.org/versions/r0.8/api_docs/python/nn.html#convolution

niketanpansare: Sure, can you please go ahead and add the description about performing cross-correlation in deep-learning.md. Bute please donot add implementation detail as it is subject to change and depends on the memory requirements, ExecType, etc. Also, in my implementation I do not perform explicit padding of the image.

dusenberrymw:
@niketanpansare I left some comments. Really awesome work here! 👍

dusenberrymw:
Also, I saw that some files have been pulled in from other commits in the project.  I've found rebasing with git to be a really helpful and clean solution when updating a branch with the latest commits from the project.  I use `git pull --rebase apache master` while on a branch, where `apache` is just my alias for the SystemML git repo.

dusenberrymw:
Also, I embedded the conv functions into the `conv` layer locally in the SystemML-NN library and successfully ran the gradient checks there too.

I'll look into the pooling functions tomorrow!

niketanpansare:
Thanks @dusenberrymw  :)

@mboehm7 :  @dusenberrymw and I have added R testcase for convolution and its backward functions.

src/test/scripts/functions/tensor/PoolTest.R
```diff
@@ -0,0 +1,98 @@
...
+  out = matrix(0, N, C*Hout*Wout, byrow=TRUE)
+
+  # Max pooling - im2col implementation
+  for (n in 1:N) {  # all examples
+    img = matrix(X[n,], C, Hin*Win, byrow=TRUE)  # reshape
+    img_maxes = matrix(0, C, Hout*Wout, byrow=TRUE)  # zeros
+
+    for (c in 1:C) {  # all channels
+      # Extract local image slice patches into columns with im2col, of shape (Hf*Wf, Hout*Wout)
+      img_slice_cols = im2col(img[c,], Hin, Win, Hf, Wf, strideh, stridew)
```
dusenberrymw: Mobile right now, but I bet the issue is due to `img[c,]` returning a vector/list in R, resulting in the subsequent `nrow(img)` in `im2col` returning NA. Pooling just reuses the `im2col` code, but does it over each channel separately, rather than over all channels collectively like in convolution, so this issue would only show up here. Can you try explicitly wrapping `img[c,]` here in a matrix with 1 row and Hin*Win columns?

niketanpansare:
@mboehm7 @dusenberrymw We now have R testcases for all the built-in functions in this PR :)

mboehm7:
awesome - thanks both of you for making it happen. I'll not come to it today but will do another full pass over the code tomorrow and then we should be good to go.

src/main/java/org/apache/sysml/api/jmlc/Connection.java
```diff
@@ -189,6 +190,10 @@ public PreparedScript prepareScript( String script, Map<String, String> args, St
 			
 			//System.out.println(Explain.explain(rtprog));
 		}
+		catch(ParseException pe) {
+			// don't chain ParseException (for cleaner error output)
```
deroneriksson: This is from previous commit.

src/main/java/org/apache/sysml/hops/DataOp.java
```diff
@@ -568,4 +568,25 @@ public boolean compare( Hop that )
 		
 		return ret;
 	}
+
```
deroneriksson: From previous commit.

src/main/java/org/apache/sysml/parser/Expression.java
```diff
@@ -30,34 +30,50 @@
 
 public abstract class Expression 
 {
+	/**
```
deroneriksson: Javadocs from previous commit.

src/main/java/org/apache/sysml/parser/Expression.java
```diff
@@ -130,30 +151,30 @@
...
 	};
 	
+	/**
+	 * External built-in function operators.
+	 */
 	public enum ExtBuiltinFunctionOp {
 		EIGEN, CHOLESKY
 	};
 
-	public enum AggOp {
```
deroneriksson: Removed unused enums in previous commit.

src/main/java/org/apache/sysml/parser/ParseException.java
```diff
@@ -148,53 +149,7 @@ public String getMessage() {
 	 * @return String representing the list of parse issues.
 	 */
 	private String generateParseIssuesMessage() {
-		if (_scriptString == null)
```
deroneriksson: From previous commit.

src/main/java/org/apache/sysml/parser/common/CustomErrorListener.java
```diff
@@ -34,7 +34,8 @@
 
 	private static final Log log = LogFactory.getLog(DMLScript.class.getName());
 
-	private boolean atleastOneError = false;
+	private boolean atLeastOneError = false;
```
deroneriksson: error/warning handling from previous commit.

src/main/java/org/apache/sysml/parser/dml/DMLParserWrapper.java
```diff
@@ -179,10 +177,15 @@ public DMLProgram doParse(String fileName, String dmlScript, String sourceNamesp
 		DmlSyntacticValidator validator = new DmlSyntacticValidator(errorListener, argVals, sourceNamespace);
 		walker.walk(validator, tree);
 		errorListener.unsetCurrentFileName();
-		if (errorListener.isAtleastOneError()) {
-			List<ParseIssue> parseIssues = errorListener.getParseIssues();
+		this.parseIssues = errorListener.getParseIssues();
```
deroneriksson: from previous commit.

src/main/java/org/apache/sysml/parser/dml/DmlSyntacticValidator.java
```diff
@@ -426,12 +428,111 @@ public void exitFunctionCallAssignmentStatement(FunctionCallAssignmentStatementC
...
+			paramExpression.add(new ParameterExpression("filter_shape2", new IntIdentifier(1, currentFile, ctx.start.getLine(), ctx.start.getCharPositionInLine(), 
+					ctx.stop.getLine(), ctx.stop.getCharPositionInLine())));
+			paramExpression = replaceListParams(paramExpression, "pool_size", "filter_shape", ctx, 3);
+			if(functionName.equals("max_pool_backward"))
+				paramExpression = orderConvolutionParams(paramExpression, 2, ctx);
+			else
+				paramExpression = orderConvolutionParams(paramExpression, 1, ctx);
+		}
+		
+		castAsScalarDeprecationCheck(functionName, ctx);
```
deroneriksson: castAsScalar deprecation from previous commit.

deroneriksson:
I noticed code included from previous commits (as @dusenberrymw mentioned). I started pointing the files/lines out but stopped after a while because I don't want to cause confusion. Please be careful merging. :-)

niketanpansare:
@deroneriksson @dusenberrymw  The files are getting added when I rebase. I will look into resolving it. Thanks.

niketanpansare:
Closing the PR as created a new PR by cherry-picking commits ... https://github.com/apache/incubator-systemml/pull/142

PR #135, opened 2016-05-03
Small fixes in Zeppelin example documentation.

ethanyxu:
Changed the link to the sample Zeppelin note.json file. The old link pointed to a file with package path `com.ibm...` leading to errors.
Slightly modified the instruction to import the Zeppelin note.json.
Fixed small typo in configuration file path.


deroneriksson:
@ethanyxu Thank you for these updates. The tar.gz was out-of-date and the newer json import technique is a better way to import into Zeppelin. LGTM.

PR #136, opened 2016-05-04
[SYSTEMML-647] Replace castAsScalar calls

deroneriksson:
Replace castAsScalar() with as.scalar() in DML.
Replace castAsScalar() with scalar() in PYDML.


deroneriksson:
All tests passed [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/315/).

dusenberrymw:
LGTM.

PR #140, opened 2016-05-06
[SYSTEMML-668] Python MLOutput.getDF() Can't Access JVM SQLContext

dusenberrymw:
In PySpark, access to the JVM SQLContext from a PySpark SQLContext instance has changed from `sqlContext._scala_SQLContext` to `sqlContext._ssql_ctx`, breaking any previous code using the former construct, such as our Python `MLOutput.getDF(...)` method.  This updates the latter method.

### DIFF
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -226,7 +226,7 @@ class MLOutput(object):
 
     def getDF(self, sqlContext, varName):
         try:
-            jdf = self.jmlOut.getDF(sqlContext._scala_SQLContext, varName)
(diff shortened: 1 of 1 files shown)


mboehm7:
@dusenberrymw could we abstract this somehow to access the sql context in a version-dependent manner for backwards compatibility?

dusenberrymw:
@mboehm7 I looked into this more.  It turns out that access to the JVM SQLContext has always been official exposed via `sqlContext._ssql_ctx` (as far back as 1.0).  However, we have been using an unofficial variable, `sqlContext._scala_SQLContext`, which has been renamed in 2.0.  So, luckily, this change will give us proper backward (and hopefully future) compatibility.

mboehm7:
excellent  - thanks @dusenberrymw

PR #141, opened 2016-05-06
[SYSTEMML-560] Frame converter between Matrix and Binary block

asurve:


mboehm7:
1) Simplifications: Could you please simplify the implementation of matrixBlockToBinaryBlockLongIndex to use mapToPair instead of mapPartitionsToPair as you don't need the full partition scope here. Also please use the existing util functions for getBlockIndex and getIndexInBlock instead of creating replicas of these functions (e.g., in FrameRDDConverterUtils, FrameReblockBuffer).

2) Code style: Please stick to conventions and capitalize class names (e.g., matrixToBinaryBlockFunction) as well as use postincrement instead of preincrement if possible.

asurve:
All review comments addressed.

mboehm7:
LGTM - thanks @asurve

PR #142, opened 2016-05-06
[SYSTEMML-540] Initial implementation of conv2d/maxpooling builtin functions (new)

niketanpansare:
2. The documentation is available in docs/devdocs folder.
3. Added R test-cases for the above built-in functions.

@dusenberrymw @deroneriksson @mboehm7 @bertholdreinwald I created a new PR and cherry-picked the commits from previous PR to avoid commit-related issues.


mboehm7:
thanks for the cleaned up pull request - @niketanpansare . Here are a couple of comments to further improve the current status:

1) Hop: Keeping the separate hop is ok for isolation, but we need to be aware that this hop then needs to be treated separately in the rewrites e.g., simplify agg over reorg. If you'd like to keep this hop, please remove the state that is just used locally (because it might confuse people why this status is not part of deep copy and compare)

2) Operator Selection: Please force the execution type cp, otherwise it would fail to compile
if we are running against hadoop/spark execution modes.

3) Simplify Parser: Could we please keep the syntactic parsers (dml, pydml) free of builtin function names and specific parameters. Please, move the entire code to validate (if really needed during parsing, at least let's rework this to a call to the respective expressions)

4) Simplify MatrixObject: in clearReusableData, the _data member is always of type MatrixBlock - so there is no need for checks or casts.

5) Simplify MatrixBlock: Please move the map for reused objects out of MatrixBlock as you anyway use it only from your CP instruction. Also let's remove the references from MatrixBlock to Statistics (having the map outside of matrixblock would simplify this too). Furthermore, you might want to think about a new operator type and methods in matrix block to surface the calls to LibMatrixDNN such that all operations are callable directly on the matrix block object.

6) Correctness Runtime NNZ Maintenance: You explicitly set outputBlock.setNonZeros(numRows * numCols) during setup of the output blocks. This is incorrect if not all cells are represented in the output (not matter if dense or not) and if your operations call a set/get which internally increments the nnz too.

7) Correctness LibMatrixDNN: In doPoolingBackward you use double maxVal = Double.MIN_VALUE; However, this is incorrect in the presence of negative values. Please use -Double.MAX_VALUE (note that Double.MIN_VALUE is the smallest double magnitude not the minimal value). Also, it would be good to use the threads as passed into the runtime, without additional calls to OptimizerUtils.

8) Cleanup Tests: The calls to bHM.clear() in finalize are unnecessary as these maps are fully garbage collected.

9) Code Style: Finally, there are some minor code style issues. First, please use auto-boxing instead of explicit new Integer(val) whenever possible. Second, don't unnecessarily separate declaration and value assignment (e.g., params in ConvolutionCPInstruction). Third, please remove commented code.

niketanpansare:
@mboehm7 Added your suggestions. Please take a look. Thanks :)

mboehm7:
thanks @niketanpansare - we just need to do some final cleanups:

see 1 above) The modified hop is already much cleaner. Let's just also completely remove ConvolutionParameters (as member and from compare/clone) as you only use is transiently anyway. 

see 2 above) The current version would work. However, the flag _etypeForced is meant to be set only from outside the hop (e.g., by the parfor optimizer). It would be great if you could use our default approach for forcing the execution type during operator selection (e.g., transformapply/decode/meta in ParameterizedBuiltinOp).

see 5 above) Could we please remove the calls from MatrixBlock to Statistics in order to make sure it is not called for distributed operations, where (other than parfor) we anyway do not propagate statistics back to the driver.

Finally, there are a couple of TODOs for documentation, pydml integration, sparse operations, distributed Spark/MR operations, etc. Could we please create jira tasks/subtask for them to keep track that these issues are resolved in the near future? Thanks.

niketanpansare:
Thanks @mboehm7 ... I have added suggested modification. Please take a look.

Also I have created following new JIRAs under epic SYSTEMML-540:
https://issues.apache.org/jira/browse/SYSTEMML-685
https://issues.apache.org/jira/browse/SYSTEMML-686
https://issues.apache.org/jira/browse/SYSTEMML-687

mboehm7:
LGTM - thanks @niketanpansare for your efforts

niketanpansare:
Thanks @mboehm7 

@dusenberrymw Do you have any suggestions ?

dusenberrymw:
@niketanpansare Overall, this is awesome!   I created new layers in my library using the new built-in functions internally for convolution and max pooling, and plugged them into all my DML test suites.  The convolution functions work great!  

I did find one problem with the built-in max pooling function: values are always compared to a lower-bounded threshold of `0`, so if a patch has all negative values, the output for the max pooling operation over that patch will be `0`, rather than the maximum negative number in that patch.  I will comment on the line, but a quick update fixes it.  Would probably be good to add a few tests over entirely negative tensors to unit test it.

Looking forward to the GPU & Spark implementations!

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
```diff
@@ -0,0 +1,564 @@
...
+		for (int p = 0; p < params.P; p++) {
+			for (int q = 0; q < params.Q; q++) {
+				int start_index_h = p * params.stride_h - params.pad_h;
+				int start_index_w = q * params.stride_w - params.pad_w;
+				int end_index_h = Math.min(start_index_h + params.R, params.H);
+				int end_index_w = Math.min(start_index_w + params.S, params.W);
+				start_index_h = Math.max(start_index_h, 0);
+				start_index_w = Math.max(start_index_w, 0);
+				int out_index = n*params.C*params.P*params.Q + c*params.P*params.Q +  p * params.Q + q;
+				outputArray[out_index] = 0;
```
dusenberrymw: If we update this to `outputArray[out_index] = -Double.MAX_VALUE;`, patches with all negative values will produce the correct result.

src/test/java/org/apache/sysml/test/integration/functions/tensor/PoolTest.java
```diff
@@ -0,0 +1,134 @@
...
+		int numImg = 2; int imgSize = 4; int numChannels = 2;  int stride = 1; int pad = 0; int poolSize1 = 3; int poolSize2 = 3;
+		runConv2DTest(ExecType.CP, imgSize, numImg, numChannels, stride, pad, poolSize1, poolSize2, "max");
+	}
+	
+	/**
+	 * 
+	 * @param et
+	 * @param sparse
+	 */
+	public void runConv2DTest( ExecType et, int imgSize, int numImg, int numChannels, int stride,
```
dusenberrymw: Nitpick if we're going to update the tests anyway: `runMaxPoolTest`.

src/test/java/org/apache/sysml/test/integration/functions/tensor/PoolBackwardTest.java
```diff
@@ -0,0 +1,129 @@
...
+		int numImg = 2; int imgSize = 7; int numChannels = 2;  int stride = 2; int pad = 0; int poolSize1 = 3; int poolSize2 = 3;
+		runConv2DTest(ExecType.CP, imgSize, numImg, numChannels, stride, pad, poolSize1, poolSize2, "max");
+	}
+	
+	/**
+	 * 
+	 * @param et
+	 * @param sparse
+	 */
+	public void runConv2DTest( ExecType et, int imgSize, int numImg, int numChannels, int stride,
```
dusenberrymw: (Nitpick if we're going to update the tests anyway: `runMaxPoolBackwardTest`)

niketanpansare:
Thanks @dusenberrymw for your comments. Please take a look at new commit.

dusenberrymw:
**LGTM**. 👍

PR #143, opened 2016-05-09
[SYSTEMML-667] Remove js and css references from standalone LICENSE

deroneriksson:
Remove js and css references in standalone LICENSE.
Remove unused antlr version reference.


deroneriksson:
The docs folder is not packaged in the standalone tar.gz and zip so the js and css references don't need to be included in the standalone LICENSE.

@lresende Please review.

dusenberrymw:
LGTM.

deroneriksson:
@dusenberrymw Thank you for reviewing.

PR #144, opened 2016-05-09
[SYSTEMML-663] Remove redundant standalone tar.gz and zip artifacts

deroneriksson:
Remove compile-scoped dependencies from standalone lib directory
since they exist in the SystemML jar.


dusenberrymw:
LGTM.

deroneriksson:
@dusenberrymw Thanks for reviewing!

PR #145, opened 2016-05-09
[SYSTEMML-672] Fix PyDML Indexing To Be 0-Based

dusenberrymw:
Currently, PyDML has a *bug* in which it uses 1-based indexing, as in R & DML, rather than the correct 0-based indexing that Python uses.  This simply aims to fix this issue in the Parser, and then update the few PyDML test scripts to use the correct 0-based indexing.


dusenberrymw:
cc @mboehm7, @deroneriksson, @gweidner

deroneriksson:
From discussions with a couple Python developers, this sounds like the best course to take WRT indexing.

@dusenberrymw Does slicing mirror the behavior of Python?

dusenberrymw:
Yeah the slicing is correct as well: lower indices are inclusive, and upper indices are exclusive. Note that the upper bounds were already correctly exclusive in PyDML prior to this PR.

dusenberrymw:
Also, I have some plans to improve the syntax for slicing as well to be closer to NumPy syntax.  It should be mostly cosmetic at the grammar level.

deroneriksson:
Thank you @dusenberrymw. The closer we mimic Python syntax, the easier it will be for Python developers to use PyDML.

dusenberrymw:
To expand on the slicing aspect a bit, the slicing correctly uses inclusive lower and exclusive upper bounds.  However, a really useful piece of slicing syntax we are currently missing from PyDML is the ability to have implicit lower or upper bounds, i.e., `X[:3,]` should return all rows up to (but not including) 3.  This means that the implicit lower bound is 0.  Similarly, `X[2:,]` should return all rows starting at row 2 (0-based), which implies that the upper bound is equal to the number of rows of X.

dusenberrymw:
I've created [SYSTEMML-676](https://issues.apache.org/jira/browse/SYSTEMML-676) to track the slicing improvements.

dusenberrymw:
Also, [test build 343](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/343/) is running.

mboehm7:
just to provide some history here: at one point we supported implicit lower and upper bounds, but removed it because (1) it is invalid R (1:3 is a shorthand for sequence, which is undefined with missing lower/upper bounds), and (2) because we cannot compile valid execution plans in case of unknown input size. However, meanwhile (with dynamic recompilation), the latter problem can be mitigated with additional runtime changes (e.g., for forced CP execution w/o recompile).

dusenberrymw:
Test build 343 was successful.

deroneriksson:
LGTM. Thank you @dusenberrymw

dusenberrymw:
@mboehm7 Great to see some of the history.  I'm going to move this conversation to the slicing JIRA and continue it there.

dusenberrymw:
Merged.

PR #147, opened 2016-05-10
[SYSTEMML-654] Support DML functions override built-in functions

gweidner:
This patch adds support for overriding built-in functions.  The approach was to track new internal and external function definition names per script and skip built-in function call handling if the function name was also defined by user (or imported when converting pydml syntax).  Also added parse error detection/reporting if user attempts to define two functions with same name in same script.  Previously, the second definition would overwrite the first.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DataExpression.java
+++ b/src/main/java/org/apache/sysml/parser/DataExpression.java
@@ -200,7 +200,7 @@ public class DataExpression extends DataIdentifier
 				String pname = currExpr.getName();
 				Expression pexpr = currExpr.getExpr();
 				if (pname == null){
-					dataExpr.raiseValidateError("for Rand Statment all arguments must be named parameters");	
+					dataExpr.raiseValidateError("for Rand Statement all arguments must be named parameters");	
 				}
 				dataExpr.addRandExprParam(pname, pexpr); 
 			}
--- a/src/main/java/org/apache/sysml/parser/common/CommonSyntacticValidator.java
+++ b/src/main/java/org/apache/sysml/parser/common/CommonSyntacticValidator.java
@@ -78,13 +78,13 @@ public abstract class CommonSyntacticValidator {
 		_scripts.get().clear();
 	}
 
-	public CommonSyntacticValidator(CustomErrorListener errorListener, Map<String,String> argVals, String sourceNamespace) {
+	public CommonSyntacticValidator(CustomErrorListener errorListener, Map<String,String> argVals, String sourceNamespace, Set<String> prepFunctions) {
 		this.errorListener = errorListener;
 		currentFile = errorListener.getCurrentFileName();
 		this.argVals = argVals;
 		this.sourceNamespace = sourceNamespace;
 		sources = new HashMap<String, String>();
-		functions = new HashSet<String>();
+		functions = (null != prepFunctions) ? prepFunctions : new HashSet<String>();
 	}
 
 	protected void notifyErrorListeners(String message, int line, int charPositionInLine) {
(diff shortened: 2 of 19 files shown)


gweidner:
Tests passed at [OnDemand build 164](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/164/) except for 1 intermittent failure not related to this PR:
FrameConverterTest.testFrameMixedCsvBinSpark:85->runFrameConverterTest:185->verifyFrameData:225 Target value for cell (0,2) is -7, is not same as original value -6.4639550422544385

Note the error found in [SYSTEMML-670](https://issues.apache.org/jira/browse/SYSTEMML-670) to be addressed separately.

gweidner:
All tests passed [PR build 347](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/347/).

dusenberrymw:
Thanks, @gweidner for this!  A few thoughts:

mboehm7:
@dusenberrymw could you please do me a favor and reproduce this issue with a function that has a scalar/matrix return type? Thanks.

dusenberrymw:
@mboehm7 Definitely.  Here are some interesting results.

1) Scalar integer return type.  Works if `rand` is defined before it's usage, but not after, just like the initial example.
```
y = rand(2)
print(y)

rand = function(int x) return (int y){
  print("hi")
  z = matrix(x, rows=4, cols=2)
  y = x*2
}
```
```
16/05/11 10:20:52 ERROR DMLScript: line 6:0 unable to process builtin function expression rand:ERROR: null -- line 6, column 0 -- for Rand Statment all arguments must be named parameters
```
2) Matrix return type.  Works if `rand` is defined before it's usage, but not after, just like the initial example.
```
z = rand(2)
print(sum(z))

rand = function(int x) return (matrix[double] z){
  print("hi")
  z = matrix(x, rows=4, cols=2)
  y = x*2
}
```
```
16/05/11 10:21:50 ERROR DMLScript: line 6:0 unable to process builtin function expression rand:ERROR: null -- line 6, column 0 -- for Rand Statment all arguments must be named parameters
```
3) Scalar integer *and* matrix return types.  This one is interesting -- it does *not* work regardless of whether it is placed before or after its usage.
```
[y, z] = rand(2)
print(y + ", " + sum(z))

rand = function(int x) return (int y, matrix[double] z) {
  print("hi")
  y = x*2
  z = matrix(y, rows=4, cols=2)
}
```
```
16/05/11 10:22:50 ERROR DMLScript: line 6:0 unable to process builtin function expression rand:ERROR: null -- line 6, column 0 -- for Rand Statment all arguments must be named parameters
```
```
rand = function(int x) return (int y, matrix[double] z) {
  print("hi")
  y = x*2
  z = matrix(y, rows=4, cols=2)
}

[y, z] = rand(2)
print(y + ", " + sum(z))
```
```
16/05/11 10:23:33 ERROR DMLScript: line 6:0 unable to process builtin function expression rand:ERROR: null -- line 6, column 0 -- for Rand Statment all arguments must be named parameters
```

gweidner:
Thank you both for the review and comments in catching additional scenarios that I missed.
I've pushed new commits as follows.
1) Rebased and updated indexing for pydml tests.
2) Expanded unit test scripts to include previously failing scenarios.
3) Added "preprocessor" class to get names of function definitions to handle case where definition occurs after the new function used in a statement.
4) Moved function name validation from the SyntacticValidator to preprocessor and updated ParserWrappers to use preprocessor.
5) Updated CommonSyntacticValidator.buildForBuiltInFunction to handle case of multiple value return (e.g., scalar and matrix).
6) Fixed typo in error message for DataExpression.

Note preprocessing (which is intended to be minimal to limit performance impact) is done both for "main" script and imported scripts such that user functions take precedence over built-in functions with same name in sourced scripts too for consistency.

gweidner:
Test in progress at [OnDemand build 167](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/167/).

dusenberrymw:
Thanks, @gweidner.  This works great now!

dusenberrymw:
LGTM.

PR #148, opened 2016-05-10
[SYSTEMML-676] Improve PyDML Slicing

dusenberrymw:
Currently, slicing correctly uses inclusive lower and exclusive upper bounds. However, a really useful piece of slicing syntax we are currently missing from PyDML is the ability to have implicit lower or upper bounds, i.e., `X[:3,]` should return all rows up to (but not including) 3. This means that the implicit lower bound is 0. Similarly, `X[2:,]` should return all rows starting at row 2 (0-based), which implies that the upper bound is equal to the number of rows of `X`.  We should update the grammar to allow these cases, row both row and column slicing.  More generally, `X[:,:]`, should also be valid, regardless if it is useful.

Example (PyDML):
```
N = 4
D = 3
X = matrix("[1 2 3; 4 5 6; 7 8 9; 10 11 12]")
X = X[:2,1:]  # slice with implicit lower row and upper column bounds

# print
for i in 0:nrow(X)-1:
  for j in 0:ncol(X)-1:
    print("X["+i+","+j+"]: " + scalar(X[i,j]))
```

```
X[0,0]: 2.0
X[0,1]: 3.0
X[1,0]: 5.0
X[1,1]: 6.0
```

### DIFF
--- a/src/main/java/org/apache/sysml/parser/pydml/Pydml.g4
+++ b/src/main/java/org/apache/sysml/parser/pydml/Pydml.g4
@@ -220,8 +220,8 @@ dataIdentifier returns [ org.apache.sysml.parser.common.ExpressionInfo dataInfo 
        // $dataInfo.expr = new org.apache.sysml.parser.DataIdentifier();
 } :
     // ------------------------------------------
-    // IndexedIdentifier
-    name=ID OPEN_BRACK (rowLower=expression (':' rowUpper=expression)?)? ',' (colLower=expression (':' colUpper=expression)?)? CLOSE_BRACK # IndexedExpression
+    // IndexedIdentifier -- allows implicit lower and upper bounds
+    name=ID OPEN_BRACK ( (rowLower=expression)? (rowImplicitSlice=':' (rowUpper=expression)?)? )? (',' (((colLower=expression)? (colImplicitSlice=':' (colUpper=expression)?)?)?)?)? CLOSE_BRACK # IndexedExpression
     // ------------------------------------------
     | ID                                            # SimpleDataIdentifierExpression
     | COMMANDLINE_NAMED_ID                          # CommandlineParamExpression
--- a/src/main/java/org/apache/sysml/parser/pydml/PydmlSyntacticValidator.java
+++ b/src/main/java/org/apache/sysml/parser/pydml/PydmlSyntacticValidator.java
@@ -36,6 +36,7 @@ import org.apache.sysml.conf.CompilerConfig.ConfigType;
 import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.parser.AssignmentStatement;
 import org.apache.sysml.parser.BinaryExpression;
+import org.apache.sysml.parser.BuiltinFunctionExpression;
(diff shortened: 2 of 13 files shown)


dusenberrymw:
@mboehm7 Can you take a look at this approach to adding implicit bounds at the parser level to **PyDML**?  Essentially, I insert a `1` int expression for implicit lower bounds, and `nrow(X)` or `ncol(X)` for implicit upper bounds.

If we think this is a decent approach, I can clean this up and add tests.

Note: We will only apply this to PyDML.

mboehm7:
in general the approach is fine - just two things to think about: 

(1) Usually we try to keep the parser(s) free of semantics regarding individual operations. It would be great to have this functionality in the actual validation step. I understand that this piece is pydml-specific and hence moved into the pydml parser. However, maybe it's possible to fill in the missing lower/upper bounds during validate (in a generic manner) but reject missing lower/upper bounds in the dml parser?

(2) Negative ranges: R supports something like X[10:5,] which returns rows 5 through 10 in reverse order. Does python support this too and if yes, how would we define the implicit lower/upper bounds there? I think we should just mention this in the documentation for now, as our runtime/compiler anyway does not support reverse ranges for indexing (yet).

dusenberrymw:
@mboehm7 I looked into doing (1), however, I think that route is going to introduce a lot more confusing logic than simply having it in the initial parser.  If we do this and place it in the validation step of `IndexedIdentifier`, we have to inject state for whether implicit bounds were used (the use of a colon in a statement, such as `X[2:,]`, then override the existing behavior of setting both the upper and lower bounds to the same value if one actual index is received, then add the custom logic for actually injecting the implicit bounds, and all of this is only applicable for PyDML.  I think simply "filling in" the implicit bounds for the user when parsing the PyDML and then sending it on to the validation step like normal keeps the code cleaner, at least in this particular example.  Thoughts?

As for (2), Python doesn't have this functionality.  However, it does have another slicing functionality for having indexes that count backwards from the bounds, i.e., `X[:-1,]` would index all rows *except* the final row.

mboehm7:
1) This is fine - let's go with the parser change for now. However, you might want to create a jira task to track a parser cleanup where we systematically go over the parser and evaluate what really needs to be at the level of the syntactic parser (e.g., I'm still wondering if we could have done the UDF namespace handling at language level after parsing w/o the second pass and independent of pydml/dml frontends).

2) Interesting - I agree that this can be easily addressed (by a separate PR). Thanks for sharing.

dusenberrymw:
Great, I'll go ahead with the small addition in the PyDML parser.  Also, just to expand a bit on reverse sequences in (2), while `X[3:0,]` doesn't work directly in Python, the addition of a third "step" parameter in the slice will allow it, i.e., `X[3:0:-1,]` will return row 3 (inclusive) back to row 0 (exclusive) correctly reversed.

dusenberrymw:
[Test build 364](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/364/) running.

dusenberrymw:
Testing successful.

PR #152, opened 2016-05-11
[SYSTEMML-683] Update LICENSE and NOTICE for in-memory jar

deroneriksson:


deroneriksson:
Note: Apache JIRA seems to be experiencing some difficulties. I created SYSTEMML-682 for this but it seems to have 'mostly' disappeared from the JIRA site. So I created a new SYSTEMML-683 to handle this. Apache JIRA is on 'lockdown mode', which probably has something to do with this strange behavior.

dusenberrymw:
LGTM.

deroneriksson:
Thank you @dusenberrymw

PR #154, opened 2016-05-12
[API DOC] add explain for addIDToDataFrame since it's special

Wenpei:
I add a explain since it's special that start from 1 but not 0.

I will try add scala version to make this parameter-able in another pr.


Wenpei:
ping @niketanpansare 
It's a simple doc change.

dusenberrymw:
@Wenpei LGTM.  I just wanted to follow the conversation on the other PR prior to merging this.

Thanks!

PR #155, opened 2016-05-12
[SYSTEMML-646][SYSTEMML-581] Make testing of MLPipeline wrappers more robust

Wenpei:
This PR include
1. Move test to test_suites as @dusenberrymw suggestion.
2. Add result validation to compare with baseline data and spark.


src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -33,6 +33,12 @@ import org.apache.spark.SparkConf
 import org.apache.spark.mllib.linalg.Vectors
 import org.apache.spark.mllib.regression.LabeledPoint
 import scala.reflect.ClassTag
+import org.apache.spark.sql.types.DataTypes
+import org.apache.spark.sql.RowFactory
+import org.apache.spark.sql.Row
+import scala.collection.mutable.ArrayBuffer
+import org.apache.spark.sql.types.StructField
+import org.apache.sysml.api.ml.util.RDDConverterUtilsExt
```
dusenberrymw: Nitpick: Can you sort these imports to have:
1. `scala.....` at the top, followed by a newline.
2. `org.apache.spark.....` next, followed by a newline.
3. `org.apache.sysml.....` next.

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -160,8 +167,8 @@ class LogisticRegressionModel(
     val rawPred = outNew.getDF(df.sqlContext, "rawPred", true).withColumnRenamed("C1", "rawPrediction").withColumnRenamed("ID", "ID2")
     var predictionsNProb = prob.join(pred, prob.col("ID").equalTo(pred.col("ID1"))).select("ID", "probability", "prediction")
     predictionsNProb = predictionsNProb.join(rawPred, predictionsNProb.col("ID").equalTo(rawPred.col("ID2"))).select("ID", "probability", "prediction", "rawPrediction")
-    val dataset1 = RDDConverterUtils.addIDToDataFrame(df, df.sqlContext, "ID")
-    dataset1.join(predictionsNProb, dataset1.col("ID").equalTo(predictionsNProb.col("ID")))
+    val dataset1 = RDDConverterUtilsExt.addIDToDataFrame(df,"dfindex",0)
```
dusenberrymw: Can you give some background on this proposed change?

dusenberrymw:
Thanks, @Wenpei, for moving and cleaning up these tests!  I left a few comments.

src/test/scala/org/apache/sysml/test/integration/applications/mlpipeline/LogisticRegressionSuite.scala
```diff
@@ -0,0 +1,147 @@
...
+    val lr_spark = new org.apache.spark.ml.classification.LogisticRegression("log")
+    val pipeline_spark = new Pipeline().setStages(Array(tokenizer, hashingTF, lr_spark))
+    val crossval_spark = new CrossValidator().setEstimator(pipeline_spark).setEvaluator(new BinaryClassificationEvaluator)
+    val paramGrid_spark = new ParamGridBuilder().addGrid(hashingTF.numFeatures, Array(10, 100, 1000)).addGrid(lr_systemml.regParam, Array(0.1, 0.01)).build()
+    crossval_spark.setEstimatorParamMaps(paramGrid_spark)
+    crossval_spark.setNumFolds(2)
+    val model_spark = crossval_spark.fit(training.map { x => LabeledDocument(x.id,x.text, x.label-1)}.toDF)
+    val scored_spark = model_spark.transform(test.toDF).orderBy("id")
+    
+    scored_systemml.select("prediction").collect() shouldEqual scored_spark.select(scored_spark("prediction")+1).collect()
```
niketanpansare: This implies that in our test suite, we are comparing the result of our wrappers with that of MLLib. It is a logical assumption and I am fine with this approach. But we should let others comment on it before merging the PR as we may want to use the same approach for other wrappers. @mboehm7 @bertholdreinwald @prithvirajsen @fmakari 

Note: an alternate approach for testing is that we create a pipeline without the model to generate the features. Save the file in csv/ijv format and test using R script. For example: https://github.com/apache/incubator-systemml/blob/master/src/test/java/org/apache/sysml/test/integration/applications/LinearRegressionTest.java#L89

Good work Wenpei :)

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
Wenpei: Original `RDDConverterUtils.addIDToDataFrame` use 1 as start index, which cause dataset join unmatch. 
so I add new function `RDDConverterUtilsExt.addIDToDataFrame` to make start value set-able.

Wenpei:
Address @dusenberrymw 's comments

Wenpei:
Build failed, let me take a look.

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
dusenberrymw: @niketanpansare Do you know if we need to use 1-based indexing in the original `RDDConverterUtils.addIDToDataFrame(...)` function?  If not, would it be reasonable to allow 0-based indexing to it, given that 1-based indexing can cause issues?

src/test/scala/org/apache/sysml/test/integration/applications/mlpipeline/LogisticRegressionSuite.scala
dusenberrymw: For these wrappers that are designed to be used as a direct replacement for Spark ML algorithms, I think that testing against those Spark ML algorithms would be appropriate.

dusenberrymw:
@gweidner Do you have any thoughts on how @Wenpei could integrate these Scala tests into our current testing framework, given the issue described?

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
niketanpansare: @dusenberrymw Yes, since the ID correspond to row indexes (i.e. X[i,]) they are one-based.

@Wenpei Can you explain where the zero-based IDs are coming in ? If one of our converter utils is returning a zero-based IDs, then its a bug and we should fix that instead of making start index settable.

Wenpei: @niketanpansare I will submit a Jira bug if needed.

niketanpansare: Thanks @Wenpei

Wenpei:
ping @gweidner

petro-rudenko:
Can you also remove spark context as a constructor parameter for LogisticRegression class:
`class LogisticRegression(override val uid: String, val sc: SparkContext)`

So it becomes compatible with Spark version. It needs only to convert dataframe to binary bloc matrix, and could be get from dataframe: `dataset.getSqlContext.getSparkContext`

niketanpansare:
Thanks @petro-rudenko  ... that's a good point. It simplifies the interface.

@Wenpei do you mind adding this to the PR ?

Wenpei:
@niketanpansare @petro-rudenko thanks, I will do this.
En.., since we can't cover scala test in current systemml test framework(for java junit only). i will change to common scala test.

Wenpei:
I rebase this PR, and add below change
1. Roll back to common scala test.
2. Address @petro-rudenko suggestion.

PS: 
Submit Jira issues for 0-base matrix issues.
https://issues.apache.org/jira/browse/SYSTEMML-765

petro-rudenko:
Can you also add `def this() = this(Identifiable.randomUID("sysml-logreg"))` to LogisticRegression class like in Spark LogReg  to instantiate object without uid param.

rightwaitforyou:
How to use the java code to save/load your trained model?????

Wenpei:
Fix build break. Add address @petro-rudenko 's suggestion.
But issues is, in spark 1.4, Identifiable is private for spark only. So a little different when implement.
@rightwaitforyou , there is no save/load for spark 1.4, I can add it when systemml support spark 1.6

niketanpansare:
LGTM. Thanks Wenpei for your efforts.

As an FYI, the tests are passed: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/435/

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -65,14 +68,18 @@ trait HasRegParam extends Params {
   final def getRegParam: Double = $(regParam)
 }
 object LogisticRegression {
-  final val scriptPath = "scripts" + File.separator + "algorithms" + File.separator + "MultiLogReg.dml"
+  final val scriptPath = "algorithms" + File.separator + "MultiLogReg.dml"
```
deroneriksson: I believe the algorithms are now packaged in scripts/algorithms/.

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -113,7 +120,7 @@ class LogisticRegression(override val uid: String) extends Estimator[LogisticReg
   }
 }
 object LogisticRegressionModel {
-  final val scriptPath = "scripts" + File.separator + "algorithms" + File.separator + "GLM-predict.dml"
+  final val scriptPath = "algorithms" + File.separator + "GLM-predict.dml"
```
deroneriksson: I believe the algorithms are now packaged in scripts/algorithms/.

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -61,14 +68,18 @@ trait HasRegParam extends Params {
   final def getRegParam: Double = $(regParam)
 }
 object LogisticRegression {
-  final val scriptPath = "scripts" + File.separator + "algorithms" + File.separator + "MultiLogReg.dml"
+  final val scriptPath = "algorithms" + File.separator + "MultiLogReg.dml"
```
deroneriksson: I believe the algorithms are now packaged in scripts/algorithms/.

src/main/scala/org/apache/sysml/api/ml/LogisticRegression.scala
```diff
@@ -108,7 +120,7 @@ class LogisticRegression(override val uid: String, val sc: SparkContext) extends
   }
 }
 object LogisticRegressionModel {
-  final val scriptPath = "scripts" + File.separator + "algorithms" + File.separator + "GLM-predict.dml"
+  final val scriptPath = "algorithms" + File.separator + "GLM-predict.dml"
```
deroneriksson: I believe the algorithms are now packaged in scripts/algorithms/.

dusenberrymw:
Any updates on this PR?

Wenpei:
@dusenberrymw Sorry, I take a look this but didn't get good resolution month ago.
Let me take a try this week.

petro-rudenko:
Any updates? Want to seamlessly integrate into my spark ml pipeline, blocked by this.

niketanpansare:
@petro-rudenko Part of code in this PR was refactored (including bug-fixes) in the commit: https://github.com/apache/incubator-systemml/commit/f02f7c018b221384b5e02b13ac181122853f65ba

niketanpansare:
@Wenpei Should we close this PR without merging ?

PR #157, opened 2016-05-13
[SYSTEMML-659] Fix LICENSE and NOTICE for main jar artifact

deroneriksson:
Main jar artifact has compile-scope depenencies, so create assembly
to reflect the included libraries in the LICENSE and NOTICE files.


deroneriksson:
This technique will basically create a new main jar that replaces the old main jar (which is actually the shaded jar which replaced the original jar), but with the right LICENSE, NOTICE, and DISCLAIMER inserted.

@lresende @frreiss Please review, maven experts.

lresende:
@deroneriksson Have you tested the resulting jar ? if it works well, the solution is ok by me.

deroneriksson:
Yes, tested the resulting jar. Thank you for reviewing. Will merge.

PR #158, opened 2016-05-13
[SYSTEMML-692][WIP] Added initial version of DML generator for Caffe proto

niketanpansare:
1. Added initial version of DML generator that supports SGD with momentum+weight decay for various learning rate schedules (const, exp, step, inv) and various initializers (constant, uniform, gaussian, xavier, msra).
2. The initial version contains layers required to implement Lenet. Also added an example Lenet.proto that is able get 98% accuracy on MNIST dataset after 1200 iteration over validation set. This can be improved by trying different learning rate schedule.

``` python
ITER=0 Validation set accuracy (%): 10.72
ITER=100 Validation set accuracy (%): 91.18
ITER=200 Validation set accuracy (%): 94.74
ITER=300 Validation set accuracy (%): 96.42
ITER=400 Validation set accuracy (%): 95.84
ITER=500 Validation set accuracy (%): 96.52
ITER=600 Validation set accuracy (%): 96.84
ITER=700 Validation set accuracy (%): 96.86
ITER=800 Validation set accuracy (%): 97.76
ITER=900 Validation set accuracy (%): 97.78
ITER=1000 Validation set accuracy (%): 98.08
```

TODOs:
1. Integrate Barista into DMLScript that way user can run proto directly. Example: hadoop jar SystemML.jar -f LenetSolver.proto -caffe
2. Discuss and reach consensus on the format of input dataset for Data layer. The current version is only a reference implementation to facilitate discussion. In addition, we need to provide preprocessing tools to convert jpeg images into our binary format, perform z-scoring, etc.
3. Also, the current version generates test DML code in the same DML script. Instead we can create two different DML scripts (for example: GLM.dml and GLM-predict.dml)
4. Consolidate the forward/backward functions in this prototype with Mike's DML bodied function that way we suggest exactly one way to implement a given layer.
5. Add additional example proto files. For example: auto-encoder.
6. Add additional layers. For example: LSTM, RNN, etc.
7. Add additional solver types. For example: Adagrad, ...
8. Add tutorial that describes end-to-end example (for example: download dataset from http://yann.lecun.com/exdb/mnist/, apply preprocessing to convert it into binary block format, ...).
9. Add support for transfer learning.

@bertholdreinwald @dusenberrymw @mboehm7 @prithvirajsen Please have a look and provide suggestions/comments.


mboehm7:
Thanks @niketanpansare for putting this up for discussion. I would, however, prefer to get SYSTEMML-685, SYSTEMML-686 (+MR), SYSTEMML-687 in first before we generate these functions automatically.

niketanpansare:
Thanks @mboehm7 ... I will try my best to get the other JIRAs in asap. I understand your concern that people who want to test SystemML on very large models might get a bad impression of SystemML as whole, based on the narrow set of functions. In that case, we have to politely explain them that this is work-in progress and we will get distributed operators implemented soon. On the flip side, this PR (which I believe is somewhat orthogonal to the above JIRAs) will allow the members of SystemML community try out various network on different datasets, refine and optimize the implementation as well as their model and enhance deep learning expertise in our community.

That said, there are few features that are extremely crucial. For examples: I would be comfortable if we merge this PR with "WIP" tag when:
1. We optimize CP convolution functions for sparse data to somewhat satisfactory degree.
2. We optimize other CP functions including ba+*, r', max, etc.
3. We add the layers required to port Caffe's AlexNet and Autoencoder (may be GoogleNet too). This will ensure that this implementation is fairly general enough and not a mock prototype.

Spark/MR and GPU backend would be nice to have but should not stop us from making progress on this front. Anyways, this is my opinion and if others disagree too (or if others won't be trying out additional models in near future), I can be convinced otherwise :)

The reason I feel this way is because a typical deep learning expert who wants to try out a new framework will use frequently used networks (AlexNet, GoogleNet, Autoencoder, LSTM, etc) on common benchmarks (ImageNet, MNIST, ..) where the model parameters of convolution layers will likely fit on single machine. They will certainly appreciate the flexibility of using larger models on SystemML using our distributed backend. But more importantly, they will also look into whether we are feature complete wrt deep learning (i.e. the layers, initializers, learning rate schedules, different optimizers, etc.) as well as how much of learning curve there is in trying out SystemML.

dusenberrymw:
@niketanpansare I think this is great, and I agree that it will facilitate others in quickly trying out SystemML using existing Caffe models.  Additionally, it could act as a stepping stone towards getting those people to use and adopt SystemML. 👍 

I'll try it out and go through the code soon.

niketanpansare:
@dusenberrymw Thanks Mike :)

dusenberrymw:
Just FYI, I'll be out of town for a few days, so I'll get back to this next week.

niketanpansare:
@dusenberrymw No worries. We need to ensure that both Deep Learning DML Library and Caffe DML generator generate same DML. Of course, that would involve lot of discussion and don't know what is the recommended forum to have that discussion (mailing list, jira or PR).

Here is an example:
Comparing forward functions in scripts/staging/SystemML-NN/nn/layers/conv_builtin.dml and src/main/java/org/apache/sysml/api/dl/layer/ConvolutionLayer.java: Both have same expression for forward for convolution but differ in bias. 
Caffe generator will output:
ones_2 = matrix(1.0, rows=1, cols=(P_2*Q_2));
H_2 = H_2 + matrix(bias_2 %*% ones_2, rows=1, cols=(32*(P_2*Q_2)));

whereas DML library will use:
for (i in 1:P_2*Q_2-1)
    bias = cbind(bias, b)  # creating shape (32, P_2*Q_2)
H_2 = H_2 + matrix(bias, rows=1, cols=32*P_2*Q_2)

Links:
https://github.com/apache/incubator-systemml/pull/160/files#diff-512c25ade02fd0dd41ee07a7b415f85bR70
https://github.com/apache/incubator-systemml/pull/158/files#diff-3fc6d70003f29a7bb91bac7dd043af30R113

Either ways, I think we are heading in right direction wrt DML library and Caffe DML generator. I can imagine both serving the needs/painpoints of different types of end-users :)

dusenberrymw:
@niketanpansare Awesome!  I plan to try this out tomorrow.  Also, I can add the padding to the `max_pool` layers -- I want to get it integrated with the testing framework, and update all the existing scripts that use `max_pool`.

niketanpansare:
@dusenberrymw  can you please hold off on the padding as I had to add it in to test VGG ? https://github.com/apache/incubator-systemml/pull/158/files#diff-ff98724ff981be4d42d6d1ae623249b6L26

niketanpansare:
Closing this PR in favor of https://github.com/apache/incubator-systemml/pull/422

PR #159, opened 2016-05-16
[SYSTEMML-294][DOC] Print matrix documentation

nakul02:
|  Argument  |  Description  |  Default  |
| ------------- | --------------- | ---------- |
| "rows" | Maximum number of rows to print | 100 |
| "cols"  | Maximum number of columns to print  | 100 |
| "decimal" | Number after decimal to print, 0 padded  | 3 |
| "sparse"  | Whether to print in **sparse** format, which is _Row Index_ _Col Index_ _Value_ | FALSE |
| "sep" | Separator String between elements in a row or between the three columns in sparse Format | " " |
| "linesep" | Separator String between rows or between each line in sparse format | "\n" |

### DIFF
--- a/docs/dml-language-reference.md
+++ b/docs/dml-language-reference.md
@@ -508,7 +508,7 @@ This will result in parser warning, but the program will run to completion. If t
     D = 1;
     foo = function(double A, double B) return (double C) {
         A = 3.0; # value of global A won’t change since it is pass by value
-        
+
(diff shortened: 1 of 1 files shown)


deroneriksson:
Thank you @nakul02 . LGTM.

Thank you for fitting it to the existing document format. At some point, we should consider reformatting the whole DML Language Reference. I really like the tabular format in your comment above.

PR #160, opened 2016-05-16
[SYSTEMML-618] Deep Learning DML Library

dusenberrymw:
This introduces a new deep learning DML library, **SystemML-NN**.  SystemML-NN is a layers-based library written in pure DML that contains layers with a simple forward/backward API for affine, spatial convolution, max-pooling, non-linearities (relu, sigmoid, softmax, etc.), dropout, loss functions, other layers, optimizers, and gradient checks.

**SystemML-NN**:
* Layers:
  * Core:
    * Affine
    * Spatial Convolution
    * Max Pooling
  * Nonlinearities:
    * ReLU
    * Sigmoid
    * Softmax
    * Tanh
  * Loss:
    * Cross-entropy loss
    * L1 loss
    * L2 loss
    * Log ("Logistic") loss
  * Regularization:
    * Dropout
    * L1 reg
    * L2 reg
* Optimizers:
  * Adagrad
  * Adam
  * RMSprop
  * SGD
  * SGD w/ Momentum
  * SGD w/ Nesterov Momentum
* Tests:
  * Gradient Checks

The upstream, experimental version of the codebase lives at [https://github.com/dusenberrymw/systemml-nn](https://github.com/dusenberrymw/systemml-nn), and this addition represents the portion of the codebase that is relatively stable.  In particular, vanilla RNNs & LSTMs are currently still in experimental status upstream, but will be added soon.  The idea is to rapidly work on the upstream repo, and then contribute stable pieces to the main SystemML repo.  For now, we will keep this in the `scripts/staging` directory.

### DIFF
--- a/scripts/staging/SystemML-NN/.gitignore
+++ b/scripts/staging/SystemML-NN/.gitignore
@@ -1,0 +1,2 @@
+scratch_space
+*.ipynb
--- a/scripts/staging/SystemML-NN/README.md
+++ b/scripts/staging/SystemML-NN/README.md
@@ -1,3 +1,22 @@
+<!--
+{% comment %}
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to you under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+{% endcomment %}
+-->
+
 # SystemML-NN
 
 ### A deep learning library for [Apache SystemML](https://github.com/apache/incubator-systemml).
--- a/scripts/staging/SystemML-NN/README.md
+++ b/scripts/staging/SystemML-NN/README.md
@@ -1,0 +1,163 @@
+# SystemML-NN
+
+### A deep learning library for [Apache SystemML](https://github.com/apache/incubator-systemml).
+
+## Examples:
+### Neural net for regression with vanilla SGD:
+```python
+# Imports
+source("nn/layers/affine.dml") as affine
+source("nn/layers/l2_loss.dml") as l2_loss
+source("nn/layers/relu.dml") as relu
+source("nn/optim/sgd.dml") as sgd
+
+# Generate input data
+N = 1024 # num examples
+D = 100 # num features
+t = 1 # num targets
+X = rand(rows=N, cols=D, pdf="normal")
+y = rand(rows=N, cols=t)
+
+# Create 2-layer network:
+## affine1 -> relu1 -> affine2
+M = 64 # number of neurons
+[W1, b1] = affine::init(D, M)
+[W2, b2] = affine::init(M, t)
+
+# Initialize optimizer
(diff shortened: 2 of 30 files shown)


deroneriksson:
:thumbsup::thumbsup: Awesome comments and encapsulation. Very clean code. :smiley:

dusenberrymw:
Thanks, @deroneriksson!

dusenberrymw:
cc @niketanpansare, @mboehm7, @prithvirajsen Interested in any thoughts you may have as well.

niketanpansare:
@dusenberrymw Awesome work. I agree with Deron regarding encapsulation and comments  :+1:

niketanpansare:
@dusenberrymw Any thoughts on the data converters/preprocessing utils for reading existing (image) datasets. For example: Mnist, ImageNet, etc.

mboehm7:
Thanks @dusenberrymw It's probably a good idea to keep this in staging until we have full builtin support for DL. Furthermore, you might want to drop the redundant SystemML-NN path prefix.

mboehm7:
since this library can also serve as an example of how to build dml script libraries, it might be worthwhile to include the vectorized functions for pad_image and im2col too.

dusenberrymw:
@niketanpansare Thanks!  As for the image utils, my preference would be to have those written in Scala/Java & Python as part of `MLContext`.  DML is great for math, but not so great for data wrangling, and especially not with raw bytes ;).  The latter point is important because I don't think we can package MNIST data files directly, and otherwise they have to be processed at the byte level.  For some MNIST examples I'm working on, I just grab them with the TensorFlow utilities.

dusenberrymw:
@mboehm7 Thanks for the feedback.  As for the `SystemML-NN` path prefix, my preference would be to keep it, as it signifies a complete neural net library that just so happens to be distributed in the main SystemML repo.  By keeping it, I'll be able to include documentation and examples directly with the rest of the code.  The goal is to promote others to develop libraries for SystemML.

As for the vectorized functions, that's still on the todo list.  I also want to make sure we improve those parfor loops within functions as well. ;)

mboehm7:
for mnist, you could directly use (1) the infinimnist data generator (http://leon.bottou.org/projects/infimnist), which generates libsvm files, (2) MLlibUtils.loadLibSVMFile to convert to LabeledPoints, and (3) finally our existing converters to convert this to binary block.

dusenberrymw:
@mboehm7 Good idea, and that will work great as part of `MLContext`.

@deroneriksson We should add this as a nice convenience method to `MLContext`.

deroneriksson:
@mboehm7 If tests pass [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/393/), can we merge this today for the 0.10.0 release?

mboehm7:
sure - since we're talking about a library of scripts in staging, this is absolutely fine. Also, there should be no need to run any tests here. Thanks for merging.

dusenberrymw:
Great! I want to check one last thing, and then I'll merge this today when I have a moment.

deroneriksson:
Thank you @mboehm7 and @dusenberrymw !

deroneriksson:
@mboehm7 Should the DML scripts in staging be included in the build artifacts or excluded?

deroneriksson:
(BTW, they will always be included in the -src.tar.gz and -src.zip artifacts.)

mboehm7:
personally, I would not include the staging scripts into the jar as 'staging' refers to experimental status. Having them in the tar is fine though.

deroneriksson:
OK I will  exclude them except from the src.tar.gz and src.zip. Thank you.

dusenberrymw:
Merged!

deroneriksson:
Congrats! This is a great addition to SystemML.

dusenberrymw:
Thanks, @deroneriksson!

PR #162, opened 2016-05-17
[SYSTEMML-656] Fix scalar input boolean addition test

deroneriksson:
Uncomment scalar input boolean addition test and set result to be "2.0" (R behavior) rather than "TRUE".


mboehm7:
LGTM - thanks @deroneriksson

deroneriksson:
Thank you for reviewing. I'll merge.

PR #166, opened 2016-05-18
[SYSTEMML-639] Can't use semicolon after user-defined function in DML

nishi-t:
This patch addressed the issue [SYTEMML-639](https://issues.apache.org/jira/browse/SYSTEMML-639). It's small fix to recognize semicolons after user-defined function in DML. I made sure that the following example is parsed successfully.
```
x=1;print(x);
printFoo=function(int y){print(y);};z=printFoo(x);
```
Please review.


deroneriksson:
All tests passed [here](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/392/).

LGTM. Thank you @nishi-t !

PR #167, opened 2016-05-19
[SYSTEMML-688] Update dml file packaging in artifacts

deroneriksson:
Move dml to scripts dir for main, in-memory, sources, and standalone jars.
Include algorithms, datagen, and utils directories.
Exclude obsolete algorithms, perftest, and staging.
Update distrib.xml and standalone.xml to reflect similar packaging.
Update ScriptUtils.scala and LogisticRegression.scala for these changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,15 +73,24 @@
 	</properties>
 
 	<build>
-	
-		<resources>
-      		<resource>
-      			<directory>scripts/algorithms</directory>
-      			<includes>
-		          <include>**/*.dml</include>
-		        </includes>
-      		</resource>
-      	</resources>
+
+	<!-- Adds scripts to main jar, in-memory jar, sources jar, and standalone jar -->
+	<resources>
+		<resource>
+			<directory>scripts</directory>
+			<excludes>
+				<exclude>algorithms/obsolete/*</exclude>
+				<exclude>algorithms/obsolete</exclude>
+				<exclude>perftest/*</exclude>
+				<exclude>perftest</exclude>
+				<exclude>staging/**/*</exclude>
+				<exclude>staging</exclude>
+				<!-- <exclude>*.sh</exclude> --> <!-- applies to sparkDML.sh -->
+			</excludes>
+			<targetPath>scripts</targetPath>
+		</resource>
+	</resources>
+
 		<plugins>
 
 			<plugin>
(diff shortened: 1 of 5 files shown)


deroneriksson:
My Eclipse currently isn't set up for Scala development. Would it be possible for someone to verify the ScriptUtils and LogisticRegression updates look good?

lresende:
Looks like we are changing to explicitly choose the scripts to put on the packages, to now put all except for the ones on specific folders. I am ok with that, but what to double check if all the scripts are ready to be first class scripts on the release package, otherwise we could move them to other folders as I do like your solution better.

@mboehm7 @niketanpansare  thoughts on the comment above ?

deroneriksson:
@lresende Thanks for checking on this. The 'first-class' algorithms go in the primary scripts/algorithms directory. The 'second-class' scripts go in scripts/staging, which is excluded from all the artifacts (except in the src artifacts). By including scripts/algorithms/* and excluding scripts/staging/*, all first-class scripts are automatically included, and all second-class scripts are excluded. When an algorithm script has graduated, it is moved from scripts/staging/ to scripts/algorithms/ and will automatically be placed in each build artifact. A benefit of this PR is that we no longer need to manually add each first-class script to the assemblies, which can be error-prone.

@mboehm7 @niketanpansare Let me know if this is correct or if anything needs to be modified.

lresende:
Thanks @deroneriksson , LGTM

niketanpansare:
Thanks Deron, LGTM :)

mboehm7:
@deroneriksson I absolutely agree. Just one thing: If I'm not mistaken it needs to be "/scripts/..." instead of "scripts/..." when loaded from the resource stream.

deroneriksson:
@mboehm7 I'll check this when I get to SF (getResourceAsStream with and without the initial "/")

deroneriksson:
@mboehm7 You are correct WRT requiring the "/" for getResourceAsStream. I'll update to prepend the "/".

deroneriksson:
@gweidner Could you check and see if this is working correctly from Scala? Thank you.

deroneriksson:
@gweidner Thank you for the help testing. Could you test the latest commit? Thanks!

gweidner:
Verified with test program that dml scripts read successfully from jar.  LGTM.
Thanks @deroneriksson !

deroneriksson:
Thanks for all the help @gweidner ! I'll merge the PR.

gweidner:
Sounds good!  Note Windows environment used.

PR #171, opened 2016-06-02
[SYSTEMML-562] Frame Left Indexing (Not for 0.10 release)

asurve:


mboehm7:
thanks @asurve - most of the following comments are due to the recent modifications of the frame integration:

1) Input/output infos: Could we consistently use binaryblock throughout the compiler, tests, and runtime for both matrices and frames (e.g., remove from Connection, DAG, CacheableData, etc). We can map it to BinaryBlockFrame input/output formats whenever required during runtime.

2) Checkpoint: Meanwhile we support spark checkpoints over frames too. Hence, there is no need to disable this rewrite for frames any longer.

3) Frame object: Why is there a need to have 'num rows per block' in the frame object - we could simply use a global optimizer constant whenever needed because this does not describe the actual data.

4) Frame block: Furthermore, why is there a need for leftindexing changes on frame block. The additional indexing operations for multiple output blocks are ok, but the rest should stay in tact, right? Also, please remove any reference to longwritable from frame block (to be replaced with long). Similarly, why do we need a merge over unaligned blocks - any operation that wants to merge blocks should align them first.

5) Frame read/write rdds: Please stick to the current version of reading/writing frame rdds; I did on purpose not implement guarded collect as the memory check needs further investigation first.

6) Execution context frame primitives: The method getFrameRDDHandleForVariable is redundant, given the existence of getRDDHandleForVariable. Also please reuse the existing toFrameBlock (note that there is no notion of bclen for frames).

7) Generalization broadcast matrix object: Could we remove the new matrix broadcast object and frame broadcast object? The valid check should be possible in a generic manner too, right? Also, please double check if we could generalize the partitioned matrix and frame blocks.

8) RDDAggregateUtils: Please revert the signature change of mergeByFrameKey and change the usage in RDDConverterUtils. If there is a key type mismatch, let's call LongWritableToLong before merge.

9) Write meta data file: Please consolidate this new method with the existing metadata write (which is already used for frames too) to avoid code duplication.

10) UtilFunctions: It would be good to move the compare functionality with tolerance out of util functions and into test utils to clearly separate between performance relevant code and tests. Of course, you could still reuse whenever existing functionality directly matches the needs of your tests.

11) Code style: Please update the exception messages of refactored code (e.g., indexing instructions), delete commented code, remove TODOs for MR (as we did not even start this yet), merge multiple static initializers, and make static constants final.

asurve:
Addressed all except following comments.
  4. Merge alignment: This has not been done for Map case, where slices are created for LHS block from multiple RHS broadcast blocks and merged together. If I shift these slices then LeftIndexing operation to be changed. If performance is an issue for merging slices from broadcast object then I can look into it.
  7. Generalization: I will investigate if generalization of Broadcast/Partitioned objects/block can be easily done.

asurve:
Addressed Generalization comment (number 7).

asurve:
All recent comments except #7 has been addressed, which I will start looking into.

asurve:
Comment #7, has been addressed (Tyrpesafe wrt generalization)

mboehm7:
1) Cache block extensions: First, estimateSizeInMemory and estimateSizeOnDisk are redundant to the existing getInMemorySize and getExactSerializedSize. Second, the interface of the overloaded sliceOperations w/ PartitionedBlock needs cleanup (return type T, both block and pBlock as inputs). I would actually be in favor of removing the second slide altogether (and keep the functionality in PartitionedBlock). Third, please also use T for CacheBlock paramters to these functions to avoid unnecessary modifications of MatrixBlock/FrameBlock. Fourth, the methods for getNewInstance are odd as they need to be called on an existing instance (if they are kept, also change the return type to T). Can't we create a simple method UtilFunctions that does the same via java reflection? 

2) CacheableData extensions: Please change members/APIs from BroadcastObject<CacheBlock> to BroadcastObject<T> to make it type-safe.

3) FrameObject: Replace the change in FrameObject.readBlobFromHDFS with current head (you need to rebase to the latest code).

4) Codestyle: First, remove unnecessary casts, e.g., after getBlock (e.g., ParameterizedBuiltinSPInstruction). The same applies to any casts like (T[]) (e.g., PartitionedBlock) - this is usually an indicator of a bad API design. Second, please generally try to use short but meaningful variable names - something like 'bDoNotUpdateZeroData' is just unnecessarily convoluted.

5) Input/OutputInfos: see 5 above (e.g., FrameReaderFactory/FrameWriterFactory/InputInfo/OutputInfo/MapReduceTool). 

6) MR Compatibility: Remove the import of scala.Tuple2. You need to be careful not to introduce Spark dependencies into the core block runtime otherwise this would break on MR clusters without Spark libs.

7) Testsuite: First, the R input file handling seems incorrect (name + "Csv") - either use a proper file extension or the same name. Second, remove the BinaryBlockFrame input/output infos from ALL tests (FrameConverterTest/FrameMatrixCasting, see 5). Third, remove commented code from your testclasses - this is more confusing than it helps anybody.

PR #173, opened 2016-06-06
[SYSTEMML-756] Adding a rewrite for aggregating the product of a binary and a scalar 

tgamal:
I added a rewrite for the pattern:  sum(lamda*X) -> lamda*sum(X).


akchinSTC:

mboehm7:
thanks @tgamal - this already looks pretty good. I would just recommend a couple of cleanups: 

1) Rewrite: Could you please apply the rewrite only if the number of parents of (lambda*X) is 1 i.e., the sum is the only consumer? This would ensure it is always beneficial to apply the rewrite.

2) Rewrite code style: First, please remove the unnecessary casts to BinaryOp for data type checks and input references. Second, could we rename the rewrite to 'pushdownSumBinaryMult' and position it after pushdownUnaryAggTransposeOperation to group these similar rewrites together?

3) Test scripts: It would be good to change the testscripts to something like "lambda=sum(X)" to test the rewrite with variables not just literals; furthermore, this would also test an interesting case of common subexpression elimination after rewrites (i.e., sum(X) should be executed just once). In addition, the R script should get only 'expectedDir' as an argument, which would allow you to change the scripts from args[2] to args[1] too.

4) Test scripts copyright: Currently, both R scripts are missing the ASF copyright header - please add this.

tgamal:
Thanks Matthias for the feedback. I applied and committed the suggested changes

mboehm7:
thanks @tgamal - could you please also add the copyright header to RewritePushdownSumBinaryMult.R?

tgamal:
Done !

mboehm7:
LGTM; thanks @tgamal

PR #175, opened 2016-06-10
Include guava in standalone distributions

deroneriksson:
Guava needs to be included in the standalone distributions (tar.gz, zip, and uberjar) in order for them to work. Thank you for finding this @dusenberrymw

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -806,6 +806,14 @@
 			<scope>provided</scope>
 		</dependency>
 
+		<!-- Adding Gauva version 14.0.1 to workaround conflict between spark and hadoop dependency -->
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>14.0.1</version>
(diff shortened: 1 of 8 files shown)


deroneriksson:
@dusenberrymw Could you see if the latest commit fixes the issue you ran into?

niketanpansare:
@deroneriksson This PR fixes the guava issue that I was facing on standalone jar. Thanks :+1:

deroneriksson:
Thank you for checking @niketanpansare. I appreciate it!

@dusenberrymw I think this also resolve the protobuf-java jar issue. When you have time, can you let me know if your mnist example with the standalone jar runs successfully for you?

dusenberrymw:
@deroneriksson Thanks, that fixes the issue!

deroneriksson:
Note: I did not commit the NOTICE updates since these are not needed. For more information, see [PR180](https://github.com/apache/incubator-systemml/pull/180).

PR #177, opened 2016-06-15
Update Beginner's Guide for toString, 0-based PyDML, and elif

deroneriksson:
Add toString() function to Beginner's Guide.
Update guide for PyDML 0-based indexing.
Add else-if (elif) to examples since it now exists in PyDML.

Preview [here](http://deroneriksson.github.io/incubator-systemml/beginners-guide-to-dml-and-pydml.html).


dusenberrymw:
LGTM.

PR #179, opened 2016-06-17
[SYSTEMML-766] Adding a rewrite for Axpy (matrix-scalar product and adds the result to a matrix)

tgamal:
Rewrite for the following pattern to iterate on X and Y simultaneously and avoid dense intermediates
  - pattern: X + lamda*Y -> X +* lambda Y


akchinSTC:

mboehm7:
thanks @tgamal - this is a great feature as the addressed pattern is indeed very common. Couple of comments: 

1) Simplify sparsity estimate to min(sp1+sp2, 1)

2) Please rename the rewrite to 'fuse binary operation chain'. Also use something like 'tmp' as the name for the new ternary operator, and add the linenumbers to the debug output.

3) Fix the lop comments of matrix/scalar operands (exchange the position scalar and second matrix)

4) Make the ValueFunctionWithConstant class member private

5) Test: Rename the test to RewriteFuseBinaryOpChainTest, incl the used dml/R scripts. Further, please decomment the actual testcases, add this class to the testsuite org.apache.sysml.test.integration.functions.misc.ZPackageSuite, remove all sysout prints from the test class, and remove the class comment copied from a different test. 

6) Why did you have to change RewriteSimplifyRowColSumMVMultTest.java?

7) Add the missing copyright in RewriteAxpy.R and RewriteAxmy.R

dusenberrymw:
add to whitelist.

tgamal:
Thanks Matthias for the feedback, I addressed the comments you specified. The only comment I am not sure I understood was "Also use something like 'tmp' as the name for the new ternary operator". Could you please elaborate more on what you meant by that?

mboehm7:
thanks @tgamal - I really just meant the "name" parameter of the newly constructed hop during rewrites - having something like short that indicates its automatic construction would simplify debugging.

tgamal:
thanks @mboehm7 - I updated the name. Kindly let me know if you suggest further changes

mboehm7:
thanks @tgamal - there are just two minor cleanups remaining:

1) Please remove the "oldFlag = OptimizerUtils.ALLOW_ALGEBRAIC_SIMPLIFICATION" and one of the two "OptimizerUtils.ALLOW_ALGEBRAIC_SIMPLIFICATION = rewrites" assignments in testRewriteAxpy as they are redundant.

2) Please add an additional check to your rewrite. We should NOT apply the rewrite if DMLScript.rtplatform=RUNTIME_PLAN.Hadoop or DMLScript.rtplatform=RUNTIME_PLAN.Hybrid as the physical operator is not implemented for hadoop yet. The check is sufficient - once this feature is in, I will add the MR support but we cannot break the testsuite.

tgamal:
Done. thanks @mboehm7.

mboehm7:
LGTM - thanks @tgamal

PR #180, opened 2016-06-20
[SYSTEMML-757] Removed abego, ANTLR and SLF4J from NOTICE files

gweidner:
Permissively-Licensed Dependencies under BSD and MIT/X11 licenses do not need to be included in NOTICE file as per instructions at http://www.apache.org/dev/licensing-howto.html#permissive-deps.  This applies to abego, ANTRL and SLF4J which are referenced in LICENSE file.

### DIFF
--- a/src/assembly/distrib/NOTICE
+++ b/src/assembly/distrib/NOTICE
@@ -3,12 +3,3 @@ Copyright [2015-2016] The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
-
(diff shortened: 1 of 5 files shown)


gweidner:
This was in response to comment during IPMC vote for release 0.10.0 requesting it be fixed in next release.  NOTICE files are now identical but left versions in distrib, standalone, etc. folders since referenced separately in corresponding build xml files in case diverge in future.

deroneriksson:
Thank you @gweidner. LGTM.

PR #182, opened 2016-06-24
[SYSTEMML-771] Fix warnings in CsplineCG.dml and CsplineDS.dml

gweidner:
Fixed 'print statement can only print scalars' warnings by replacing 1x1 wrapper matrix with actual scalar variable for algorithm and corresponding dml and pydml test scripts.

### DIFF
--- a/scripts/algorithms/CsplineCG.dml
+++ b/scripts/algorithms/CsplineCG.dml
@@ -102,7 +102,7 @@ y_mat = matrix(y, 1, 1)
 if (fileO != " ") {
   write (y_mat, fileO);
 } else {
-  print(y_mat)
+  print(y)
 }
 
 print ("END CUBIC SPLINE REGRESSION SCRIPT");
--- a/scripts/algorithms/CsplineDS.dml
+++ b/scripts/algorithms/CsplineDS.dml
@@ -82,7 +82,7 @@ y_mat = matrix(y, 1, 1)
 if (fileO != " ") {
    write (y_mat, fileO);
  } else {
-   print(y_mat)
+   print(y)
 }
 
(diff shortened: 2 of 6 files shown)


gweidner:
Note chose not to use new toString or as.scalar since the matrix simply wrapped a single scalar value.  Another option was to delete the print which was invalid for matrix if executed.

deroneriksson:
LGTM. Thank you @gweidner for fixing the warnings in the main algorithms. This is a big win for new users since we don't want to scare them away with unnecessary warnings.

By any chance, if you are running through the algorithms and testing them, would you mind saving (1) the data generation invocation, and (2) the algorithm invocation (such as $ ./runStandaloneSystemML.sh scripts/algorithms/Univar-Stats.dml -nvargs X=data/haberman.data TYPES=data/types.csv STATS=data/univarOut.mtx)? I would really like our documentation to include examples that a user can copy/paste from the documentation and run ALL our algorithms without having to do any additional work.

gweidner:
Thank you for the review and suggestion!  That's good idea so I'll capture that information and also compare with Algorithms Reference.  For these 2 main algorithms, I ran from unit test cases since the test scripts were identical to scripts/algorithms.

deroneriksson:
The Algorithms Reference is quite verbose. My hope is to create a small single-paragraph summary of each main algorithm and have an accompanying invocation line that can be copy/pasted and it will run the algorithm with absolutely no tweaking required by the user.

gweidner:
That's even better and would definitely be helpful to end users.  I'll take that approach and provide information as it will be good exercise for me to get familiar with other algorithms too :)

PR #183, opened 2016-06-27
[SYSTEMML-562]Frame Right Indexing

asurve:


mboehm7:
thanks @asurve - just two required fixes; once they are done you could directly push this change.

1) Partitioning-preserving: please change the condition to entire columns only (but arbitrary many, e.g., X[,3:1328] is still partitioning preserving); for rows it would only apply to the first row but that's not worth including.

2) Simplifications Block Filter: The function IsFrameBlockInRange only requires [rl,ru], please remove the cl/cu parameters. Also, remove all block sizes (they are not required and might be even incorrect in case of variable length blocks). Let's remove bclen from UtilFunctions.isInFrameBlockRange and pass frame.getNumRows() as brlen.

PR #186, opened 2016-07-06
[SYSTEMML-774] sparse matrix concurrency fix

nakul02:
Added checks threadSafe implementation of MatrixBlock in operations.
@mboehm7 - could you please take a look at these additions?

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/io/MatrixReaderFactory.java
+++ b/src/main/java/org/apache/sysml/runtime/io/MatrixReaderFactory.java
@@ -24,6 +24,8 @@ import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.runtime.DMLRuntimeException;
 import org.apache.sysml.runtime.matrix.data.CSVFileFormatProperties;
 import org.apache.sysml.runtime.matrix.data.InputInfo;
+import org.apache.sysml.runtime.matrix.data.MatrixBlock;
+import org.apache.sysml.runtime.matrix.data.SparseBlock;
 
 /**
  * 
@@ -44,14 +46,14 @@ public class MatrixReaderFactory
 		
 		if( iinfo == InputInfo.TextCellInputInfo || iinfo == InputInfo.MatrixMarketInputInfo )
 		{
-			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_TEXTFORMATS) )
+			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_TEXTFORMATS) && MatrixBlock.DEFAULT_SPARSEBLOCK == SparseBlock.Type.MCSR )
 				reader = new ReaderTextCellParallel( iinfo );
 			else
 				reader = new ReaderTextCell( iinfo );	
 		}
 		else if( iinfo == InputInfo.CSVInputInfo )
 		{
-			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_TEXTFORMATS) )
+			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_TEXTFORMATS) && MatrixBlock.DEFAULT_SPARSEBLOCK == SparseBlock.Type.MCSR )
 				reader = new ReaderTextCSVParallel(new CSVFileFormatProperties());
 			else
 				reader = new ReaderTextCSV(new CSVFileFormatProperties());
@@ -59,7 +61,7 @@ public class MatrixReaderFactory
 		else if( iinfo == InputInfo.BinaryCellInputInfo ) 
 			reader = new ReaderBinaryCell();
 		else if( iinfo == InputInfo.BinaryBlockInputInfo ) {
-			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_BINARYFORMATS) )
+			if( ConfigurationManager.getCompilerConfigFlag(ConfigType.PARALLEL_CP_READ_BINARYFORMATS) && MatrixBlock.DEFAULT_SPARSEBLOCK == SparseBlock.Type.MCSR )
 				reader = new ReaderBinaryBlockParallel( false );
 			else
 				reader = new ReaderBinaryBlock( false );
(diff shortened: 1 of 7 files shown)


nakul02:
Launched [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/182/) build.

mboehm7:
thanks @nakul02 - couple of comments:

1) Matrix readers: is there a reason why you only modified the parallel binary block reader but not csv, textcell, and mm? I would  recommend to simply modify the MatrixReaderFactory and select the sequential readers if DEFAULT_SPARSEBLOCK is not MCSR. Yes, it might be too pessimistic if the data is dense, but we anyway use MCSR to overcome CSR performance issues on incremental construction (e.g., read of textcell with unordered input cells). 

2) LibMatrixDNN: This is not related to the PR at hand, but I would also call the sequential version for numThreads=1 because there is no point in creating the threadpool for single-threaded execution. Furthermore, right now any exceptions within those  tasks get discarded - please get the set of futures from invokeall and explicitly call get (which will propagate any occurred exceptions). @niketanpansare 

3) LibMatrixDatagen/LibMatrixMult: Please fold the thread-safe condition into the main fallback to sequential execution in order to simplify potential code paths.

4) Matrix Block: We need to make this isThreadSafe method robust against non-allocated sparse blocks (sparseBlock==null).

5) Tests: It would be actually a great idea to use CSR as default sparse block type and run our entire testsuite to reveal any open issues. We've never done that so far. Anyway, if we find anything, we don't need to delay this PR but can solve it separately.

nakul02:
thanks for the comments @mboehm7,
Fixed 1 & 4.
5) I can send in a test on Jenkins once this PR is in.

2) LibMatrixDNN - Correct me if I am wrong, but I think sequential version is being called for when numThreads is 1. Also the Exceptions are being propagated up. 

3) LibMatrixDataGen - fixed, LibMatrixMult - The check is done when the ``output.sparse`` flag has been set. The ``isThreadSafe()`` check cannot be folded into the main fallback for sequential execution in those cases.

Launched [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/186/) Build.

mboehm7:
thanks @nakul02 

ad 2) good - then, I must have overlooked it.
ad 3) well, it would be good if we could also do it for matrix mult; the conditions for sparse output are very simple, meaning we could easily fold them into the fall back condition instead of creating the instance first and checking the instance. If there is a good reason fine, otherwise I would really prefer to have it as simple as possible, especially for something like matrix mult.

nakul02:
@mboehm7 
3) Fixed
I had this question about LibMatrixMult, related to this PR:

In [LibMatrixMult.matrixMultPermute](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L488), the output is specifically being set to dense. Thereafter, in [MatrixMultPermuteTask.call](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L4294), a check is being made for m2 or ret1 being sparse. If the code in [matrixMultPermuteSparse](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L2104) is executed, it would crash because of [this](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L2108) line, since before this call, ret1 is already set to dense. `matrixMultPermuteDenseSparse` would never be called because re1 was set to dense.
I understand that the code structure in [MatrixMultPermuteTask.call](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L4294) mimcs the code structure in the [serial version](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L445). Is there any other reason for the code to look like that?
I ask because, in [matrixMultPermuteDense](https://github.com/nakul02/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixMult.java#L2013), ret2 is begin allocated and being operated on. As it stands right now, this will not crash because it allocates a Dense block, which is thread safe, but it is something to be aware of for future code modifications.

Launched [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/187/) build.

mboehm7:
that's a great catch @nakul02 - we should fix the sparsity condition in matrixMultPermute(..., int k). 

The reason why it never showed up so far is that we primarily compile these pmm operators in MR/Spark to avoid materializing diag(v) and there pmm is always executed as single-threaded block operations. However, there are potential corner cases where after recompile pmm is pushed into CP and hence we should fix that asap. Thanks.

nakul02:
@mboehm7 - is this PR good to go?

mboehm7:
ad 3) LibMatrixAgg and LibMatrixMult: Did you update the PR? There are still several instances in LibMatrixAgg and LibMatrixMult with multiple fallbacks to sequential execution - let's please consolidate them. For simplicity don't add these checks whenever it is always dense. Also, I would recommend some static check in MatrixBlock like isThreadSafe(boolean sparse) that would allow you to directly fold these checks into the existing fallback conditions.

6) The License header of LibMatrixDNN has been corrupted. Also the formatting of some conditions have been unnecessarily changed (e.g., ReaderBinaryBlockParallel) - please simply revert with head.

7) MatrixBlock: Please delete DEFAULT_SPARSEBLOCK_THREADSAFE as it is a potential source of inconsistencies and check instead SparkBlock.TypeMCSR. Thanks.

nakul02:
3) Fixed, I think having the isThreadSafe check even when dense is alright. If later for performance reasons, we wanted to make dense blocks thread unsafe, these checks would help.

6) Fixed

7) IMHO, DEFAULT_SPARSEBLOCK_THREADSAFE is the better approach than to always check against MCSR. 
If someone were to change the DEFAULT_SPARSEBLOCK type, he would correspondingly change the DEFAULT_SPARSEBLOCK_THREADSAFE since they are very closely located.
Also, later if someone were to make COO or CSR threadsafe, or introduce another SparseBlock implementation which is thread safe and change the default to that, a check against MCSR would be useless.

Also, a little comment - on examining LibMatrixMult, there seem be many "Magic Numbers".
blocks sizes and cache line sizes and such are hard-coded in guard statements. Maybe as a later PR, these could be made class level final static integers?

mboehm7:
8) LibMatrixDatagen: Please, fix the check in LibMatrixDataGen. This check needs to happen either after reset(..., lsparse) or much better directly on lsparse and the static isThreadSafe.

Remark) I still don't see a good usecase for all this unnecessary complexity (indirections via redundant constants, separated sparse=false + check for isThreadSafe(sparse) + allocation). While CSR is great in terms of memory efficiency for read-only data, it has severe performance problems on incremental construction (read and any operations that appends cells to the output). Accordingly, we will not change the default sparse representation.

nakul02:
Started [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/189/) build.
@mboehm7 - Does this look good

mboehm7:
LGTM - thanks @nakul02 (there are still some stylistic issues but I'll address them on the next modification)

nakul02:
Tests passed. 
Thanks @mboehm7.
@deroneriksson - could you please merge this in?

PR #190, opened 2016-07-08
[SYSTEMML-562] Frame Append operation

asurve:


mboehm7:
thanks @asurve - couple of comments:

1) Fix frame mappend rbind: This function would create incorrect results for multiple rhs blocks (rhs blocks never emitted). Also, remove the unncessary code paths (e.g., MapSideAppendFunction else)

2) Fix frame mappend comments: The comment on partitioning preserving operations is misleading, the number of blocks are just indirectly relevant - the actual question is if keys change or not

3) Simplification frame mappend: Since rbind and cbind have different functions, let's simplify the code and remove any cbind checks from these functions. You might also want to rename them accordingly.

4) Fix frame rappend rbind: This instruction only supports cbind (would create incorrect results on rbind). Further, in case of misaligment (variable sized lhs/rhs blocks) it would also create incorrect results for cbind.

5) For frame rbind, we should always compile an rappend (union/and remapping of rhs blocks). Or do you have a good usecase where we want to still compile a broadcast based plan?

6) Tests: Please, pick fixed sparsity values for both lhs and rhs - otherwise the tests become non-deterministic which created problems in the past. Also the formatting of commonAppendTest needs fixing (mix of space/tab indentation)

7) Test scripts: There are only tests for cbind - please add the tests for rbind as well (which will help to fix the issues of incorrect results).

mboehm7:
thanks @asurve - this looks already much better - just some minor cleanups.

8) moveData in CacheableData: Please simplify this functionality (the writeBlobFromRDDToHDFS should already be covered by export).

9) VariableCPInstruction: Also simply the unnecessary casts and branches when calling moveData (simply getCachableData and dirctly call moveData on it).

10) FrameAppendMSPInstruction.processInstruction: Modify the runtime exception to use opcodes; otherwise this exception message is incomprehensible for users.

11) MapSideAppendPartitionFunction: Please, simplify the implementation, i.e., remove the _cbind member variable and its use.
 
NOTE: In the future, please rename the existing instruction to the new matrix instruction and create a new base class. By deleting the code from the existing instruction and reinserting it into a new matrix instruction, you detach the file history. I'm not asking you to modify this in the current PR, just something to keep in mind.

asurve:
Review comments addressed, waiting for build to get completed.

mboehm7:
LGTM, but I would appreciate if you could fix (10) as well. Thanks @asurve 

ad 10) You did change the message to "Opcode " + instOpcode + " not supported/required with FrameAppendM, instead use FrameAppendR for this opcode processing." which would resolve to "Opcode mappend not supported/required with FrameAppendM, instead use FrameAppendR for this opcode processing." - please change to "Append type rbind not supported for frame mappend, instead use rappend".

asurve:
done.

PR #193, opened 2016-07-17
Correct broken links.

ahmaurya:


akchinSTC:

nakul02:
LGTM

deroneriksson:
Welcome to Apache SystemML, @ahmaurya! Thank you for working on this.  Thank you @nakul02 for reviewing.

Unfortunately, I don't think I will be able to merge this. Changing the extensions from .html to .md will allow a user to browse through the raw md files on the GitHub project site (https://github.com/apache/incubator-systemml/blob/master/docs/algorithms-reference.md). However, it will break the links on the generated GitHub documentation site (http://apache.github.io/incubator-systemml/algorithms-reference.html) and any locally generated documentation site (docs/_site/algorithms_reference.html). Since jekyll uses the raw md files to generate a static html documentation site, we need to keep the extensions as html.

ahmaurya:
Since Jekyll generates HTML files, it should be its job to link them properly as well, perhaps using a plugin as advised here: https://stackoverflow.com/questions/25826770/jekyll-using-links-to-internal-markdown-files Anyway, this issue is not important to consume your time and mine. I am closing this thread out. Thanks.

deroneriksson:
Thank you @ahmaurya. I think it is a bit frustrating that this issue is not handled automatically by Jekyll and GitHub. If you would like to do a contribution that allows links to work both in the raw project view and the generated documentation site, I would be very happy to merge. Note that a plugin requires updating of the licenses in the project artifacts that would contain the plugin.

PR #195, opened 2016-07-21
[SYSTEMML-560] Frame DataFrame converters

asurve:
SystemML Frame to Spark DataFrame converters and vice versa.


mboehm7:
thanks @asurve - let's do just some additional cleanups: 

1) Avoid shuffle: Please remove the sorting from binaryBlockToDataFrame. A data frame is just a bag of rows, which allows us to do a simple map-side conversion without shuffle.

2) Efficiency row deep copy: Use a RowFactory.create(tmp.clone) instead of array allocation plus loop-based copy. This applies to BinaryBlockToDataFrameFunction and RowGenerator. Also, please double check that this is really required, i.e., have a look at RowFactory and see if the passed arrays are not already copied under the covers.

3) Code style: Globally change pre-increments ++i to post-increments i++ for consistency with the rest of the code base!

NOTE: I'll deliver some changes to the csvToBinaryBlock converter tomorrow - so please merge carefully.

asurve:
RowFactory keeps references to array, so need to copy. I cloned it now instead of copy per comments. Addressed all comments.

mboehm7:
LGTM - thanks @asurve for the cleanup, including the other code style issues in related files.

asurve:
done.

PR #196, opened 2016-07-23
[SYSTEMML-701][[SYSTEMML-702] sparse matrix multiplication for gpu

nakul02:
Here is a preliminary support of Sparse matrix multiplication on GPU.
With this PR, sparse * sparse matrix multiplication support on the GPU is added.
sparse * dense support will be added after discussion here, if at all.

Here are some performance benchmarks I ran on a very beefy node with NVIDIA Tesla K80 GPUs. 
The benchmark compares a matrix multiply in CP with the same matrix multiply on GPU.
``-gpu force=true`` option was used on the command line.

Only one GPU was used.
The test machine had 24 cores (E5-2620), 128GB.
CUDA version - 7.5
OS - RHEL6

In the graphs, the blue line is for the GPU. Please note that the graphs are in log-scale on the y-axis.
The x-axis represents the values of M,N & K where a block of size (M*K) is multiplied with a block of size (K*N). Sparsity was set to ``0.3`` for both matrices.
For missing data points, 0 was inserted.
There were crashes with higher sizes; investigation on these is pending.

matmult
execution
elapse

Also note that the wallclock/elapsed time for the GPU run had a large compilation time (about 5 seconds). This is to be investigated also.

[Here](https://gist.github.com/nakul02/3a7f76c27c8581d3aa51cbeac435ab34) are SVG Versions of the same graphs. (Github won't let me insert them on this page).

@mboehm7, @niketanpansare, @bertholdreinwald - Comments, suggestions?


mboehm7:
thanks for sharing the numbers @nakul02 

1) Could you please provide the numbers for both CPU_CSR and CPU_MCSR, incl sparse matrix-vector products as well? Also do these numbers show the time after JIT (how many warmup runs, how many repetitions)?
2) It would be good to separate the experiments into two scenarios: transfer+compute as well as transfer + repeated compute on device (especially for the memory-bandwidth-bound sparse matrix-vector multiply).
3) Regarding the actual pull-request: the introduced getValueArray() and getColumnIndexArray() in the individual sparse block implementations are redundant to the existing values() and indexes().

nakul02:
@mboehm7 -
My experiments were run like so:
A simple dml file was written to do sparse matrix multiplication
sparse.dml -
```
m=$1
k=$2
n=$3
X = rand(rows=m, cols=k, min=0, max=4, pdf="uniform", sparsity=0.3)
Y = rand(rows=k, cols=n, min=0, max=4, pdf="uniform", sparsity=0.3)
if (1 == 1) {
    S <- X %*% Y
}
print (as.scalar(S[1,1]))
```

A simple bash script was used to run tests for various values of M=N & K
```
#!/bin/bash

M="100 200 400 800 1600 3200 6400 12800 25600 51200"
K="100 200 400 800 1600 3200 6400 12800 25600 51200"


for i in $M
    do
        for j in $K
            do
                echo "M = $i, K = $j, N = $i" >> OUT
                bin/systemml sparse.dml -stats -args $i $j $i 2>> OUT
                bin/systemml sparse.dml -stats -gpu force=true -args $i $j $i 2>> OUT
            done
    done
```

I then process the `OUT` file with a python script and generate the graphs using matplotlib.

1) The numbers in the graph do not have any repetition runs and do not have warmup runs.
But I can modify the sparse.dml file to put the matrix multiplication in a loop. 
Also, I can do this for a sparse matrix vector product (by setting N=1).
Will this be enough?

2) With this particular experiment, what exactly are we trying to answer?
In this PR, there is instrumentation to measure the transfer of data from device to host and vice versa. Will that suffice?

3) The` values(int r)` and `indexes(int r)` in `SparseBlockCSR` and SparseBlockCOO` are supposed to return the column indices/values of row r, but they don't. Instead they return the entire array itself. The argument `r` is ignored.
Assuming this is fixed in a later version, I though introducing a dedicated getter method to get the internal members would be safer.

mboehm7:
1) ok, then I would strongly recommend to run multiple iterations (depending on datasize, at least 20) as the difference between initial performance and after JIT is roughly 2-3x for dense, 5x for sparse. Usually, it's a good idea to instrument the core block multiply and run it with gc:verbose to exclude buffer pool times and runs were garbage collection interfered with the timing.

2) well, right now we're still in a stage were we first need to really understand the basic characteristics. Personally, I would be interested in a couple of things: How close to peak PCIe transfer speeds to we get? How close to peak GPU memory bandwidth do we get with a sparse/dense matrix-vector multiply. Where is the break even point between compute/memory bandwidth for common operations like XtX and XY (with increasing number of columns for Y). 

3) The methods values(r), indexes(r) return the full arrays in order to avoid copy - together with pos(r) this allowed us to implement all sparse operations efficiently against CSR and MCSR at the same time. However, making it more explicitly is fine; I would recommend to follow the same pattern and call them values() and indexes().

nakul02:
Some performance numbers after running tests:
(All run on single node, Graphs are not log-scale like before)

# DENSE MATRIX-MATRIX MULTIPLY

gc
gpu_overhead
matmult


[SVGs](https://gist.github.com/nakul02/d6b893dd8d51f6d234a5c67e9ffc140c)

nakul02:
# Dense Matrix-Matrix Multiply

gc
gpu_overhead
matmult

[SVGs](https://gist.github.com/nakul02/80128edd357f52e74d85fb6488194011)

nakul02:
# Sparse Matrix Vector Multiply

gc
gpu_overhead
matmult

[SVGs](https://gist.github.com/nakul02/64829118ac6e5b1975809df690145372)

nakul02:
# Sparse Matrix Matrix Multiply

gc
gpu_overhead
matmult

[SVGs](https://gist.github.com/nakul02/ded8f1df1b941cbc0f711614dde30ce9)

nakul02:
For the sparse matrix tests, this was the DML script:

```r
m=$1
k=$2
n=$3

I=$4
X = rand(rows=m, cols=k, min=0, max=4, pdf="uniform", sparsity=0.3)
Y = rand(rows=k, cols=n, min=0, max=4, pdf="uniform", sparsity=0.3)
if (1 == 1) {
    for (i in 1:I) {
        S <- X %*% Y
    }
}
print (as.scalar(S[1,1]))
```

For the dense matrix tests, this was the DML script:

```r
m=$1
k=$2
n=$3

I=$4
X = rand(rows=m, cols=k, min=0, max=4, pdf="uniform", sparsity=1.0)
Y = rand(rows=k, cols=n, min=0, max=4, pdf="uniform", sparsity=1.0)

if (1 == 1) {
    for (i in 1:I) {
        S <- X %*% Y
    }
}
print (as.scalar(S[1,1]))
```

The tests were run with the following script:
```bash
#!/bin/bash

M=(100  200  400  800  1600  3200  6400  12800  25600)
K=(100  200  400  800  1600  3200  6400  12800  25600)
I=(20   20   20   20   20    10    10    5      5)


length=${#K[@]}

OUT="out.sparsemm"
for i in ${M[@]}
    do
        for ((l=0;l<length;l++))
            do
                j="${K[$l]}"
                q="${I[$l]}"
                echo "M = $i, K = $j, N = $i" >> $OUT
                echo "I = $q" >> $OUT
                bin/systemml_gputest sparse.dml -stats -args $i $j $i $q 2>> $OUT
                bin/systemml_gputest sparse.dml -stats -gpu force=true -args $i $j $i $q 2>> $OUT
            done
    done


OUT="out.sparsemv"
for i in ${M[@]}
    do
        for ((l=0;l<length;l++))
            do
                j="${K[$l]}"
                q="${I[$l]}"
                echo "M = $i, K = $j, N = 1" >> $OUT
                echo "I = $q" >> $OUT
                bin/systemml_gputest sparse.dml -stats -args $i $j 1 $q 2>> $OUT
                bin/systemml_gputest sparse.dml -stats -gpu force=true -args $i $j 1 $q 2>> $OUT
            done
    done

OUT="out.densemm"
for i in ${M[@]}
    do
        for ((l=0;l<length;l++))
            do
                j="${K[$l]}"
                q="${I[$l]}"
                echo "M = $i, K = $j, N = $i" >> $OUT
                echo "I = $q" >> $OUT
                bin/systemml_gputest dense.dml -stats -args $i $j $i $q 2>> $OUT
                bin/systemml_gputest dense.dml -stats -gpu force=true -args $i $j $i $q 2>> $OUT
            done
    done

OUT="out.densemv"
for i in ${M[@]}
    do
        for ((l=0;l<length;l++))
            do
                j="${K[$l]}"
                q="${I[$l]}"
                echo "M = $i, K = $j, N = 1" >> $OUT
                echo "I = $q" >> $OUT
                bin/systemml_gputest dense.dml -stats -args $i $j 1 $q 2>> $OUT
                bin/systemml_gputest dense.dml -stats -gpu force=true -args $i $j 1 $q 2>> $OUT
            done
    done

OUT="out.densevv"
for ((l=0;l<length;l++))
     do
         j="${K[$l]}"
         q="${I[$l]}"
         echo "M = 1, K = $j, N = 1" >> $OUT
         echo "I = $q" >> $OUT
         bin/systemml_gputest dense.dml -stats -args 1 $j 1 $q 2>> $OUT
         bin/systemml_gputest dense.dml -stats -gpu force=true -args 1 $j 1 $q 2>> $OUT
     done
```

nakul02:
An explanation of the _GPU Overhead_ graph:

**Alloc** and **Dealloc** are times spent in calls made to `cudaMalloc` and `cudaFree`.
**ToDev** and **FromDev** are times spent in memory transfer to the device and from the device respectively.
**Conversion** is the time spent in converting dense to sparse, or sparse to dense format in host or device memory.


The point of having the Garbage Collection graph is to show that it was insignificant compared to the time spent in matrix multiplication.

Some points are missing in the "GPU" lines. Those were from when the program crashed because of there not being enough memory or some other reason. I haven't fully investigated this.

@niketanpansare, @mboehm7 - thoughts?

src/main/java/org/apache/sysml/utils/Statistics.java
```diff
@@ -640,14 +642,16 @@ public static String display()
 			sb.append("CUDA/CuLibraries init time:\t" + String.format("%.3f", cudaInitTime*1e-9) + "/"
 					+ String.format("%.3f", cudaLibrariesInitTime*1e-9) + " sec.\n");
 			sb.append("Number of executed GPU inst:\t" + getNoOfExecutedGPUInst() + ".\n");
-			sb.append("GPU mem tx time (alloc/dealloc/toDev/fromDev):\t" 
+			sb.append("GPU mem tx time (alloc/dealloc/conversion/toDev/fromDev):\t"
```
niketanpansare: I like the idea of having conversion time :+1:

nakul02:
Speedup Plots

# Dense Vector-Vector
speedup
speedup_l
[SVGs](https://gist.github.com/nakul02/82b05ab65e35ff812a8b0a7fbd8101eb)

# Dense Matrix Vector
speedup_l
speedup
[SVGs](https://gist.github.com/nakul02/6310433a98cc71ced085aaf7a81597c2)

# Dense Matrix Matrix
speedup_l
speedup
[SVGs](https://gist.github.com/nakul02/02cefb45464c5b17aaa032dd83ced950)

# Sparse Matrix Vector
speedup_l
speedup
[SVGs](https://gist.github.com/nakul02/75e3f9b6281e1dbe998d9247a8e2bde8)

# Sparse Matrix Matrix
speedup_l
speedup
[SVGs](https://gist.github.com/nakul02/821978c19413e40296644ae4124487b6)

@niketanpansare

nakul02:
I tested all possible combinations of matrix multiplies with tests.
Here are the scripts I used.


```bash
#!/bin/bash

X1=(1 100 200)
X2=(100 200)
Y2=(1 100 200)
SX=(0.3 1.0)
SY=(0.3 1.0)

for x1 in ${X1[@]}
do
    for x2 in ${X2[@]}
    do
        y1=$x2
        for y2 in ${Y2[@]}
        do
            for sX in ${SX[@]}
            do
                for sY in ${SY[@]}
                do
                    echo "matmul_1 =>  x1=$x1, x2=$x2, y1=$y1, y2=$y2 sX=$sX, sY=$sY"
                    bin/systemml write_random.dml -nvargs x1=$x1 x2=$x2 y1=$y1 y2=$y2 sX=$sX sY=$sY Xf="X" Yf="Y" 2>/dev/null
                    bin/systemml matmul_1.dml -nvargs Xf="X" Yf="Y" OUT="A" 2>/dev/null
                    bin/systemml matmul_1.dml -gpu force=true -nvargs Xf="X" Yf="Y" OUT="B"  2>/dev/null
                    bin/systemml threshold_test.dml -nvargs A="A" B="B"
                done
            done
        done
    done
done


X1=(1 100 200)
X2=(100 200)
Y1=(1 100 200)

for x1 in ${X1[@]}
do
    for x2 in ${X2[@]}
    do
        y2=$x2
        for y1 in ${Y1[@]}
        do
            for sX in ${SX[@]}
            do
                for sY in ${SY[@]}
                do
                    echo "matmul_2 =>  x1=$x1, x2=$x2, y1=$y1, y2=$y2 sX=$sX, sY=$sY"
                    bin/systemml write_random.dml -nvargs x1=$x1 x2=$x2 y1=$y1 y2=$y2 sX=$sX sY=$sY Xf="X" Yf="Y" 2>/dev/null
                    bin/systemml matmul_2.dml -nvargs Xf="X" Yf="Y" OUT="A" 2>/dev/null
                    bin/systemml matmul_2.dml -gpu force=true -nvargs Xf="X" Yf="Y" OUT="B"  2>/dev/null
                    bin/systemml threshold_test.dml -nvargs A="A" B="B"
                done
            done
        done
    done
done



X1=(100 200)
X2=(1 100 200)
Y2=(1 100 200)

for x1 in ${X1[@]}
do
    y1=$x1
    for x2 in ${X2[@]}
    do
        for y2 in ${Y2[@]}
        do
            for sX in ${SX[@]}
            do
                for sY in ${SY[@]}
                do
                  echo "matmul_3 =>  x1=$x1, x2=$x2, y1=$y1, y2=$y2 sX=$sX, sY=$sY"
                    bin/systemml write_random.dml -nvargs x1=$x1 x2=$x2 y1=$y1 y2=$y2 sX=$sX sY=$sY Xf="X" Yf="Y" 2>/dev/null
                    bin/systemml matmul_3.dml -nvargs Xf="X" Yf="Y" OUT="A"  2>/dev/null
                    bin/systemml matmul_3.dml -gpu force=true -nvargs Xf="X" Yf="Y" OUT="B"  2>/dev/null
                    bin/systemml threshold_test.dml -nvargs A="A" B="B"
                done
            done
        done
    done
done


X1=(1 100 200)
X2=(1 100 200)
Y1=(1 100 200)

for x1 in ${X1[@]}
do
    y2=$x1
    for x2 in ${X2[@]}
    do
        for y1 in ${Y1[@]}
        do
            for sX in ${SX[@]}
            do
                for sY in ${SY[@]}
                do
                  echo "matmul_4 =>  x1=$x1, x2=$x2, y1=$y1, y2=$y2 sX=$sX, sY=$sY"
                    bin/systemml write_random.dml -nvargs x1=$x1 x2=$x2 y1=$y1 y2=$y2 sX=$sX sY=$sY Xf="X" Yf="Y" 2>/dev/null
                    bin/systemml matmul_4.dml -nvargs Xf="X" Yf="Y" OUT="A" 2>/dev/null
                    bin/systemml matmul_4.dml -gpu force=true -nvargs Xf="X" Yf="Y" OUT="B"  2>/dev/null
                    bin/systemml threshold_test.dml -nvargs A="A" B="B"
                done
            done
        done
    done
done
```

The random matrices are written out using write_random.dml:
```r
x1 = $x1 # rows in X
x2 = $x2 # cols in X
y1 = $y1 # rows in Y
y2 = $y2 # cols in Y
sX = $sX # sparsity for X
sY = $sY # sparsity for Y

X = rand(rows=x1, cols=x2, min=0, max=4, pdf="uniform", sparsity=sX)
Y = rand(rows=y1, cols=y2, min=0, max=4, pdf="uniform", sparsity=sY)

write(X, $Xf)
write(Y, $Yf)
```

Here are matmul_1,2,3,4:
```r
X = read ($Xf)
Y = read ($Yf)

O = X %*% Y # <- matmul_1
O = X %*% t(Y) # <- matmul_2
O = t(X) %*% Y # <- matmul_3
O = t(X) %*% t(Y) # <- matmul_4

write (O, $OUT)
```

Matmul_1-4 are run with CPU and GPU and the results are compared:

```r
A = read($A)
B = read($B)

compare = function (matrix[double] A, matrix[double] B) return (boolean out) {
  Ax = abs(A - B)
  T = Ax > 10e-6
  s = sum(T)
  out = s
}

x <- compare(A, B)
if (x){
  print (" MATRICES NOT MATCHED ")
}
```

All the tests passed.

niketanpansare:
@nakul02 Please change LOG.info to LOG.debug in LibMatrixCUDA.

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUContext.java
```diff
@@ -36,24 +36,21 @@ protected GPUContext() { }
...
-				public void run() {
-					// Lazy GPU context creation
-					synchronized(isGPUContextCreated) {
-						currContext = new JCudaContext();
-						OptimizerUtils.GPU_MEMORY_BUDGET = ((JCudaContext)currContext).getAvailableMemory();
-						isGPUContextCreated = true;
-					}
-				}
-			}).start();
+			synchronized(isGPUContextCreated) {
```
niketanpansare: Won't this make createGPUContext a blocking call and add additional 5 seconds in the execution time ?

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -300,57 +327,416 @@ public static void matmultTSMM(MatrixObject left, MatrixObject output,
 	    if(!output.getGPUObject().isAllocated)
 	            throw new DMLRuntimeException("Output is not allocated:" + output.getGPUObject().isAllocated);
 	
-	    Pointer A = ((JCudaObject)left.getGPUObject()).jcudaPointer;
-	    Pointer C = ((JCudaObject)output.getGPUObject()).jcudaPointer;
+	    Pointer A = ((JCudaObject)left.getGPUObject()).jcudaDenseMatrixPtr;
```
niketanpansare: We need to do sparse to dense conversion here if the GPU object is in sparse format.

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUContext.java
nakul02: Yes, but there is no join on the thread that creates the context. This is potentially an incorrect program. 
In a multi-threaded program, relying on "delays" for program correctness is a bad idea.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
nakul02: Let us fix this in the another PR.
My tests won't cover for this change.

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUContext.java
niketanpansare: Ok ... we can keep this logic (i.e. blocking call) for now and try to take care of additional 5 seconds in subsequent PR.

nakul02: Ok,
Also, Log.info -> Log.debug

niketanpansare:
@deroneriksson @dusenberrymw LGTM. Please feel free to merge.

@nakul02 This is good work  :+1: 

I have also tested this branch on Lenet for correctness (end-to-end):
``` bash
EPOCH=9.883720930232558 ITER=8500 Validation set accuracy (%): 99.1 step-size=0.006302494097246091
Number of executed GPU inst:    97877.
GPU mem tx time (alloc/dealloc/conversion/toDev/fromDev):       29.002/19.461/0.325/122.697/303.068 sec.
GPU mem tx count (alloc/dealloc/conversion/toDev/fromDev/evict):        294886/208883/17200/162609/97841/0.
```

deroneriksson:
Great @nakul02 and @niketanpansare! I'll go ahead and merge.

nakul02:
Updated graphs, just so that there is a record of the performance information:

## Dense VV
composite
[SVG](https://gist.github.com/nakul02/92c96bb63181a8f4f172b0d4b85ac48c)

## Dense MM
composite
[SVG](https://gist.github.com/nakul02/76ebb531b29e5cab560070fc6dd3f9c0)

## Dense MV
composite
[SVG](https://gist.github.com/nakul02/9bf8c5250757015e959149a52548eaa5)

## Sparse MV
composite
[SVG](https://gist.github.com/nakul02/d1813c0cd04d3cbebf0f036b9618f69b)

## Sparse MM
composite
[SVG](https://gist.github.com/nakul02/bd41ab5b75ae4e6a882736e82683d6eb)

PR #197, opened 2016-07-28
[SYSTEMML-451] Python embedded DSL

niketanpansare:
To make progress on adoption front, I think we need to support embedded Python DSL. So, I have created this PR as an initial proposal. I have added subset of builtin functions (namely, full/rand.*/mlprint/dot), binary/unary operators for matrices and parfor as an example language level construct. 

Please use this PR to facilitate further discussion on this topic. Here are some initial points:

1. What is preferred usage scenario for a Python data scientist ?
a. Scikit-learn like-library. In this case, we can call the DML algorithms using MLContext.
b. External DSL approach: The data scientists will write her code in PyDML and then use MLContext.
c. Embedded DSL approach.

2. Even in embedded DSL approach, there are two possible implementation choices:
a. Execute code with a context (as in this PR). 
Pros: Simple and elegant push-down mechanism and no redundant computation.
Cons: Difficult to implement mixed code.
b. Add lazy data structures which will be executed only when certain actions are invoked on them.

3. Push-down of language-level constructs (**Will skip this in this commit**):
a. No modification to Python implementation, but slightly inelegant usage through use of functions (as in this PR). See parfor invocation in the below code.
b. Modify Python parser to support pushdown. This requires users to rebuilt oru version of python from scratch. 

4. Whether to add standalone support as it requires reimplementing Py4J bridge of PySpark.

5. APIs of the built-in functions: Should it be similar to PyDML ?

If after the discussion, we decide to go in different direction, I will delete this PR.

If we agree to go in this direction, I can create following tasks on JIRA:
1. Make Python embedded DSL feature complete.
2. Add documentation and examples for Python embedded DSL.
3. Create .py files for all our algorithms using Python embedded DSL. We can keep interface similar to scikit-learn.
4. Add additional input/output mechanism (other than DataFrame), for example: binary blocked RDD, string rdd, MLLib's BlockMatrix, NumPy arrays, etc.
5. Add Py4J support to allow standalone usage and create pip installer for systemml.

Please note: I am using older MLContext approach and we can update it to newer MLContext once it is delivered. 

An example script supported by this commit:
``` python
# wget https://sparktc.ibmcloud.com/repo/latest/systemml-0.11.0-incubating-SNAPSHOT.jar
# pyspark --master local[*] --driver-class-path systemml-0.11.0-incubating-SNAPSHOT.jar

>>> import SystemML as sml
>>> import numpy as np
>>> sml.setSparkContext(sc)

Welcome to Apache SystemML!

>>> m1 = sml.matrix(np.ones((3,3)) + 2)
>>> m2 = sml.matrix(np.ones((3,3)) + 3)
>>> m2 = m1 * (m2 + m1)
>>> m4 = 1.0 - m2
>>> m4
# This matrix (mVar5) is backed by below given PyDML script (which is not yet evaluated). To fetch the data of this matrix, invoke toNumPyArray() or toDataFrame() or toPandas() methods.
mVar1 = load(" ", format="csv")
mVar2 = load(" ", format="csv")
mVar3 = mVar2 + mVar1
mVar4 = mVar1 * mVar3
mVar5 = 1.0 - mVar4
save(mVar5, " ")

<SystemML.defmatrix.matrix object>
>>> m2.eval()
>>> m2
# This matrix (mVar4) is backed by NumPy array. To fetch the NumPy array, invoke toNumPyArray() method.
<SystemML.defmatrix.matrix object>
>>> m4
# This matrix (mVar5) is backed by below given PyDML script (which is not yet evaluated). To fetch the data of this matrix, invoke toNumPyArray() or toDataFrame() or toPandas() methods.
mVar4 = load(" ", format="csv")
mVar5 = 1.0 - mVar4
save(mVar5, " ")

<SystemML.defmatrix.matrix object>
>>> m4.sum(axis=1).toNumPyArray()
array([[-60.],
       [-60.],
       [-60.]])
>>>
```

@mboehm7  @frreiss @bertholdreinwald @nakul02 @dusenberrymw  @deroneriksson


dusenberrymw:
@niketanpansare This is great!  I'm definitely in favor of creating a Python DSL.  Here are my thoughts on the proposed questions:

1. What is preferred usage scenario for a Python data scientist ?
a. Scikit-learn like-library. In this case, we can call the DML algorithms using MLContext.
b. External DSL approach: The data scientists will write her code in PyDML and then use MLContext.
**c. Embedded DSL approach.**  This is where we need to move to if we want data scientists to start using the project since they will want to write code in Python itself.

2. Even in embedded DSL approach, there are two possible implementation choices:
a. Execute code with a context (as in this PR). 
Pros: Simple and elegant push-down mechanism and no redundant computation.
Cons: Difficult to implement mixed code.
**b. Add lazy data structures which will be executed only when certain actions are invoked on them.**  Ideally, we will want the user to be able to mix in NumPy and other Python code seamlessly.  Option (a) is still a decent approach.

3. Push-down of language-level constructs:
**a. No modification to Python implementation, but slightly inelegant usage through use of functions (as in this PR). See parfor invocation in the below code.**  We definitely do *not* want to ask the user to create a custom Python build.  We should simply support PySpark with Python 2 or 3.
b. Modify Python parser to support pushdown. This requires users to rebuilt oru version of python from scratch.

4. Whether to add standalone support as it requires reimplementing Py4J bridge of PySpark.  **No, using PySpark on a laptop is already extremely simple.**

5. APIs of the built-in functions: Should it be similar to PyDML ? **I would vote for it to be equivalent to NumPy.**

niketanpansare:
Thanks @dusenberrymw for your thoughts :+1: 

Here are my comments:
**Add lazy data structures which will be executed only when certain actions are invoked on them. Option (a) is still a decent approach.**
My only worry with Option (2.b) is that adding lazy data structure would make Python API complex and would make it difficult to debug errors/exceptions. We can start by getting feedback about Option (2.a) from few data scientists and go to Option (2.b) only if absolutely necessary.

**Push-down of language-level constructs**
I agree with you about Option (3a). Also, cleaned up the API for language-level constructs. So Option (3.b) is not required.

**Ideally, we will want the user to be able to mix in NumPy and other Python code seamlessly.** 
There are two options here:
1. We ask users to isolate the mixed code themselves:
``` python
with context(sqlCtx):
	m1 = random.uniform(rows=3, cols=2)
	m2 = random.normal(rows=2, cols=3)
	m3 = dot(m1, m2)
	registerOutput([m3])

# mixed code:
numpyM3 = m3.getNumPyArray()
numpyM4 = do certain numpy operations
m4 = Matrix(numpyM4)

with context(sqlCtx):
	some DML that uses m4
```
The above option requires writing appropriate input/output wrappers to convert binary blocked RDD into NumPy arrays and viceversa.

2. We do callbacks to Python using external builtin functions. Here the user will have to wrap up mixed code into Python UDF and then we provide some construct to push it down as external UDF. The external UDF will do necessary conversions and call Python UDF using Py4J framework.

iyounus:
I like the idea of writing everything in python. My suggestion would be to make the API more python like, so that for a python user it doesn't feel unnatural. I would like the syntax to be something like the following, but I don't know if it is technically possible.

```python
import numpy as np
import SystemML as sml
from pyspark.sql import SQLContext

mlContext = MLContext(sc)

m1 = sml.full(1, rows=3, cols=3)
m1 = m1 + 1

tmp = np.ones((3,3))  # numpy matrix
m2 = sml.matrix(tmp)  # ability to convert numpy to systemml matrix (???)
m2 += 2
m3 = m1 * m2
m3 /= m1
m1 = sml.random.uniform(rows=3, cols=2)
m2 = sml.random.normal(rows=2, cols=3)
m4 = sml.dot(m1, m2)

mlContext.registerOutput([m3, m4])
mlContext.execute()

m3.getDF().show()
m4.getDF().show()
```

niketanpansare:
Thanks @iyounus ... I have delivered minor fixes and conversion to/from numpy arrays to current prototype.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +313,39 @@ def __init__(self, sc, *args):
...
         except Py4JError:
             traceback.print_exc()
+            
+    def start(self, sqlCtx):
+		try:
+			prevGlobalMLContext = MLContext.globalMLContext
+		except AttributeError:
+			pass
+		self.numTabs = 0
+		self.sqlCtx = sqlCtx
```
dusenberrymw: We can just use `self.sqlCtx = SQLContext.getOrCreate(sc)`, assuming a `from pyspark.sql import SQLContext` import statement up top.  That way, we can get rid of the need to explicitly create and pass around a SQLContext object (which is going away anyways in 2.0).

dusenberrymw:
@niketanpansare The latest update is awesome!  I really like not having to use with `with ...` construct for the main part of the code, as well as not having to explicitly call any of the `registerInput` stuff.

dusenberrymw:
@niketanpansare Also, as a side note, can you replace the tabs in the file with spaces so that we can use this with Python 3?  I made a commit on Friday addressing this with a couple of tabs accidentally added into the existing MLContext.  Additionally, I made another slight change to `executeScript` on Friday as well for Python 3.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Good point. I will update my PR. For some reason `self.sqlCtx = SQLContext.getOrCreate(sc)` is throwing an error. Instead am creating `self.sqlCtx = SQLContext(sc)` for now and we can replace it with cleaner call.

niketanpansare:
@dusenberrymw Replaced tabs with spaces :+1:

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -25,10 +25,264 @@
...
+        return ', ' + name + '=' + str(val)
+
+class Matrix(object):
+    systemmlVarID = 0
+    def __init__(self, value):
+        if isinstance(value, DataFrame):
+            self.df = df
+        Matrix.systemmlVarID += 1
+        self.ID = 'mVar' + str(Matrix.systemmlVarID)
+    def getID(self):
```
MechCoder: stylistic comment: Can you space the methods out with a space in between, so that it is easier to read?

MechCoder:
I am not a SystemML developer but I am looking forward to testing this out.

@niketanpansare I still get errors on using this with Python3

```python
  File "/home/manoj/incubator-systemml/SystemML.py", line 318
    try:
        ^
TabError: inconsistent use of tabs and spaces in indentation
```

MechCoder:
Oh, I used the version before the last commit was made. Please ignore my previous comment.

dusenberrymw:
Yeah, looks like there are still a bunch of tabs in the file -- I think more were actually introduced.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Done :+1:

niketanpansare:
@MechCoder Can you please reply with your thoughts/preference (as @dusenberrymw did in his first comment) on the questions in the first comment ? Also, any additional comments about high-level design  as well as usage wrt embedded Python API is welcomed :)

MechCoder:
@niketanpansare I really like the script in the PR description, especially after the with constraints are done with. I think this is a good starting point.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+            pass
+        self.numTabs = 0
+        self.sqlCtx = SQLContext(self.sc)
+        MLContext.globalMLContext = self
+        self.reset()
+        self.pydml = []
+        self.matrixOutput = []
+        
+    
+    def stop(self, vals):
```
MechCoder: Can this be `stop(self, *vals)` or `stop(self, vals=None)`?

It is odd that I need to provide an argument to stop the `MLContext`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -25,10 +25,327 @@
...
+def toString(val):
+    if isinstance(val, Scalar):
+        return val.getID()
+    elif isinstance(val, Matrix):
+        return 'toString(' + val.getID() + ')'
+    else:
+        return str(val)
+        
+@contextmanager
+def parfor(s, start, end, check=None, par=None, mode=None, taskpartitioner=None, tasksize=None, datapartitioner=None, resultmerge=None, opt=None):
```
MechCoder: What is your opinion on making this a generator? It feels more natural to write

```python
for x in parfor(s, start, end)
```

rather than a `with`?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
             self.ml = sc._jvm.org.apache.sysml.api.MLContext(sc._jsc, monitorPerformance, setForcedSparkExecType)
         except Py4JError:
             traceback.print_exc()
+            
+    def start(self):
+        try:
+            prevGlobalMLContext = MLContext.globalMLContext
+        except AttributeError:
+            pass
```
MechCoder: `prevGlobalMLContext` seems unused right?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
             self.ml = sc._jvm.org.apache.sysml.api.MLContext(sc._jsc, monitorPerformance, setForcedSparkExecType)
```
MechCoder: style: Why not change the class instantiation to

```python
def __init__(self, sc, monitor=False, forceSpark=False):
```
Basically makes sure that passing args of `length>2` is not silently ignored.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+        except AttributeError:
+            pass
+        self.numTabs = 0
+        self.sqlCtx = SQLContext(self.sc)
+        MLContext.globalMLContext = self
+        self.reset()
+        self.pydml = []
+        self.matrixOutput = []
+        
+
```
MechCoder: xtra space. ;)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
             traceback.print_exc()
+            
+    def start(self):
+        try:
+            prevGlobalMLContext = MLContext.globalMLContext
+        except AttributeError:
+            pass
+        self.numTabs = 0
+        self.sqlCtx = SQLContext(self.sc)
+        MLContext.globalMLContext = self
```
MechCoder: Is this assignment necessary to do? What happens if we replace every usage of `MLContext.globalMLContext` with `self`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+        self.matrixOutput = []
+        
+    
+    def stop(self, vals):
+        if isinstance(vals, Matrix) or isinstance(vals, Scalar):
+            vals.registerAsOutput()
+        else:
+            for i in range(0, len(vals)):
+                if isinstance(vals[i], Matrix) or isinstance(vals[i], Scalar):
+                    vals[i].registerAsOutput()
```
MechCoder: Are you planning to do anything more complicated with `registerAsOutput` apart from appending to `self.matrixOutput`?

If not, I would just do away with `registerAsOutput` and directly append to `matrixOutput` here.

MechCoder: Is the `matrixOutput` attribute used anywhere else, except during when `stop` is called?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+        self.numTabs = 0
+        self.sqlCtx = SQLContext(self.sc)
+        MLContext.globalMLContext = self
+        self.reset()
+        self.pydml = []
+        self.matrixOutput = []
+        
+    
+    def stop(self, vals):
+        if isinstance(vals, Matrix) or isinstance(vals, Scalar):
```
MechCoder: nitpick: `isinstance(vals, (Matrix, Scalar))`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+            vals.registerAsOutput()
+        else:
+            for i in range(0, len(vals)):
```
MechCoder: nitpick: 0 is redundant

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -59,6 +376,40 @@ def __init__(self, sc, *args):
...
+            addPyDMLStmt('save', args=[o.getID(), ', \" \"'])
+        pydml = ''.join(self.pydml)
+        # print('Execution following pydml script:\n' + pydml)
+        out = self.executeScript(pydml, isPyDML=True)
+        for o in self.matrixOutput:
+            o.setDF(out.getDF(self.sqlCtx, o.getID()))
+        MLContext.globalMLContext.reset()
+        try:
+            MLContext.globalMLContext = prevGlobalMLContext
+        except NameError:
```
MechCoder: This will always raise a `NameError` no? `prevGlobalMLContext` is not defined.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -25,10 +25,327 @@
...
+    else:
+        pydmlStmt = pydmlStmt + [  obj, '.', fnName, '(']
+    if args is not None:
+        pydmlStmt = pydmlStmt + args
+    MLContext.globalMLContext.pydml.append(''.join(pydmlStmt + [')\n']))
+    
+# Constructor         
+def full(init, rows=None, cols=None):
+    if rows is None or cols is None:
+        raise Exception('rows and cols are mandatory arguments')
```
MechCoder: If they are mandatory, you should not make them optional in the signature :p

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: Ooh, is this done because of the `scopeCheck` call?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -25,10 +25,327 @@
...
+    #    pass
+    if len(kwargs) > 0:
+        raise Exception('named argument not supported for mlprint inside context')
+    if len(args) > 0:
+        pydmlStmt = [getTabs(), 'print(', toString(args[0])]
+        for i in range(1, len(args)):
+            pydmlStmt = pydmlStmt + [' + ', toString(args[i])]
+        MLContext.globalMLContext.pydml.append(''.join(pydmlStmt + [')\n']))
+    else:
+            raise Exception('print cannot be empty')
```
MechCoder: awkward tab.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -25,10 +25,327 @@
...
+    else:
+        return ', ' + name + '=' + str(val)
+
+class Matrix(object):
+
+    systemmlVarID = 0
+    
+    def __init__(self, value):
+        if isinstance(value, DataFrame):
+            self.df = df
```
MechCoder: Should this be `self.df = value`?

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: Yeah I think it might also be interesting to think about removing the `start(...)` function, and renaming this `stop(...)` function to something like `eval(...)`.

dusenberrymw:
I agree with @MechCoder that this is a great starting point.  In general, we should really continue to focus on solid design for this API so that it is correctly Pythonic, and extremely well integrated with NumPy.

frreiss:
Is it appropriate to use SYSTEMML-451 as the locus for this work? It seems like the Python DSL is going to proceed more or less independently of work in Scala and Java, at least in the short term.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: @MechCoder Good point. Will update the PR with your suggestion :)

@dusenberrymw  The only reason I added start and stop is specify the scope that SystemML will try to exploit as in some cases it might have adverse performance impact due to internal book-keeping. I will keep start for now, but will rename the stop to eval :+1:

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: It was added to support nested start/eval.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: @MechCoder Good suggestion. Let's improve the API of MLContext in a separate PR as part of https://issues.apache.org/jira/browse/SYSTEMML-234. A good task for someone to be a SystemML contributor :)

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: No. Just during stop/eval call. I also thought of supporting matrix.registerAsOutput. But we can revisit this

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Not always. Will work if we intend to support nested start/eval

dusenberrymw:
By the way, let's keep this open for a bit as a WIP to continue thinking carefully about the overall design and trying it out.  I also want to look around at other distributed NumPy-like projects.

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: One possibility is to start with a "fresh" scope on creation of the original `MLContext` object, and then "restart" the scope on every call to `eval(...)`.

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: Better yet, we should just copy the syntax of [`numpy.full`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.full.html).

niketanpansare:
Here is my proposed solution for @MechCoder's question `What are you planning to do with the previous PR? In other words, what are the use-cases of exposing SystemML datastructures to the not-so-advanced user?` (see https://github.com/apache/incubator-systemml/pull/204):

We stick with `start ... eval` construct for now as it has simplest logic and focus on the design of built-in functions (i.e. to make `full` API compatible with NumPy). When we are relatively happy with the design, we push this PR in as WIP and mention in our doc that `start ... eval` will be removed in future. The advanced push-down logic and lazy data structures will be implemented in Scala DSL. Once this DSL is part of master, we replace [addPyDMLStmt and related functions](https://github.com/apache/incubator-systemml/pull/197/files#diff-2c4346a06f8454b330603d8a0dee33b7R37) with calls to Scala lazy structures. With this, we can then remove `start ... eval` logic :) 

@dusenberrymw @frreiss @mboehm7 @bertholdreinwald  Please feel free to provide your suggestions.

MechCoder:
Would you want to refactor this over so that it uses the new MLContext?

MechCoder:
Feel free to merge this, as I would not be able to make a detailed pass at least till next week and maybe refactor in another PR.

niketanpansare:
@MechCoder I think its best to merge the new Python MLContext first and then refactor the current PR to use new MLContext.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        pydmlStmt = pydmlStmt + [  obj, '.', fnName, '(']
+    if args is not None:
+        pydmlStmt = pydmlStmt + args
+    MLContext.globalMLContext.pydml.append(''.join(pydmlStmt + [')\n']))
+ 
+def addPyDMLString(pydmlStmt):
+    MLContext.globalMLContext.pydml.append(''.join(pydmlStmt))
+   
+# Constructor         
+def full(init, rows=None, cols=None):
```
dusenberrymw: Can you update this to match `numpy.full` \[1] to the extent possible?  We should accept a `shape` that can be either a single `int` referring the number of rows (with cols implicitly set to 1) or a `tuple` referring to the rows and cols, as well as an `fill_value`.

\[1]: http://docs.scipy.org/doc/numpy/reference/generated/numpy.full.html

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+    if rows is None or cols is None:
+        raise Exception('rows and cols are mandatory arguments')
+    scopeCheck('full')
+    return simpleAddPyDMLStmt('full', args=[str(init), ', rows=', str(rows), ', cols=', str(cols)])
 
 SUPPORTED_TYPES = (np.ndarray, pd.DataFrame, spmatrix)
 
-class MLContext(object):
+def matrix(numPyArr):
```
dusenberrymw: This function should also accept Python lists.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+    scopeCheck('matrix')
+    if isinstance(numPyArr, np.ndarray):
+        colNames = []
+        numCols = len(numPyArr[0,:])
```
dusenberrymw: This will blow up if a user passes in a vector, such as `np.array([1,2,3])`, which has shape `(3,)`, thus no explicit number of columns.  Instead, let's just create the Pandas DataFrame first via `pdf = pd.DataFrame(numPyArr)`, which will have a full, 2-dimensional shape such as `(3,1)`.  Then, we can get the number of columns with `pdf.shape[1]`.  The column names list can be created using a Pythonic for-comprehension via `columns = ["C" + str(i) for i in range(pdf.shape[1])]`.  (The `0` in `range(0, numCols)` is not necessary).  Then, the Pandas DataFrame columns can be updated via `pdf.columns = columns`.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        for i in range(0, numCols):
+            colNames = colNames + [ str('C' + str(i))]
+        pdf = pd.DataFrame(numPyArr, columns=colNames)
+        df = MLContext.globalMLContext.sqlCtx.createDataFrame(pdf, colNames)
+    elif isinstance(numPyArr, pd.core.frame.DataFrame):
```
dusenberrymw: Combine this with the NumPy if clause above, and just simply convert the `numPyArr` to a Pandas DataFrame in either case, regardless if it is a Numpy array, or Pandas DataFrame.  `pd.DataFrame` won't make a copy of the data for Numpy or Pandas DataFrame inputs anyways, so no harm done.  Also, we should add Python lists as well, and also convert those.  So, basically, this eliminates this additional clause.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        pdf = numPyArr
+        numCols = pdf.shape[1]
+        colNames = []
+        for i in range(0, numCols):
+            colNames = colNames + [ str('C' + str(i))]
+        df = MLContext.globalMLContext.sqlCtx.createDataFrame(pdf, colNames)
+    elif hasattr(numPyArr, '_jdf'):
```
dusenberrymw: Can we check if this is an instance of a PySpark DataFrame?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+    return m1.transpose()
+    
+def mlprint(*args, **kwargs):
+    #try:
+    #    if MLContext.globalMLContext is None:
+    #        return __builtin__.print(*args, **kwargs)
+    #except AttributeError:
+    #    pass
+    if len(kwargs) > 0:
+        raise Exception('named argument not supported for mlprint inside context')
```
dusenberrymw: Let's just eliminate the `**kwargs` from this function if they're not allowed.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+    if len(args) > 0:
```
dusenberrymw: Let's change this to `if args:`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        pydmlStmt = [getTabs(), 'print(', toString(args[0])]
+        for i in range(1, len(args)):
+            pydmlStmt = pydmlStmt + [' + ', toString(args[i])]
+        addPyDMLString(pydmlStmt + [')\n'])
+    else:
+            raise Exception('print cannot be empty')
```
dusenberrymw: How about just not doing anything `mlprint` is called without any arguments?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+    elif isinstance(value, Scalar):
+        return value.getID()
+    else:
+        raise Exception('Incorrect type: expected float or int or long or Scalar')
+        
+def scopeCheck(fnName):
+    if MLContext.globalMLContext is None:
+        raise Exception('Incorrect usage: \"' + fnName + '\" can only run if MLContext.start has been called')
+
+class random(object):
```
dusenberrymw: We should update these to be like NumPy.  Additionally, I would recommend using `np.random.randn` for normal distributions, and `np.random.rand` for uniform distributions.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+            raise Exception('rows and cols are mandatory arguments')
+        return simpleAddPyDMLStmt('random.normal', args=['rows=', str(rows), ', cols=', str(cols), ', sparsity=', getDefault(sparsity, 1.0)])
+        
+    @staticmethod
+    def uniform(rows=None, cols=None, sparsity=None, min=None, max=None):
+        if rows is None or cols is None:
+            raise Exception('rows and cols are mandatory arguments')
+        return simpleAddPyDMLStmt('random.uniform', args=['rows=', str(rows), ', cols=', str(cols), ', sparsity=', getDefault(sparsity, 1.0), ", min=", getDefault(min, 0.0), ", max=", getDefault(max, 1.0) ])
+
+def getDefault(val, defaultVal):
```
dusenberrymw: This should be eliminated in favor of just supplying default arguments in other functions.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+def scopeCheck(fnName):
+    if MLContext.globalMLContext is None:
+        raise Exception('Incorrect usage: \"' + fnName + '\" can only run if MLContext.start has been called')
+
+class random(object):
+    @staticmethod
+    def normal(rows=None, cols=None, sparsity=None):
+        if rows is None or cols is None:
+            raise Exception('rows and cols are mandatory arguments')
+        return simpleAddPyDMLStmt('random.normal', args=['rows=', str(rows), ', cols=', str(cols), ', sparsity=', getDefault(sparsity, 1.0)])
```
dusenberrymw: `Remove the `getDefault` call, and just set `sparsity=1.0` in the function declaration for a default value for `sparsity`.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        
+    @staticmethod
+    def uniform(rows=None, cols=None, sparsity=None, min=None, max=None):
+        if rows is None or cols is None:
+            raise Exception('rows and cols are mandatory arguments')
+        return simpleAddPyDMLStmt('random.uniform', args=['rows=', str(rows), ', cols=', str(cols), ', sparsity=', getDefault(sparsity, 1.0), ", min=", getDefault(min, 0.0), ", max=", getDefault(max, 1.0) ])
```
dusenberrymw: Remove the `getDefault` call, and just set `sparsity=1.0` in the function declaration for a default value for `sparsity`.

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: Yeah this needs to be updated.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -37,11 +37,329 @@
...
+        raise Exception('Not yet implemented')
+
+def getTabs():
+    ret = ''
+    for i in range(0, MLContext.globalMLContext.numTabs):
+        ret = ret + '\t'
+    return ret
+                
+class MLContext(object):
+
```
dusenberrymw: Can you remove this extra newline?

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: It would be cleaner to set `globalMLContext = None` and `prevGlobalMLContext = None` as proper class variables at the top of the class so that there is never an `AttributeError`.

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: Also, below the comments, can you add the following:

```
globalMLContext = None
prevGlobalMLContext = None
```

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -73,6 +391,42 @@ def __init__(self, sc, *args):
...
+            pass
+        self.numTabs = 0
+        self.sqlCtx = SQLContext(self.sc)
+        MLContext.globalMLContext = self
+        self.reset()
+        self.pydml = []
+        self.matrixOutput = []
+        
+    
+    def eval(self, outputs=None):
```
dusenberrymw: Can you update `outputs=None` to `*outputs` to allow for a variable number of outputs, which could be empty, and will always be suppled to the body of the function as a list?  This would mean that this would be called as `ml.eval(m1, m2)`, for example, or as `ml.eval()`.  Then you can do the following:

```
for output in outputs:
  output.registerAsOutput()

for o in self.matrixOutput():
  self.registerOutput(o.getID())
  addPyDMLStmt('save', args=[o.getID(), ', \" \"'])
```

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -73,6 +391,42 @@ def __init__(self, sc, *args):
...
+                if isinstance(outputs[i], Matrix) or isinstance(outputs[i], Scalar):
+                    outputs[i].registerAsOutput()
+        if not outputs is None:
+            for o in self.matrixOutput:
+                self.registerOutput(o.getID())
+                addPyDMLStmt('save', args=[o.getID(), ', \" \"'])
+        pydml = ''.join(self.pydml)
+        # print('Execution following pydml script:\n' + pydml)
+        out = self.executeScript(pydml, isPyDML=True)
+        if not outputs is None:
```
dusenberrymw: This check isn't necessary -- just start with the `for o in self.matrixOutput:` part, which will be fine even if `self.matrixOutput` is empty.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -73,6 +391,42 @@ def __init__(self, sc, *args):
...
+            for o in self.matrixOutput:
+                o.setDF(out.getDF(self.sqlCtx, o.getID()))
+        MLContext.globalMLContext.reset()
+        try:
+            MLContext.globalMLContext = prevGlobalMLContext
+        except NameError:
```
dusenberrymw: This can be removed once the `globalMLContext` is defined as a class variable at the top of `MLContext`.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -73,6 +391,42 @@ def __init__(self, sc, *args):
             self.sqlCtx = SQLContext(sc)
         except Py4JError:
             traceback.print_exc()
+            
+    def start(self):
```
dusenberrymw: Can you define all of the instance variables here in the `__init__` function above with initial defaults?

dusenberrymw:
@niketanpansare I've left additional comments from another full pass through the code!

niketanpansare:
@MechCoder @dusenberrymw Cleanedup the implementation by not requiring start/eval.

dusenberrymw:
Awesome! I'll check this out again later today.

docs/beginners-guide-python.md
```diff
@@ -0,0 +1,219 @@
...
+pip install SystemML
+wget https://sparktc.ibmcloud.com/repo/latest/SystemML.jar
+</div>
+</div>
+
+**Congrats! You can now use SystemML!**
+
+## Invoke SystemML's algorithms
+
+{% highlight python %}
```
MechCoder: You can wrap the code around with triple backticks and it will show syntax highlighting in the README as well. (See: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting)

docs/beginners-guide-python.md
```diff
@@ -0,0 +1,219 @@
...
+from pyspark.sql import SQLContext
+sqlCtx = SQLContext(sc)
+digits = datasets.load_digits()
+X_digits = digits.data
+y_digits = digits.target + 1
+n_samples = len(X_digits)
+X_train = X_digits[:.9 * n_samples]
+y_train = y_digits[:.9 * n_samples]
+X_test = X_digits[.9 * n_samples:]
+y_test = y_digits[.9 * n_samples:]
```
MechCoder: You have reimplemented `sklearn.model_selection.train_test_split(X_digits, y_digits, train_size=0.9)` ;) . But that's fine.

docs/beginners-guide-python.md
```diff
@@ -0,0 +1,219 @@
...
+test = sqlCtx.createDataFrame([
+    (12L, "spark i j k"),
+    (13L, "l m n"),
+    (14L, "mapreduce spark"),
+    (15L, "apache hadoop")], ["id", "text"])
+prediction = model.transform(test)
+prediction.show()
+{% endhighlight %}
+
+## Matrix operations
```
MechCoder: Should matrix operations come above the description of machine learning algorithms?

docs/beginners-guide-python.md
```diff
@@ -0,0 +1,219 @@
...
+
+{% highlight python %}
+import SystemML as sml
+import numpy as np
+sml.setSparkContext(sc)
+m1 = sml.matrix(np.ones((3,3)) + 2)
+m2 = sml.matrix(np.ones((3,3)) + 3)
+m2 = m1 * (m2 + m1)
+m4 = 1.0 - m2
+m4.sum().toNumPyArray()
```
MechCoder: This definitely looks much better :-)

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,284 @@
...
+    script = None
+    ml = None
+    visited = []
+    def __init__(self, data, op=None):
+        """
+        Constructs a lazy matrix
+        
+        Parameters
+        ----------
+        data: NumPy ndarray, Pandas DataFrame, scipy sparse matrix or PySpark DataFrame
```
MechCoder: you should also document what happens when "data" is None

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,284 @@
...
+            m.data = results.getNumPyArray(m.ID)
+    
+# Instead of inheriting from np.matrix
+class matrix(object):
+    systemmlVarID = 0
+    dml = []
+    script = None
+    ml = None
+    visited = []
+    def __init__(self, data, op=None):
```
MechCoder: op is not documented

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,284 @@
...
+def setSparkContext(sc):
+    """
+    Before using the matrix, the user needs to invoke this function.
+    
+    Parameters
+    ----------
+    sc: SparkContext
+        SparkContext
+    """
+    matrix.ml = MLContext(sc)
```
MechCoder: WDYT about having a separate class and setting the static variables `ml` and `sc` of that class? The current approach works but means that if you have a new SystemML datatype that relies on the MLContext, you would still have to check `matrix.ml` which seems a bit odd.

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        checkIfMLContextIsSet()
+        self.visited = False 
+        matrix.systemmlVarID += 1
+        self.output = False
+        self.ID = 'mVar' + str(matrix.systemmlVarID)
+        if isinstance(data, SUPPORTED_TYPES):
+            self.data = data
+        elif hasattr(data, '_jdf'):
+            self.data = data
```
MechCoder: nitpick: You can refactor the above four lines into one statement.

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,284 @@
...
+        elif data is None and op is not None:
+            self.data = None
+            self.op = op
+        else:
+            raise Exception('Unsupported input type')
```
MechCoder: Should be a TypeError.

MechCoder:
@niketanpansare Sorry, is this still WIP? I can have a more detailed look tomorrow.

niketanpansare:
@MechCoder No, was merging @dusenberrymw's changes from yesterday.

docs/beginners-guide-python.md
niketanpansare: For some reason, sklearn is throwing an error for model_selection:
```bash
>>> from sklearn.model_selection import train_test_split
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named model_selection
```

I tried to update scikit-learn using `conda install scikit-learn`, but still the same error :(

MechCoder: Oh yes, the rename has not yet been released. You could do `from sklearn.cross_validation` instead.

docs/beginners-guide-python.md
dusenberrymw: Yeah this is looking great!

src/main/java/org/apache/sysml/api/mlcontext/MLResults.java
```diff
@@ -115,7 +120,21 @@ public Data getData(String outputName) {
 	 */
 	public MatrixObject getMatrixObject(String outputName) {
 		Data data = getData(outputName);
-		if (!(data instanceof MatrixObject)) {
+		if(data instanceof ScalarObject) {
```
dusenberrymw: What is this being used for?

src/main/python/SystemML/converters.py
```diff
@@ -0,0 +1,80 @@
...
+        numRows = src.shape[0]
+        numCols = src.shape[1]
+        data = src.data
+        row = src.row.astype(np.int32)
+        col = src.col.astype(np.int32)
+        nnz = len(src.col)
+        buf1 = bytearray(data.tostring())
+        buf2 = bytearray(row.tostring())
+        buf3 = bytearray(col.tostring())
+        return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertSciPyCOOToMB(buf1, buf2, buf3, numRows, numCols, nnz)
```
dusenberrymw: Curious, would it be possible to return a new [`MLContext Matrix`](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/mlcontext/Matrix.java#L36) class here instead, i.e. the `MatrixBlock` wrapped in a `MatrixObject` and then in the `Matrix` (and that Java object wrapped on the Python side)?  I'll leave more comments below, but it would be cool to fuse together that new `Matrix` class on the Java/Scala side, and the new Python `Matrix` and `matrix` classes (using the later name in Python).  If so, then we could include a new set of methods for this conversion in the `MLContext` on both the Python and Java sides, which would keep the main entry point in that API.  Basically, it would be nice to anchor interactions via the `MLContext`.

src/main/python/SystemML/converters.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        raise TypeError('sc needs to be of type SparkContext') # TODO: We can generalize this by creating py4j gateway ourselves
+
+
+def convertToPandasDF(X):
+    if not isinstance(X, pd.DataFrame):
+        return pd.DataFrame(X, columns=['C' + str(i) for i in range(getNumCols(X))])
+    return X
+    
+__all__ = [ 'getNumCols', 'convertToMatrixBlock', 'convertToNumpyArr', 'convertToPandasDF', 'SUPPORTED_TYPES' ]
```
dusenberrymw: Nitpick: Can you move these `__all__` statements to the top underneath the license comments?  https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

src/main/python/SystemML/defmatrix.py
dusenberrymw: How about just setting a `ml` class variable on `MLContext` itself?

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,285 @@
...
+        m._visit(execute=execute)
+    if not execute:
+        return ''.join(matrix.dml)
+    matrix.script.scriptString = ''.join(matrix.dml)
+    results = matrix.ml.execute(matrix.script)
+    for m in outputs:
+        if outputDF:
+            m.data = results.getDataFrame(m.ID)
+        else:
+            m.data = results.getNumPyArray(m.ID)
```
dusenberrymw: First, this lazy evaluation of the new `matrix` class is awesome!

Continuing with my thought on fusing the Python MLContext `Matrix` and `matrix` classes.  The current `Matrix` class has a few cool properties -- it can be used directly in other scripts as inputs (i.e. it represents a fully evaluated matrix), and at the same time it can be converted to other types as desired by the user, i.e. to a `DataFrame` via `toDF()`, and to NumPy arrays once I add the function.  If we merge that `Matrix` class with the new `matrix` class (using the later name), these lazy matrices could be used in either future direct matrix manipulation via this new DSL, *as well as* future scripts directly via `MLContext`.  Once a user wanted the results, he/she simply could call `m1.toDF()`, `m1.toNumPyArray()`, etc.

This lazy matrix DSL is exciting!

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,285 @@
...
+    
+# Instead of inheriting from np.matrix
+class matrix(object):
```
dusenberrymw: Just saying, this is great!

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,285 @@
...
+    script = None
+    ml = None
+    visited = []
+    def __init__(self, data, op=None):
+        """
+        Constructs a lazy matrix
+        
+        Parameters
+        ----------
+        data: NumPy ndarray, Pandas DataFrame, scipy sparse matrix or PySpark DataFrame. (data cannot be None for external users, 'data=None' is used internally for lazy evaluation).
```
dusenberrymw: Could we also accept a Python list?

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,285 @@
...
+        return self.data
+    
+    def toDataFrame(self):
+        if self.data is None:
+            self.eval(outputDF=True)
+        if not isinstance(self.data, DataFrame):
+            if MLResults.sqlContext is None:
+                MLResults.sqlContext = SQLContext(matrix.sc)
+            self.data = sqlContext.createDataFrame(self.toPandas())
+        return self.data
```
dusenberrymw: Yeah if we merge this with the `Matrix` class we could merge this method with `toDF()` using the later name to be consistent with Spark.

src/main/python/SystemML/defmatrix.py
```diff
@@ -0,0 +1,285 @@
...
+        
+    def _visit(self, execute=True):
+        if self.visited:
+            return self
+        self.visited = True
+        # for cleanup
+        matrix.visited = matrix.visited + [ self ]
+        if self.data is not None:
+            matrix.dml = matrix.dml + [ self.ID,  ' = load(\" \", format=\"csv\")\n']
```
dusenberrymw: The new `MLContext` no longer requires that the explicit read/load or write/save statements be added for inputs and outputs, so we could remove this.

src/main/python/SystemML/defmatrix.py
MechCoder: sounds good

src/main/python/SystemML/mlcontext.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/python
```
dusenberrymw: This can be removed.

src/main/python/SystemML/mlcontext.py
```diff
@@ -20,20 +21,23 @@
...
 from pyspark import SparkContext
 import pyspark.mllib.common
-
+from pyspark.sql import DataFrame, SQLContext
+from .converters import *
 
 def dml(scriptString):
     """
     Create a dml script object based on a string.
-
```
dusenberrymw: Nitpick: Can you add back these missing newlines from this file?  Just looks like an artifact from the merge.

src/main/python/SystemML/mlcontext.py
```diff
@@ -122,22 +121,49 @@ def toDF(self):
...
+        ----------
+        outputs: string, list of strings
+            Output variables as defined inside the DML script.
+        """
+        outs = [convertToNumpyArr(self.sc, self._java_results.getMatrix(out).asBinaryBlockMatrix().getMatrixBlock()) for out in outputs]
+        if len(outs) == 1:
+            return outs[0]
+        return outs
+    
+    def getDataFrame(self, *outputs):
```
dusenberrymw: I would remove this and just call `toDF()` on the resulting `Matrix` (or `matrix` if we merge them) from `get()`.

src/main/python/SystemML/mlcontext.py
```diff
@@ -157,11 +183,8 @@ class Script(object):
 
     Parameters
     ----------
-    scriptString: string
+    path: string
```
dusenberrymw: Can you swap these lines?  Just another artifact from the merge.

src/main/python/SystemML/mlcontext.py
```diff
@@ -157,11 +183,8 @@ class Script(object):
...
         Can be either a file path to a DML script or a DML script itself.
-
-    scriptType: string
-        Script language, either "dml" for DML (R-like) or "pydml" for PyDML (Python-like).
```
dusenberrymw: Can you add these lines back?  Just another artifact from the merge.

src/main/python/uploadToPyPI.sh
```diff
@@ -0,0 +1,34 @@
...
+tar -xzf target/systemml-*-SNAPSHOT.tar.gz -C src/main/python/SystemML
+
+cd src/main/python/SystemML
+mv systemml-*-incubating-SNAPSHOT SystemML-java
+
+cd ..
+echo "Preparing to upload to PyPI ...."
+python setup.py register sdist upload
+
+rm -r SystemML/SystemML-java
```
dusenberrymw: Can you move this file from `src/main/python/` to `dev/release/`?

dusenberrymw:
@niketanpansare This is awesome!!  Just tried it out with Python 3 -- works great! 👍 

I left some thoughts, mostly regarding the cool potential to merge `Matrix` with `matrix` which would allow this DSL to work seamlessly with direct script invocation via `MLContext`.

Also, since we're creating a package structure here, can we go ahead and rename the top-level package from `SystemML` to `systemml`?

dusenberrymw:
Also, we could save the merging of `Matrix` and `matrix` for a future PR, and brainstorm some plans and infrastructure for it in the meantime.

src/main/java/org/apache/sysml/api/mlcontext/MLResults.java
niketanpansare: This unifies returning scalar as well as matrix and simplifies the python interface. Currently, if the registered output variable turns out to be scalar, then this change will allow it to be returned as 1 X 1 matrix.

src/main/python/SystemML/defmatrix.py
niketanpansare: Sure, I was also thinking of extending to accept filenames. But, may be we should have a seperate PR where we unify the supported types for matrix, mllearn as well as mlcontext :)

src/main/python/SystemML/defmatrix.py
niketanpansare: Good point, will clean it up.

niketanpansare:
Updated the code as per your comments. 

Also, I am fine with renaming the top-level package from SystemML to systemml. I kept package naming for backward compatibility. Let's get feedback from others before making this change.

I agree we should merge Matrix and matrix in seperate PR :+1:

niketanpansare:
Updated the documentation: http://niketanpansare.github.io/incubator-systemml/beginners-guide-python.html

niketanpansare:
Unless there are major concerns, I think we should merge this PR now and create new PR for additional fixes. This is because any additional Python related changes (eg: improvements to MLContext, merging Matrix & matrix classes, release-related changes (SystemML v/s systemml, pip installer), examples, etc) are dependent on merging this PR.

src/main/java/org/apache/sysml/api/mlcontext/MLResults.java
dusenberrymw: Okay, we may not need this anymore since the new `MLContext

dusenberrymw: Okay, we may not need this anymore since the new `MLContext` can natively return scalar values.  Additionally, a simple call to `get("varName1", "varName2"...)` will return the correct types.  Checkout out this example where a scalar and a matrix from a script are returned in the correct formats:

```python
from SystemML import MLContext, dml

rdd1 = sc.parallelize(["1.0,2.0", "3.0,4.0"])

sums = """
s1 = sum(m1)
m2 = m1 * 2
"""

ml = MLContext(sc)
script = dml(sums).input(m1=rdd1).out("s1", "m2")
s1, m2 = ml.execute(script).get("s1", "m2")
s1, m2
```
Returns:
```python
(10.0, Matrix)
```

Then, 
```python
m2.toDF()
```
Returns
```python
DataFrame[ID: double, C1: double, C2: double]
````

File: src/main/java/org/apache/sysml/api/python/SystemML.py [hunk 56]
```diff
@@ -1,482 +0,0 @@
-#!/usr/bin/python
```
dusenberrymw: I don't think we should remove this entire file in this PR as it contains the the *old* `MLContext` that was in the previous release.  I would just remove the `mllearn` and `mlpipeline` stuff from it, as that was added after the last official release.

src/main/python/SystemML/mlcontext.py
```diff
@@ -20,20 +21,24 @@
...
 def dml(scriptString):
     """
     Create a dml script object based on a string.
-
+    
     Parameters
     ----------
     scriptString: string
         Can be a path to a dml script or a dml script itself.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -47,12 +52,10 @@ def dml(scriptString):
 def pydml(scriptString):
     """
     Create a pydml script object based on a string.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -47,12 +52,10 @@ def dml(scriptString):
...
     Parameters
     ----------
     scriptString: string
         Can be a path to a pydml script or a pydml script itself.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -86,12 +89,10 @@ def _py2java(sc, obj):
 class Matrix(object):
     """
     Wrapper around a Java Matrix object.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -86,12 +89,10 @@ def _py2java(sc, obj):
...
     Parameters
     ----------
     javaMatrix: JavaObject
         A Java Matrix object as returned by calling `ml.execute().get()`.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -105,7 +106,6 @@ def __repr__(self):
     def toDF(self):
         """
         Convert the Matrix to a PySpark SQL DataFrame.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -122,22 +122,49 @@ def toDF(self):
 class MLResults(object):
     """
     Wrapper around a Java ML Results object.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -122,22 +122,49 @@ def toDF(self):
...
     Parameters
     ----------
     results: JavaObject
         A Java MLResults object as returned by calling `ml.execute()`.
-
```
dusenberrymw: Add this line back.

src/main/python/SystemML/mlcontext.py
```diff
@@ -158,9 +185,6 @@ class Script(object):
     Parameters
     ----------
     scriptString: string
-        Can be either a file path to a DML script or a DML script itself.
-
-    scriptType: string
```
dusenberrymw: Add these 3 lines back.

dusenberrymw:
@niketanpansare I agree that we should get this PR in and then create a set of JIRAs/PRs to address a bunch of the other concerns.  Before we merge though, I left comments for a few things that should be addressed now:

1. I don't think we should completely remove the old `src/main/java/org/apache/sysml/api/python/SystemML.py` file in this PR, as it contains the *old* `MLContext`.  We should just remove the `mllearn` and `mlpipeline` stuff from it, as those components were only added recently.  That way we can deprecate both the Python & Java old `MLContext` APIs, and then remove them in the future.
2. There are still several places in `mlcontext.py` in which the docstrings were messed up in the merge, including missing comments.  Can you add those back in before merging?

dusenberrymw:
For the **next set** of JIRAs/PRs, we should aim to complete the following **before** the upcoming official release:

1. Update the package from `SystemML` to `systemml`.  The entire `src/main/python` folder was added from scratch very recently with a new API that is not backwards compatible anyway, and the `mllearn` and `mlpipeline` components were also added recently after the previous release, we won't break anything, and will be more Pythonic in the process.
2. Remove the `MLResults.getDataFrame` and replace usages of `getDataFrame("matrixName")` with `get("matrixName").toDF()`.  This way, the API stays clean and simple for an end user, and will help with the upcoming merge of `Matrix` and `matrix`.  Keeping only `get()` makes it simple to get a mixture of datatypes out of a DML script at the same time. [See below example]
3. Remove the `MLResults.getNumPyArray`, add a `Matrix.toArray()` function on the Java and Python sides (returning a list in Python), add a `Matrix.toNumPyArray()` on the Python side, and replace usages of `getNumPyArray("matrixName")` with `get("matrixName").toNumPyArray()`.  @deroneriksson should be able to assist with adding this on the Java side.  Once again, this will keep the API clean, help the upcoming merge, and keeping only `get()` makes it simple to get a mixture of datatypes out of a DML script at the same time.
4.  Merge `Matrix` with `matrix`, keeping the later name.


Overall, I'm excited for the direction the Python API is moving! 👍 


Example of using `get()` for mixed datatypes (scalars, matrices, strings, etc):

```python
from SystemML import MLContext, dml

rdd1 = sc.parallelize(["1.0,2.0", "3.0,4.0"])

sums = """
s1 = sum(m1)
m2 = m1 * 2
t = "whatever"
"""

ml = MLContext(sc)
script = dml(sums).input(m1=rdd1).out("s1", "m2", "t")
s1, m2, t = ml.execute(script).get("s1", "m2", "t")
s1, m2, t
```
Returns:
```python
(10.0, Matrix, 'whatever')
```

Then, 
```python
m2.toDF()
```
Returns
```python
DataFrame[ID: double, C1: double, C2: double]
````

niketanpansare:
+1 for all 4 points. Two comments on "Update the package from SystemML to systemml": 

- Let's send an email on the dev mailing list to notify everyone of this change once the PR is created.
- The PR should also update the existing documentation and examples.

niketanpansare:
@dusenberrymw Updated mlcontext.py as per your suggestions. Please let me know if there are any more changes.

dusenberrymw:
@niketanpansare Awesome, **LGTM**.  Let's merge this, and create JIRAs + PRs to cover the remaining 4 items above, and the remaining set of miscellaneous comments throughout this PR.  After those two steps, let's focus on the important aspect of expanding and hardening each of the APIs (matrix, mllearn, mlpipeline, MLContext), i.e. getting complete feature parity with NumPy matrices, etc.

+1 for the documentation page you created too 👍 

Also, when you merge this, can you append the line `Closes #197.` to the end of the commit message so that this PR and the commit will be automatically linked together?

PR #201, opened 2016-08-01
[SYSTEMML-838] Spark Shell docs for new MLContext

deroneriksson:
Updated Spark Shell documentation for new MLContext API.

Preview [here](http://deroneriksson.github.io/incubator-systemml/spark-mlcontext-programming-guide.html).

### DIFF
--- a/docs/spark-mlcontext-programming-guide.md
+++ b/docs/spark-mlcontext-programming-guide.md
@@ -58,6 +58,14 @@ should see a "`Welcome to Apache SystemML!`" message.
 
 <div class="codetabs">
 
+<div data-lang="Scala" markdown="1">
+{% highlight scala %}
+import org.apache.sysml.api.mlcontext._
+import org.apache.sysml.api.mlcontext.ScriptFactory._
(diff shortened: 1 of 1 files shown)


dusenberrymw:
In general, this looks great!  Two quick things:

1. Can we swap the "Spark Shell" and "Statements" tabs, and show the "Statements" content by default?  Additionally, can we rename "Statements" to "Scala", in anticipation or also including Python syntax at some point?
2. Let's remove the Java example -- data scientists won't be using Java, and so it would be a good opportunity to slim down the docs.

deroneriksson:
Thank you for reviewing @dusenberrymw 

1. I agree, Statements tab as the default increases readability. I'll make this change. Also I'll change Statements to Scala.
2. Agreed about the Java example. I'll remove it for now. We can create a new Java page if users express a need for it.

deroneriksson:
Made requested updates. Preview [here](http://deroneriksson.github.io/incubator-systemml/spark-mlcontext-programming-guide.html).

dusenberrymw:
LGTM.

PR #204, opened 2016-08-05
[SYSTEMML-234] [SYSTEMML-208] Added mllearn library to support scikit-learn and MLPipeline

niketanpansare:
1. Fixed bugs in scala LogisticRegression wrapper (handling of raw predictions and passing dfam).

2. Extended java MLContext to accept MatrixBlock. Also added utility function in Python file. (Also created https://issues.apache.org/jira/browse/SYSTEMML-846. It should be a good migration task to learn Python MLContext).

3. Added mllearn class to allow scikit-learn and MLPipeline users to use SystemML. 

### Using SystemML's Logistic Regression (scikit-learn way):
``` python
from sklearn import datasets, neighbors
import SystemML as sml
from pyspark.sql import SQLContext
sqlCtx = SQLContext(sc)
digits = datasets.load_digits()
X_digits = digits.data
y_digits = digits.target + 1
n_samples = len(X_digits)
X_train = X_digits[:.9 * n_samples]
y_train = y_digits[:.9 * n_samples]
X_test = X_digits[.9 * n_samples:]
y_test = y_digits[.9 * n_samples:]
import time
t1 = time.time()
logistic = sml.mllearn.LogisticRegression(sqlCtx)
print('LogisticRegression score: %f' % logistic.fit(X_train, y_train).score(X_test, y_test))
t2 = time.time()
# Convert to DataFrame for i/o: current way to transfer data
logistic = sml.mllearn.LogisticRegression(sqlCtx, transferUsingDF=True)
print('LogisticRegression score: %f' % logistic.fit(X_train, y_train).score(X_test, y_test))
t3 = time.time()
print('Execution time: without DF: %f and with DF: %f' % (t2-t1, t3-t2))
```

Points to note:

- The execution time without DF is 1-2 seconds whereas with DF is 18-20 seconds.
- The current version only supports numeric labels/features.
- The above interface is especially useful if the input/output fit on the node, but the intermediate data doesn't.

### Using SystemML's Logistic Regression (MLPipeline way):
``` python
from pyspark.ml import Pipeline
import SystemML as sml
from pyspark.ml.feature import HashingTF, Tokenizer
from pyspark.sql import SQLContext
sqlCtx = SQLContext(sc)
training = sqlCtx.createDataFrame([
    (0L, "a b c d e spark", 1.0),
    (1L, "b d", 2.0),
    (2L, "spark f g h", 1.0),
    (3L, "hadoop mapreduce", 2.0),
    (4L, "b spark who", 1.0),
    (5L, "g d a y", 2.0),
    (6L, "spark fly", 1.0),
    (7L, "was mapreduce", 2.0),
    (8L, "e spark program", 1.0),
    (9L, "a e c l", 2.0),
    (10L, "spark compile", 1.0),
    (11L, "hadoop software", 2.0)
], ["id", "text", "label"])
tokenizer = Tokenizer(inputCol="text", outputCol="words")
hashingTF = HashingTF(inputCol="words", outputCol="features", numFeatures=20)
lr = sml.mllearn.LogisticRegression(sqlCtx)
pipeline = Pipeline(stages=[tokenizer, hashingTF, lr])
model = pipeline.fit(training)
test = sqlCtx.createDataFrame([
    (12L, "spark i j k"),
    (13L, "l m n"),
    (14L, "mapreduce spark"),
    (15L, "apache hadoop")], ["id", "text"])
prediction = model.transform(test)
prediction.show()
```

### Using SystemML's Linear Regression (scikit-learn way):
``` python
import numpy as np
from sklearn import datasets
import SystemML as sml
from pyspark.sql import SQLContext
# Load the diabetes dataset
diabetes = datasets.load_diabetes()
# Use only one feature
diabetes_X = diabetes.data[:, np.newaxis, 2]
# Split the data into training/testing sets
diabetes_X_train = diabetes_X[:-20]
diabetes_X_test = diabetes_X[-20:]
# Split the targets into training/testing sets
diabetes_y_train = diabetes.target[:-20]
diabetes_y_test = diabetes.target[-20:]
# Create linear regression object
regr = sml.mllearn.LinearRegression(sqlCtx)
# Train the model using the training sets
regr.fit(diabetes_X_train, diabetes_y_train)
# The mean square error
print("Residual sum of squares: %.2f" % np.mean((regr.predict(diabetes_X_test) - diabetes_y_test) ** 2))
```

### Using SystemML's SVM (scikit-learn way):
``` python
from sklearn import datasets, neighbors
import SystemML as sml
from pyspark.sql import SQLContext
sqlCtx = SQLContext(sc)
digits = datasets.load_digits()
X_digits = digits.data
y_digits = digits.target 
n_samples = len(X_digits)
X_train = X_digits[:.9 * n_samples]
y_train = y_digits[:.9 * n_samples]
X_test = X_digits[.9 * n_samples:]
y_test = y_digits[.9 * n_samples:]
svm = sml.mllearn.SVM(sqlCtx, is_multi_class=True)
print('LogisticRegression score: %f' % svm.fit(X_train, y_train).score(X_test, y_test))
```

### Using SystemML's SVM (MLPipeline way):
``` python
from pyspark.ml import Pipeline
import SystemML as sml
from pyspark.ml.feature import HashingTF, Tokenizer
from pyspark.sql import SQLContext
sqlCtx = SQLContext(sc)
training = sqlCtx.createDataFrame([
    (0L, "a b c d e spark", 1.0),
    (1L, "b d", 2.0),
    (2L, "spark f g h", 1.0),
    (3L, "hadoop mapreduce", 2.0),
    (4L, "b spark who", 1.0),
    (5L, "g d a y", 2.0),
    (6L, "spark fly", 1.0),
    (7L, "was mapreduce", 2.0),
    (8L, "e spark program", 1.0),
    (9L, "a e c l", 2.0),
    (10L, "spark compile", 1.0),
    (11L, "hadoop software", 2.0)
], ["id", "text", "label"])
tokenizer = Tokenizer(inputCol="text", outputCol="words")
hashingTF = HashingTF(inputCol="words", outputCol="features", numFeatures=20)
svm = sml.mllearn.SVM(sqlCtx, is_multi_class=True)
pipeline = Pipeline(stages=[tokenizer, hashingTF, svm])
model = pipeline.fit(training)
test = sqlCtx.createDataFrame([
    (12L, "spark i j k"),
    (13L, "l m n"),
    (14L, "mapreduce spark"),
    (15L, "apache hadoop")], ["id", "text"])
prediction = model.transform(test)
prediction.show()
```

### Using SystemML's Naive Bayes (Scikit learn way):
``` python
from sklearn.datasets import fetch_20newsgroups
from sklearn.feature_extraction.text import TfidfVectorizer
import SystemML as sml
from sklearn import metrics
from pyspark.sql import SQLContext
sqlCtx = SQLContext(sc)
categories = ['alt.atheism', 'talk.religion.misc', 'comp.graphics', 'sci.space']
newsgroups_train = fetch_20newsgroups(subset='train', categories=categories)
newsgroups_test = fetch_20newsgroups(subset='test', categories=categories)
vectorizer = TfidfVectorizer()
# Both vectors and vectors_test are SciPy CSR matrix
vectors = vectorizer.fit_transform(newsgroups_train.data)
vectors_test = vectorizer.transform(newsgroups_test.data)
nb = sml.mllearn.NaiveBayes(sqlCtx)
nb.fit(vectors, newsgroups_train.target)
pred = nb.predict(vectors_test)
metrics.f1_score(newsgroups_test.target, pred, average='weighted')
```

### DIFF
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -35,7 +35,8 @@ from pyspark.ml.feature import VectorAssembler
 from pyspark.mllib.linalg import Vectors
 import sys
 from pyspark.ml import Estimator, Model
-
+from scipy.sparse import spmatrix
+from scipy.sparse import coo_matrix
 
 class MLContext(object):
 
@@ -269,7 +270,19 @@ def getNumCols(numPyArr):
         return numPyArr.shape[1]
        
 def convertToMatrixBlock(sc, src):
-    if isinstance(sc, SparkContext):
+    if isinstance(src, spmatrix):
+        src = coo_matrix(src,  dtype=np.float64)
+        numRows = src.shape[0]
+        numCols = src.shape[1]
+        data = src.data.astype(np.float64)
+        row = src.row.astype(np.int32)
+        col = src.col.astype(np.int32)
+        nnz = len(src.col)
+        buf1 = bytearray(data.tostring())
+        buf2 = bytearray(row.tostring())
+        buf3 = bytearray(col.tostring())
+        return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertSciPyCOOToMB(buf1, buf2, buf3, numRows, numCols, nnz)
+    elif isinstance(sc, SparkContext):
         src = np.asarray(src)
         numCols = getNumCols(src)
         numRows = src.shape[0]
@@ -319,7 +332,7 @@ class mllearn:
         def fit(self, X, y=None, params=None):
             if y is None:
                 return self._fit(X)
-            elif y is not None and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame)):
+            elif y is not None and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame) or isinstance(X, spmatrix)):
                 if self.transferUsingDF:
                     pdfX = convertToPandasDF(X)
                     pdfY = convertToPandasDF(y)
@@ -346,7 +359,7 @@ class mllearn:
             return self.predict(X)
             
         def predict(self, X):
-            if isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame):
+            if isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame) or isinstance(X, spmatrix):
                 if self.transferUsingDF:
                     pdfX = convertToPandasDF(X)
                     df = assemble(self.sqlCtx, pdfX, pdfX.columns, 'features').select('features')
@@ -442,5 +455,15 @@ class mllearn:
             self.estimator.setTol(tol)
             self.estimator.setIcpt(int(fit_intercept))
             self.transferUsingDF = transferUsingDF
-            self.setOutputRawPredictionsToFalse = False            
-                
+            self.setOutputRawPredictionsToFalse = False    
+
+    class NaiveBayes(BaseSystemMLEstimator):
+
+        def __init__(self, sqlCtx, laplace=1.0, transferUsingDF=False):
+            self.sqlCtx = sqlCtx
+            self.sc = sqlCtx._sc
+            self.uid = "nb"
+            self.estimator = self.sc._jvm.org.apache.sysml.api.ml.NaiveBayes(self.uid, self.sc._jsc.sc())
+            self.estimator.setLaplace(laplace)
+            self.transferUsingDF = transferUsingDF
+            self.setOutputRawPredictionsToFalse = False                
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -259,24 +259,19 @@ class MLOutput(object):
         #    traceback.print_exc()
 
 def getNumCols(numPyArr):
-	if len(numPyArr.shape) == 1:
+	if numPyArr.ndim == 1:
 		return 1
 	else:
 		return numPyArr.shape[1]
        
 def convertToJavaMatrix(sc, src):
-	if isinstance(src, np.ndarray):
-		from array import array
-		numCols = getNumCols(src)
-		numRows = src.shape[0]
-		if src.dtype.type is np.float64:
-			arr = src.reshape(-1)
-		else:
-			arr = array('d', src.reshape(-1))
-		buf = bytearray(arr.tostring())
-		return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertPy4JArrayToMB(buf, numRows, numCols)
-	else:
-		raise Exception('Type is not supported')
+	src = np.asarray(src)
+	numCols = getNumCols(src)
+	numRows = src.shape[0]
+	arr = src.ravel().astype(np.float64)
+	buf = bytearray(arr.tostring())
+	return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertPy4JArrayToMB(buf, numRows, numCols)
+	
 
 def convertToNumpyArr(sc, mb):
 	numRows = mb.getNumRows()
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -33,6 +33,7 @@ from pyspark.ml.feature import VectorAssembler
 from pyspark.mllib.linalg import Vectors
 import sys
 from pyspark.ml import Estimator, Model
+from __future__ import division
 
 class MLContext(object):
 
@@ -303,7 +304,7 @@ class mllearn:
             self.tol = tol
             if C < 0:
                 raise Exception('C has to be positive')
-            reg = 1/C
+            reg = 1.0 / C
             self.reg = reg
             self.updateLog()
             if solver != 'newton-cg':
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -302,7 +302,7 @@ class mllearn:
             self.max_iter = max_iter
             self.max_inner_iter = max_inner_iter
             self.tol = tol
-            if C < 0:
+            if C <= 0:
                 raise Exception('C has to be positive')
             reg = 1.0 / C
             self.reg = reg
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -288,83 +288,55 @@ def convertToNumpyArr(sc, mb):
     else:
         raise TypeError('sc needs to be of type SparkContext') # TODO: We can generalize this by creating py4j gateway ourselves
 
-class mllearn:
-    # Or we can create new Python project with package structure
-    class LogisticRegression(Estimator):
+def convertToPandasDF(X):
+    if not isinstance(X, pd.DataFrame):
+        return pd.DataFrame(X, columns=['C' + str(i) for i in range(getNumCols(X))])
+    return X
+            
+def tolist(inputCols):
+    if isinstance(inputCols, pd.indexes.base.Index):
+        return inputCols.get_values().tolist()
+    elif isinstance(inputCols, list):
+        return inputCols
+    else:
+        raise Exception('inputCols should be of type pandas.indexes.base.Index or list')
 
-        def __init__(self, sqlCtx, penalty='l2', fit_intercept=True, max_iter=100, max_inner_iter=0, tol=0.000001, C=1.0, solver='newton-cg', transferUsingDF=False):
-            self.sqlCtx = sqlCtx
-            self.sc = sqlCtx._sc
-            self.log = self.sc._jvm.org.apache.sysml.api.ml.LogisticRegression("lr", self.sc._jsc.sc())
-            self.transferUsingDF = transferUsingDF
-            if penalty != 'l2':
-                raise Exception('Only l2 penalty is supported')
-            self.icpt = int(fit_intercept)
-            self.max_iter = max_iter
-            self.max_inner_iter = max_inner_iter
-            self.tol = tol
-            if C <= 0:
-                raise Exception('C has to be positive')
-            reg = 1.0 / C
-            self.reg = reg
-            self.updateLog()
-            if solver != 'newton-cg':
-                raise Exception('Only newton-cg solver supported')
-             
-        def updateLog(self):
-            self.log.setMaxOuterIter(self.max_iter)
-            self.log.setMaxInnerIter(self.max_inner_iter) 
-            self.log.setRegParam(self.reg)
-            self.log.setTol(self.tol)
-            self.log.setIcpt(self.icpt)
-            
-        def convertToPandasDF(self, X):
-            if not instance(X, pd.DataFrame):
-                return pd.DataFrame(X, columns=['C' + str(i) for i in range(numCols)])
-            return X
-            
-        def tolist(self, inputCols):
-            if isinstance(inputCols, pd.indexes.base.Index):
-                return inputCols.get_values().tolist()
-            elif isinstance(inputCols, list):
-                return inputCols
-            else:
-                raise Exception('inputCols should be of type pandas.indexes.base.Index or list')
-                
-        def assemble(self, pdf, inputCols, outputCol):
-            tmpDF = self.sqlCtx.createDataFrame(pdf, self.tolist(pdf.columns))
-            assembler = VectorAssembler(inputCols=self.tolist(inputCols), outputCol=outputCol)
-            return assembler.transform(tmpDF)
-            
+def assemble(sqlCtx, pdf, inputCols, outputCol):
+    tmpDF = sqlCtx.createDataFrame(pdf, tolist(pdf.columns))
+    assembler = VectorAssembler(inputCols=tolist(inputCols), outputCol=outputCol)
+    return assembler.transform(tmpDF)
+
+class mllearn:
+    class BaseSystemMLEstimator(Estimator):
         def _fit(self, X):
             if hasattr(X, '_jdf') and 'features' in X.columns and 'label' in X.columns:
-                self.model = self.log.fit(X._jdf)
+                self.model = self.estimator.fit(X._jdf)
                 return self
             else:
                 raise Exception('Incorrect usage: Expected dataframe as input with features/label as columns')
             
         def fit(self, X, y=None, params=None):
-            self.updateLog()
             if y is None:
                 return self._fit(X)
             elif y is not None and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame)):
                 if self.transferUsingDF:
-                    pdfX = self.convertToPandasDF(X)
-                    pdfY = self.convertToPandasDF(y)
+                    pdfX = convertToPandasDF(X)
+                    pdfY = convertToPandasDF(y)
                     if getNumCols(pdfY) != 1:
                         raise Exception('y should be a column vector')
                     if pdfX.shape[0] != pdfY.shape[0]:
                         raise Exception('Number of rows of X and y should match')
                     colNames = pdfX.columns
                     pdfX['label'] = pdfY[pdfY.columns[0]]
-                    df = self.assemble(pdfX, colNames, 'features').select('features', 'label')
-                    self.model = self.log.fit(df._jdf)
+                    df = assemble(self.sqlCtx, pdfX, colNames, 'features').select('features', 'label')
+                    self.model = self.estimator.fit(df._jdf)
                 else:
                     numColsy = getNumCols(y)
                     if numColsy != 1:
                         raise Exception('Expected y to be a column vector')
-                    self.model = self.log.fit(convertToMatrixBlock(self.sc, X), convertToMatrixBlock(self.sc, y))
-                self.model.setOutputRawPredictions(False)
+                    self.model = self.estimator.fit(convertToMatrixBlock(self.sc, X), convertToMatrixBlock(self.sc, y))
+                if self.setOutputRawPredictionsToFalse:
+                    self.model.setOutputRawPredictions(False)
                 return self
             else:
                 raise Exception('Unsupported input type')
@@ -375,8 +347,8 @@ class mllearn:
         def predict(self, X):
             if isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame):
                 if self.transferUsingDF:
-                    pdfX = self.convertToPandasDF(X)
-                    df = self.assemble(pdfX, pdfX.columns, 'features').select('features')
+                    pdfX = convertToPandasDF(X)
+                    df = assemble(self.sqlCtx, pdfX, pdfX.columns, 'features').select('features')
                     retjDF = self.model.transform(df._jdf)
                     retDF = DataFrame(retjDF, self.sqlCtx)
                     retPDF = retDF.sort('ID').select('prediction').toPandas()
@@ -405,5 +377,49 @@ class mllearn:
                 raise Exception('Unsupported input type')
                 
         def score(self, X, y):
-            return sk.metrics.accuracy_score(y, self.predict(X))
+            return sk.metrics.accuracy_score(y, self.predict(X))    
+    
+    # Or we can create new Python project with package structure
+    class LogisticRegression(BaseSystemMLEstimator):
+
+        def __init__(self, sqlCtx, penalty='l2', fit_intercept=True, max_iter=100, max_inner_iter=0, tol=0.000001, C=1.0, solver='newton-cg', transferUsingDF=False):
+            self.sqlCtx = sqlCtx
+            self.sc = sqlCtx._sc
+            self.uid = "logReg"
+            self.estimator = self.sc._jvm.org.apache.sysml.api.ml.LogisticRegression(self.uid, self.sc._jsc.sc())
+            self.estimator.setMaxOuterIter(max_iter)
+            self.estimator.setMaxInnerIter(max_inner_iter)
+            if C <= 0:
+                raise Exception('C has to be positive')
+            reg = 1.0 / C
+            self.estimator.setRegParam(reg)
+            self.estimator.setTol(tol)
+            self.estimator.setIcpt(int(fit_intercept))
+            self.transferUsingDF = transferUsingDF
+            self.setOutputRawPredictionsToFalse = True
+            if penalty != 'l2':
+                raise Exception('Only l2 penalty is supported')
+            if solver != 'newton-cg':
+                raise Exception('Only newton-cg solver supported')
+
+    class LinearRegression(BaseSystemMLEstimator):
+
+        def __init__(self, sqlCtx, fit_intercept=True, max_iter=100, tol=0.000001, C=1.0, solver='newton-cg', transferUsingDF=False):
+            self.sqlCtx = sqlCtx
+            self.sc = sqlCtx._sc
+            self.uid = "lr"
+            if solver == 'newton-cg' or solver == 'direct-solve':
+                self.estimator = self.sc._jvm.org.apache.sysml.api.ml.LinearRegression(self.uid, self.sc._jsc.sc(), solver)
+            else:
+                raise Exception('Only newton-cg solver supported')
+            self.estimator.setMaxIter(max_iter)
+            if C <= 0:
+                raise Exception('C has to be positive')
+            reg = 1.0 / C
+            self.estimator.setRegParam(reg)
+            self.estimator.setTol(tol)
+            self.estimator.setIcpt(int(fit_intercept))
+            self.transferUsingDF = transferUsingDF
+            self.setOutputRawPredictionsToFalse = False
+            
                 
--- a/src/main/java/org/apache/sysml/api/python/SystemML.py
+++ b/src/main/java/org/apache/sysml/api/python/SystemML.py
@@ -321,13 +321,15 @@ class mllearn:
     class BaseSystemMLEstimator(Estimator):
     # TODO: Allow users to set featuresCol (with default 'features') and labelCol (with default 'label')
     
+        # Returns a model after calling fit(df) on Estimator object on JVM    
         def _fit(self, X):
             if hasattr(X, '_jdf') and 'features' in X.columns and 'label' in X.columns:
                 self.model = self.estimator.fit(X._jdf)
                 return self
             else:
                 raise Exception('Incorrect usage: Expected dataframe as input with features/label as columns')
-            
+        
+        # Returns a model after calling fit(X:MatrixBlock, y:MatrixBlock) on Estimator object on JVM  
         def fit(self, X, y=None, params=None):
             if y is None:
                 return self._fit(X)
@@ -356,7 +358,8 @@ class mllearn:
         
         def transform(self, X):
             return self.predict(X)
-            
+        
+        # Returns either a DataFrame or MatrixBlock after calling transform(X:MatrixBlock, y:MatrixBlock) on Model object on JVM    
         def predict(self, X):
             if isinstance(X, SUPPORTED_TYPES):
                 if self.transferUsingDF:
@@ -389,12 +392,23 @@ class mllearn:
             else:
                 raise Exception('Unsupported input type')
                 
+    class BaseSystemMLClassifier(BaseSystemMLEstimator):
+
+        # Scores the predicted value with ground truth 'y'
         def score(self, X, y):
             return metrics.accuracy_score(y, self.predict(X))    
     
-    # Or we can create new Python project with package structure
-    class LogisticRegression(BaseSystemMLEstimator):
+    class BaseSystemMLRegressor(BaseSystemMLEstimator):
 
+        # Scores the predicted value with ground truth 'y'
+        def score(self, X, y):
+            return metrics.r2_score(y, self.predict(X), multioutput='variance_weighted')
+
+    
+    # Or we can create new Python project with package structure
+    class LogisticRegression(BaseSystemMLClassifier):
+
+        # See https://apache.github.io/incubator-systemml/algorithms-reference for usage
         def __init__(self, sqlCtx, penalty='l2', fit_intercept=True, max_iter=100, max_inner_iter=0, tol=0.000001, C=1.0, solver='newton-cg', transferUsingDF=False):
             self.sqlCtx = sqlCtx
             self.sc = sqlCtx._sc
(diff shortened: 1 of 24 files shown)


niketanpansare:
@mboehm7 @frreiss @bertholdreinwald @dusenberrymw @deroneriksson @MechCoder @iyounus

MechCoder:
Hey, the script looks great! But I may be biased as a sklearn user and developer.

What are you planning to do with the previous PR? In other words, what are the use-cases of exposing SystemML datastructures to the not-so-advanced user?

niketanpansare:
@MechCoder Please see https://github.com/apache/incubator-systemml/pull/197 for answer to your question. Also, since you are biased sklearn user/developer, you will be the best person to critique the API of our library :) ... The high-level pitch we can make as SystemML community is that we support subset of sklearn algorithm and if your application is using these sklearn algorithms, you can replace `sk._` call with `sml._` call and everything should work as expected.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
         #    return rdd
         #except Py4JJavaError:
         #    traceback.print_exc()
+
+def getNumCols(numPyArr):
+	if len(numPyArr.shape) == 1:
```
MechCoder: `np.ndim` is the preferred way to do this. Btw, in 1-D arrays, the number of elements are generally regarded as the number of features.

(Also in sklearn, we deprecated the use of 1-D arrays. All data should be provided as a 2-D array)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+		numCols = getNumCols(src)
+		numRows = src.shape[0]
+		if src.dtype.type is np.float64:
+			arr = src.reshape(-1)
+		else:
+			arr = array('d', src.reshape(-1))
+		buf = bytearray(arr.tostring())
+		return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertPy4JArrayToMB(buf, numRows, numCols)
+	else:
+		raise Exception('Type is not supported')
```
MechCoder: You can do `src = np.asarray(src)` in the beginning to make sure data structures like list of lists etc are supported.

MechCoder: Also, you should use more specific errors like `ValueError` etc.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+       
+def convertToJavaMatrix(sc, src):
+	if isinstance(src, np.ndarray):
+		from array import array
+		numCols = getNumCols(src)
+		numRows = src.shape[0]
+		if src.dtype.type is np.float64:
+			arr = src.reshape(-1)
+		else:
+			arr = array('d', src.reshape(-1))
```
MechCoder: You can just do

```python
src.ravel().astype(np.float64)
```
(assuming that was your intention)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+		buf = bytearray(arr.tostring())
```
MechCoder: The `tobytes` method does not work?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
         #except Py4JJavaError:
         #    traceback.print_exc()
+
+def getNumCols(numPyArr):
+	if len(numPyArr.shape) == 1:
+		return 1
+	else:
+		return numPyArr.shape[1]
+       
+def convertToJavaMatrix(sc, src):
```
MechCoder: Would you want to check if sc is a SparkContext instance to raise a more meaningful error?

MechCoder: Looking at the return type, do you think `convertToMatrixBlock` might be more appropriate for the function name?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+                raise Exception('Only newton-cg solver supported')
+             
+        def updateLog(self):
+            self.log.setMaxOuterIter(self.max_iter)
+            self.log.setMaxInnerIter(self.max_inner_iter) 
+            self.log.setRegParam(self.reg)
+            self.log.setTol(self.tol)
+            self.log.setIcpt(self.icpt)
+            
+        def convertToPDF(self, X):
```
MechCoder: A slightly more explicit name might be better. It kind of reminds me of [this](https://www.freepdfconvert.com/) ;)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+            if penalty != 'l2':
+                raise Exception('Only l2 penalty is supported')
+            if fit_intercept:
+                self.icpt = 1
+            else:
+                self.icpt = 0
+            self.max_iter = max_iter
+            self.max_inner_iter = max_inner_iter
+            self.tol = tol
+            if C == 0:
```
MechCoder: `C < 0`?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,163 @@ def getStringRDD(self, varName, format):
...
+            self.sqlCtx = sqlCtx
+            self.sc = sqlCtx._sc
+            self.log = self.sc._jvm.org.apache.sysml.api.ml.LogisticRegression("lr", self.sc._jsc.sc())
+            self.transferUsingDF = transferUsingDF
+            if penalty != 'l2':
+                raise Exception('Only l2 penalty is supported')
+            if fit_intercept:
+                self.icpt = 1
+            else:
+                self.icpt = 0
```
MechCoder: `self.icpt = int(fit_intercept)

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Done.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Done.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +257,158 @@ def getStringRDD(self, varName, format):
...
+            self.max_iter = max_iter
+            self.max_inner_iter = max_inner_iter
+            self.tol = tol
+            if C == 0:
+                raise Exception('C cannot be 0')
+            reg = 1/C
```
MechCoder: `reg = 1.0 / C` just in case the user uses Python2 and decides to give `C` as an integer.

dusenberrymw: As an alternative, we could use `from __future__ import division` to prevent this issue from creeping in.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -290,15 +297,12 @@ def __init__(self, sqlCtx, penalty='l2', fit_intercept=True, max_iter=100, max_i
...
-                self.icpt = 1
-            else:
-                self.icpt = 0
+            self.icpt = int(fit_intercept)
             self.max_iter = max_iter
             self.max_inner_iter = max_inner_iter
             self.tol = tol
-            if C == 0:
-                raise Exception('C cannot be 0')
+            if C < 0:
```
dusenberrymw: I think this needs to be `if C <= 0` to prevent division by 0.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -259,25 +260,31 @@ def getStringRDD(self, varName, format):
...
+def convertToMatrixBlock(sc, src):
+    if isinstance(sc, SparkContext):
+        src = np.asarray(src)
+        numCols = getNumCols(src)
+        numRows = src.shape[0]
+        arr = src.ravel().astype(np.float64)
+        buf = bytearray(arr.tostring())
+        return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertPy4JArrayToMB(buf, numRows, numCols)
+    else:
+        raise TypeError('sc needs to be of type SparkContext') # TODO: We can generalize this by creating py4j gateway ourselves
```
dusenberrymw: I say we stay tied to Spark for now as the end user doesn't gain anything using SystemML in a standalone mode, as opposed to running it in Spark local mode on a laptop.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+            self.max_inner_iter = max_inner_iter
+            self.tol = tol
+            if C < 0:
+                raise Exception('C has to be positive')
+            reg = 1.0 / C
+            self.reg = reg
+            self.updateLog()
+            if solver != 'newton-cg':
+                raise Exception('Only newton-cg solver supported')
+
```
MechCoder: I think we can start refactoring this into a `BaseSystemMLEstimator`

1. We can rename `self.log` to `self.model` and make it a property in the `BaseEstimator` and in the inherited classes we can define a setter that sets this property (http://stackoverflow.com/a/1684851/1170730)
2. Is it possible to set all these parameters of the base SystemML java model together right now? If yes, then we should define a `get_params` and `set_params` that updates the base model with these parameters.

MechCoder: Note that I am reading from top to bottom, so not sure if this is applicable or not.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+            self.sc = sqlCtx._sc
+            self.log = self.sc._jvm.org.apache.sysml.api.ml.LogisticRegression("lr", self.sc._jsc.sc())
+            self.transferUsingDF = transferUsingDF
+            if penalty != 'l2':
+                raise Exception('Only l2 penalty is supported')
+            self.icpt = int(fit_intercept)
+            self.max_iter = max_iter
+            self.max_inner_iter = max_inner_iter
+            self.tol = tol
+            if C < 0:
```
dusenberrymw: I think this needs to be `if C <= 0` to prevent division by 0.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: I agree. The first iteration will support on Spark.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Oops. Somehow slipped through, even though the message is correct :)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+            self.log.setTol(self.tol)
+            self.log.setIcpt(self.icpt)
+            
+        def convertToPandasDF(self, X):
+            if isinstance(X, np.ndarray):
+                colNames = []
+                numCols = getNumCols(X)
+                for i in range(0, numCols):
+                    colNames = colNames + [ str('C' + str(i))]
+                pdfX = pd.DataFrame(X, columns=colNames)
```
MechCoder: Are there reasons why you prefer to have each column name starting with `C`? If not, then this whole method seems like just `pd.DataFrame(x)` here.

Also, generally list comprehensions are preferred in Python
`['C' + str(i) for i in range(numCols)]`

MechCoder: It is safe to say that you can move this method out of this class, nothing in the implementation seems to suggest it belongs to LogisticRegression

niketanpansare: For consistency with the naming convention of Java MLContext.

MechCoder: All right. :)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+        def _fit(self, X):
+            if hasattr(X, '_jdf') and 'features' in X.columns and 'label' in X.columns:
+                self.model = self.log.fit(X._jdf)
+                return self
+            else:
+                raise Exception('Incorrect usage: Expected dataframe as input with features/label as columns')
+            
+        # TOOD: Ignoring kwargs
+        def fit(self, X, *args, **kwargs):
+            self.updateLog()
```
MechCoder: Why `update` again?

MechCoder:
Would you be able to add some minor tests? Thanks!

mboehm7:
could we please batch these comments?

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+            numArgs = len(args) + 1
+            if numArgs == 1:
+                return self._fit(X)
+            elif numArgs == 2 and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame)):
```
MechCoder: Why not just change the signature to `fit(X, y=None)` and remove `args`?

MechCoder:
@mboehm7 Sorry, but what are batched comments?

dusenberrymw:
@mboehm7 Inline comments at specific lines of code in the PR are super useful for reviewing and discussing the code without any confusion.  Unless you're "Watching" the entire repo, "Unsubscribing" from this particular PR should limit the inbox noise. :)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+                colNames = []
+                numCols = getNumCols(X)
+                for i in range(0, numCols):
+                    colNames = colNames + [ str('C' + str(i))]
+                pdfX = pd.DataFrame(X, columns=colNames)
+            elif isinstance(X, pd.core.frame.DataFrame):
+                pdfX = X
+            else:
+                raise Exception('The input type not supported')
+            return pdfX
```
MechCoder: I would refactor this entire method this way in any case

```python
# Let Pandas handle the conversion error internally and allow other array-like formats
if not instance(X, pd.DataFrame):
    return pd.DataFrame(X, columns=['C' + str(i) for i in range(numCols)])
return X
```

mboehm7:
thanks @dusenberrymw.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +259,161 @@ def getStringRDD(self, varName, format):
...
+                y = args[0]
+                if self.transferUsingDF:
+                    pdfX = self.convertToPandasDF(X)
+                    pdfY = self.convertToPandasDF(y)
+                    if getNumCols(pdfY) != 1:
+                        raise Exception('y should be a column vector')
+                    if pdfX.shape[0] != pdfY.shape[0]:
+                        raise Exception('Number of rows of X and y should match')
+                    colNames = pdfX.columns
+                    pdfX['label'] = pdfY[pdfY.columns[0]]
```
MechCoder: The current behaviour explicitly renames the columns of `y` to label even if the user provides `y` as a DataFrame. Are we okay with that?

niketanpansare:
>> Would you be able to add some minor tests? Thanks!
Do you have recommendations of how we can add Python tests along with JUnit ?

>>  Why update again?
I added that to test MLPipeline's CrossValidator.

>> Why not just change the signature to fit(X, y=None) and remove args?
Done.

>> Let Pandas handle the conversion error internally and allow other array-like formats
Good point.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +260,150 @@ def getStringRDD(self, varName, format):
...
+                self.model = self.log.fit(X._jdf)
+                return self
+            else:
+                raise Exception('Incorrect usage: Expected dataframe as input with features/label as columns')
+            
+        def fit(self, X, y=None, params=None):
+            self.updateLog()
+            if y is None:
+                return self._fit(X)
+            elif y is not None and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame)):
```
MechCoder: Seems like the check for X is done internally in `convertToPandasDF`

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: Sorry, I read `pdfX` as `pdY`. Deleting this comment.

MechCoder:
@niketanpansare Thanks for addressing my comments!  I made a first pass. Hope to get back to it on Monday.

niketanpansare:
Thanks @MechCoder for your help and suggestions :+1:

niketanpansare:
See documentation for mllearn: http://niketanpansare.github.io/incubator-systemml/algorithms-classification.html#usage

MechCoder:
@niketanpansare Is it possible to just keep one ML model (LogisticRegression) in this PR and keep the rest for another PR after this has been merged. So that the reviewing can be focused on just the design?

niketanpansare:
@MechCoder I would prefer to add other ML models as well in this PR for three reasons:

1. All the other models (including LogisticRegression) inherit their implementation from BaseSystemMLEstimator. So, the focus should be on design of BaseSystemMLEstimator. If we are comfortable with that, it doesn't matter whether the implemented ML model is LogisticRegression or NaiveBayes or any other.

2. Due to overdesigning the initial implementation, very little progress has been made. As a side note, java LogisticRegression API has been there for almost a year and no model has been added since then (java or python).

3. The hope of this API is to increase adoption at least in algorithm front. As a first step towards that, we need to update the Algorithm documentation with interesting examples that motivates people to try SystemML. Adding just one algorithm (without strong reason) seems to defeat the purpose of that.

Please note: the API is WIP, so you are welcome to modify the added ML models or add new ML models once this PR is in :)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+        return pd.DataFrame(X, columns=['C' + str(i) for i in range(getNumCols(X))])
+    return X
+            
+def tolist(inputCols):
+    if isinstance(inputCols, pd.indexes.base.Index):
+        return inputCols.get_values().tolist()
+    elif isinstance(inputCols, list):
+        return inputCols
+    else:
+        raise Exception('inputCols should be of type pandas.indexes.base.Index or list')
```
MechCoder: This whole method is just `list(inputCols)`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+            self.uid = "logReg"
+            self.estimator = self.sc._jvm.org.apache.sysml.api.ml.LogisticRegression(self.uid, self.sc._jsc.sc())
+            self.estimator.setMaxOuterIter(max_iter)
+            self.estimator.setMaxInnerIter(max_inner_iter)
+            if C <= 0:
+                raise Exception('C has to be positive')
+            reg = 1.0 / C
+            self.estimator.setRegParam(reg)
+            self.estimator.setTol(tol)
+            self.estimator.setIcpt(int(fit_intercept))
```
MechCoder: This need not be addressed right now, but in the future it might be good to set all these parameters at once. (I assume that was your initial thought while inheriting from `pyspark.ml.Estimator`, to avoid rewriting the methods provided by `Params`)

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+        return inputCols
+    else:
+        raise Exception('inputCols should be of type pandas.indexes.base.Index or list')
+
+def assemble(sqlCtx, pdf, inputCols, outputCol):
+    tmpDF = sqlCtx.createDataFrame(pdf, tolist(pdf.columns))
+    assembler = VectorAssembler(inputCols=tolist(inputCols), outputCol=outputCol)
+    return assembler.transform(tmpDF)
+
+class mllearn:
```
MechCoder: Let us start having a package structure with this moved to a directory called `mllearn` and the `BaseSystemMLEstimator` moved to `base.py`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+    class BaseSystemMLEstimator(Estimator):
+        def _fit(self, X):
+            if hasattr(X, '_jdf') and 'features' in X.columns and 'label' in X.columns:
```
MechCoder: You should add a TODO here for users to set the name of the `featuresCol` and `labelCols`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+                self.model = self.estimator.fit(X._jdf)
+                return self
+            else:
+                raise Exception('Incorrect usage: Expected dataframe as input with features/label as columns')
```
MechCoder: Can we rename `Exception` to a more useful error everywhere?

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: Also not sure there should be the condition for `label` because `Estimator` is a general name and can include unsupervised algorithms.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Will do.

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: `X` can be a Pandas Dataframe as well, right?

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: I was planning to do that which is why I created `mllearn`. Not sure what is the best way to put it in repo. Infact, I would prefer to package our python project as `pip`. Any thoughts ?

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Good point. I was going by the convention in MLPipeline.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+            
+        def fit(self, X, y=None, params=None):
+            if y is None:
+                return self._fit(X)
+            elif y is not None and (isinstance(X, np.ndarray) or isinstance(X, pd.core.frame.DataFrame) or isinstance(X, spmatrix)):
```
MechCoder: Minor: I suppose we should start having a SUPPORTED_TYPES list defined at the top with = [np.ndarray, pd.DataFrame, spmatrix] and check `isinstance(X, SUPPORTED_TYPES)`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+                if self.transferUsingDF:
```
MechCoder: If your benchmarks suggests that this is 20x slower always, then I believe there is no use of exposing this option to the user.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+        return 1
+    else:
+        return numPyArr.shape[1]
+       
+def convertToMatrixBlock(sc, src):
+    if isinstance(src, spmatrix):
+        src = coo_matrix(src,  dtype=np.float64)
+        numRows = src.shape[0]
+        numCols = src.shape[1]
+        data = src.data.astype(np.float64)
```
MechCoder: minor: data is already casted to `float64`

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+        row = src.row.astype(np.int32)
+        col = src.col.astype(np.int32)
+        nnz = len(src.col)
```
MechCoder: COO matrix has a nnz attribute (http://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.nnz.html#scipy.sparse.coo_matrix.nnz) which I bet just does len(cols) internally :P

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+        buf1 = bytearray(data.tostring())
+        buf2 = bytearray(row.tostring())
+        buf3 = bytearray(col.tostring())
+        return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertSciPyCOOToMB(buf1, buf2, buf3, numRows, numCols, nnz)
```
MechCoder: Correct me if I am wrong but this methods just seems to create a `DenseMatrix` internally. In that case it is better to just raise an Error on the Python side, saying that sparse matrices are not supported or a warning that the sparse matrices are being densified, to reduce the surprise of the user.

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+                    df = assembler.transform(X)
+                retjDF = self.model.transform(df._jdf)
+                retDF = DataFrame(retjDF, self.sqlCtx)
+                # Return DF
+                return retDF.sort('ID')
+            else:
+                raise Exception('Unsupported input type')
+                
+        def score(self, X, y):
+            return metrics.accuracy_score(y, self.predict(X))
```
MechCoder: This suggests that we should maybe start having a `BaseRegressor` and a `BaseClassifier`

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: No, it creates Sparse matrix block at Java side. Please look at https://github.com/apache/incubator-systemml/pull/204/files#diff-bf3788ec75ea2e179c7bc02086e875f2R271

src/main/java/org/apache/sysml/api/python/SystemML.py
```diff
@@ -247,3 +262,208 @@ def getStringRDD(self, varName, format):
...
+                    colNames = pdfX.columns
+                    pdfX['label'] = pdfY[pdfY.columns[0]]
+                    df = assemble(self.sqlCtx, pdfX, colNames, 'features').select('features', 'label')
+                    self.model = self.estimator.fit(df._jdf)
+                else:
+                    numColsy = getNumCols(y)
+                    if numColsy != 1:
+                        raise Exception('Expected y to be a column vector')
+                    self.model = self.estimator.fit(convertToMatrixBlock(self.sc, X), convertToMatrixBlock(self.sc, y))
+                if self.setOutputRawPredictionsToFalse:
```
MechCoder: I really think this should be a parameter to `predict`, because I do not want do another costly fit just to obtain predictions instead of raw labels etc.

Alternatively this can return a wrapper to a wrapper around SystemML's java model, and I should be able to `self.model.setRawPredictions`

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: Oh sorry then.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: The Python part of the code is a thin API and I think having BaseClassifier and BaseRegressor might be overkill. But, I agree we need to have BaseClassifier and BaseRegressor at Scala side as it implements the core logic and have updated code accordingly.

src/main/java/org/apache/sysml/api/python/SystemML.py
niketanpansare: Currently, none of our models return raw predictions. In future, the raw predictions will be computed as part of `fit`. MLPipeline's `fit` will return DataFrame with raw predictions as one of the column. For scikit-learn `fit`, we can either skip computing it or follow your suggestion of wrapping the java model.

niketanpansare: On second thought, it's good idea to create baseclassifier n baseregressor hierarchy and put getprob function into baseclassifier :)

src/main/java/org/apache/sysml/api/python/SystemML.py
MechCoder: I was suggesting this because if we inherit from `BaseSystemMLEstimator`, the default score is the accuracy score which does not apply for non-classifiers.
We should either make this an abstract method and reimplement the score logic for all the models, or make a BaseRegressor etc and avoid reimplenting the logic.

niketanpansare: Fair point. I will do the necessary change in the next commit :+1:

MechCoder:
Before I proceed any further:

1. Can you run PEP* on all the python files?
2. Can you add documentation to all of the methods? Please explicitly document the expected return types of all the methods?
3. With regards to making it pip-installable or even having a `setup.py`, I would postpone that to another PR. There is still some discussion (or non-discussion rather) in making pyspark pip installable. (https://github.com/apache/spark/pull/8318/). Right now how would just focus on organizing the project structure. We can just have it similar to pyspark's project structure.

niketanpansare:
Added documentation as well as created BaseSystemMLClassifier and BaseSystemMLRegressor classes in Python. 

Let's address remaining comments in next PR. For now, I believe this API is in reasonably stable state. Also Spark 2.0 support is dependent on this.

niketanpansare:
https://github.com/apache/incubator-systemml/commit/f02f7c018b221384b5e02b13ac181122853f65ba closes this PR.

src/main/java/org/apache/sysml/api/python/SystemML.py
dusenberrymw: Yeah packaging it up via `pip` would be fantastic.  @deroneriksson, @MechCoder, and I were discussing about updating the current Python `MLContext` to target the new `MLContext` on the JVM side.  As part of that update, it would be nice to restructure our Python code into a proper package.

niketanpansare: Here is the list of related JIRAs. Can we please move the discussion to the JIRA ?
1. https://issues.apache.org/jira/browse/SYSTEMML-854
2. https://issues.apache.org/jira/browse/SYSTEMML-846
3. https://issues.apache.org/jira/browse/SYSTEMML-855

PR #211, opened 2016-08-12
[SYSTEMML-853] Python API for new MLContext

MechCoder:
This is the Python API for the new MLContext

Here is a quick example

```python
sums = """
s1 = sum(m1)
s2 = sum(m2)
s3 = 'whatever'
"""
ml = MLContext(sc)
rdd1 = sc.parallelize(["1.0,2.0", "3.0,4.0"])
rdd2 = sc.parallelize(["5.0,6.0", "7.0,8.0"])
script = dml(sums).input(m1=rdd1).input(m2=rdd2).out("s1", "s2", "s3")
ml.execute(script).get("s1", "s2", "s3")
[10.0, 26.0, "whatever"]
```

The tests cases should explain the other use cases.


python/SystemML.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns
+        -------
+        ml_results: MLResults
+            MLResults instance.
+        """
+        if not isinstance(dml, Dml):
+            raise ValueError("Expected dml to be an instance of Dml")
+        path = dml.path
+        if isinstance(path, str):
+            if os.path.exists(path):
```
frreiss: You might want to add some error-handling code here in case the user mistypes a path. The code could, for example, use a regex to match strings that can only be file paths (because they won't parse as DML), and print "path ${path} not found" if the path doesn't exist.

src/main/java/org/apache/sysml/api/mlcontext/MLResults.java
```diff
@@ -330,9 +330,19 @@ public double getDouble(String outputName) {
 		return so.getDoubleValue();
 	}
 
+	public Object get(String outputName) {
```
frreiss: If this method is intended as a public API method, can you please add appropriate JavaDoc?

python/SystemML.py
```diff
@@ -0,0 +1,154 @@
...
+        outputs: string, list of strings
+            Output variables as defined inside the DML script.
+        """
+        outs = [_java2py(self.sc, self._java_results.get(out)) for out in outputs]
+        if len(outs) == 1:
+            return outs[0]
+        return outs
+
+
+class Dml(object):
```
frreiss: I don't see a way to set the `ScriptType` flag (for telling the internal MLContext that a string/file contains PYDml or "R-like" DML). Am I missing something?

src/main/java/org/apache/sysml/api/mlcontext/MLResults.java
MechCoder: It is meant to be private but exposed on the Python side to avoid using `getDouble` etc. Do you still want me to write documentation?

python/SystemML.py
MechCoder: I'll fix this.

MechCoder:
@frreiss I believe, I have addressed your comments right now. The API should be consistent with that on the Java side. Could you make another pass?

python/SystemMLtests.py
```diff
@@ -0,0 +1,87 @@
...
+        rdd2 = sc.parallelize(["5.0,6.0", "7.0,8.0"])
+        script = dml(sums).input(m1=rdd1).input(m2=rdd2).out("s1", "s2", "s3")
+        self.assertEqual(
+            ml.execute(script).get("s1", "s2", "s3"), [10.0, 26.0, "whatever"])
+
+    def test_pydml(self):
+        script = "A = full('1 2 3 4 5 6 7 8 9', rows=3, cols=3)\nx = toString(A)"
+        script = pydml(script).out("x")
+        self.assertEqual(
+            ml.execute(script).get("x")
```
dusenberrymw: Need a comma at the end of the line here.

deroneriksson:
@frreiss @dusenberrymw @niketanpansare 
Is this ready to merge?

deroneriksson:
Thank you for looking at this @dusenberrymw.

dusenberrymw:
@MechCoder This is great!  Here's a list of things I tried out, along with some additional thoughts:

1. Tried this out the example on Python 3.
  * Can you update the example in the initial post to reflect the recent update, as in:
    
      ```
      sums = """
      s1 = sum(m1)
      s2 = sum(m2)
      s3 = 'whatever'
      """
      ml = MLContext(sc)
      rdd1 = sc.parallelize(["1.0,2.0", "3.0,4.0"])
      rdd2 = sc.parallelize(["5.0,6.0", "7.0,8.0"])
      dml = Dml(sums).input(m1=rdd1).input(m2=rdd2).out("s1", "s2", "s3")
      ml.execute(dml).get("s1", "s2", "s3")
      [10.0, 26.0, "whatever"]
      ```

2. Ran the Python tests with `spark-submit` (after adding the missing comma).
  * I used `PYSPARK_PYTHON=python3 spark-submit --master local[*] --driver-class-path $SYSTEMML_HOME/target/SystemML.jar --py-files=$SYSTEMML_HOME/python/SystemML.py $SYSTEMML_HOME/python/SystemMLtests.py`
3. Tried to alter the script so that `s2` would return a matrix, i.e. `s2 = m1 * 3`.
  * Right now, this just returns a `Java object` reflecting a `mlcontext.Matrix` type that was added (  @deroneriksson correct if this is wrong).  We should add a `Matrix` type to our Python API here, and then add some logic to convert this to a DataFrame, RDD, etc. as is done on the Java/Scala side.  Should just be another simple wrapper over \[1].

Some other thoughts:

1. Can you move the files from `python/` to `src/main/python` just for consistency with the Java and Scala code?  I know Spark has a `python/` folder, but they also have the project split into multiple modules, so they have to consolidate the Python code into a single top-level `python/` folder to allow for imports to work properly.  We only have one single module, so that won't be a problem.
2. In a follow-up PR, could you look into adding a `setup.py` file and any other requirements so that we could create a `pip` package?  Then we could allow for `pip` installs directly from the git repo, using a subdirectory flag pointing to `src/main/python` \[2].

\[1]: http://apache.github.io/incubator-systemml/spark-mlcontext-programming-guide#matrix-output
\[2]: http://stackoverflow.com/a/36074026

deroneriksson:
@dusenberrymw 
I believe right now get() returns a SystemML MatrixObject if dealing with a matrix. We should probably have it return a Matrix (very easy to do in a couple lines of code). However, perhaps we could do this in a separate followup JIRA/PR, along with adding a Matrix type to the Python API?

dusenberrymw:
@deroneriksson Good idea -- @MechCoder Can you (1) update the given example, (2) add the missing comma to that test, and (3) move the source code to the other location for this PR?  Then you could add the `Matrix` logic in a followup PR.

Also, you write great, clean Python code, @MechCoder! 👍

deroneriksson:
@dusenberrymw @MechCoder 
I believe the get method can return the 'right' type (Matrix) with something like the following (off the top of my head):
```
public Object get(String outputName) {
      Data data = getData(outputName);
      if (data instanceof ScalarObject) {
            ScalarObject so = (ScalarObject) data;
            return so.getValue();
      else if (data instanceof MatrixObject) {
            return getMatrix(outputName);
      } else {
            return data;
      }
}
```

MechCoder:
Thanks @dusenberrymw @deroneriksson! I've moved it. We can address other stuff in followup PR's

dusenberrymw:
Awesome, **LGTM**.  Thanks, @MechCoder!

dusenberrymw:
Merged.  Thanks, @MechCoder!

PR #212, opened 2016-08-15
[SYSTEML-860] SparkR/HydraR integration with SystemML

aloknsingh:


deroneriksson:
LGTM. Thank you @aloknsingh!

Since this utilizes the older MLContext API, we will most likely want to eventually move to the newer API.

PR #217, opened 2016-08-23
[SYSTEMML-865] Add Matrix Wrapper to Python MLContext

dusenberrymw:
This allows the Java API to return a `Matrix` type, with the Python side creating a simple wrapper around it.  Then, the user can either use this `Matrix` object in a future SystemML script, or can convert it to a DataFrame with `toDF()`.  For the latter conversion, I've opted to go ahead and sort by the row index, and then drop the index column by default, although the option exists for the user to instead keep the row index ("ID") column to preserve order through any future operations on the DataFrame.

cc @MechCoder, @deroneriksson

### DIFF
--- a/src/main/python/SystemML.py
+++ b/src/main/python/SystemML.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 #-------------------------------------------------------------
 #
 # Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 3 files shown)


src/main/python/SystemML.py
```diff
@@ -22,21 +21,112 @@
...
+        Instance of a script object.
+    """
+    if not isinstance(scriptString, str):
+        raise ValueError("scriptString should be a string, got %s" % type(scriptString))
+    return Script(scriptString, scriptType="pydml")
+
+
+def _java2py(sc, obj):
+    """ Convert Java object to Python. """
+    obj = pyspark.mllib.common._java2py(sc, obj)
```
MechCoder: In future, it might be better to port this method rather than relying on a private method in pyspark.

src/main/python/SystemML.py
```diff
@@ -22,21 +21,112 @@
...
+    def __repr__(self):
+        return "Matrix"
+
+    def toDF(self, keepIndex=False):
+        """
+        Convert the Matrix to a DataFrame.
+
+        Parameters
+        ----------
+        keepIndex: Boolean
```
MechCoder: Just curious to know what is the behaviour on the Java side. Does doing `_java_matrix.asDataFrame` also return a column with "ID"?

src/main/python/SystemML.py
```diff
@@ -22,21 +21,112 @@
...
+    def __init__(self, javaMatrix, sc):
+        self._java_matrix = javaMatrix
+        self.sc = sc
+
+    def __repr__(self):
+        return "Matrix"
+
+    def toDF(self, keepIndex=False):
+        """
+        Convert the Matrix to a DataFrame.
```
MechCoder: to a Spark SQL DataFrame. (DataFrame can also be a Pandas dataframe)

src/main/python/SystemML.py
```diff
@@ -22,21 +21,112 @@
...
+
+        Parameters
+        ----------
+        keepIndex: Boolean
+            Either keep the row index in the DataFrame as a column named "ID",
+            or sort the DataFrame by the row index, then drop the index column.
+            Note: If the index is dropped, the DataFrame will be returned ordered,
```
MechCoder: Should the default behaviour be `keepIndex=True` then? In that way we don't lose any information regarding the order.

MechCoder:
Could you add a couple of tests? Thanks!

niketanpansare:
@MechCoder @dusenberrymw Is this replicating the functionality of https://github.com/apache/incubator-systemml/pull/197/files#diff-0a0277e5cf166adf136c0d02b45427cfR54 ?

src/main/python/SystemML.py
dusenberrymw: Yeah I thought about this more, and I'm going to remove it for now, and we can revisit it later starting on the Java/Scala side.

dusenberrymw:
@MechCoder Thanks for the review!

dusenberrymw:
@niketanpansare No, this just exposes the `Matrix` type that is present in the new Java `MLContext`.  Here, we return all of the correct types via a simple call to `get()` on a `MLResults` object, including returning the existing `Matrix` type from the new `MLContext` on the Java side.  On the Java side, the new `Matrix` type can already be used for future SystemML scripts, and contains methods for converting to other types, such as a Spark DataFrame.  We simply expose that functionality on the Python side here.  By only using a single `get()`, we reduce the complexity.

niketanpansare:
@dusenberrymw Cool. Since I have refactored the python code into package structure, either I can merge once ur PR is in or you can merge your code in the file https://github.com/niketanpansare/incubator-systemml/blob/python/src/main/python/SystemML/mlcontext.py once it is merged :)

dusenberrymw:
@niketanpansare Awesome, let's go ahead and merge this PR in (once I add a few updates), and then we can merge your PR.  I want to go ahead and update some example code that relies on this PR to go along with a presentation from today.

niketanpansare:
@dusenberrymw sounds good. Please go ahead and merge when you are ready :+1:

src/main/python/SystemML.py
MechCoder: great

src/main/python/SystemML.py
dusenberrymw: Added a TODO note.

src/main/python/SystemML.py
dusenberrymw: Updated.

src/main/python/SystemML.py
dusenberrymw: Removed.

dusenberrymw:
@MechCoder Updated the PR to remove the `keepIndex` logic, update the comments, and add test cases.  All tests pass on Python 2 & 3.

MechCoder:
Awesome. Feel free to merge. (Y)

PR #218, opened 2016-08-24
[SYSTEMML-834] Improve MLContext DataFrame support

deroneriksson:
Adds support for 4 types of DataFrame inputs and outputs via MLContext (doubles with ID column, doubles with no ID column, vectors with ID column, vectors with no ID column).

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
@@ -129,7 +129,7 @@ public class Matrix {
 	 */
 	public DataFrame asDataFrameDoubleNoIDColumn() {
 		DataFrame df = MLContextConversionUtil.matrixObjectToDataFrame(matrixObject, sparkExecutionContext, false);
-		df = df.drop("ID");
+		df = df.sort("ID").drop("ID");
 		return df;
 	}
 
@@ -150,7 +150,7 @@ public class Matrix {
 	 */
 	public DataFrame asDataFrameVectorNoIDColumn() {
 		DataFrame df = MLContextConversionUtil.matrixObjectToDataFrame(matrixObject, sparkExecutionContext, true);
-		df = df.drop("ID");
+		df = df.sort("ID").drop("ID");
 		return df;
 	}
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
```diff
@@ -103,16 +103,58 @@ public MatrixObject asMatrixObject() {
...
+	}
+
+	/**
+	 * Obtain the matrix as a {@code DataFrame} of doubles with no ID column
+	 * 
+	 * @return the matrix as a {@code DataFrame} of doubles with no ID column
+	 */
+	public DataFrame asDataFrameDoubleNoIDColumn() {
+		DataFrame df = MLContextConversionUtil.matrixObjectToDataFrame(matrixObject, sparkExecutionContext, false);
+		df = df.drop("ID");
```
dusenberrymw: I think we need to sort before dropping the "ID" column to ensure the order is correct, i.e. `df = df.sort("ID").drop("ID")`.

src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
```diff
@@ -103,16 +103,58 @@ public MatrixObject asMatrixObject() {
...
+	}
+
+	/**
+	 * Obtain the matrix as a {@code DataFrame} of vectors with no ID column
+	 * 
+	 * @return the matrix as a {@code DataFrame} of vectors with no ID column
+	 */
+	public DataFrame asDataFrameVectorNoIDColumn() {
+		DataFrame df = MLContextConversionUtil.matrixObjectToDataFrame(matrixObject, sparkExecutionContext, true);
+		df = df.drop("ID");
```
dusenberrymw: I think we need to sort before dropping the "ID" column to ensure the order is correct, i.e. `df = df.sort("ID").drop("ID")`.

deroneriksson:
@dusenberrymw Added sorts before the drops

dusenberrymw:
LGTM.

deroneriksson:
Thank you for reviewing @dusenberrymw! I will merge.

PR #225, opened 2016-08-29
[SYSTEMML-568] Frame New MLContext support

asurve:
At this point changes will support Frame MLContext with 
   Input: JavaRDD<String>, CSVFile, IJVFile, DataFrame
   Output: Frame

Other changes will be added soon.

There are other pending changes such as 
    Schema input, separator support.


src/main/java/org/apache/sysml/api/mlcontext/MLContextConversionUtil.java
```diff
@@ -309,6 +360,52 @@ public static MatrixObject dataFrameToMatrixObject(String variableName, DataFram
...
+	 * 
+	 * @param variableName
+	 *            name of the variable associated with the frame
+	 * @param dataFrame
+	 *            the Spark {@code DataFrame}
+	 * @return the {@code DataFrame} matrix converted to a converted to a
+	 *         {@code FrameObject}
+	 * @throws DMLRuntimeException 
+	 */
+	public static FrameObject dataFrameToFramexObject(String variableName, DataFrame dataFrame) throws DMLRuntimeException {
```
dusenberrymw: I think this should be `dataFrameToFramexObject` -> `dataFrameToFrameObject`

src/main/java/org/apache/sysml/api/mlcontext/Script.java
```diff
@@ -325,6 +330,23 @@ public Script input(String name, Object value) {
...
+	 *            name of the input
+	 * @param value
+	 *            value of the input
+	 * @param matrixMetadata
+	 *            optional matrix metadata
+	 * @param bFrame
+	 *            if input is of type frame
+	 * @return {@code this} Script object to allow chaining of methods
+	 */
+	public Script in(String name, Object value, MatrixMetadata matrixMetadata, boolean bFrame) {
```
dusenberrymw: So one of the cool new features in the new `MLContext` is that the API seamlessly allows various input types and output types to flow through without the use of any extra boolean flags.  This makes the overall API really clean and great to use.  Could you explore removing the boolean flags and keeping the same new API style?  It's really nice from a user's perspective.

cc @deroneriksson Do you have any clever thoughts on this?  I know a lot of work went into this design, and it turned out great.

deroneriksson:
Hi @asurve. Adding Frame support to the MLContext API is a great addition to the project! I just finished reviewing the PR and made some notes below. Overall this is great.

LGTM. I think the best strategy is to merge and then maybe we can address some of the potential issues and improvements? Perhaps you could fix 2 and 5 below and then merge?

-----------------------

1) MLContext InternalProxy (lines 380:425)
Could you make a note to check with Matthias or Niketan to see if any updates need to be made here for frames?

2) MLContextConversionUtil line 373
dataFrameToFramexObject to dataFrameToFrameObject (as Mike mentioned).

3) MLContextConversionUtil
I think we will probably need to create something like a FrameMetadata class and a FrameFormat enum. MatrixMetadata is being used for Frame metadata, and since a Frame is a different entity, it should probably have its own associated metadata and format classes.

Does Frame internally use MatrixCharacteristics or does it need to have something like its own FrameCharacteristics class? If it uses something like MatrixCharacteristics, we could still have a FrameMetadata act as its wrapper at the API level.

4) MLContextConversionUtil
For frameObjectToListStringIJV and frameObjectToListStringCSV, since types such as Strings are allowed, there might need to be some fancier character escaping (or surrounding double quotes with quotes escaped inside) or something like that.

5) MLContextConversionUtil
Could getSparkContext and getJavaSparkContext methods be private scope? Either that or move them to MLContextUtil?

6) MLContextUtil
The convertInputType takes a boolean flag but I've tried to avoid boolean flags in the API. Perhaps we can solve this using a FrameMetadata class in the future.

7) MLResults
We might want to create a Frame wrapper class similar to the Matrix wrapper class. We could have MLResults getFrame() method return a Frame object, similar to the getMatrix() method that returns a Matrix object.

8) MLResults
We need to be careful with methods such as the ones that take in delimiter strings. If we have multiple values like this and we don't encapsulate them, the API potentially becomes more difficult to use. It is a nice feature to have though.

10) Script
The inputVariablesType map and the boolean flag on the in() method are functional but I think a cleaner option may be to have classes/enums for Frame, FrameMetadata, and FrameFormat.

11) MLContextTest
Since the additional tests all concern frame tests and the tests are coded in a different format than all the existing tests, perhaps they could be moved to a new MLContextFrameTest class?

cc @dusenberrymw @bertholdreinwald @frreiss

mboehm7:
ad 3) The meta data is by design shared across frame and matrix. The same applies to the input/output formats. Please keep the meta data and formats orthogonal to the data type. Thanks.

ad 8) You can use CSVFileFormatProperties (subclass of FileFormatProperties) which already encapsulates parameters such header, delimiter, fill, and fillvalue.

in addition: Please be more careful when constructing the filenames for matrix and frame objects. Make sure that the process_id is appended as the suffix to the scratch_space already and use the appropriate sequence ids (NOT time) as done in Lops.DAG. Otherwise there is a risk that multiple users running in different processes on the same node affect each other.

deroneriksson:
@mboehm7 At some point when you have time, I'd really appreciate it if you could review the references to MLContextUtil.scratchSpace() in MLContextConversionUtil and make corrections as needed for the scratch space paths.

mboehm7:
sure - once I'm back, I'll create some abstractions which can be used by all APIs as well as related tests.

deroneriksson:
I believe this PR can be closed now.
@asurve If you amend the commit message with "Closes #225", when you push to apache master, GitHub will automatically close the PR for you.

dusenberrymw:
Ping @asurve Can you close this PR?

PR #226, opened 2016-08-29
[SYSTEMML-886] Add setStatistics and setExplain to the PythonAPI

MechCoder:


MechCoder:
cc: @nakul02 @niketanpansare @dusenberrymw

mboehm7:
thanks @MechCoder - one request; could you please make the input argument consistent with our command line flags, i.e., lower case?

MechCoder:
Were you referring to this (https://github.com/MechCoder/incubator-systemml/blob/59299d84c87fca085081fb3abc309c7d73a58244/src/main/python/SystemML/mlcontext.py#L327) or something else?

I have kept them consistent with what is on the Scala side. Let me know if you would still like to change those.

deroneriksson:
Great job @MechCoder!

I think @mboehm7 would like lowercase like line 135 of DMLScript.java:
```
+ "   -explain: <type> (optional) explain plan (hops, [runtime], recompile_hops, recompile_runtime)\n"
```

deroneriksson:
Whatever 'feels' right from the perspective of expectations of a Python developer is probably the best approach to use.

MechCoder:
I can do that but it will be inconsistent with the enum `explainLevel` here (https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java#L120)

mboehm7:
thanks for clarifying @deroneriksson - yes, I would prefer to have all warnings and errors consistent in lower case. It might be a good idea to simply call ExplainType.parseExplainType from all locations to have equivalent error handling. While you're at it, could you also change the parseExplainType implementation to valueof. Thanks.

deroneriksson:
@MechCoder @mboehm7 @dusenberrymw 
Just to mention a thing about the MLContext design. The MLContext class contains an ExplainLevel enum that is basically a wrapper of the Explain class's ExplainType enum. The reason that ExplainLevel exists is so that it is in the same package as all the other main MLContext classes. Because of this, it allows a single import (import org.apache.sysml.api.mlcontext._) to handle all MLContext classes, rather than needing to do multiple imports to use the MLContext API. This is very handy from a Scala perspective using SystemML from a Spark Shell.

The ExplainLevel values are uppercased as a Java enum convention. This allows Java/Scala developers to recognize these as 'constant' values. So, the uppercase format is useful when doing enums from Java or Scala, mostly as a convention.

However, for Python, if these are being treated as strings, maybe you could support both uppercase and lowercase strings to prevent case errors? That way you can match both the command-line flags and have symmetry with the Java/Scala.

dusenberrymw:
+1 for allowing any case format, but encourage the lowercase format with the allowed types list.

+1 for only touching the `MLContext` API directly, rather than reaching down into the system.

mboehm7:
good point @deroneriksson - I'm not in favor of this redundancy but I understand your reasoning behind it. Let's leave the MLContext ExplainLevel for now but let's ensure consistency across all entry points for the following aspects: (1) consistency of defaults (explain=true it's 'runtime' not 'hops'), (2) string specification in lower case, (3) consistent warning/error messages. Thanks.

niketanpansare:
@nakul02 Should we also enable -gpu via MLContext ?

MechCoder:
I've addressed the "lower-case handling" on the Java side. Let me know if that is ok.

src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
```diff
@@ -365,6 +365,27 @@ public void setExplainLevel(ExplainLevel explainLevel) {
 	}
 
 	/**
+	 * Set the level of program explanation that should be displayed if explain
+	 * is set to true.
+	 *
+	 * @param explainLevel
+	 *            string denoting program explanation
+	 */
+  public void setExplainLevel(String explainLevel) {
```
dusenberrymw: nitpick: Can you fix the formatting here?

dusenberrymw:
@MechCoder This looks good -- can you fix the Java formatting, and then I can merge this in.

MechCoder:
done and rebased

dusenberrymw:
Merged. Thanks, @MechCoder!

PR #227, opened 2016-08-29
[SYSTEMML-884] Fix Scala Map MLContext input

deroneriksson:
Update ```scala.collection.Map<String, Object>``` to ```scala.collection.Map```.
to avoid asInstanceOf casts of ints/doubles/etc from Scala.
Add comments to input methods of Script.
Create output methods for Script.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/Script.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/Script.java
@@ -232,6 +232,13 @@ public class Script {
 		return this;
 	}
 
+	/**
+	 * Pass a map of inputs to the script.
+	 *
+	 * @param inputs
+	 *            map of inputs (parameters ($) and variables).
+	 * @return {@code this} Script object to allow chaining of methods
+	 */
 	public Script input(Map<String, Object> inputs) {
 		return in(inputs);
 	}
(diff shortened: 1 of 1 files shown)


deroneriksson:
This allows the following code to work from Spark Shell:
```
val addScript = dml("print(a + b)")
addScript.in(Map("a"->1,"b"->2))
```
rather than needing to cast (as previously required), as in the following:
```
addScript.in(Map("a"->1.asInstanceOf[AnyRef],"b"->2.asInstanceOf[AnyRef]))
```

src/main/java/org/apache/sysml/api/mlcontext/Script.java
```diff
@@ -243,14 +250,23 @@ public Script input(Map<String, Object> inputs) {
 	 *            Scala Map of inputs (parameters ($) and variables).
 	 * @return {@code this} Script object to allow chaining of methods
 	 */
-	public Script in(scala.collection.Map<String, Object> inputs) {
-		Map<String, Object> javaMap = JavaConversions.mapAsJavaMap(inputs);
+	@SuppressWarnings({ "rawtypes", "unchecked" })
```
dusenberrymw: Can you add a quick comment here when you merge indicating that we updated this to all Scala Maps to be passed in that would otherwise have the type `[String, Any]`, preventing it from being used from Java?

dusenberrymw:
Awesome, thanks for finding a fix for this, @deroneriksson.  LGTM.

deroneriksson:
Sure I'll add a comment to describe the reasoning.

Thanks for reviewing.

PR #228, opened 2016-08-30
[SYSTEMML-887] Automatic DataFrame type determination in MLContext

deroneriksson:
Automatically determines if input DataFrame is one of the following 4 types (although metadata can optionally be specified):
1) ID column with doubles
2) No ID column with doubles
3) ID column with vector
4) No ID column with vector

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextConversionUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextConversionUtil.java
@@ -38,6 +38,7 @@ import org.apache.spark.rdd.RDD;
 import org.apache.spark.sql.DataFrame;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.types.StructType;
 import org.apache.sysml.api.MLContextProxy;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/sysml/api/mlcontext/MLContextConversionUtil.java
```diff
@@ -361,13 +363,57 @@ public static MatrixObject dataFrameToMatrixObject(String variableName, DataFram
...
+		boolean hasID = false;
+		try {
+			schema.fieldIndex("ID");
+			hasID = true;
+		} catch (IllegalArgumentException iae) {
+		}
+		Row firstRow = dataFrame.first();
+		MatrixFormat mf = null;
+		if (hasID) {
+			Object object = firstRow.get(1);
```
dusenberrymw: I would stick with this implementation for now, but at some point if this happens to become a bottleneck, we should explore determining the type directly from the schema.

dusenberrymw:
LGTM.

deroneriksson:
Great feedback @dusenberrymw. I believe that StructType is made of StructFields and that StructField has a DataType that could probably be used. If there is some type of bottleneck experienced, I will update the code to examine the StructField values.

Thank you for reviewing.

PR #229, opened 2016-08-31
[SYSTEMML-889] Remove MLContext Java input output methods

dusenberrymw:
This removes the `Script.input` and `Script.output` methods on the JVM side, hacks the Python side so that it is able to call the `Script.in` method on the JVM side, and updates `Script.out` to `Script.output` on the Python side (calling `Script.out` on the JVM side still).

This means that a Scala/Java user would use the clean `in(...)` and `out(...)` syntax, while a Python user would use the `input(...)` and `output(...)` syntax.


dusenberrymw:
cc @deroneriksson, @niketanpansare, @MechCoder

PR #231, opened 2016-09-01
[SYSTEMML-878] Update the Python package from SystemML to systemml

niketanpansare:
- Updated Python package name from SystemML to systemml
- Moved uploadToPyPI.sh script to dev/release
- Updated the documentation

@dusenberrymw


dusenberrymw:
Awesome, LGMT.

PR #232, opened 2016-09-06
[SYSTEMML-897] Add old MLContext Spark Shell examples to docs

deroneriksson:
Until versioned documentation is available on the SystemML main website, we need to have available the old MLContext Spark Shell examples. The old API works with SystemML 0.10.0 and earlier, but the new API does not yet work with an official SystemML release (until the next release). Therefore, if someone downloads a SystemML release (ie, 0.10.0) and wants to use it with Spark Shell, they need the old examples to be available online.

Preview here: http://deroneriksson.github.io/incubator-systemml/spark-mlcontext-programming-guide.html


dusenberrymw:
LGTM as a temporary fix, but we should drop this from the official 1.0 docs.  At that point, we should create versioned documentation.

deroneriksson:
Definitely agree. Part of the process of our web upgrade and next release should be to ensure we have versioned documentation online so that we can remove this.

PR #235, opened 2016-09-06
[SYSTEMML-891] Update MLContext Matrix and Frame 'as' methods to 'to'

deroneriksson:
Update 'as' methods to 'to' methods to more closely follow Spark conventions.
Remove 'String' from rdd and JavaRDD access methods for conciseness.

### DIFF
--- a/docs/spark-mlcontext-programming-guide.md
+++ b/docs/spark-mlcontext-programming-guide.md
@@ -463,8 +463,8 @@ n = sum(m)
 val scr = dml(s).out("m", "n");
 val res = ml.execute(scr)
 val (x, y) = res.getTuple[Matrix, Double]("m", "n")
-x.toRDDIJV.collect.foreach(println)
-x.toRDDCSV.collect.foreach(println)
+x.toRDDStringIJV.collect.foreach(println)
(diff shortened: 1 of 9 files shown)


nakul02:
Pending [Jenkins](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/203/) Job

deroneriksson:
@dusenberrymw Would you mind trying out this update and see if it has a better feel to Spark developers?

deroneriksson:
cc @bertholdreinwald

dusenberrymw:
The `toDF` looks good.  I might recommend using `toRDDStringIJV` and related variants to make it clear that those methods are for getting a string representation of a row of numbers.  This way, we can add a `toRDDIJV` function that returns a proper tuple of numbers.   The string formats are old anyways, and we should probably deprecate them at some point.

Also, can you double check that the Python MLContext `toDF` function is updated to use this new Java `toDF` function?  Should be a quick change confirmed with tests.

deroneriksson:
Thanks for the feedback @dusenberrymw. I will add the 'String' to the RDD methods to avoid confusion in the future.

deroneriksson:
Requested changes made. I will merge.

PR #236, opened 2016-09-08
[SYSTEMML-900] Rename DataFrame ID column

deroneriksson:
The use of "ID" as the name of DataFrame column for ordering can cause confusion and unexpected results because ID is a name that can commonly be used. By changing the column name to a more unique name such as "__INDEX", we can avoid such conflicts.

### DIFF
--- a/src/main/java/org/apache/sysml/api/MLOutput.java
+++ b/src/main/java/org/apache/sysml/api/MLOutput.java
@@ -101,7 +101,7 @@ public class MLOutput {
 	
 	/**
 	 * Note, the output DataFrame has an additional column ID.
-	 * An easy way to get DataFrame without ID is by df.sort("ID").drop("ID")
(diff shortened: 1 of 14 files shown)


deroneriksson:
All tests passed: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/646/console

deroneriksson:
cc @dusenberrymw @bertholdreinwald

dusenberrymw:
This LGTM.  We might also want to think about adding an optional `set_index(...)` function, like Pandas has \[1].

\[1]: http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.set_index.html

dusenberrymw:
Oh also, can you check that the Python test still run, and that the Python examples are still correct?

deroneriksson:
Verified that test_mlcontext.py still works. I've never been able to run test_mllearn.py on my machine.

PR #238, opened 2016-09-09
[SYSTEMML-901] Improve unavailable MLContext message

deroneriksson:
Add option to suppress MLContextException stacktrace.

### DIFF
--- a/src/main/java/org/apache/sysml/api/MLContextProxy.java
+++ b/src/main/java/org/apache/sysml/api/MLContextProxy.java
@@ -21,6 +21,7 @@ package org.apache.sysml.api;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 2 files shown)


deroneriksson:
If an MLContext hasn't been created and it is required, here is an example of the new behavior:
```
scala> val LinRegCg = LinRegCgDML.in("X", X).in("y", y).out("beta_out")
No MLContext object is currently active. Have you created one? Hint: in Scala, 'val ml = new MLContext(sc)'
```
Stacktrace is suppressed in this situation since it is irrelevant for the user.

dusenberrymw:
LGTM.

deroneriksson:
Thank you @dusenberrymw. I will merge.

PR #239, opened 2016-09-13
[SYSTEMML-879] Remove the Python MLResults.getDataFrame & MLResults.getNumpyArray functions.

dusenberrymw:
* Remove the `MLResults.getDataFrame` function and replace usages of `getDataFrame("matrixName")` with `get("matrixName").toDF()`.
* Remove the `MLResults.getNumpyArray` function and replace usages of `getNumpyArray("matrixName")` with `get("matrixName").toNumPyArray()`.

All tests run.

cc @niketanpansare, @MechCoder

### DIFF
--- a/docs/beginners-guide-python.md
+++ b/docs/beginners-guide-python.md
@@ -354,5 +354,5 @@ y_df = sqlCtx.createDataFrame(pd.DataFrame(y_digits[:.9 * n_samples]))
 ml = sml.MLContext(sc)
 script = os.path.join(os.environ['SYSTEMML_HOME'], 'scripts', 'algorithms', 'MultiLogReg.dml')
 script = sml.dml(script).input(X=X_df, Y_vec=y_df).output("B_out")
-beta = ml.execute(script).getNumPyArray('B_out')
+beta = ml.execute(script).get('B_out').toNumPyArray()
 ```
--- a/src/main/python/systemml/converters.py
+++ b/src/main/python/systemml/converters.py
@@ -19,7 +19,7 @@
 #
 #-------------------------------------------------------------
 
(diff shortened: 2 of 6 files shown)


dusenberrymw:
Actually, I'd be interested in changing `toNumPyArray()` to `toNumPy()`.  Then we could have `toDF()`, `toNumPy()`, `toPandas()`, `toList()`, etc.

Thoughts?

dusenberrymw:
cc @nakul02

nakul02:
I am not much of a python programmer, but here are my 2 cents:
`toDF()` might be ok
`toNumPy` and `toPandas()` make sense if **NumPy** and **Pandas** offer exactly on data type/data structure.

dusenberrymw:
Yeah in this case we're following Spark conventions by using `toDF()` & `toPandas()`, which return a Spark DataFrame and a Pandas DataFrame respectively in Spark right now.  `toNumPy()` would return a NumPy array specifically, and would follow along with the same conventions.

If there are no objections, I will merge this tomorrow.  Also, @nakul02 this may require a slight function change in the demo you've been working on.

niketanpansare:
I like the idea of having `toDF()` and `toPandas()` as they are consistent with PySpark's API: https://spark.apache.org/docs/1.3.0/api/python/pyspark.sql.html

I would refer to @dusenberrymw and @MechCoder regarding `toNumPy()` or `toNumPyArray` or `toNumpyMatrix` or `as_matrix`: http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.as_matrix.html

dusenberrymw:
Okay great.  I'm going to go with `toNumPy()` for now as it is short and clean, and prevents any confusion between SystemML "matrices" and NumPy "matrices".

PR #241, opened 2016-09-13
[DOC] Updated brew install instructions to install spark1.6.

nakul02:
@deroneriksson @niketanpansare @dusenberrymw

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,10 @@ Before you get started on SystemML, make sure that your environment is set up an
   brew install Caskroom/cask/java
   ```
 
-  3. **Install Spark.**
+  3. **Install Spark 1.6.**
   ```
-  brew install apache-spark
+  brew tap homebrew/versions
+  brew install apache-spark16
   ```
 
   4. **Download SystemML.**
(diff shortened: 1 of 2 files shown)


deroneriksson:
LGTM. Thank you @nakul02.

dusenberrymw:
Good call, @nakul02.  What are the executable names now?  Before, it would have installed `spark-shell`, `pyspark`, etc.  Does it have the version appended now?

nakul02:
I think they're still called `spark-shell`, `pyspark`.
The [homebrew/versions tap](https://github.com/Homebrew/homebrew-versions/blob/master/apache-spark16.rb) recipe just gets Spark 1.6.2 instead of the [default homebrew-core](https://github.com/Homebrew/homebrew-core/blob/master/Formula/apache-spark.rb) recipe

niketanpansare:
Thanks Nakul for looking into this. We should also consider adding Window's instruction in subsequent PR :)

nakul02:
@niketanpansare - Could you please merge this in?

dusenberrymw:
@MadisonJMyers Can you make this change in your blog post?

nakul02:
@deroneriksson or @dusenberrymw - Could you please merge this in, @niketanpansare has a permission issue.

MadisonJMyers:
@dusenberrymw yes! will do.

PR #242, opened 2016-09-14
[SYSTEMML-906] [SYSTEMML-907] Update df first to df schema

deroneriksson:
Utilize DataFrame schema to determine matrix/frame characteristics
rather than calling first().

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -38,10 +38,13 @@ import org.apache.spark.SparkContext;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
-import org.apache.spark.mllib.linalg.Vector;
+import org.apache.spark.mllib.linalg.VectorUDT;
 import org.apache.spark.rdd.RDD;
 import org.apache.spark.sql.DataFrame;
-import org.apache.spark.sql.Row;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
 import org.apache.sysml.conf.CompilerConfig;
 import org.apache.sysml.conf.CompilerConfig.ConfigType;
 import org.apache.sysml.conf.ConfigurationManager;
@@ -494,8 +497,7 @@ public final class MLContextUtil {
 						(FrameMetadata) metadata);
 				return frameObject;
 			} else if (!hasMetadata) {
-				Row firstRow = dataFrame.first();
-				boolean looksLikeMatrix = doesRowLookLikeMatrixRow(firstRow);
+				boolean looksLikeMatrix = doesDataFrameLookLikeMatrix(dataFrame);
 				if (looksLikeMatrix) {
 					MatrixObject matrixObject = MLContextConversionUtil.dataFrameToMatrixObject(name, dataFrame);
 					return matrixObject;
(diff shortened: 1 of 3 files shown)


deroneriksson:
All tests pass here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/205/console

deroneriksson:
cc @dusenberrymw

src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
```diff
@@ -585,24 +587,24 @@ public static boolean isCSVLineAllNumbers(String line) {
...
-			} catch (NumberFormatException e) {
+	public static boolean doesDataFrameLookLikeMatrix(DataFrame df) {
+		StructType schema = df.schema();
+		StructField[] fields = schema.fields();
+		if (fields == null) {
+			return true;
+		}
+		for (StructField field : fields) {
+			DataType dataType = field.dataType();
+			if ((dataType != DataTypes.DoubleType) && (dataType != DataTypes.IntegerType)
```
dusenberrymw: We may need to add the `BigInt` type here as well.  At least from Python, when an RDD of integers is converted to a DataFrame with `toDF()`, the integers are set to type `BigInt`.  I'm testing the behavior of this in the current master to see if the conversion happens smoothly even with `BigInt`.

src/test/java/org/apache/sysml/test/integration/mlcontext/MLContextTest.java
```diff
@@ -561,13 +561,13 @@ public void testDataFrameSumDMLDoublesWithIDColumn() {
...
 
-		JavaRDD<Row> javaRddRow = javaRddString.map(new CommaSeparatedValueStringToRow());
+		JavaRDD<Row> javaRddRow = javaRddString.map(new CommaSeparatedValueStringToDoubleArrayRow());
 		SQLContext sqlContext = new SQLContext(sc);
 		List<StructField> fields = new ArrayList<StructField>();
-		fields.add(DataTypes.createStructField("__INDEX", DataTypes.StringType, true));
-		fields.add(DataTypes.createStructField("C1", DataTypes.StringType, true));
-		fields.add(DataTypes.createStructField("C2", DataTypes.StringType, true));
-		fields.add(DataTypes.createStructField("C3", DataTypes.StringType, true));
+		fields.add(DataTypes.createStructField("__INDEX", DataTypes.DoubleType, true));
```
dusenberrymw: I think we should set to this to `IntegerType` -- perhaps even throw an error if the `__INDEX` is a double.

src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
dusenberrymw: Confirmed that DataFrames with a `BigInt` type (i.e. a bigint for the `__INDEX`) will get converted correctly on the master branch.  Let's just add an additional `BigInt` type here.

deroneriksson:
@dusenberrymw On the Java side, looks like the bigint gets converted to a long, so I added a check for DataTypes.LongType. Can be verified with:
```
import org.apache.spark.mllib.linalg._
val row1 = (1L, Vectors.dense(1.0,2.0,3.0))
val row2 = (2L, Vectors.dense(4.0,5.0,6.0))
val rdd = sc.parallelize(Array(row1, row2))
val df = rdd.toDF("__INDEX", "C1")
df.show

val ml = new MLContext(sc)
val scr = ScriptFactory.dml("print(toString(M))").in("M", df)
ml.execute(scr)
```

I also updated the tests to use DataTypes.IntegerType.

dusenberrymw:
The following is now confirmed to work on PySpark:

```python
from systemml import MLContext, dml
from pyspark.mllib.linalg import Vectors

X = sc.parallelize([(1, Vectors.dense([3.4,4.3,5.3])), (2, Vectors.dense([5.2,6.5,7.0]))]).toDF(["__INDEX", "sample"])

ml = MLContext(sc)
script = """
s = sum(X)
"""
script = dml(script).input(X=X).output("s")
s = ml.execute(script).get("s")
```

deroneriksson:
Thank you for including the PySpark example. Very helpful if needed later.

PR #247, opened 2016-09-17
[SYSTEMML-446] [WIP] Exploit cublas libraries for transpose and certain cases of binary operations + add support for invoking custom kernels

niketanpansare:
1. Tanuj surveyed the CuBLAS libraries and implemented LibMatrixCUDA methods. These methods map SystemML operations to CuBLAS calls similar to LibMatrixCUDA.matmult(). Here are the list of methods: transpose, cellwiseMatMatAddSub, matScalarElementwiseMultDiv and vectorScalarMult. In subsequent PR, we will implement remaining binary operations using custom kernels (i.e. phase 2). 
2. Also, added initial support for invoking custom kernels in preparation for phase 2. This relies of custom kernels compiled into ptx format and included as resource file. Since ptx format are target-independent, the kernels can be supported for different GPU versions. The utility
   org.apache.sysml.runtime.instructions.gpu.context.JCudaKernels simplifies the launching of the kernels. For example: to launch a kernel `copyUpperToLowerTriangleDense<<1,1,32,32>>(jcudaDenseMatrixPtr, dim, dim*dim)`, the user has to call: `kernels.launchKernel("copyUpperToLowerTriangleDense", new
   ExecutionConfig(1,1,32,32), jcudaDenseMatrixPtr, dim, dim*dim)`.
3. Also fixed a GPU bufferpool bug (setting of dirty flag).
4. Implemented sparseToDense() and denseToSparse() methods in JCudaObject similar to MatrixBlock.

@nakul02 @bertholdreinwald @taasawat


src/main/cpp/kernels/SystemML.ptx
```diff
@@ -0,0 +1,76 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
nakul02: You might not want to commit this, this is generated code.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -948,4 +983,173 @@ public static boolean isInSparseFormat(MatrixObject mo) {
...
+	 * Performs elementwise multiplication of input vector in and constant using cublasDaxpy
+	 * 
+	 * @param ec
+	 * @param in
+	 * @param constant
+	 * @param outputName
+	 * @param op
+	 * @throws DMLRuntimeException
+	 */
+	public static void vectorScalarMult(ExecutionContext ec, MatrixObject in, double constant, String outputName, GPUEnabledElementwiseOp op) throws DMLRuntimeException {
```
nakul02: rename vectorScalarMult -> vectorMatrixElementWise
Also you might want to stick with naming convention in matrix consistent with [cuBlasDaxpy](http://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-axpy) since this function is a wrapper to that.

Also, this function seems to be take in a "control parameter", namely the "op".
My preference matches this [guy](http://programmers.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior), but then again, this is a matter of design and is subjective. Also, since an element wise divide can be done, the name of the function does not make sense anymore.
I would split it up into 2 functions - _vectorScalarElementWiseMult_ & _vectorScalarElementWiseDiv_

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -948,4 +983,173 @@ public static boolean isInSparseFormat(MatrixObject mo) {
...
+	 * 
+	 * @param ec
+	 * @param in
+	 * @param constant
+	 * @param outputName
+	 * @param op
+	 * @param isTransposed
+	 * @throws DMLRuntimeException
+	 */
+	public static void matScalarElementwiseMultDiv(ExecutionContext ec, MatrixObject in,
```
nakul02: Look at the "control parameter" argument left at vectorScalarMult.
Also, names of the arguments should match [cublasDgeam](http://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-geam) since this is a wrapper over it. (**alpha** for constant, **A** for the matrix)

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -948,4 +983,173 @@ public static boolean isInSparseFormat(MatrixObject mo) {
...
+	 * @param ec
+	 * @param in1
+	 * @param in2
+	 * @param outputName
+	 * @param isLeftTransposed
+	 * @param isRightTransposed
+	 * @param isAdd
+	 * @throws DMLRuntimeException
+	 */
+	public static void cellwiseMatMatAddSub(ExecutionContext ec, MatrixObject in1, MatrixObject in2,
```
nakul02: Match parameters to [cublasDgeam](http://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-geam) (transa, transb, A, B...).
Rename to **matMatAdd**. The "element-wise" part is implied.

src/main/cpp/kernels/SystemML.ptx
niketanpansare: There are following options for supporting custom kernels:

1. Compile .ptx file on machine where nvcc is installed and include it as resource file (Current option). This will use CUDA driver-side API (similar to cublas calls) and has no dependency on whether nvcc + Visual Studio or g++.
**Pros:** Will work in distributed setting, no installation of SystemML libraries.
**Cons:** Will require .ptx to be committed to git.

2. Include .cu file as resource file and compile it during execution. Here are two options to do so:
a. Include jcuda utility into the classpath and use KernelLauncher: http://www.jcuda.org/utilities/KernelLauncherSample.java
b. Run nvcc: http://jcuda.org/samples/JCudaVectorAdd.java
**Pros:** Will not require .ptx to be committed to git.
**Cons:** GPU backend dependent on System setup. For example: nvcc 7.5 requires Visual Studio 10 on Windows.

4. Add gpu profile in pom.xml, which invokes nvcc during `mvn package`
**Pros:** Will not require .ptx to be committed to git.
**Cons:** Will increase the number of artifacts. 

5. Compile dll/so files and use JNI. 
**Cons:** Build complexity. Will require shipping and testing SystemML on different OS.

Including @lresende and @deroneriksson since this is related to build issues.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
niketanpansare: I plan to generalize these method by adding custom kernels and have one single entry point: `bincellOp` similar to LibMatrixBincell.

nakul02: This is a good idea. The GPU matrix libraries will resemble the CP matrix libraries. Much like how LibMatrixCUDA (for now) looks like LibMatrixCUDA.
However, there is nothing stopping us from making all the functions well named and well engineered :)

src/main/cpp/kernels/SystemML.ptx
niketanpansare: In recent commit, I enabled option 3 by not requiring .ptx file to be included in git. To use GPU kernels, the user will have to compile using the command `mvn package -DenableGPU=true`. This option expects that the user will have to setup nvcc on his dev environment to use GPU. To reduce the number of artifacts, we can include the GPU kernels by default in our releases ...  TODO: set `-DenableGPU=true` in dev/release scripts.

niketanpansare:
Created JIRAs SYSTEMML-931 to SYSTEMML-939 of issues that are not covered by this PR.

nakul02:
LGTM, there are some tests failing marked by @niketanpansare in JIRAS  SYSTEMML-[931](https://issues.apache.org/jira/browse/SYSTEMML-931) to [939](https://issues.apache.org/jira/browse/SYSTEMML-939).
Those JIRAs also contain performance enhancements.

niketanpansare:
@deroneriksson @lresende As mentioned earlier, I believe the build related concerns are addressed. Like Java, .cu files (corresponding to .java files) are checked into git and .ptx files (corresponding to .class files) are compiled into the jar using mvn:
```bash
mvn clean prepare-package -DenableGPU=true
mvn package -Pdistribution
```

The first command invokes `nvcc -ptx SystemML.cu -o SystemML.ptx` and the next command simply puts SystemML.ptx into the jar as resource :)

src/test/java/org/apache/sysml/test/integration/AutomatedTestBase.java
```diff
@@ -95,6 +95,7 @@ public String lowerCase() {
 	
 	// By default: TEST_GPU is set to false to allow developers without Nvidia GPU to run integration test suite 
 	public static final boolean TEST_GPU = false;
+	public static final double GPU_TOLERANCE = 1e-6;
```
nakul02: @niketanpansare - this is the threshold that i used for my tests. But I'd recommend using as **small** a threshold as possible (maybe start with 10e-10?)

niketanpansare:
As of 9/19/2016, following application junit test cases are successful with TEST_GPU set in AutomatedTestBase:
org.apache.sysml.test.integration.applications.dml.{ApplyTransformDMLTest, CsplineCGDMLTest, CsplineDSDMLTest, GNMFDMLTest, HITSDMLTest, L2SVMDMLTest, LinearRegressionDMLTest, NaiveBayesDMLTest, NaiveBayesParforDMLTest, PageRankDMLTest, WelchTDMLTest}

Following application junit test cases are failing:

- org.apache.sysml.test.integration.applications.dml.{GLMDMLTest, LinearLogRegDMLTest} failing because of tolerance/incorrect results.
Example: 
```bash
(25,1): 0.21514032737322933 <--> 0.215139025489853
(1,1): -0.047827077493918596 <--> -0.04780980775776103
(35,1): -0.27654559523937755 <--> -0.2765466094740588
(2,1): 0.13255497059193416 <--> 0.1325471025480187
(3,1): -0.08823584272224108 <--> -0.08823474387632056
(101,1): -0.6240092982055565 <--> -0.624016229774438
```

-  org.apache.sysml.test.integration.applications.dml.{ID3DMLTest, MDABivariateStatsDMLTest, ArimaDMLTest} fails with following exception:
```bash
Caused by: jcuda.CudaException: cudaErrorIllegalAddress
	at jcuda.runtime.JCuda.checkResult(JCuda.java:437)
	at jcuda.runtime.JCuda.cudaMemGetInfo(JCuda.java:3415)
	at org.apache.sysml.runtime.instructions.gpu.context.JCudaContext.getAvailableMemory(JCudaContext.java:89)
	at org.apache.sysml.runtime.instructions.gpu.context.GPUObject.getAvailableMemory(GPUObject.java:193)
	at org.apache.sysml.runtime.instructions.gpu.context.JCudaObject.ensureFreeSpace(JCudaObject.java:526)
	at org.apache.sysml.runtime.instructions.gpu.context.JCudaObject.copyFromHostToDevice(JCudaObject.java:582)
	at org.apache.sysml.runtime.instructions.gpu.context.JCudaObject.prepare(JCudaObject.java:339)
	at org.apache.sysml.runtime.instructions.gpu.context.JCudaObject.acquireDeviceRead(JCudaObject.java:363)
	at org.apache.sysml.runtime.controlprogram.context.ExecutionContext.getMatrixInputForGPUInstruction(ExecutionContext.java:317)
	at org.apache.sysml.runtime.instructions.gpu.ReorgGPUInstruction.processInstruction(ReorgGPUInstruction.java:77)
	at org.apache.sysml.runtime.controlprogram.ProgramBlock.executeSingleInstruction(ProgramBlock.java:305)
```
Once https://issues.apache.org/jira/browse/SYSTEMML-931 is resolved, these testcases will pass.
	
-  org.apache.sysml.test.integration.applications.dml.MultiClassSVMDMLTest failed with following core dump:
```bash
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x000007fed5afe549, pid=7840, tid=2792
# Problematic frame:
# C  [cusparse64_75.dll+0x1e549]
```
Let's try to reproduce this on different machine after fixing SYSTEMML-931 as the dump is related to cusparse.

src/test/java/org/apache/sysml/test/integration/AutomatedTestBase.java
niketanpansare: Done.

niketanpansare:
@nakul02 Can you please review this PR ?
@deroneriksson Once @nakul02 gives his LGTM, we should merge this PR. Please note that (unlike other PRs) this contains 3 commits (one by Nakul and two by me) to ensure that Nakul's work is acknowledged.

nakul02:
This LGTM.
Once the [tests](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/216/) go through, we should merge this in.
@deroneriksson, @dusenberrymw - Can you please take a look?

niketanpansare:
@deroneriksson @nakul02 @dusenberrymw The tests failed due to license issue. I have added license to the generated ptx: https://github.com/niketanpansare/incubator-systemml/blob/7e127422b85396543998fad3ef767d2cb5463d7a/src/main/cpp/kernels/SystemML.ptx#L1. Can you please restart the tests ?

nakul02:
Launched [tests](https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/217/), lets merge it in if it goes through

nakul02:
So the tests have gone through, @deroneriksson - could you please take a look?

deroneriksson:
LGTM.

PR #249, opened 2016-09-18
[SYSTEMML-930] Set unlimited result size when call through  MLContext

asurve:


asurve:
Its not working in some cases, trying to debug it further.

mboehm7:
thanks @asurve - could you also comment on how you tested it (the most important question to answer is how to do this configuration management on an already created spark context that is passed from outside into mlcontext).

asurve:
It seems setting maxResultSize after SparkContext started does not get reflected. It needs to be set before starting SparkContext. User needs to set it through configuration like
  On command line by adding following for unlimited size (which SystemML prefers)
     --conf spark.driver.maxResultSize=0

mboehm7:
well, in that case, I would recommend to throw a warning if a spark context with maxResultSize < local_memory_budget is passed in. 

An alternative would be to automatically lower the local memory budget to the max result size but this has many negative implications for local operations and broadcasts and hence is no real solution.

asurve:
Requested changes added.

mboehm7:
could we please simplify this a little bit (remove redundant copies of the warning code, e.g., by moving it into SparkExecutionContext, remove the redundant setting of unlimited result size, and simplify the warning code, e.g., by using conf.get directly with defaults)

mboehm7:
LGTM

deroneriksson:
Hi @asurve. Is this PR ready to be closed?

PR #250, opened 2016-09-20
[SYSTEMML-568] Frame Schema support through MLContext

asurve:
Schema propagation through DataFrame/JavaRDD/RDD has been handled.


mboehm7:
please fix the modified dataFrameToBinaryBlock - if the schema is given but no column names, you miss to extract the column names.

asurve:
ok, added.

mboehm7:
thanks - one thing you might wanna double check though, is the integration of this extended converter. Right now there is not a single invocation with a schema other than DF schemas (which was the functionality provided before).

mboehm7:
LGTM - thanks @asurve

PR #256, opened 2016-09-23
[SYSTEMML-954] PyDML can be called with two arguments

sandeep-n:
* Changed argument checking in PyDmlSyntacticValidator
* Added unit tests for range() with corresponding PyDML test scripts


src/main/java/org/apache/sysml/parser/pydml/PydmlSyntacticValidator.java
```diff
@@ -959,9 +959,11 @@ else if(namespace.equals(DMLProgram.DEFAULT_NAMESPACE) && functionName.equals("r
 			namespace = DMLProgram.DEFAULT_NAMESPACE;
 		}
 		else if(namespace.equals(DMLProgram.DEFAULT_NAMESPACE) && functionName.equals("range")) {
-			if(paramExpression.size() != 3) {
```
nakul02: @niketanpansare - can you please take a look at this change?
The semantics of the `range` function in pydml are being changed here. The change makes sense to me, but seeing as how you spent the most amount of time working on PyDML, do you think this is ok?

src/test/java/org/apache/sysml/test/integration/functions/data/RangeTest.java
```diff
@@ -0,0 +1,69 @@
+package org.apache.sysml.test.integration.functions.data;
```
gweidner: It's good you've included license headers in new test script pydml files.
Please also include license header in new RangeTest.java to resolve PR build failure:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/692/

niketanpansare:
Thanks Sandeep for fixing the range-related bug in this PR. For consistency, I would recommend keeping the logic of `RangeTest`  similar to  https://github.com/apache/incubator-systemml/blob/master/src/test/java/org/apache/sysml/test/integration/functions/data/SequenceTest.java and parameterizing Range1.pydml and Range2.pydml (similar to Sequence.dml and Sequence2inputs.dml), i.e.:

```bash
# license ...
A = range($1, $2, $3);
save(A, $4);
```

The code of RangeTest would be similar to SequenceTest except following changes to :

1. Test-case naming: https://github.com/apache/incubator-systemml/blob/master/src/test/java/org/apache/sysml/test/integration/functions/data/SequenceTest.java#L47

2. Invoking the .pydml https://github.com/apache/incubator-systemml/blob/master/src/test/java/org/apache/sysml/test/integration/functions/data/SequenceTest.java#L108

This way, you check with R for sanity testing rather than constant data. Also, other developers are free to try testing range with different parameters: https://github.com/apache/incubator-systemml/blob/master/src/test/java/org/apache/sysml/test/integration/functions/data/SequenceTest.java#L69

niketanpansare:
For other committers, I have sanity tested this PR with following code and it works as expected:
``` bash
X = range(5, 2) # range(5, 2, -2), range(1, 5), ...
# Also, range(5,1,1) throws LanguageException
str = toString(X)
print(str)
```

src/test/java/org/apache/sysml/test/integration/functions/data/RangeTest.java
sandeep-n: Thanks for the heads-up!

niketanpansare:
@sandeep-n @nakul02 @gweidner If my comments are holding this PR, then please go ahead and merge as this fix would be important for 0.11 release. I can fix the testcases in a separate PR :)

sandeep-n:
@niketanpansare I'm working on the testcases right now. I'll update my PR in a few hours. But please feel free to merge earlier if it's urgent :)

niketanpansare:
@sandeep-n awesome ... In that case, I would recommend to wait until you add the testcases. Thanks for taking care of it :+1:

gweidner:
Thanks @sandeep-n for the updates!  All tests passed in [PR build 701](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/701/).

niketanpansare:
LGTM. Thanks @sandeep-n :)

gweidner:
Thanks for reviewing @niketanpansare - I will merge.

PR #257, opened 2016-09-23
[SYSTEMML-956]Casting issue with scalar assignment in Frame

asurve:


mboehm7:
that's a good catch, but could you please use 'input2.getValueType()' instead of the schema - the value is passed as string and the target schema is handled within the target block.

mboehm7:
LGTM - thanks @asurve

PR #259, opened 2016-09-23
[SYSTEMML-957] Bug fix for seq() in for loop

niketanpansare:
To reproduce this bug, please use following scripts:

Works as expected:
```bash
for(i in seq(1,5,1)) {
        print(i)
}
for(i in 1:5) {
        print(i)
}
```

Doesnot work:
```bash
for(i in seq(1,5)) {
        print(i)
}
```


mboehm7:
that's a great catch. It used to work with our old javacc parser. Seeing it show up now means there is not a single related testcase in our testsuite yet. Could you please extend ForLoopPredicateTest in test.integration.functions.parfor by one or two tests for this pattern? Thanks.

niketanpansare:
@mboehm7  Done.

PR #260, opened 2016-09-27
[SYSTEMML-961]Scalar object assigned as Matrix to Frame Object

asurve:
This problem occurs in Spark env.


mboehm7:
LGTM - thanks @asurve

asurve:
thanks

PR #262, opened 2016-09-28
[SYSTEMML-964] Do not use FastBufferedDataIOStream for frame

asurve:


mboehm7:
LGTM - thanks @asurve; just one minor thing: could you please add a comment that this is a temporary workaround - once frames are subject to eviction, we would need to handle UTF in both FastBufferedDataInputStream and FastBufferedDataOutputStream anyway and could generalize this code here again.

mboehm7:
what's the status here - I'd like to get this in to unblock our release. Also note that SYSTEMML-968 has been pushed, which means you can revert the writeExternal handling in this PR as our FastBufferedDataOutputStream now supports writeUTF.

PR #263, opened 2016-09-28
[SYSTEMML-967] Add abego/stringtemplate licenses for standalone tar.gz and zip

deroneriksson:
The antlr4-4.5.3.jar library includes the abego and stringtemplate libraries within
the jar file, so the LICENSE file for the standalone tar.gz and zip artifacts needs
to contain license information for abego and stringtemplate since antlr4-4.5.3.jar
is included in the packaged lib directory.

### DIFF
--- a/src/assembly/standalone/LICENSE
+++ b/src/assembly/standalone/LICENSE
@@ -286,6 +286,81 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 ================================================================================
 
+The abego TreeLayout Core dependency is distributed under the BSD license.
+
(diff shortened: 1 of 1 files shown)


deroneriksson:
cc @gweidner

gweidner:
Hi Deron - thanks for handling this artifact.  Could you try changing the standalone.xml so only antlr4-runtime included and ANTLR4 Tool jar (antlr4-4.5.3.jar) doesn't get pulled-in?

deroneriksson:
After an examination, it appears that the antlr4-4.5.3.jar (tool jar) is not needed for the standalone tar.gz and zip artifacts. Therefore, I will close this JIRA/PR and create a new JIRA/PR to remove the antlr4 tool jar from the standalone tar.gz and zip artifacts.

deroneriksson:
Thank you for looking at this too @gweidner. I investigated and came to the same conclusion.

PR #265, opened 2016-09-30
[SYSTEMML-998] Overview page as a design document for developers

niketanpansare:
Advantage of this approach:
- Will be helpful for a new contributor to learn SystemML internals
- Hyperinks to classes
- Code and documentation in the same file. Makes it easy to enforce "Making documentation part of a PR that affects user  APIs might be a good idea"
- Potentially could be exposed through systemml.apache.org
- Simplified documentation versioning

Note: this is just a developer documentation and is not a substitute for user guide/documentation.

@mboehm7 @fschueler @nakul02 @deroneriksson @dusenberrymw @lresende

### DIFF
--- a/src/main/java/org/apache/sysml/overview.html
+++ b/src/main/java/org/apache/sysml/overview.html
@@ -1,0 +1,99 @@
+<!--
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+-->
+
+<!-- 
+`mvn javadoc:javadoc` will compile the javadocs in the folder target/site/apidocs 
+-->
+<html>
+<body>
+<h1>SystemML Architecture</h1>
+Algorithms in Apache SystemML are written in a high-level R-like language called Declarative Machine learning Language (DML)
+or a high-level Python-like language called PyDML.
+SystemML compiles and optimizes these algorithms into hybrid runtime 
+plans of multi-threaded, in-memory operations on a single node (scale-up) and distributed MR or Spark operations on
+a cluster of nodes (scale-out). SystemML's high-level architecture consists of the following components:
+
+<h2>Language</h2>
+DML (with either R- or Python-like syntax) provides linear algebra primitives, a rich set of statistical
+functions and matrix manipulations, 
+as well as user-defined and external functions, control structures including parfor loops, and recursion. 
+The user provides the DML script through one of the following APIs:
+<ul>
+  <li>Command-line interface ( {@see org.apache.sysml.api.DMLScript} )</li>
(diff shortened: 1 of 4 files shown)


niketanpansare:
javadoc

nakul02:
After talking with @niketanpansare and looking at the documentation for the [javadoc](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/javadoc-resources.html) phase, I think the file `overview.html` should be in `src/main/javadoc/overview.html` and not where its currently put.

src/main/java/org/apache/sysml/overview.html
```diff
@@ -0,0 +1,99 @@
...
+  <li>Command-line interface ( {@see org.apache.sysml.api.DMLScript} )</li>
+  <li>Convenient programmatic interface for Spark users ( {@see org.apache.sysml.api.mlcontext.MLContext} )</li>
+  <li>Java Machine Learning Connector API ( {@see org.apache.sysml.api.jmlc.Connection} )</li>
+</ul>
+
+{@see org.apache.sysml.parser.AParserWrapper} performs syntatic validation and 
+parses the input DML script using ANTLR into a 
+a hierarchy of {@see org.apache.sysml.parser.StatementBlock} and 
+{@see org.apache.sysml.parser.Statement}  as defined by control structures.
+Another important class of the language component is {  @see org.apache.sysml.parser.DMLTranslator}
```
nakul02: Extra space here - `{  @see org.apache.sysml.parser.DMLTranslator}` causes it to not be rendered correctly.

nakul02:
@MadisonJMyers, @deroneriksson - do you think this content would be useful to have in the github wiki and link to it in the `README.md` ?

nakul02:
and @niketanpansare - thank you!
this is very good work and a much needed effort in the direction of documentation that helps newcomer.

src/main/java/org/apache/sysml/overview.html
niketanpansare: Good catch ... I will make the change :+1:

deroneriksson:
@nakul02 When we have the javadocs deployed on the main website, this documentation will appear when a user goes to the javadocs, so I believe I would use this path to get to the overview.

niketanpansare:
@deroneriksson @nakul02 I will move the overview.html to the directory that nakul recommended in my next commit :) ... Is there an umbrella JIRA for improving the documentation ?

deroneriksson:
@niketanpansare there are probably 200 things needed for improving the documentation, so i'm not sure if a single umbrella JIRA handles it all.

niketanpansare:
I understand. It just feels little awkward to push a PR without JIRA associated with it :(

deroneriksson:
Please feel free to create a JIRA for it

deroneriksson:
@niketanpansare Before merging please also verify that the overview renders correctly when 'mvn javadoc:javadoc ...' is run and also when 'mvn clean package -P distribution' generates the javadoc jar.

niketanpansare:
Both  'mvn javadoc:javadoc ...' and 'mvn clean package -P distribution' are tested and works correctly on my setup after recent commit.

deroneriksson:
@niketanpansare Great! I verified that both commands work correctly for me too with Java 8.

LGTM.

niketanpansare:
To compile Python doc, you need to install Sphinx
```bash
pip install -U Sphinx
```

Then prepare docs:
```bash
cd src/main/python
sphinx-apidoc --full --force --output-dir docs systemml
```
Add following lines to src/main/python/docs/conf.py
```python
import sys
sys.path.append('/home/biuser/nike/python/apache/incubator-systemml/src/main/python')
autodoc_mock_imports = ['pyspark', 'py4j.java_gateway', 'pyspark.mllib.common', 'pyspark.context', 'pyspark.sql', 'pyspark.ml', 'pyspark.ml.feature']
```

Make the doc
```bash
cd docs
make singlehtml
or 
make html
```

Here is a sample document generated by above steps:
[python_systemml_documentation.pdf](https://github.com/apache/incubator-systemml/files/503900/python_systemml_documentation.pdf)

niketanpansare:
Closed by the commit https://github.com/apache/incubator-systemml/commit/c7d990ce6f94b5ceae2ed37e62b66164cb15b8c8

PR #268, opened 2016-10-06
[SYSTEMML-565][WIP] Document frame support in dml language reference

gweidner:
Added separate section for Frame support in dml language reference manual.

View new content:
http://gweidner.github.io/incubator-systemml/dml-language-reference.html#frames


gweidner:
Any comments for corrections, additions or deletions are welcomed.  MLContext guide to be addressed separately.

deroneriksson:
cc @asurve @mboehm7

mboehm7:
thanks @gweidner - this looks great and it already revealed a shortcoming of our transformdecode: https://issues.apache.org/jira/browse/SYSTEMML-1018.

Some additional comments for further improvement:

* creating frames: you might want to rephrase the default schema a bit: schema="*" resolves to a string schema and if no schema is specified, the default is "*"

* casting: frame to matrix is a best effort operation, which tries to parse doubles - if there are for example strings that cannot be parsed, this operation produces errors.

* schema impliciations on casting: as.frame(matrix) produces a double schema, as.scalar(frame) produces of scalar of value type given by the frame schema.

* transformdecode: please mention that not all transformation are reversible (recoding, dummy coding and pass-through are reversible, i.e., subject to transformdecode, whereas binning, missing value imputation, and omit are not)

* transformapply: please add that this function applies existing meta data. In order to show the relationship to transformencode, you can mention that transformencode := build+apply, whereas transformapply := apply

* meta data as frames: please highlight (maybe in the transformencode example) that the meta data is simply a frame and hence all operations and read/write can similarly be applied to the meta data too.

* toString: it would be good to either extend the description of toString (to support frame inputs) or include it under frames too (its included in the examples but the tables are the actual language reference)

gweidner:
Thank you very much @mboehm7  for your input and time!  I will incorporate each item and push new commits to this PR.

asurve:
Thank you very much @gweidner for willing to  volunteer on documenting this work, really appreciate it.
It looks good for me. I have minor comments if you willing.
- Indexing Frames
  Please use B, C, D, E instead of G, H, I, J in right indexing assignments to show relevance. Please put  left/right indexing sub-section for those examples.

- Casting Frames:
   Please be consistent with inA/inputA, please describe what is inA/inputA etc
   Provide an additional statement to write with different format csv, text, binary 
- Manipulating Frames:
   Not sure how length() is going to work, never used.

- Thanks

gweidner:
Thank you @asurve for feedback.  I've incorporated all comments.

gweidner:
I've refreshed my gh-pages so the latest changes can be viewed at http://gweidner.github.io/incubator-systemml/dml-language-reference.html.  Summary of changes:
1) Rephrased default schema in Creating Frames section.
2) Added description in casting frames for potential errors since Java's Double.parseDouble() used internally.
3) Expanded all transform-related sections including adding table showing supported transform-types for encode, decode, apply.
4) Added toString description to Manipulating Frames table.
5) Clarified Indexing Frames example.
6) Added statements to Casting example.
7) Removed length from Manipulating Frames table although it seemed to work at least for one simple scenario.

gweidner:
Based on additional input, I've integrated toString, nrow, ncol function descriptions for frame's into existing matrix function tables (15 and 3).  Removed section/table 'Manipulating Frames'.

gweidner:
I've pushed new commit which updates both Creating Frames and transformencode sections to incorporate additional input.  Tabs were added to show impact of extra white space before column name in header.  Verified the given transform specifications matched the corresponding headers by running testHomesRecodeColnamesSingleNodeCSV with this example data.

I've refreshed my gh-pages so the latest changes can be viewed at http://gweidner.github.io/incubator-systemml/dml-language-reference.html.

deroneriksson:
LGTM. Thank you @gweidner for adding Frame support to our documentation. Great job!

deroneriksson:
I will merge this. Thank you @gweidner.

PR #270, opened 2016-10-13
[SYSTEMML-880] [WIP] Initial version of pushdown loop for Python DSL

niketanpansare:
Design decisions:
1.  In the initial version, I decided not to parse and walk the Python AST. The key assumption is that the **code to be pushed-down has to map to PyDML**. This will help us to identify the gaps in PyDML as well (for example: `dot(H, H.transpose())` instead of `H.dot(H.transpose))`. Also, cases such as `sml.matrix` is not handled.
2. The inputs and outputs need to be specified. There is a tradeoff here between **simplicity of the API+Implementation and optimization scope**. 
- Alternative to the current implementation is that inputs+outputs are not specified and the function to be pushed-down has access to all the variables in the global scope. In this case, we walk down the AST and identify all the variables on RHS and treat them as outputs and all other variables not created as input (hand-waving lil bit here). 
- Another alternative is to not require providing variable name in string format: `'H'`, but this was difficult  to implement due to pass-by-value and point 1.
- Also, I could not find a way to enable the decorator (`@parallelize`) on statement block instead of function.

``` python
import systemml as sml
from systemml import random
sml.setSparkContext(sc)

m, n = 100, 20
k = 40
V = sml.random.uniform(size=(m, n))
W = sml.random.uniform(size=(m, k))
H = sml.random.uniform(size=(k, n))

@sml.parallelize
def push_down_loop(inputs, outputs):
    max_iters = 200
    for i in range(0, max_iters):
        H = H * dot(W.transpose(), V)/dot(W.transpose(), dot(W, H))
        W = W * dot(V, H.transpose())/dot(W, dot(H, H.transpose()))

# The returned values as Matrix class
ret = push_down_loop(inputs={'H': H, 'W':W, 'V': V}, outputs=['H', 'W'])
H = ret[0].toNumPy()
W = ret[1].toNumPy()
```

@iyounus @dusenberrymw @mboehm7 @frreiss @bertholdreinwald @nakul02  @deroneriksson  @asurve

### DIFF
--- a/src/main/python/systemml/__init__.py
+++ b/src/main/python/systemml/__init__.py
@@ -22,7 +22,9 @@
 from .mlcontext import *
 from .defmatrix import *
 from .converters import *
+from .pushdown import *
 
 __all__ = mlcontext.__all__
 __all__ += defmatrix.__all__
 __all__ += converters.__all__
+__all__ += pushdown.__all__
--- a/src/main/python/systemml/mlcontext.py
+++ b/src/main/python/systemml/mlcontext.py
@@ -87,6 +87,7 @@ def _py2java(sc, obj):
     if isinstance(obj, Matrix):
         obj = obj._java_matrix
     # TODO: Port this private PySpark function.
+    # TODO: Handle Numpy Array
     obj = pyspark.mllib.common._py2java(sc, obj)
     return obj
 
--- a/src/main/python/systemml/pushdown.py
+++ b/src/main/python/systemml/pushdown.py
@@ -1,0 +1,62 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+from .defmatrix import *
+from . import MLContext, pydml
+
+from dill.source import getsource
+from textwrap import dedent
+
+__all__ = ['parallelize']
+
+def parallelize(fn):
+    """
+    The current implementation can only pushdown subset of Python that maps to PyDML.
+    Also, the inputs and outputs need to be explicitly specified.
+    We will relax this constraints in future version by performing AST analysis.
+    """
+    def inner(*args, **kwargs):
+        if matrix.ml is None:
+            raise Exception('Expected setSparkContext(sc) to be called.')
+        # ------------------------------------------------------------------------------------
+        # Assumption that @parallelize and function definition has one line. Fix this by having a class that extends ast.NodeVisitor
+        if 'inputs' not in kwargs or 'outputs' not in kwargs:
+            raise ValueError('The function to parallelize should contains exactly two named arguments with keys: inputs, outputs')
+        codeWithParallelizeHeader = getsource(fn)
+        # ------------------------------------------------------------------------------------
+        codeWithoutParallelizeHeader = ''.join(['\n', dedent(codeWithParallelizeHeader.split("\n",2)[2]), '\n'])
+        # print "Executing code as pydml:\n" + codeWithoutParallelizeHeader
+        script = pydml(codeWithoutParallelizeHeader)
+        for name, value in kwargs['inputs'].items():
+            if isinstance(value, matrix):
+                # TODO: Uncomment this after _py2java allow NumPy array
+                # value.eval()
+                # script.input(name, value.data)
+                script.input(name, value.toDataFrame())
+            else:
+                script.input(name, value)
+        map(lambda o: script.output(o), kwargs['outputs'])        
+        matrix.ml = MLContext(matrix.sc)
+        results = matrix.ml.execute(script)
+        return map(lambda o: results.get(o), kwargs['outputs'])
+    return inner
+
+        


frreiss:
This looks ok as an initial prototype cut. This change set really should include some explicit documentation on the limitations of the current implementation.

dusenberrymw:
In general, I think it is great that we are continuing to think about how to solve these kinds of problems in the DSL.  For this particular case pushing down for-loops, I'd like to step back and discuss other possible ideas.

In this case, we are assuming that we need to push the actual loop down into DML.  Overall, I'm concerned that the current prototype could end up causing problems due to users not understanding what subset of Python could be used within the function.  As an alternative, what if we instead only defined the graph of operations for a single iteration, and then evaluated that graph repeatedly in a Python for-loop?  For example:

```r
// Starting values
V = .... // some input data converted to a SystemML Matrix
W = sml.random.uniform(size=(m, k))
H = sml.random.uniform(size=(k, n))

// Train in a loop
for i in range(0, 200):
  H = H * dot(W.transpose(), V)/dot(W.transpose(), dot(W, H))
  W = W * dot(V, H.transpose())/dot(W, dot(H, H.transpose()))
  sml.eval(W, H)  // Eval the graph so that W & H are evaluated, starting with their previous values

// Now, W & H have the final values in SystemML matrix objects, which we can convert to NumPy:
W = W.toNumPy()
H = H.toNumPy()
```

We could also define the graph of operations for a single iteration outside of the for loop:

```r
// Starting values
V = .... // some input data converted to a SystemML Matrix
W = sml.random.uniform(size=(m, k))
H = sml.random.uniform(size=(k, n))

// Graph of operations for a single iteration
H = H * dot(W.transpose(), V)/dot(W.transpose(), dot(W, H))
W = W * dot(V, H.transpose())/dot(W, dot(H, H.transpose()))

// Train in a loop
for i in range(0, 200):
  sml.eval(W, H)  // Eval the graph so that W & H are evaluated, starting with their previous values

// Now, W & H contain their final values in SystemML matrix objects, which we can convert to NumPy, for example:
W = W.toNumPy()
H = H.toNumPy()
```

Thoughts?

niketanpansare:
@dusenberrymw @frreiss Thanks for your feedback.

We should support following two approaches as they are not mutually exclusive:

A. "the graph of operations for a single iteration, and then evaluated that graph repeatedly in a Python for-loop": This is currently supported without PR 270.

Pros: 
- Simple and user-friendly API especially to the Spark users who are familiar with lazy evaluation.

Cons: 
- Doesnot exploit loop optimization mentioned in the VLDB paper. Imran said that this approach didnot yield satisfactory performance for NMF script.

B. Push-down of loop via parallelize (this PR):

Pros: 
- Exploits loop optimization.
- If we plan reimplement our DML algorithms using Python DSL and ship it as part of mllearn, we should take this approach as it would have same performance as invoking the algorithm from commandline.

Cons:
- Limited by what PyDML supports -> It important to note that this is limitation of PyDML (and by transitive property DML) not parallelize construct. As we add more functionality to SystemML, this limitation will be addressed over time without a single line of change in parallelize construct.

May be I am missing something, but the "graph of operations for a single iteration outside of the for loop" option will not work as expected, you had already evaluated the operation during the first iteration. The remaining iteration will turn out to be "no op".

iyounus:
From API point of view, I would go with Mike's suggestions. The decorators are not used in general as part of APIs. At least I have not seen any. In my opinion, this will add a level complexity that many data scientist may not like deal with.

niketanpansare:
Just to be clear, then we will be making conscious choice that our Python API will:

- not be able to exploit any loop optimizations and 
- be significantly slower than simply converting the code to PyDML and invoking it from commandline.

dusenberrymw:
Yeah that's a good breakdown of the possible options.  I should have clarified it before, but the code snippets I gave don't necessarily work, but reflect different possible designs and what they might look like in practice.

"It important to note that this is limitation of PyDML (and by transitive property DML) not parallelize construct." -> Yes, you are technically correct that this is a limitation of PyDML.  The problem I was identifying is just that -- a user of this parallelize construct would have to know the exact subset of Python that would work with PyDML, and any deviation from it would cause errors.  Thus, although we are allowing the user to literally use Python, in reality, we are forcing them to learn which subset of Python corresponds to the underlying PyDML.

With regards to "Imran said that this approach did not yield satisfactory performance for NMF script", the problem was not that the performance was slightly unsatisfactory, but rather that it was essentially broken.  Evaluation of `H` was quick as it should be, but then the subsequent evaluation of `W` took over an hour to finish, despite using almost the same operations.  This occurred when the graph was constructed within the loop and then executed once at the end.  I would suggest we first explore the root cause of that issue.  Once we figure out that issue, the remaining step would be to identify the magnitude of the performance improvements gained via loop optimizations.

Regarding the graph of operations outside of the for loop, yes that won't work today (noted above -- should have added that to the original post).  It represents another possible design, where the idea would be to define what the operations would be, starting with an initial set of values, and then allow for that graph to be executed as many times as desired.  Therefore, we start with initial values of `H` and `W`, then define update operations, then evaluate a single set of update steps repeatedly within the loop.  Basically, this would be similar to a TensorFlow style.

fschueler:
If Imran's concern is about the decorator, maybe we could think about going back to making the iteration a (set of) special function(s) like `sml.iterate(...)`? 
This way the user wouldn't have to define her own function (`def push_down_loop(inputs, outputs)`) where for me it is not clear right now if the function signature is unrestricted or if I could annotate arbitrary functions with @sml.parallelize.
For convenience it might also be possible to inspect the closure of a `sml.iterate(...)` function and set inputs automatically? Not sure about the output handling though...

niketanpansare:
@dusenberrymw  With respect to PyDML, my point is: Any API can only support subset of features that DML/PyDML supports. Any deviation will result in an ERROR, either thrown by that API or by the language layer of SystemML. The user will not care where the error comes from, just that he cannot use Python DSL for his task. If on the other hand, we want to support a given deviation, I would prefer to extend PyDML/DML itself than make Python DSL more complex. 

To be fair, one kind of deviation that is not handled (but potentially possible): subset of operations in loop to be pushed down to SystemML and subset of operations executed in Python as external builtin function. This requires traversal of AST and I decided not to go in that direction for now; instead I added a restriction that we cannot only push-down the subset of Python which maps to PyDML.

My understanding is that Imran's code didnot break down after adding `eval(H, W)` in the for loop, but was much slower. If that is not the case, then I agree with you, we need to understand the root cause of the problem.

@fschueler The signature of push-down loop is restricted. In current setup, we can examine the closure too by traversing the AST without a need for new construct. In my earlier PR, I had proposed adding a new construct `FOR` instead of `iterate` and it was not well-received.

iyounus:
eval(H,W) did not have any effect on the calculation of W. It still takes one hour.

frreiss:
Ok, so it looks at this point like this PR does not actually fix the bug that it's associated with (SYSTEMML-1043). But it does represent a useful prototype developer API that future work could build on. How about if we open a new JIRA of type "improvement" to cover something like "initial support for pushing down the body of a function from Python DSL" and associate this change set with that JIRA? Then we can do something more simple and stable to cover SYSTEMML-1043.

dusenberrymw:

deroneriksson:
6 month check-in. What is the status of this PR? Merge or close?

niketanpansare:
I guess Mike had concerns about this PR. If he is OK, we can merge this as our dsl is still experimental.

deroneriksson:
@niketanpansare @dusenberrymw 
Is this PR is still under active development?

niketanpansare:
No, this PR is not under active development. Since we didnot come to a consensus on the high-level approach for pushing down loop in Python DSL, we can close this PR.

deroneriksson:
@niketanpansare Thank you

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1601/

PR #272, opened 2016-10-20
ALS_topk_predict.dml: fix typo

jdyer1:
fix a typo in ALS_topk_predict.dml


akchinSTC:

deroneriksson:
Welcome to SystemML, @jdyer1! It looks like this bug was introduced when the ppred function was removed from the script.

Thank you. LGTM.

PR #274, opened 2016-10-22
[SYSTEMML-1056] TransformEncode fails in metadata creation

asurve:


mboehm7:
copy from the dev list discussion - (2) and (3) are relevant to this PR:

1) Escaping: transformencode, transformdecode, and transformapply do not remove quotes to provide easy to understand semantics. If users want to match strings with different escaping policies to the same entry it's the user's responsibility to handle the unquoting. The nice side effect is that transformencode/transformapply and transformdecode are truly inverse operations, at least for reversible transformations like recoding and dummy coding.

2) Metadata frames: The schema for meta data frames is a string column per original column where each transformation type has its special serialization format. For example, for recoding, we serialize distinct <token><delim><code> (one entry per row). The reason why we use the quote-aware splitting on parsing this meta data is a best effort to handle cases where delim occurs inside the quoted token. A simply splitting on <delim> (as done in the "fix" by PR 274) would fail in this situation.

3) Solution: We could, however, simply flip the serialization format to <code><delim><token> which allows splitting on the first occurrence of <delim> because <code> is guaranteed not to include <delim>. Note that this would loose binary backwards compatibility to existing meta data frames though.

asurve:
1. I don't believe we are suggesting to remove quotes. If we are following RFC 4180 based string format, then input coming from any source to be converted to RC 4180 based string format before any processing should occur. This will avoid exception if string is not compliant to RFC 4180 based format. We may have at least three options:    

a. Make input string compliant to RFC 4180 format within code by escaping quotes etc. (No burden on user, but we need to be consistent to avoid different output based on input source.    
b. Ask user (programmer) to put data in RFC 4180 based string format (Either through some utility we add in SystemML but user will leverage this utility on needed input data. (User may not be happy with, as they have to do additional steps)   
c.  Output error message and exit processing (Again user may not be happy with)   

2. We know limitation of fix which is documented in code. Its unicode which may occur in original string (token), very uncommon scenario, but possible. 

3. WRT backward compatibility of existing meta data frames. Are these meta data frames in SystemML earlier released code? File based transform generates such meta data frames?

-Arvind

bertholdreinwald:
Let's keep this Jira and the discussion on the mailing list separate although the points raised in the discussion cause this issue here. 

Keep in mind that this issue only came about as strings where ingested that do NOT comply with the RFC spec.

I like the suggestion of switching to {code}{delim}{token} as it provides a more robust solution independent of the outcome of the discussion and the risk of people having the unicode delimiter in their string. In the code however, I would still not use the split.csv() but the much simpler strIndex(delim) because (1) we are only looking for 1 split, and (2) in this part of the code we do not care about the format of the token but only want the token. Quote-awareness etc. is dealt with in other parts of the system, and would only introduce unnecessary processing here.

As persisting the meta data frames were not released prior, we do not have to worry about backward compatibility.

Arvind, you should add the points a., b., and c. to the discussion in the mailing list.

src/main/java/org/apache/sysml/runtime/transform/RecodeAgent.java
```diff
@@ -543,13 +543,13 @@ public void initMetaData( FrameBlock meta ) {
...
-	 * @param code
-	 * @return
+	 * Returns the Recode map entry which consists of concatenation of code, delimiter and token. 
+	 * @param token	is part of Recode map
+	 * @param code  is code for token 
+	 * @return the concatenation of code and token with delimiter in between
 	 */
 	public static String constructRecodeMapEntry(String token, Long code) {
-		return token + Lop.DATATYPE_PREFIX + code.toString();
+		return code.toString() + Lop.DATATYPE_PREFIX + token;
```
frreiss: There's no need to change the format of the strings. Just search *backwards* for `Lop.DATATYPE_PREFIX` when deserializing the recode map in `getRecodeMap()`

bertholdreinwald: Awesome!

PR #276, opened 2016-10-28
[SYSTEMML-1073] Remove maven-deploy-plugin from pom

deroneriksson:
Move maven-deploy-plugin out of distribution profile so that double
deploy doesn't happen when deploy to snapshot repository.

See https://issues.apache.org/jira/browse/SYSTEMML-1073

cc @lresende

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -673,6 +673,28 @@
 		</profile>
 
 		<profile>
+			<id>deploy-release</id>
+			<build>
+				<plugins>
+					<plugin>
+						<artifactId>maven-deploy-plugin</artifactId>
+						<version>2.8.2</version>
+						<configuration>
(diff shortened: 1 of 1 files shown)


lresende:
@deroneriksson What happen if you just get rid of the profile and then run the script to deploy the snapshot ? If it works, we can basically eliminate this from the pom as the necessary config is already configured in Apache parent pom.

deroneriksson:
Hi @lresende. I think eliminating this from the pom would be best. I was keeping it in a profile in case it was needed to do the releases. I would be happy to remove this profile/plugin. Thanks for the feedback.

lresende:
LGTM

deroneriksson:
Thank you for reviewing @lresende. I'll merge.

PR #277, opened 2016-10-28
[SYSTEMML-1074] Update external function example to TimeWrapper

deroneriksson:
The DML Language Reference contains an external function example using JLapackEigenWrapper. This class does not exist in the project. Therefore, update example to an existing class, such as TimeWrapper.


niketanpansare:
LGTM. Thanks for taking care of it :+1:

deroneriksson:
Thank you for reviewing @niketanpansare. I'll merge.

PR #278, opened 2016-10-30
[SYSTEMML-1075] Added sizzle license for src artifact

gweidner:
Added license text from https://github.com/jquery/sizzle/blob/master/LICENSE.txt since jquery includes sizzle https://github.com/jquery/jquery/blob/1.12.0/dist/jquery.js in response to https://www.mail-archive.com/general@incubator.apache.org/msg56892.html.  The src distribution artifact (systemml-0.11.0-incubating-src) contains jquery-1.12.0.min.js at docs\js\vendor which has minimized sizzle.


deroneriksson:
LGTM.

gweidner:

PR #285, opened 2016-11-10
[SYSTEMML-1099] Mavenize the creation of python package

niketanpansare:
To create a python package,
```bash
 mvn post-integration-test -P distribution -DskipTests
```
If we want to use creation of python distribution to `package` phase instead we need to reorder the plugin such that it is called last as it depends on `target/SystemML.jar`.

This command will create `systemml-incubating-0.11.0.dev1.tar.gz` in the directory `src/main/python/dist/`. The content of this tar.gz file is as follows:
```bash
DISCLAIMER  LICENSE  MANIFEST.in  NOTICE  PKG-INFO  setup.cfg  setup.py  systemml  systemml_incubating.egg-info

```

To install this package, 
```bash
 pip install systemml-incubating-0.11.0.dev1.tar.gz
```

@lresende @deroneriksson @bertholdreinwald Let's use this PR to prepare the python distribution for 0.12 release.


niketanpansare:
Closed by the commit https://github.com/apache/incubator-systemml/commit/c94e86cfe6e78e00391df4967c83ac4ef2b259a0

deroneriksson:
@niketanpansare @lresende @bertholdreinwald 

Here are some issues to be aware of with this approach:

1) Binding to the "post-integration-test" phase of the maven lifecycle.
The maven standard is to have build artifacts with the "package" phase of the maven lifecycle.

2) All other artifacts (besides main jar, sources jar, and javadoc jar) use the assembly plugin to create them. This uses exec-maven-plugin to execute python, which is a totally different pathway. It also means that if you don't have python installed, you can't build this artifact.

3) The other artifacts contain their LICENSEs at src/assembly/*. This artifact has its LICENSE at src/main/python/. This spreads build information to different locations rather than keeping it localized in src/assembly.

4) If you run "mvn post-integration-test -P distribution -DskipTests", this artifact does not end up in the target directory like a normal artifact. It ends up in src/main/python/dist. This is not standard.

5) As a result of 4, if you do a "mvn clean", this artifact does not get deleted. Not being able to clean the artifact is confusing to maven users. Also, it can typically lead to artifacts that aren't built correctly.

6) If I build, the artifact generated is called "systemml-incubating-0.11.0.dev1.tar.gz". However, all the other artifacts in target are called "systemml-0.12.0-incubating-SNAPSHOT*". This means the version number for this artifact must not be coming from the pom. This makes it hard to maintain.

7) What is "dev1"? Is that a python standard? I do not understand what "dev1" means.

8) The "systemml-0.11.0.dev6.tar.gz" that was on pypi contained the "SystemML.jar" in it at systemml/systemml-java/. This "systemml-incubating-0.11.0.dev1.tar.gz" is 23K and does not contain the systemml jar file. Is it supposed to contain the SystemML jar file like it did before?

9) If the artifact is supposed to contain the SystemML jar file, then the LICENSE needs to be updated to reflect that it contains the antlr runtime and wink dependencies in it. If it is not supposed to contain the SystemML jar file, then it is fine.

niketanpansare:
Thanks @deroneriksson  for detailed comments. I would highly appreciate if you could help with issues 1, 5, 6, 8 and 9.

> 1) Binding to the "post-integration-test" phase of the maven lifecycle.
> The maven standard is to have build artifacts with the "package" phase of the maven lifecycle.
Yes, I could not find an easy way to ensure that SystemML.jar is created before the `generate-python-package` is invoked in package phase. 

> 2) All other artifacts (besides main jar, sources jar, and javadoc jar) use the assembly plugin to create them. This uses exec-maven-plugin to execute python, which is a totally different pathway. It also means that if you don't have python installed, you can't build this artifact.
That is correct. The profile `distribution` now requires that python is installed because SystemML python package is now a part of the distribution.

> 3) The other artifacts contain their LICENSEs at src/assembly/*. This artifact has its LICENSE at src/main/python/. This spreads build information to different locations rather than keeping it localized in src/assembly.
This was to address the issue raised by Apache mentors, they explictly wanted the LICENSE to be included in the root folder of tgz. @lresende : Can you please double-check this ?

> 4) If you run "mvn post-integration-test -P distribution -DskipTests", this artifact does not end up in the target directory like a normal artifact. It ends up in src/main/python/dist. This is not standard.
That is python's standard location for building the artifact. I understand here we have to chose whether to follow maven standard location or python standard location and I am fine with either.

> 5) As a result of 4, if you do a "mvn clean", this artifact does not get deleted. Not being able to clean the artifact is confusing to maven users. Also, it can typically lead to artifacts that aren't built correctly.
Good point. 

> 6) If I build, the artifact generated is called "systemml-incubating-0.11.0.dev1.tar.gz". However, all the other artifacts in target are called "systemml-0.12.0-incubating-SNAPSHOT*". This means the version number for this artifact must not be coming from the pom. This makes it hard to maintain.
Good catch. We need to update the setup.py to get 0.12.0

> 7) What is "dev1"? Is that a python standard? I do not understand what "dev1" means.
Yes.

> 8) The "systemml-0.11.0.dev6.tar.gz" that was on pypi contained the "SystemML.jar" in it at systemml/systemml-java/. This "systemml-incubating-0.11.0.dev1.tar.gz" is 23K and does not contain the systemml jar file. Is it supposed to contain the SystemML jar file like it did before?
Yes, the new tgz should have same layout as systemml-0.11.0.dev6.tar.gz.

> 9) If the artifact is supposed to contain the SystemML jar file, then the LICENSE needs to be updated to reflect that it contains the antlr runtime and wink dependencies in it. If it is not supposed to contain the SystemML jar file, then it is fine.

deroneriksson:
WRT 9
If the python tar.gz contains the main SystemML jar, it needs to reference the licenses for antlr runtime and wink, which are included in the SystemML jar. The LICENSE at src/assembly/jar/ contains the correct license information for the main jar.

WRT 1, 4, and 5
This can be quite tricky, since python is building the artifact in question. However, the way things are now, "package" doesn't generate the artifact, and if the artifact is generated, it doesn't go to where it's supposed to go using maven, which in addition to other issues means a default clean doesn't hit it.

Maven follows "convention over configuration", where in general a serious effort is made to conform to the maven standards wherever possible. Straying from those conventions greatly increases the difficulty of managing the build process.

These issues are further complicated because we need to be able to do things such as deploy our releases.  As @lresende can attest, this can be a difficult process even when maven conventions are followed. With the current set-up, it could require significant 'non-standard' modifications to the deployment process, whereas if "package" generates the artifact and the artifact goes to target/, it probably only requires a line or two of additions to the release script that Luciano has created.

Concerning 5, the maven-clean-plugin lets you do things like delete files in non-standard locations, but I don't recommend that approach. The 'proper' maven way to do things is to get the artifact to be generated by "package" and to have it go to "target".

However, I don't think this is an easy problem. Usually with maven, it can do what you want it to do, but it can take days of tinkering to find a clean solution. I would recommend looking for other maven projects that execute python to build artifacts.

WRT 2
An issue I see with requiring python to build the distribution profile artifacts is that if someone wants to clone from GitHub and build the artifacts such as the standalones, the person needs python. That would mean an R-centric data scientist would need to install python to build a standalone artifact. That could potentially drive off some potential users. In addition, this needs to integrate cleanly with the deploy process, so running things by Luciano is very important since he is our Apache deployment expert. cc @frreiss 

WRT 6
Off the top of my head, I do not know a simple solution. However, you need maven to be more involved. One potential thing might be to generate an artifact with a default name and see if you can have maven rename the artifact using the project version in the pom. The assembly plugin and antrun plugin might offer potential solutions or parts of a solution.

WRT 8
I believe what is going into the python artifact is being managed by python. I don't know python, so I probably am not a good person to ask about this one.

niketanpansare:
> WRT 9
> If the python tar.gz contains the main SystemML jar, it needs to reference the licenses for antlr runtime and wink, which are included in the SystemML jar. The LICENSE at src/assembly/jar/ contains the correct license information for the main jar.

That makes sense. I have updated the LICENSE in the PR https://github.com/apache/incubator-systemml/pull/290
 
> WRT 1, 4, and 5
> This can be quite tricky, since python is building the artifact in question. However, the way things are now, "package" doesn't generate the artifact, and if the artifact is generated, it doesn't go to where it's supposed to go using maven, which in addition to other issues means a default clean doesn't hit it.
> 
> Maven follows "convention over configuration", where in general a serious effort is made to conform to the maven standards wherever possible. Straying from those conventions greatly increases the difficulty of managing the build process.
> 
> These issues are further complicated because we need to be able to do things such as deploy our releases. As @lresende can attest, this can be a difficult process even when maven conventions are followed. With the current set-up, it could require significant 'non-standard' modifications to the deployment process, whereas if "package" generates the artifact and the artifact goes to target/, it probably only requires a line or two of additions to the release script that Luciano has created.

Yes, I understand and agree with those points, but with my limited mvn knowledge, I could not come up with better way to deal with following restrictions:
- generate-python-package expects SystemML.jar in the target folder.
- pre_setup.py script is executed before setup.py (both are invoked via exec-maven-plugin).

If you or @lresende can come up with better alternative, I would highly appreciate your help. 

> Concerning 5, the maven-clean-plugin lets you do things like delete files in non-standard locations, but I don't recommend that approach. The 'proper' maven way to do things is to get the artifact to be generated by "package" and to have it go to "target".
> 
> However, I don't think this is an easy problem. Usually with maven, it can do what you want it to do, but it can take days of tinkering to find a clean solution. I would recommend looking for other maven projects that execute python to build artifacts.

That is fair point. I looked around for other projects that build python artifacts and unfortunately, this is best I could come up with. Maven and pip are not well integrated yet :( 

> WRT 2
> An issue I see with requiring python to build the distribution profile artifacts is that if someone wants to clone from GitHub and build the artifacts such as the standalones, the person needs python. That would mean an R-centric data scientist would need to install python to build a standalone artifact. That could potentially drive off some potential users. In addition, this needs to integrate cleanly with the deploy process, so running things by Luciano is very important since he is our Apache deployment expert. cc @frreiss

I thought we agreed upon that python packages should be built using mvn. If this is not the case, then a simple shell script mentioned in my email would work (and we can simply comment out the generate-python-package from our pom.xml) :) @frreiss @bertholdreinwald @lresende @dusenberrymw 

Since python files are also part of SystemML project (like our scala files), requiring a standard python installation for build is a fair ... according to me. 

> WRT 6
> Off the top of my head, I do not know a simple solution. However, you need maven to be more involved. One potential thing might be to generate an artifact with a default name and see if you can have maven rename the artifact using the project version in the pom. The assembly plugin and antrun plugin might offer potential solutions or parts of a solution.

Sure, we can explore that as part of separate PR.

deroneriksson:
BTW I was getting the following warning:
```
[WARNING] Some problems were encountered while building the effective model for org.apache.systemml:systemml:jar:0.12.0-incubating-SNAPSHOT
[WARNING] 'profiles.profile[distribution].plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.codehaus.mojo:exec-maven-plugin @ line 833, column 14
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

I just committed a fix to reduce the two exec-maven-plugin declarations to a single one with two executions. On my machine, this happened to fix 8, so the SystemML jar now appears in the python tar.gz.

PR #286, opened 2016-11-10
[SYSTEMML-1102] Argument to print top n heavy hitters

nakul02:
Added an argument for "-stats" option to specify the top n heavy hitters.
@deroneriksson - is this useful in general?
@niketanpansare, @bertholdreinwald - this will help with the gpu performance investigation.


nakul02:
An example of this - 

image

deroneriksson:
LGTM. Great idea @nakul02!

We might also want to eventually add something like a setStatisticsCount() to the new MLContext API.

nakul02:
Thanks @asurve & @deroneriksson !

nakul02:
Thanks @deroneriksson

PR #287, opened 2016-11-11
[SYSTEMML-1104] Can set max heavy hitters to be printed with statistics from MLContext

nakul02:
Builds on PR #286 
Ping @deroneriksson

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
@@ -100,6 +100,11 @@ public class MLContext {
 	private boolean statistics = false;
 
 	/**
+	 * The number of heavy hitters that are printed as part of the statistics option
+	 */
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
```diff
@@ -100,6 +100,11 @@
 	private boolean statistics = false;
 
 	/**
+	 * The number of heavy hitters that are printed as part of the statistics option
+	 */
+	private int statisticsHeavyHitters = 10;
```
deroneriksson: Hi @nakul02. Would you mind updating the field **statisticsHeavyHitters** to **statisticsMaxHeavyHitters** to match the setStatisticsMaxHeavyHitters setter name and statisticsMaxHeavyHitters in ScriptExecutor?

deroneriksson:
Thank you @nakul02. LGTM.

PR #288, opened 2016-11-14
[SYSTEMML-1112] Add Scala Algorithms API For Spark ML Pipeline

asurve:
This change covers following and need more changes in later PRs to follow
  - Working algorithms for Cox, DecisionTree, Means, L2NormSVM, LinearRegressionCG, LinearRegressionDS, MultiClassSVM, NaiveBayes, RandomForest
  - Scala test cases
  - Examples for algorithms wrappers added

Tasks to be finished which will be added in future PRs
   - Running multiple suites (or consolidating scala test suites)
   - Supporting Python wrappers
   - Moving from old MLContext to new MLContext
   - Editor setting to avoid display false error messages


deroneriksson:
Wow that's a huge amount of work @asurve! I noticed https://github.com/apache/incubator-systemml/pull/269 of @gweidner updates the Spark version to 1.6.2 and this updates the Spark version to 1.6.0. Should we go with 1.6.2 @asurve @gweidner?

Also, can this be tested with "mvn clean verify"?

asurve:
We should stay with lower version at possible to be compatible unless there are known issues with lower version. WE should start with 1.6.0 instead of 1.6.2.

niketanpansare:
@asurve First of all, I really appreciate you putting so much effort in this PR, but I will have to give a -1.

I would highly recommend:
1. Writing wrappers in Scala for uniformity.
2. Using new MLContext because the old one is going to get deprecated.
3. Maximizing code reuse by deriving from BaseSystemMLClassifier or BaseSystemMLRegressor (or you can add a new Base class).

The above 3 points simplifies the integration with Python wrappers (i.e. mllearn - supports not just DataFrame but also NumPy/SciPy/Pandas DF) and will also help in maintaining the wrappers. For each algorithm, the core logic rests in two methods:
- `def getTrainingScript(isSingleNode:Boolean):(Script, String, String)`
- `def getPredictionScript(mloutput: MLResults, isSingleNode:Boolean): (Script, String)`

dusenberrymw:

niketanpansare:
@asurve  Also, when you are back, can you please replace this with your PL_Scala2 branch to avoid confusion ?

asurve:
Separate PR will be opened to add wrappers per new wrapper framework.

PR #290, opened 2016-11-18
[SYSTEMML-1116] Make SystemML Python DSL NumPy-friendly

niketanpansare:
- Bumped up the Python package version number @deroneriksson 

- Allow SystemML matrix to be invoked by numpy methods ( @dusenberrymw  ):
```python
import systemml as sml
import numpy as np
m1 = sml.matrix(np.ones((3,3)) + 2)
m2 = sml.matrix(np.ones((3,3)) + 3)
np.add(m1, m2).toNumPyArray()
```

- Allow users to control the depth of lazy evaluation (as per @frreiss  suggestion).
```python
>>> import systemml as sml
>>> import numpy as np
>>> m1 = sml.matrix(np.ones((3,3)) + 2)

Welcome to Apache SystemML!

>>> m2 = sml.matrix(np.ones((3,3)) + 3)
>>> np.add(m1, m2) + m1
# This matrix (mVar4) is backed by below given PyDML script (which is not yet evaluated). To fetch the data of this matrix, invoke toNumPyArray() or toDataFrame() or toPandas() methods.
mVar2 = load(" ", format="csv")
mVar1 = load(" ", format="csv")
mVar3 = mVar1 + mVar2
mVar4 = mVar3 + mVar1
save(mVar4, " ")


>>> sml.set_max_depth(1)
>>> m1 = sml.matrix(np.ones((3,3)) + 2)
>>> m2 = sml.matrix(np.ones((3,3)) + 3)
>>> np.add(m1, m2) + m1
# This matrix (mVar8) is backed by NumPy array. To fetch the NumPy array, invoke toNumPyArray() method.
```

Few caveats:
- The current version of NumPy explicitly disables overriding `ufunc`, but this should be enabled in next release. Until then to test above code, please use:

```bash
git clone https://github.com/niketanpansare/numpy.git
cd numpy
python setup.py install
```

- Following NumPy ufunc that are not yet implemented in SystemML's Python DSL:
```bash
conj, hyperbolic/inverse-hyperbolic functions(i.e. sinh, arcsinh, cosh, ...), bitwise operator, xor operator, isreal, iscomplex, isfinite, isinf, isnan, copysign, nextafter, modf, frexp, trunc
```


dusenberrymw:
This is quite interesting!  Integrating the DSL more tightly with NumPy in this manner is great. I haven't tried it yet, but +1 to the example I see.

As for the depth of evaluation, what it we changed it to something simpler such as `sml.set_lazy(False)` to disable lazy evaluation?

niketanpansare:
Thanks @dusenberrymw ... My intent is that if we can even scale subset of external libraries that uses numpy/scipy with minimal change (for eg: scikit-learn), it will be a great selling point for  SystemML :)

`sml.set_lazy(False)` is much cleaner interface. I can update it in my next commit :+1:

niketanpansare:
Completed tasks in this PR:
1. Added python test cases for matrix.
2. Added web documentation for all the Python APIs: http://niketanpansare.github.io/incubator-systemml/python-reference
3. Much tighter integration with NumPy (and several additional operators supported) via https://docs.scipy.org/doc/numpy-dev/reference/ufuncs.html
4. Facility to enable and disable lazy evaluation via `set_lazy` method.
5. matrix class itself has almost all basic linear algebra operators supported by DML. 

Remaining tasks:
1. Update SystemML.jar to *-incubating.jar
2. Integrate python testcases with maven and jenkins.
3. Large-scale testing: The current matrix has a huge performance issue (due to py4j conversion) especially in non-lazy setting or when evaluated too often. 
4. Pushdown-able loop structures.
5. Explore integration of non-unfuncs in Numpy (such as concatenate)
6. Refactor scikit-learn to make it scalable (by above point 5 and also by not redundantly forcing numpy array conversion).

@bertholdreinwald @dusenberrymw @deroneriksson  @lresende  @asurve If it's OK with you all, I can merge this in.

niketanpansare:
Also, updated the issue https://issues.apache.org/jira/browse/SYSTEMML-1043

src/main/python/systemml/classloader.py
```diff
@@ -51,8 +51,12 @@ def createJavaObject(sc, obj_type):
...
         return _createJavaObject(sc, obj_type)
     except (py4j.protocol.Py4JError, TypeError):
-        import imp
-        jar_file_name = os.path.join(imp.find_module("systemml")[1], "systemml-java", 'SystemML.jar')
+        import imp, fnmatch
+        jar_file_name = '_ignore.jar'
+        java_dir = os.path.join(imp.find_module("systemml")[1], "systemml-java")
+        for file in os.listdir(java_dir):
+            if fnmatch.fnmatch(file, 'systemml-*-incubating-SNAPSHOT.jar'):
+                jar_file_name = os.path.join(java_dir, file)
```
dusenberrymw: Have we tested if this JAR will be distributed to the worker nodes on a cluster as well?

src/main/python/tests/test_matrix_agg_fn.py
```diff
@@ -0,0 +1,95 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+# To run:
+#   - Python 2: `PYSPARK_PYTHON=python2 spark-submit --master local[*] --driver-class-path SystemML.jar test_mlcontext.py`
+#   - Python 3: `PYSPARK_PYTHON=python3 spark-submit --master local[*] --driver-class-path SystemML.jar test_mlcontext.py`
```
dusenberrymw: Can you update these two test invocation lines with this filename?

src/main/python/tests/test_matrix_binary_op.py
```diff
@@ -0,0 +1,138 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+# To run:
+#   - Python 2: `PYSPARK_PYTHON=python2 spark-submit --master local[*] --driver-class-path SystemML.jar test_mlcontext.py`
+#   - Python 3: `PYSPARK_PYTHON=python3 spark-submit --master local[*] --driver-class-path SystemML.jar test_mlcontext.py`
```
dusenberrymw: Can you update these two test invocation lines with this filename?

src/main/python/tests/test_mllearn_df.py
```diff
@@ -0,0 +1,108 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+# To run:
+#   - Python 2: `PYSPARK_PYTHON=python2 spark-submit --master local[*] --driver-class-path SystemML.jar test_mllearn.py`
+#   - Python 3: `PYSPARK_PYTHON=python3 spark-submit --master local[*] --driver-class-path SystemML.jar test_mllearn.py`
```
dusenberrymw: Can you update these two test invocation lines with this filename?

src/test/java/org/apache/sysml/test/integration/functions/python/PythonTestRunner.java
```diff
@@ -0,0 +1,119 @@
...
+import java.lang.ProcessBuilder.Redirect;
+import java.util.Map;
+import org.apache.sysml.runtime.DMLRuntimeException;
+import org.apache.sysml.test.integration.AutomatedTestBase;
+import org.apache.sysml.test.integration.TestConfiguration;
+import org.junit.Test;
+
+/**
+ * To run Python tests, please:
+ * 1. Set the RUN_PYTHON_TEST flag to true.
```
dusenberrymw: Let's aim to run this test by default on Jenkins.

src/test/java/org/apache/sysml/test/integration/functions/python/PythonTestRunner.java
```diff
@@ -0,0 +1,119 @@
...
+//		org.apache.spark.deploy.SparkSubmit.main(args);
+		Map<String, String> env = System.getenv();
+		if(!env.containsKey("SPARK_HOME")) {
+			throw new DMLRuntimeException("Please set the SPARK_HOME environment variable");
+		}
+		String spark_submit = env.get("SPARK_HOME") + File.separator + "bin" + File.separator + "spark-submit";
+		if (System.getProperty("os.name").contains("Windows")) {
+			spark_submit += ".cmd";
+		}
+		Process p = new ProcessBuilder(spark_submit, "--master", "local[*]",
```
dusenberrymw: Let's test with Python 2 & 3 here.  The `PYSPARK_PYTHON` environment variable can be used, i.e., `PYSPARK_PYTHON=python3 spark-submit .....`.

src/main/python/systemml/defmatrix.py
```diff
@@ -60,24 +67,42 @@ def __init__(self, inputs, dml=None):
...
 
+    MAX_DEPTH = 0
+    
     def _visit(self, execute=True):
         matrix.dml = matrix.dml + self.dml
 
-    # Don't use this method instead use matrix's printAST()
-    def printAST(self, numSpaces):
+    # Don't use this method instead use matrix's print_ast()
+    def print_ast(self, numSpaces):
```
dusenberrymw: Let's rename this to `_print_ast` to indicate that it is a private method not for use by the user.

src/main/python/systemml/defmatrix.py
```diff
@@ -116,50 +165,74 @@ def performDFS(outputs, execute):
...
-    return constructIntermediateNode([X], [OUTPUT_ID, ' = ', fnName,'(', getValue(X), ')\n'])
+    inputs = []
+    lhsStr, inputs = _matricize(X, inputs)
+    return construct_intermediate_node(inputs, [OUTPUT_ID, ' = ', fnName,'(', lhsStr, ')\n'])
 
+def seq(start=None, stop=None, step=1):
+    """
+    Creates a single column vector with values starting from <start>, to <stop>, in increments of <step>.
+    Note: Unlike Numpy's arange which returns a row-vector, this returns a column vector.
+    Also, Unlike Numpy's arange which doesnot include stop, this method includes stop in the interval.
```
dusenberrymw: I would prefer to only support the equivalent of `np.arange` in this Python DSL, rather than introduce our R-like `seq` function.  The users of the Python DSL will be Python users, so they will expect every aspect of our Python DSL to be Pythonic.

src/main/python/systemml/defmatrix.py
```diff
@@ -555,66 +616,296 @@ def __repr__(self):
...
+            multiline_dml = multiline_dml + [ OUTPUT_ID, '[1,0] = ', clen_ID, '\n' ]
+            ret = construct_intermediate_node(inputs, multiline_dml).toNumPyArray()
+            self._shape = tuple(np.array(ret, dtype=int).flatten())
+        return self._shape 
+    
+    shape = property(fget=get_shape, fset=set_shape)
+    
+    def __numpy_ufunc__(self, func, method, pos, inputs, **kwargs):
+        """
+        This function enables systemml matrix to be compatible with NumPy's ufuncs.
```
dusenberrymw: Can you add a bit of documentation to the parameters here?

src/main/python/tests/test_matrix_agg_fn.py
```diff
@@ -0,0 +1,95 @@
...
+class TestMatrixAggFn(unittest.TestCase):
+
+    def test_sum1(self):
+        self.assertTrue(np.allclose(sml.matrix(m1).sum(), m1.sum()))
+
+    def test_sum2(self):
+        self.assertTrue(np.allclose(sml.matrix(m1).sum(axis=0), m1.sum(axis=0)))
+    
+    def test_sum3(self):
+        self.assertTrue(np.allclose(sml.matrix(m1).sum(axis=1), m1.sum(axis=1).reshape(dim, 1)))
```
dusenberrymw: Can we add a quick note to the reference guide that explains that we always return a 2d matrix (ex. (3,1)), while NumPy can return a 1d vector (i.e. (3,))?

dusenberrymw: Also, ditto for the other functions below with this behavior.

src/main/python/tests/test_mllearn_df.py
```diff
@@ -0,0 +1,108 @@
...
+from systemml.mllearn import LinearRegression, LogisticRegression, NaiveBayes, SVM
+
+sc = SparkContext()
+sqlCtx = SQLContext(sc)
+
+# Currently not integrated with JUnit test
+# ~/spark-1.6.1-scala-2.11/bin/spark-submit --master local[*] --driver-class-path SystemML.jar test.py
+class TestMLLearn(unittest.TestCase):
+
+    def testLogisticSK2(self):
```
dusenberrymw: Nitpick: To be in line with the Pythonic style of the rest of the DSL codebase, can we update these functions from CamelCase to underscore_case?

niketanpansare:
Thanks @dusenberrymw for your review. I will incorporate your suggestion in the next commit :)

niketanpansare:
> Can you implement np.dot(...)?
> Can we update toNumPyArray() to toNumPy() for simplicity?
> Can we update toDataFrame() to toDF() for simplicity and to be the same as Spark?
Addressed in the commit https://github.com/apache/incubator-systemml/pull/290/commits/afac2b2eeca75b7fd9474220d31caf7f357d60ec

src/main/python/systemml/classloader.py
niketanpansare: Yes. Please try following script on your spark cluster:

```python
import systemml as sml
from systemml.random import uniform
m1 = uniform(size=(10000,10000))
m2 = m1.dot(m1).sum().toNumPy()
print(m2)
```

niketanpansare:
Addressed remaining comments in the commit: https://github.com/apache/incubator-systemml/pull/290/commits/b4b8394d3aa46c0976aceee85db85529eab67bd0 

@dusenberrymw can you please review ? ... Let's try to push this PR in today if possible as https://github.com/apache/incubator-systemml/pull/305 is dependent on this PR.

src/main/python/systemml/defmatrix.py
```diff
@@ -413,7 +417,7 @@ class matrix(object):
        Then the left-indexed matrix is set to be backed by DMLOp consisting of following pydml:
        left-indexed-matrix = new-deep-copied-matrix
        left-indexed-matrix[index] = value
-    8. Please use m.print_ast() and/or  type `m` for debugging. Here is a sample session:
+    8. Please use m._print_ast() and/or  type `m` for debugging. Here is a sample session:
```
dusenberrymw: The function name change for this internal function looks good, but I think the documentation is wrong now -- the user should still use the matrix `print_ast()` function, right?

src/main/python/systemml/defmatrix.py
```diff
@@ -60,24 +67,42 @@ def __init__(self, inputs, dml=None):
...
 
+    MAX_DEPTH = 0
+    
     def _visit(self, execute=True):
         matrix.dml = matrix.dml + self.dml
 
-    # Don't use this method instead use matrix's printAST()
-    def printAST(self, numSpaces):
+    # Don't use this method instead use matrix's _print_ast()
+    def _print_ast(self, numSpaces):
```
dusenberrymw: Okay the name change for this function `def _print_ast(self, numSpaces)` is good since it shouldn't be called by the user.  However, the other function that is supposed to be used instead should still be called `print_ast()` without an underscore.  Otherwise, they both seem like internal functions.

src/main/python/systemml/defmatrix.py
```diff
@@ -504,23 +569,23 @@ def _visit(self, execute=True):
...
         else:
             raise Exception('Expected either op or data to be set')
-        if self.data is None and self.output:
-            self._registerAsOutput(execute)
+        if self.eval_data is None and self.output:
+            self._register_as_output(execute)
         return self
 
-    def printAST(self, numSpaces = 0):
+    def _print_ast(self, numSpaces = 0):
```
dusenberrymw: Yeah let's make this function named `print_ast()` since it is intended to be called by the user, and the other variant that is not supposed to be called by the user can still be named `_print_ast(numSpaces)`.

src/main/python/systemml/defmatrix.py
```diff
@@ -555,66 +620,304 @@ def __repr__(self):
...
+        """
+        if not isinstance(self.eval_data, SUPPORTED_TYPES):
+            # Only warn if there is an unevaluated operation (which could potentially generate large matrix or if data is non-supported singlenode formats)
+            import inspect
+            frame,filename,line_number,function_name,lines,index = inspect.stack()[1]
+            msg = 'Conversion from SystemML matrix to NumPy array (occurs in ' + str(filename) + ':' + str(line_number) + ' ' + function_name + ")"
+            if matrix.THROW_ARRAY_CONVERSION_ERROR:
+                raise Exception('[ERROR]:' + msg)
+            else:
+                print '[WARN]:' + msg
```
dusenberrymw: Can you update this to Python 3 compatible syntax, as this currently throws an error as soon as the `systemml` package is imported.

src/main/python/systemml/defmatrix.py
```diff
@@ -764,33 +1184,31 @@ def _aggFn(self, fnName, axis):
...
 
     def __getitem__(self, index):
         """
         Implements evaluation of right indexing operations such as m[1,1], m[0:1,], m[:, 0:1]
         """
-        dmlOp = DMLOp([self])
-        out = matrix(None, op=dmlOp)
-        dmlOp.dml = [out.ID, ' = ', self.ID ] + getIndexingDML(index) + [ '\n' ]
-        return out
+        print '__getitem__'
```
dusenberrymw: Can you update this to Python 3 compatible syntax, as this currently throws an error as soon as the `systemml` package is imported.

niketanpansare:
@dusenberrymw addressed above comments in the commit https://github.com/apache/incubator-systemml/pull/290/commits/ecc242b2175ffd1ee230b60c9add4c344b4fbf35

dusenberrymw:
Also, we should closely follow the NumPy ufunc stuff.  Looks like they are revisiting the idea, but may be renaming the function to __array_ufunc__ \[1].  The ideal case would be to not have to rely on a custom version of NumPy.

\[1]: https://github.com/numpy/numpy/pull/8247

niketanpansare:
Once https://github.com/numpy/numpy/pull/8247 is merged and numpy is released, we can update our setup to have that version as dependency :)

deroneriksson:
Thank you @niketanpansare for all your hard work on this PR!

PR #291, opened 2016-11-18
[SYSTEMML-1118]Updated to use JCuda 0.8.0 

nakul02:
This PR updates systemml to use cuda 8 and cudnn 5.1.
Use has been significantly simplified compared to the steps listed [here](https://github.com/apache/incubator-systemml/blob/master/docs/devdocs/gpu-backend.md).

The user need only invoke maven to make the package and the cuda native shared libraries will be available wrapped in jars in the classpath without the need to set any environment variables or extra java options. (tested on a Centos 7 x86_64 system).

mavenized-jcuda and jcuda packages have been excluded in the shade plugin to stop them from being included in the package. This decision can be revisited in the future.

Note - This is a WIP. We still await the availability of a repo from where we can fetch mavenized-jcuda with cuda 8.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,10 +71,12 @@
 		<scala.test.version>2.2.6</scala.test.version>
 		<maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss z</maven.build.timestamp.format>
 		<enableGPU>false</enableGPU>
+		<jcuda.scope>provided</jcuda.scope>
+		<jcuda.version>0.8.0</jcuda.version>
 		<!-- OS-specific JVM arguments for running integration tests -->
 		<integrationTestExtraJVMArgs />
 	</properties>
-	
+
 	<repositories>
         <repository>
           <id>central</id>
@@ -82,14 +84,6 @@
           <releases>
             <enabled>true</enabled>
           </releases>
-        </repository>
-        <repository>
-            <id>mavenized-jcuda-mvn-repo</id>
-            <url>https://raw.github.com/niketanpansare/mavenized-jcuda/mvn-repo/</url>
-            <snapshots>
-                <enabled>true</enabled>
-                <updatePolicy>always</updatePolicy>
-            </snapshots>
         </repository>
     </repositories>
 
@@ -169,6 +163,13 @@
 						<goals>
 							<goal>shade</goal>
 						</goals>
+						<configuration>
+						  <artifactSet>
+							  <!--<excludes>
+								<exclude>org.jcuda:*</exclude>
+							</excludes>-->
+						  </artifactSet>
+						</configuration>
 					</execution>
 				</executions>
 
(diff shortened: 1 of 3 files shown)


nakul02:
Ping @niketanpansare

dusenberrymw:
+1 on the update to the latest cuda + cudnn.

niketanpansare:
@nakul02 Can we please close this PR and use https://github.com/apache/incubator-systemml/pull/307 instead ?

bertholdreinwald:
but let's keep native BLAS and GPU backend separate.

niketanpansare:
can we please move this discussion to https://github.com/apache/incubator-systemml/pull/307 ?

niketanpansare:
@nakul02 I take back my earlier comment. For now, let's use the approach in your PR.

LGTM after resolving conflicting files :+1:

nakul02:
This PR is now updated to use jcuda8 artifacts available on maven central.
@niketanpansare - could you please try this on your windows machine?
@asurve - for now, I am including the jcuda* artifacts in the uber/package jar. If this is not ok, I can change this.

niketanpansare:
LGTM :+1:

deroneriksson:
@nakul02 Thank you for helping getting jcuda deployed to maven central! This will be of great assistance to many people out there.

asurve:
I would not recommend to have non-SystemML related code in Uber jar file. We do have AntLR and wink files in Uber jar, but I am not sure if thats right approach. Unless you convince other way, I would prefer to have jcuda files outside Uber jar.

pom.xml
```diff
@@ -1032,50 +1086,96 @@
...
-		    </exclusions>
-		</dependency>
-		<!-- Since there is no mvn repo for jcuda
+
 		<dependency>
 			<groupId>org.jcuda</groupId>
 			<artifactId>jcuda</artifactId>
-			<version>0.7.5b</version>
-			<scope>provided</scope>
+			<version>${jcuda.version}</version>
```
niketanpansare: @nakul02  can you please mark all jcuda related artifacts as "provided" dependency ? The PR increases the size of SystemML.jar from 4.7 MB to 6.4 MB.

nakul02:
@asurve- I suggested adding the jcuda jars to the Uber jar because those jars may not be available in the environment they are run in (Spark) 
The obvious disadvantages are related to release process (Per platform, testing on those platforms, licenses).

nakul02:
If the `scope` is set to `provided`, it becomes the user's responsibility to download the jcuda jars (there are 16 jars). This makes using a deployment a pain for the user.

On the other hand, (as @niketanpansare and I spoke offline about), the jar becomes bigger and for a feature that is not used every single time, maybe making the jar bigger is not a good idea.

@lresende, @deroneriksson, @bertholdreinwald - do you have any comments/thoughts?
If not, I'll do what @niketanpansare suggested.

lresende:
Assuming jcuda is compatible with Apache License, I see two possible solutions :
- in the binary distro, have something like lib-cuda and have the necessary libs for platforms there
- the other approach, might be just help the users with a simple script that might download the necessary dependencies and put on the right place for the user.

I would prefer not to have this in the "fat-jar"  particularly because of size and platform dependency. But then we might need to add things to classpath in order to run systemml with GPU support.

nakul02:
@asurve, @lresende - excluded jcuda from jar and uber jar. The user will be responsible for providing them.

asurve:
Thanks @nakul02 LGTM

nakul02:
Docs will be available in https://github.com/apache/incubator-systemml/blob/master/docs/devdocs/gpu-backend.md.

Will merge.

PR #292, opened 2016-11-23
[SYSTEMML-1125] Enable REMOTE_SPARK in constrained optimizer

fschueler:
Forced Spark mode in parfor (mode=REMOTE_SPARK) was not used because the optimizer did not check for Spark execution mode.
This commit adds Spark to the optimizer and enables setting the forced parameter.


fschueler:
It seems that Intellij removed all the whitespaces at the end of lines and tabs in empty lines. While I think that's a good thing (why do we need tabs in empty lines) I would remove them if needed.

deroneriksson:
@fschueler I really like the idea of removing the endline whitespaces since this tends to be the git standard and `git diff` highlights the whitespaces as potential issues. Doing a regular expression search for `[ \t]+$`, it looks like the systemml project has ~75K lines with endline whitespaces!

deroneriksson:
LGTM. Updates appear logically correct and symmetrical between MR and Spark.

See conversation at https://www.mail-archive.com/dev@systemml.incubator.apache.org/msg01158.html

PR #295, opened 2016-11-29
[SYSTEMML-1127] Synchronize creation of the cache directory

fschueler:
This fixes a race condition that occurred while initializing the cache directory on remote spark workers. By synchronizing this, the problem should be resolved.

As per @mboehm7 's suggestion I synchronized the creation of the temporary cache directory. Initialization of the cache was already synchronized.

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/parfor/util/IDHandler.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/parfor/util/IDHandler.java
@@ -122,9 +122,8 @@ public class IDHandler
 		    //get ip address
 		    InetAddress addr = InetAddress.getLocalHost();
 		    String host = addr.getHostAddress();
-		    //addr.getHostName()
-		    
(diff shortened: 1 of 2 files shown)


deroneriksson:
LGTM. Thank you @fschueler!

j143:
Has the synchronization enabled in other way, than this PR. I believe there is no `synchronized` in the [file](https://github.com/apache/systemml/blob/master/src/main/java/org/apache/sysml/runtime/util/LocalFileUtils.java#L313).

cc @deroneriksson

mboehm7:
@j143 yes, I had to change this to make both creation and cleanup robust enough. We now synchronize in parfor workers because multiple setups and cleanup registrations needs to happen atomically. The cleanup is handled via a shutdown hook of the executor process.

The relevant commit and jira are here
https://github.com/apache/systemml/commit/b78c125934fa7a947a5118f0c08473afa926fa5d
https://issues.apache.org/jira/browse/SYSTEMML-1309

PR #303, opened 2016-11-30
[SYSTEMML-1131] Check hops for null before recompiling

fschueler:
This makes sure that we check for hops that are null when recompiling dynamically. Hops can be null in the case of simple literals that are not encoded as hop dags (e.g. predicates).

### DIFF
--- a/src/main/java/org/apache/sysml/hops/recompile/Recompiler.java
+++ b/src/main/java/org/apache/sysml/hops/recompile/Recompiler.java
@@ -404,8 +404,8 @@ public class Recompiler
 		ArrayList<Instruction> newInst = null;
 
 		//need for synchronization as we do temp changes in shared hops/lops
-		synchronized( hops ) 
-		{	
+		synchronized( hops )
+		{
 			LOG.debug ("\n**************** Optimizer (Recompile) *************\nMemory Budget = " + 
 					   OptimizerUtils.toMB(OptimizerUtils.getLocalMemBudget()) + " MB");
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/sysml/hops/recompile/Recompiler.java
```diff
@@ -1220,11 +1220,11 @@ else if (pb instanceof ForProgramBlock) //includes ParFORProgramBlock
 			ForProgramBlock pbTmp = (ForProgramBlock)pb;	
 			ForStatementBlock sbTmp = (ForStatementBlock) pbTmp.getStatementBlock();
 			//recompile predicate
-			if( sbTmp!=null &&!(et==ExecType.CP && !OptTreeConverter.containsMRJobInstruction(pbTmp.getFromInstructions(), true, true)) )			
+			if( sbTmp.getFromHops() != null && sbTmp!=null &&!(et==ExecType.CP && !OptTreeConverter.containsMRJobInstruction(pbTmp.getFromInstructions(), true, true)) )
```
deroneriksson: Should the `sbTmp!=null` condition be moved to the left of `sbTmp.getFromHops()!=null` condition to avoid calling `getFromHops()` on a potentially null `sbTmp`? (same goes for lines 1225 and 1227)

fschueler: Yes, absolutely! Changed it.

deroneriksson:
LGTM. Thank you @fschueler

PR #306, opened 2016-12-05
[SYSTEMML-769] Improved performance of im2col and some bug fixes

niketanpansare:
1. Improved the performance of im2col.
2. Added flags to disable SystemML's caching and also to disable sparse operations. By default, we enable both caching and sparse operations. However, having this flag allows an educated user to disable them (for example: deep learning).
3. Bug fix for single-threaded execution for conv2d_backward_filter.
4. Cleaned up the code for reuse non-zeroed array and invocation of ConvTasks. 
5. Updated developer documentation for LibMatrixDNN.
6. Added sparse test cases for conv2d. Will add similar testcases for other LibMatrixDNN operations  soon.


mboehm7:
I'm very concerned with these additional flags as (1) it is very difficult (even for devs) to judge if cache evictions are too conservative, and (2) completely disabling sparse operations is almost never a good idea. I would strongly recommend to revert this change!

niketanpansare:
@mboehm7  I think we are tying our hands if we don't give the users the freedom to disable sparse and caching (either via flags or some other mechanism). Almost all deep learning frameworks (including Tensorflow, Caffe, Torch and Theano) store their tensors in the dense format and don't support multilevel caching. So, disabling sparse operations may not be such a bad idea if the user wants to run a deep learning algorithm. Due to the characteristics of mini-batch gradient descent, disabling caching might be a good idea when the user is providing large enough memory budget (which is evident from the below experiment).

I ran some preliminary performance experiments comparing conv2d with and without sparse/caching. I provided fairly large memory budget (-Xmx20g -Xms20g -Xmn2048m -server) and used Open JDK 1.8 64-Bit Server VM. The script tested the performance of conv2d using four commonly used setups for 1000 iterations:

```python
max_iterations = 1000
setup = $2
numFilters = -1
numChannels = -1
filterSize = -1
pad = -1
if(setup == 1) {
        numFilters = 20
        numChannels = 1
        filterSize = 5
        pad = 0
}
else if(setup == 2) {
        numFilters = 50
        numChannels = 20
        filterSize = 5
        pad = 0
}
else if(setup == 3) {
        numFilters = 20
        numChannels = 1
        filterSize = 3
        pad = 1
}
else if(setup == 4) {
        numFilters = 50
        numChannels = 20
        filterSize = 3
        pad = 1
}
else {
        stop('Incorrect setup (needs to be [1, 4]).')
}
imgSize = 28
n = 60000
X = rand(rows=n, cols=numChannels*imgSize*imgSize)
batch_size = 64
w = rand(rows=numFilters, cols=numChannels*filterSize*filterSize)
P = (imgSize + 2 * pad - filterSize)  + 1
foo = matrix(0, rows=n, cols=numFilters*P*P)
for(iter in 1:max_iterations) {
        beg = (iter * batch_size) %% n + 1
        end = min(n, beg + batch_size)
        X_batch = X[beg:end, ]
        n_batch = nrow(X_batch)
        convOut_1 = conv2d(X_batch, w, input_shape=[n_batch,numChannels,imgSize,imgSize], filter_shape=[numFilters,numChannels,filterSize,filterSize], padding=[pad,pad], stride=[1,1])
        foo = convOut_1
}
print(sum(foo))
```

Here are the numbers with and without caching/sparse enabled:
``` bash
,Number of Iterations, Setup, Time in seconds
SystemML_CP,1000,1,6.498525426
SystemML_CP_WithCacheNSparseEnabled,1000,1,7.195620854
SystemML_CP,1000,2,81.769984552
SystemML_CP_WithCacheNSparseEnabled,1000,2,101.274622939
SystemML_CP,1000,3,6.355272119
SystemML_CP_WithCacheNSparseEnabled,1000,3,7.607337158
SystemML_CP,1000,4,49.716594505
SystemML_CP_WithCacheNSparseEnabled,1000,4,71.542244484
```

Here are the Cache statistics for the four cases when caching is enabled:
```bash
Cache times (ACQr/m, RLS, EXP): 0.004/0.005/0.918/0.000 sec.
Cache times (ACQr/m, RLS, EXP): 0.007/0.012/15.410/0.000 sec.
Cache times (ACQr/m, RLS, EXP): 0.005/0.005/1.318/0.000 sec.
Cache times (ACQr/m, RLS, EXP): 0.008/0.013/16.795/0.000 sec.
```

If you prefer, I can also run the CNN algorithm with mnist dataset to justify the need for these flags.

mboehm7:
well, I think disabling caching and sparse matrices only hides potential performance issues, which is dangerous because they're ignored during development but end users will ultimately run into them.  I would strongly recommend to first understand what is happening here and if required address the underlying issues. To do that, it would be necessary to separate caching (potential serialization and evictions) and sparse matrices (incl their operations) and investigate these aspects individually.

niketanpansare:
@mboehm7 I agree with the points you made (especially "they're ignored during development but end users will ultimately run into them"). Please review https://github.com/apache/incubator-systemml/pull/308. I have also created a blocker issue https://issues.apache.org/jira/browse/SYSTEMML-1140 which needs to be addressed before we claim deep learning as "non-experimental". I am spread too thin for next few months, so won't be able to look SYSTEMML-1140 right now. Please do look into it when you have time. If the above experimental setup in bit unclear, I can send you the shell script in a separate email.

PR #308, opened 2016-12-06
[SYSTEMML-769] [SYSTEMML-1140] Removed -disable-caching and -disable-sparse flags.

niketanpansare:
@mboehm7 I can merge this PR in if you are OK with it.

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -109,8 +109,19 @@ public class DMLScript
 	public static boolean USE_ACCELERATOR = false;
 	public static boolean FORCE_ACCELERATOR = false;
 	
+	// ------------------------------------------------------------------------
+	// We have identified two performance bugs that frequently occurs in deep learning script.
+	//
+	// First, we repeatedly perform unnecessary conversion to sparse format.
+	// Also, the operations such as matrix multiplication (including BLAS and CuBLAS) are 
+	// optimized for dense.
+	//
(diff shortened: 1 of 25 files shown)


mboehm7:
thanks @niketanpansare - I do appreciate this and yes it could be merged in. Personally, I would further also remove the constants from DMLScript as they would (without further integration) anyway break several places throughout the system that depend on sparsity estimates.

niketanpansare:
Thanks @mboehm7 ... For now, I would prefer to have these flags for performance testing and also for data scientists who want to try deep learning algorithms, until we resolve SYSTEMML-1140. 

Based on your suggestion, I have added the integration logic (as well as test cases) for DISABLE_SPARSE, rather than remove the constants. Instead of replicating the entire test suite, I have only added tests for our applications so that our PR builder completes in time. Please let me know if I am missing any case or any test case which is critical to these flags.

mboehm7:
the change seems to be ok, but remember that any performance experiments with these flags are completely meaningless because our mainstream algorithms would run into OOMs or use too conservative plans with this configuration.

niketanpansare:
Thanks @mboehm7 , I will merge. I differ with you on the utility of the performance experiments with the flags enabled. I think they highlight two critical performance issues that affects a large class of algorithms (i.e. deep learning). Also, the performance experiments show that it is mainly because of those two issues (and native BLAS vs LibMatrixMult) that we differ with other deep learning frameworks. Sure, these issues don't affect our batched mainstream algorithms to that extent, but nonetheless, we need to address them if SystemML aims to optimize general ML.

PR #309, opened 2016-12-06
[SYSTEMML-1141] Enable configurable testTemp location

kevin-bates:
Added the ability to override the default testTemp directory location of "target/testTemp" by setting JVM property `systemml.testTemp.root.dir` when invoking the maven command that produces the test failure.  (e.g., `mvn test -e -Dtest=CSVReadUnknownSizeTest -Dsystemml.testTemp.root.dir=/tmp/foo/testTemp`)

This enables formal build environments (typically larger enterprises) to better protect their assets while allowing tests to complete within less-protected hierarchies.


deroneriksson:
LGTM. Thank you @kevin-bates!

PR #311, opened 2016-12-07
[SYSTEMML-1144] Fix PCA documentation for principal

deroneriksson:
Update 'principle' to 'principal'.


gweidner:
Thanks for catching this!  LGTM

deroneriksson:
Thanks for reviewing Glenn! I updated the docs but David P. Nichols was the statistician who noticed this and alerted us to the typos.

PR #314, opened 2016-12-09
[SYSTEMML-1148] Explicitly set value types for MLContext scalar inputs

deroneriksson:
Not explicitly setting the scalar value type of a basic input type
through the MLContext API can cause issues since the value type can
default to double. For example, "print(!b)" where b is a boolean can
throw an error since the NOT operation can be performed on a boolean
but not a double.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -50,6 +50,7 @@ import org.apache.sysml.conf.CompilerConfig.ConfigType;
 import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.conf.DMLConfig;
 import org.apache.sysml.parser.ParseException;
+import org.apache.sysml.parser.Statement;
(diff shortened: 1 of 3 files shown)


deroneriksson:
Sure, sounds good.

deroneriksson:
Ran tests with:
```
PYSPARK_PYTHON=python3 spark-submit --master local[*] --driver-class-path "target/classes:target/lib/antlr4-runtime-4.5.3.jar:target/lib/wink-json4j-1.4.jar" test_mlcontext.py
```

Looks like python mlcontext tests pass:
```
----------------------------------------------------------------------
Ran 9 tests in 4.135s

OK
```

Regular test suite also passes:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/913/

Including new MLContextTest test cases:
```
15:58:00 Tests run: 140, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 70.283 sec - in org.apache.sysml.test.integration.mlcontext.MLContextTest
```

deroneriksson:
Thanks for reviewing @dusenberrymw!

PR #315, opened 2016-12-09
[SYSTEMML-1149] Remove unused constructors from project

deroneriksson:
Unused constructors in the project are essentially dead code and make project maintenance more difficult. They should be removed unless there is a compelling reason to keep them.


mboehm7:
please be super careful about removing the copy constructors from the package org.apache.sysml.runtime as they are frequently used for debugging.

deroneriksson:
Thank you for reviewing @mboehm7 and @dusenberrymw! I'll restore the copy constructors and merge after that.

PR #320, opened 2016-12-19
[SYSTEMML-769] [SYSTEMML-1140] Disable sparse for Deep Learning related script

niketanpansare:
@bertholdreinwald @mboehm7 @dusenberrymw @prithvirajsen This will disable sparse for convolutional neural network. The autoencoder will still have sparse enabled.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/AParserWrapper.java
+++ b/src/main/java/org/apache/sysml/parser/AParserWrapper.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import org.apache.commons.logging.Log;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.parser.common.CommonSyntacticValidator;
 import org.apache.sysml.parser.common.CustomErrorListener.ParseIssue;
@@ -59,6 +60,9 @@ public abstract class AParserWrapper
 	public static AParserWrapper createParser(boolean pydml)
 	{
 		AParserWrapper ret = null;
+		
+		// Always enable sparse before parsing phase (works for all the APIs).
+		DMLScript.DISABLE_SPARSE = false;
 		
 		//create the parser instance
 		if( pydml )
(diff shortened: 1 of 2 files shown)


dusenberrymw:
Yeah I completely agree that dense makes more sense particularly for the mini-batch cases, and especially for those involving GPU.  I'm just curious if the sparse format could still be useful for full-batch, distributed, non-GPU prediction cases.  Once we have a distributed implementation of convolution and max pooling it will be an interesting experiment. :)

In the meantime, I'm not opposed to going ahead and making use of this.  As long as we keep the change simple, we can always come back and reevaluate it for the case above.

niketanpansare:
Cool. Let's wait to hear back from @mboehm7  and @bertholdreinwald 

Regarding distributed implementation of convolution and max pooling, there are following options:
1. Current approach (Mini-batch training): for loop { CP conv + CP max_pool }
2. Batched approach (applicable for batched training as well as batched prediction):  for loop { SPARK conv + SPARK max_pool }
3. Parfor Prediction: remote SPARK parfor loop { CP conv + CP max_pool }

@dusenberrymw  Which case do you think we should address first ?

mboehm7:
As I already commented on a previous PR, I'm completely against these flags to disable sparse matrices and caching. From my perspective, it only works around underlying issues that should be addressed at runtime level.

For example, for unary operations like relu, we could scan the input twice in order to determine the exact nnz of the output, and thus the correct sparse/dense format, before allocation. Having the nnz per row would also replace the log M allocations per row with a single allocation. Due to copy-on-write the sparse intermediate would then anyway not be updated, except for in-place indexing. Note that we do similar things for sparse matrix transpose as well.

Furthermore, there is also a technical issue. Having the static flag would affect all operations in a session; coming through MLContext, even across scripts.

niketanpansare:
@mboehm7  I thought you had an issue with externalizing the decision of disabling sparse. In this PR, I was disabling sparse (not caching) only when we observe a conv2d-related script. My concern is that sparsity threshold of 0.4 is way too optimistic for allocating vectors (for example: mini-batch deep learning script). I can try other approaches to reduce the impact of this threshold (for example: https://github.com/apache/incubator-systemml/pull/329), but eventually we will have to systematically address whether the sparsity threshold of 0.4 is appropriate choice for all our algorithms (including lenet, autoencoder, lstm, etc). Until then, I can close this PR as we do not have consensus.

>> Having the static flag would affect all operations in a session; coming through MLContext, even across scripts.
That's a good point. I was not aware we decided to make MLContext multi-threaded.

PR #322, opened 2016-12-20
[SYSTEMML-1161][WIP] recursive qr factorization

iyounus:
This implements QR factorization using the Block Recursive algorithm create by Elmroth and Gustavson. A simple representation of the algorithm is presented by Golub (Algorithm 5.2.3).

Reference:
"Applying recursion to serial and parallel QR factorization leads to better performance.", E. Elmroth and F Gustavson, IBM J. RES. DEVELOP. VOL 44 JULY 2000
"Matrix Computations", Golub and Van Loan, 4th Edition, Chapter 5.

jira:
https://issues.apache.org/jira/browse/SYSTEMML-1161

### DIFF
--- a/scripts/staging/QR_recursive.dml
+++ b/scripts/staging/QR_recursive.dml
@@ -24,7 +24,7 @@
 # NAME   TYPE   DEFAULT  MEANING
 # -----------------------------------------------------------------------------
 # X      matrix  ---     input matrix to be factorized
-# q      Int     1000    smalles column block of input matrix
+# k      Int     1000    smalles column block of input matrix
 # OUTDIR String  ---     output directory to save the results
 # OFMT   String  CSV     output format
 # -----------------------------------------------------------------------------
@@ -38,7 +38,7 @@
 # -----------------------------------------------------------------------------
 
 X = read($X);
-q = ifdef($q, 1000);
+k = ifdef($q, 1000);
 # k is used for the base case in the recursive algorithm.
 # If X has very large number of rows, choose k to be smaller
 # to fit the base case on single machine.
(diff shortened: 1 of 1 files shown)


iyounus:
This algorithm is supposed to work fine. But there is one weird issue. In line 54, I am defining `Q = eye[,1:n]`. So it is supposed to be a `m x n` matrix (the same size as the `H` matrix). In the first iteration it is `m x n`, but in the successive iterations, `Q` becomes `m x m`. And because of this the algorithm breaks down. This is definitely a bug. Any suggestions/fixes/workarounds? @nakul02 @niketanpansare @mboehm7

nakul02:
@iyounus - are you running this in spark mode or standalone?
I can get it run on standalone. I needed to change how "k" is accessed in the function "QR_recursive". I needed to pass it in as a parameter.

[Here](https://gist.github.com/nakul02/651a139d53b929d08355fb447c489caf) is the script I ran.

iyounus:
I'm running it in standalone, no spark at this time.
There was a mistake in the algorithm. I didn't have this problem in my test code. I've fixed the script. Here is how you can reproduce the problem. Try running the script like this (in dml):
```
X = rand(rows=15, cols=10)
print(toString(X))
[Q, R] = QR_recursive(X, 3)

A = Q %*% R
print(toString(A))
```
`X` and `A` should be identical.
If the number of columns in your input matrix is less than k (now I called it q in new script), then there is no recursion and you go directly to base case. In that case there is not problem.

nakul02:
Ok, so with this I am able to reproduce the error.
There is something wrong with the optimizer. When I set optimization level to 0, the problem doesn't occur.
To set the optimization level to 0, set the appropriate property in the `SystemML-config.xml` file.
If you are running systemml from the `bin/systemml` script, just change the `conf/SystemML-config.xml` file.

nakul02:
@iyounus, you might want to open a JIRA about this.

deroneriksson:
@iyounus @nakul02 @dusenberrymw 
Any chance we can go ahead and merge this to staging? If there are optimizer issues, perhaps they will be more likely to be dealt with if the algorithm has been merged to the staging directory.

iyounus:
I think this can be merged to staging. Initially there didn't allow this algorithm to work but that has been fixed. It still needs some testing but I don't think any major changes are required here.

deroneriksson:
Great! Thank you @iyounus. I'll merge to staging.

PR #323, opened 2016-12-21
[HOTIFX] Fixes for RAT checks which fail as a result of SYSTEMML-1038

nakul02:


deroneriksson:
LGTM. Thank you @nakul02

PR #326, opened 2017-01-05
[SYSTEMML-1039] Added uark+/uar+

nakul02:
With this PR, the minimum CUDA compute capability required is 3.0.

@niketanpansare - Could you please review?


mboehm7:
@nakul02 just a side comment on naming conventions: the opcode uark+ stands for unary aggregate row kahan plus. Could you please either rename this operation to uar+ or (consistently with the other backends) implement the kahan summation?

nakul02:
@mboehm7 - Thanks, I've updated rerouting calls of `uark+`, `uar+` to the GPU summation reduction implementation (when applicable).
For the time being, we are redirecting kahan sum reductions to regular summation reduction implementations on the GPU. This is documented [here](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java#L928).

PR #327, opened 2017-01-05
[SYSTEMML-1164] Fix IterablePredicate.toString NPE for incrementExpr

deroneriksson:
Add a null check for _incrementExpr to IterablePredicate's toString method,
since otherwise a NullPointerException can occur, such as when DMLProgram's
toString method is called, which can call IterablePredicate's toString.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/IterablePredicate.java
+++ b/src/main/java/org/apache/sysml/parser/IterablePredicate.java
@@ -56,8 +56,10 @@ public class IterablePredicate extends Expression
 		sb.append(_fromExpr.toString());
 		sb.append(",");
 		sb.append(_toExpr.toString());
-		sb.append(",");
-		sb.append(_incrementExpr.toString());
+		if (_incrementExpr != null) {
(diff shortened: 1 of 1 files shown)


deroneriksson:
This can be tested with:
```
val ml = new MLContext(sc)
val habermanUrl = "http://archive.ics.uci.edu/ml/machine-learning-databases/haberman/haberman.data"
val typesRDD = sc.parallelize(Array("1.0,1.0,1.0,2.0"))
val scriptUrl = "https://raw.githubusercontent.com/apache/incubator-systemml/master/scripts/algorithms/Univar-Stats.dml"
val uni = dmlFromUrl(scriptUrl).in("A", new java.net.URL(habermanUrl)).in("K", typesRDD).in("$CONSOLE_OUTPUT", true)
ml.execute(uni)
val se = uni.getScriptExecutor
val program = se.getDmlProgram
```

deroneriksson:
@dusenberrymw Thank you for reviewing!

PR #328, opened 2017-01-05
[SYSTEMML-540] Support bias_add operation as well as faster sgd nesterov updates

niketanpansare:
@dusenberrymw Can you please review this PR ?


niketanpansare:
That makes sense. I will look into runtime fused operators for such patterns.

niketanpansare:
Results of experiments: The improvement with sgd_nesterov updates was only 1.07x (less that what I had expected). The average time per call for update_nesterov is comparable to that of +*. 

For now, I will keep the udf, but revert back sgd_nesterov to previous implementation and explore other improvements.

PR #329, opened 2017-01-05
[SYSTEMML-540] Added fused relu_maxpooling

niketanpansare:
- Fused relu_maxpooling reduces the unnecessary dense-to-sparse-to-dense conversion.
- Note: fused relu_maxpooling is only supported in CP, not on GPU as both relu and maxpooling invoke CuDNN functions.
- Also, improve the performance of maxpooling by computing indexes apriori.

@mboehm7 @bertholdreinwald @dusenberrymw

### DIFF
--- a/src/main/java/org/apache/sysml/hops/ConvolutionOp.java
+++ b/src/main/java/org/apache/sysml/hops/ConvolutionOp.java
@@ -25,6 +25,7 @@ import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.hops.Hop.MultiThreadedHop;
 import org.apache.sysml.lops.ConvolutionTransform;
+import org.apache.sysml.lops.ConvolutionTransform.OperationTypes;
 import org.apache.sysml.lops.Lop;
 import org.apache.sysml.lops.LopsException;
 import org.apache.sysml.lops.LopProperties.ExecType;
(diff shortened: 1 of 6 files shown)


niketanpansare:
"relu-maxpooling fusion" especially makes sense becomes relu becomes a no-op. We can definitely consider "conv-relu-conv" pattern and "relu-dropout" pattern. It is important to note that we **support sparse operators** for most of these cases including sparse matrix multiplication. However, there is a tradeoff in the runtime performance of sparse operators v/s memory consumption (i.e. cusparse matrix multiplication with sparsity 0.4 will be slower than cublas matrix multiplication, assuming no conversion). Based on previous machine learning workloads, we decided to keep threshold of conversion 0.4, and that threshold is respected throughout the system (i.e. via examine sparsity). I see only three possible solutions that respect this design:
1. Operator fusion (where temporary data is not converted to sparse). For example: this PR
2. Reduce the threshold globally. For example: https://github.com/apache/incubator-systemml/pull/320
3. Code generation.

dusenberrymw:
I think that given the newer & expanded range of ML workloads that are now being run on SystemML, we should go back through the system and challenge previous assumptions that were made, such as the sparsity threshold.  We've already started this by breaking previous assumptions of tall-skinny matrices, for example.  A big issue we're seeing now is models that contain several computational transformation ("layers") rather than just a couple.

Let's pause on (1) & (3) for a bit, and focus on (2) and a more general approach of reevaluating previous assumptions in the system.  Sparse support, caching, and native BLAS come to mind immediately.

dusenberrymw:
[SYSTEMML-1140](https://issues.apache.org/jira/browse/SYSTEMML-1140) would be a good place to start.

PR #336, opened 2017-01-09
[SYSTEMML-1171] Clean Up Python Distribution

dusenberrymw:
This cleans up the Python distribution package by removing the `scripts` folder
that is included, as it is not used by internal code, will be confusing
to users, and will cause confusion during release reviews.

Also includes some minor code formatting improvements.

### DIFF
--- a/src/main/python/MANIFEST.in
+++ b/src/main/python/MANIFEST.in
@@ -21,8 +21,4 @@
 include LICENSE
 include NOTICE
 include DISCLAIMER
-include systemml/systemml-java/scripts/sparkDML.sh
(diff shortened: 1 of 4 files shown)


dusenberrymw:
cc @deroneriksson, @niketanpansare Minor update that should reduce confusion during the upcoming release review.

deroneriksson:
LGTM. Thanks @dusenberrymw! The scripts directory is essentially redundant so removing it makes sense to me.

dusenberrymw:
@deroneriksson Thanks!

PR #337, opened 2017-01-10
[SYSTEMML-540] Improved the performance of bias_add and added CP relu_backward

niketanpansare:


niketanpansare:
The [tests](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1006/) passed. Since the performance improvement after relu_backward is significant, merging this PR.

PR #338, opened 2017-01-10
[HOTFIX] for bias_add

nakul02:
Fixed gpu bias_add function for GPU. Metadata was being set incorrectly.

@niketanpansare


niketanpansare:
Thanks @nakul02 ... nice catch. I will merge :)

PR #339, opened 2017-01-10
[SYSTEMML-540] Improved dropout performance

niketanpansare:
The modified script avoids unnecessary dense intermediate and is especially useful in case of low p value

@dusenberrymw can you please review ? We can either fix the rand bug as part of this PR or in separate PR.

### DIFF
--- a/scripts/staging/SystemML-NN/nn/layers/dropout.dml
+++ b/scripts/staging/SystemML-NN/nn/layers/dropout.dml
@@ -49,7 +49,6 @@ forward = function(matrix[double] X, double p, int seed)
   else {
     mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=0.5, seed=seed)
   }
-  mask = rand(rows=nrow(X), cols=ncol(X), min=0, max=1, seed=seed) <= p
   out = X * mask / p
 }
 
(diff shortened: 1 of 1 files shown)


niketanpansare:
Updated the PR to support generic p values. This allows us to continue making progress in optimizing MNIST demo and when https://issues.apache.org/jira/browse/SYSTEMML-515 is resolved, we can update dropout too.

niketanpansare:
Forgot to handle an if case, will update the PR soon :)

dusenberrymw:
@niketanpansare I'm still not a fan of this update.  I think it's great to explore areas of improvement, but optimizing only for single, specific usages (i.e. only for `p=0.5`) is why SystemML is in its current, fragile state.  I.e. `dropout` needs to have the same level of performance for all levels of `p`,  rather than exhibit great performance properties *only* for specific values of `p`.  For more concrete evidence, that demo will involve a hyperparameter search over `p`, rather than always selecting `p=0.5`.

-1 for the temporary patch.  Let's just fix `rand` and fix the issue entirely.

niketanpansare:
Ok, I will close this PR for now. Once https://issues.apache.org/jira/browse/SYSTEMML-515 is resolved, please use `mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=p)` rather than `mask = rand(rows=nrow(X), cols=ncol(X), min=0, max=1) <= p` as the former has much better performance.

PR #345, opened 2017-01-15
[SYSTEMML-1174] Improve OutputStatement toString

deroneriksson:
Fix case, spacing, and parameter issues.


deroneriksson:
Example OutputStatement (using univariate stats):

Before PR:
```
write ( baseStats,   );
```

After PR:
```
WRITE(id=baseStats, iofilename=" ");
```

PR #349, opened 2017-01-18
Allow Scala UDF to be passed to SystemML via external UDF mechanism

niketanpansare:
The registration mechanism is inspired from Spark SQLContext's UDF. The key construct is `ml.udf.register("fn to be used in DML", scala UDF)`.

```scala
val ml = new MLContext(sc)

// Demonstrates how to pass a simple scala UDF to SystemML
def addOne(x:Double):Double = x + 1
ml.udf.register("addOne", addOne)
val script1 = dml("v = addOne(2.0); print(v)")
ml.execute(script1)

// Demonstrates operation on local matrices (double[][])
def addOneToDiagonal(x:Array[Array[Double]]):Array[Array[Double]] = {  for(i <- 0 to x.length-1) x(i)(i) = x(i)(i) + 1; x }
ml.udf.register("addOneToDiagonal", addOneToDiagonal)
val script2 = dml("m1 = matrix(0, rows=3, cols=3); m2 = addOneToDiagonal(m1); print(toString(m2));")
ml.execute(script2)


// Demonstrates multi-return function
def multiReturnFn(x:Double):(Double, Int) = (x + 1, (x * 2).toInt)
ml.udf.register("multiReturnFn", multiReturnFn)
val script3 = dml("[v1, v2] = multiReturnFn(2.0); print(v1)")
ml.execute(script3)

// Demonstrates multi-argument multi-return function
def multiArgReturnFn(x:Double, y:Int):(Double, Int) = (x + 1, (x * y).toInt)
ml.udf.register("multiArgReturnFn", multiArgReturnFn _)
val script4 = dml("[v1, v2] = multiArgReturnFn(2.0, 1); print(v2)")
ml.execute(script4)
```

The restrictions for Scala UDF are as follows:
1. Only types specified by DML language is supported for parameters and return types (i.e. Int, Double, Boolean, String, double[][]).
2. At minimum, the function should have 1 argument and 1 return value.
3. At max, the function can have 10 arguments and 10 return values.

@dusenberrymw @deroneriksson @bertholdreinwald @nakul02 @mboehm7 @fschueler @frreiss

### DIFF
--- a/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
+++ b/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 package org.apache.sysml.udf.lib;
 
 import java.io.IOException;
--- a/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
+++ b/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
@@ -1,31 +1,42 @@
 package org.apache.sysml.udf.lib;
 
 import java.io.IOException;
-import java.util.ArrayList;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.sysml.api.ExternalUDFRegistration;
 import org.apache.sysml.runtime.DMLRuntimeException;
 import org.apache.sysml.udf.FunctionParameter;
 import org.apache.sysml.udf.Matrix;
 import org.apache.sysml.udf.PackageFunction;
 import org.apache.sysml.udf.Scalar;
+
 import scala.Function0;
 
 public class GenericFunction extends PackageFunction {
 	private static final long serialVersionUID = -195996547505886575L;
 	String [] fnSignature;
 	FunctionParameter [] returnVals;
-	Function0<FunctionParameter []> scalaUDF; 
+	Function0<FunctionParameter []> scalaUDF;
+	public String _functionName;
+	public String _namespace;
 	
-	public GenericFunction(String [] fnSignature) {
-		this.fnSignature = fnSignature;
-	}
-	public void setScalaUDF(Function0<FunctionParameter []> scalaUDF) {
-		this.scalaUDF = scalaUDF;
+	public void initialize() {
+		if(_namespace != null && !_namespace.equals(".defaultNS")) {
+			throw new RuntimeException("Expected the function in default namespace");
+		}
+		if(_functionName == null) {
+			throw new RuntimeException("Expected the function name to be set");
+		}
+		if(fnSignature == null) {
+			fnSignature = ExternalUDFRegistration.fnSignatureMapping().get(_functionName);
+			scalaUDF = ExternalUDFRegistration.fnMapping().get(_functionName);
+			ExternalUDFRegistration.udfMapping().put(_functionName, this);
+		}
 	}
 	
 	@Override
 	public int getNumFunctionOutputs() {
+		initialize();
 		String retSignature = fnSignature[fnSignature.length -1];
 		if(!retSignature.startsWith("("))
 			return 1;
@@ -36,6 +47,7 @@ public class GenericFunction extends PackageFunction {
 
 	@Override
 	public FunctionParameter getFunctionOutput(int pos) {
+		initialize();
 		if(returnVals == null || returnVals.length <= pos)
 			throw new RuntimeException("Incorrect number of outputs or function not executed");
 		return returnVals[pos];
@@ -43,21 +55,9 @@ public class GenericFunction extends PackageFunction {
 
 	@Override
 	public void execute() {
+		initialize();
 		returnVals = scalaUDF.apply();
 	}
-	
-//	public Object constructOutput(Object ret) throws DMLRuntimeException, IOException {
-//		if(ret instanceof Integer)
-//			return new Scalar(ScalarValueType.Integer, String.valueOf(ret));
-//		else if(ret instanceof Double)
-//			return new Scalar(ScalarValueType.Double, String.valueOf(ret));
-//		else if(ret instanceof String)
-//			return new Scalar(ScalarValueType.Text, String.valueOf(ret));
-//		else if(ret instanceof Boolean)
-//			return new Scalar(ScalarValueType.Boolean, String.valueOf(ret));
-//		else if(ret instanceof Boolean)
-//			
-//	}
 	
 	public Object getInput(String type, int pos) throws DMLRuntimeException, IOException {
 		if(type.equals("Int") || type.equals("java.lang.Integer")) {
--- a/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
+++ b/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.sysml.api.ExternalUDFRegistration;
+import org.apache.sysml.parser.DMLProgram;
 import org.apache.sysml.runtime.DMLRuntimeException;
 import org.apache.sysml.udf.FunctionParameter;
 import org.apache.sysml.udf.Matrix;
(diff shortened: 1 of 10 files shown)


niketanpansare:
@fschueler Please see the design documentation for this feature: https://github.com/niketanpansare/incubator-systemml/blob/00fab86aa6a7577b6efe9aa0f0ad249b10e1b1c9/src/main/scala/org/apache/sysml/api/ExternalUDFRegistration.scala#L36

fschueler:
I like the idea of this function registry but I am a little bit concerned with the current approach. We might want to coordinate the further development of any Scala parts so that it fits nicely into a whole concept and we don't end up with three half-finished prototypes that have all incompatible types. 
That includes defining a `Matrix` type that is used wherever possible in Scala (in the mlcontext, macro DSL and as a return type of any `MLContext.get(...)` calls). This could then also be used instead of Array[Double].

One way to do this is to do the definition of functions outside the macro and generate the dml external function stubs inside. To do this, we might need to add macro annotations - but I was planning on this anyways.

For now the registration on the mlcontext might work. In the macro version, the defined scala function could actually be called directly and so we would at least get the typechecking in Scala and not in DML (I think that's possible).

I think we should outline a general roadmap for our Scala interface and make sure that we remove everything else that is not really supported. I am in the process of preparing a PR for the macro DSL and once that is ready, we could think about how to integrate it with all the other things that are currently there in Scala (especially unifying the Matrix types).

niketanpansare:
Thanks for your comments @fschueler 

>> We might want to coordinate the further development of any Scala parts so that it fits nicely into a whole concept and we don't end up with three half-finished prototypes that have all incompatible types.
That includes defining a Matrix type that is used wherever possible in Scala (in the mlcontext, macro DSL and as a return type of any MLContext.get(...) calls). This could then also be used instead of Array[Double].

Infact, I delibrately decided not to add any additional types and kept the focus of the PR very limited: "Allow external builtin function invocation via scala udf". Since double[][] is the default type of our [external function framework](https://github.com/niketanpansare/incubator-systemml/tree/00fab86aa6a7577b6efe9aa0f0ad249b10e1b1c9/src/main/java/org/apache/sysml/udf), I decided to stick with it. When we integrate Matrix type in our external function framework, then you can support it with only two lines of change: [Line 1](https://github.com/niketanpansare/incubator-systemml/blob/00fab86aa6a7577b6efe9aa0f0ad249b10e1b1c9/src/main/scala/org/apache/sysml/api/ExternalUDFRegistration.scala#L459) and [Line 2](https://github.com/niketanpansare/incubator-systemml/blob/00fab86aa6a7577b6efe9aa0f0ad249b10e1b1c9/src/main/scala/org/apache/sysml/api/ExternalUDFRegistration.scala#L125). 

I don't fully understand the macro implementation. But if you feel that is more suitable approach, I am OK with closing this PR in favor of your approach :)

Also, +1 for unifying the various Matrix representation in SystemML.

fschueler:
I'm not 100% sure at this time what the best approach would be. 

For the future work on the Scala part we should probably think about drafting some design doc or talk about a general direction and features. Another side of this is always consistency with potential Python APIs/DSLs.

I think your approach is very non-invasive and if you need that feature now it can probably be merged.

niketanpansare:
Since this is a user facing API, I am OK with keeping this PR open until we draft the design document :) You might want to take this opportunity and extend the external functions framework to add Spark related data structures (RDD[(MI, MB)], DataFrame, or the unified Matrix, ...) in your design doc.

niketanpansare:
@fschueler As per our offline discussion, I guess you are OK with merging this PR in.
@deroneriksson Since this affects MLContext, can you please give this PR a LGTM before I merge it in ?

deroneriksson:
@niketanpansare 
I do see an issue here that might require some further discussion. I believe this is the first time that Java classes in the project have referenced a Scala-based class (org.apache.sysml.api.ExternalUDFRegistration). This could potentially cause some issues for Eclipse users (in the MLContext API, I wrote everything in Java rather than Java and Scala to avoid confronting this issue). For example, I use the maven-eclipse-plugin to manage my classpath in Eclipse. If I run "eclipse:eclipse", my classpath breaks here because this plugin doesn't generate the Scala-based classes in the target directory. I can get around this problem by creating a folder and placing the Scala classes in the folder and referencing the folder on the classpath, although this is a bit of a hack. Do you have any advice for Eclipse users how to cleanly manage the classpath here? I ran into issues the last time I tried the Eclipse Scala IDE.

@mboehm7 @frreiss @bertholdreinwald @gweidner

niketanpansare:
@deroneriksson Good point. Though SystemML have both scala and java files, we somehow avoided this issue by never invoking scala classes from java; thereby requiring only scala developers to modify the IDE setup. May be, we ought to revisit this restriction as the project evolves given that our primary backend (i.e. Spark) is written in Scala. For time being, if this is a blocking issue, we can still support UDF functionality by decoupling UDF and MLContext with a different API (not so elegant as the current one):
  
```scala
import org.apache.sysml.api.mlcontext.UDFRegistry
val ml = new MLContext(sc)
val udfRegistry = new UDFRegistry(ml)

// Demonstrates how to pass a simple scala UDF to SystemML
def addOne(x:Double):Double = x + 1
udfRegistry.register("addOne", addOne)
val script1 = dml("v = addOne(2.0); print(v)")
ml.execute(script1)
```

My setup:
I am using [Scala IDE](http://scala-ide.org/) and it works out of the box (If not, all you have to do is Right click on the project -> Configure -> Add Scala Nature. This will add the Scala nature declaration to your .project file: <nature>org.scala-ide.sdt.core.scalanature</nature>).

deroneriksson:
@niketanpansare I agree that this move towards Scala support is really a necessary step in the evolution of the project since we run on Spark, and it is a step that we need to take. I just wanted other Eclipse developers to be aware that this PR represents taking that step.

Also, thank you for the Scala IDE setup information. It is good to hear that the integration is clean.

mboehm7:
well, just to clarify: we invoke, and inherit from, scala classes in quite a number of places already. If I remember correctly, the problem is the automatic build of our own scala code under eclipse.

deroneriksson:
@mboehm7 Yes, that is correct. I should have said calling Scala classes within the project from Java classes within the project. We do extensively use Scala already. Also, Scala integration in the maven builds is handled perfectly.

src/main/java/org/apache/sysml/udf/lib/GenericFunction.java
```diff
@@ -0,0 +1,101 @@
...
+public class GenericFunction extends PackageFunction {
+	private static final long serialVersionUID = -195996547505886575L;
+	String [] fnSignature;
+	FunctionParameter [] returnVals;
+	Function0<FunctionParameter []> scalaUDF;
+	public String _functionName;
+	public String _namespace;
+	
+	public void initialize() {
+		if(_namespace != null && !_namespace.equals(".defaultNS")) {
```
deroneriksson: @niketanpansare Could you substitute DMLProgram.DEFAULT_NAMESPACE for the String constant here?

deroneriksson:
@niketanpansare LGTM. Thank you.

mboehm7:
Why does this PR introduce an additional mechanism for UDF meta data directories? I would strongly recommend to remove ExternalUDFRegistration and rather fix the integration with the existing directory in DMLProgram/Program. There are a number of places (IPA, parfor, explain, etc) that explicitly have to deal with UDFs; if not integrated, we would have to systematically test and potentially modify them. Furthermore, it would also solve the issue that these Scala UDFs currently cannot be registered on script level nor used in non-default namespaces.

niketanpansare:
@mboehm7  I don't fully understand your concern. ExternalUDFRegistration is only an API level class (similar to MLContext) that simplifies usage of our external built-in function. All it does it constructs the [external UDF headers](https://github.com/apache/incubator-systemml/commit/5d84a095aadc1d6f1e2e4c06439589591ca24aec#diff-aa48616e7f61866cff633a265f15b48cR457) and wraps the Scala UDF into [GenericFunction which extends PackageFunction](https://github.com/apache/incubator-systemml/blob/5d84a095aadc1d6f1e2e4c06439589591ca24aec/src/main/java/org/apache/sysml/udf/lib/GenericFunction.java). After parsing phase, all the existing UDF mechanism (IPA, parfor, explain, etc) remains unchanged.

mboehm7:
1) Since you call ExternalUDFRegistration from within this GenericFunction, these Scala UDFs cannot be used on script level as they're never registered.
2) Right now, we assume external functions to be in the class path, in a parfor context even on all workers (which is the case for functions that we package into our jar). However, this will obviously fail for user-provided functions. Here you would have to serialize the UDF, distribute it via task closures, and deserialize them at the executors as Spark's task closures refer to classes by name. Extending these kinds of features now would get really ugly because we would have to handle two different directories for functions.

niketanpansare:
>> Since you call ExternalUDFRegistration from within this GenericFunction, these Scala UDFs cannot be used on script level as they're never registered.

I do register "GenericFunction" [here](https://github.com/apache/incubator-systemml/blob/master/src/main/scala/org/apache/sysml/api/ExternalUDFRegistration.scala#L152) which calls ExternalUDFRegistration, else the above scala example will not work.

>> Right now, we assume external functions to be in the class path, in a parfor context even on all workers (which is the case for functions that we package into our jar). However, this will obviously fail for user-provided functions. Here you would have to serialize the UDF, distribute it via task closures, and deserialize them at the executors as Spark's task closures refer to classes by name. Extending these kinds of features now would get really ugly because we would have to handle two different directories for functions.

Since the closure is already created and put into the HashMap, wouldn't it be as simple as performing a broadcast of the [HashMaps](https://github.com/apache/incubator-systemml/blob/master/src/main/scala/org/apache/sysml/api/ExternalUDFRegistration.scala#L94) and setting a static variable [here](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/runtime/controlprogram/parfor/RemoteParForSpark.java#L72). If this works for you, I can add it to the parfor. If not, I can remove the UDF registration code.

mboehm7:
1) What I'm saying is that a proper integration would allow the registration of Scala UDFs on script level without the need to go through MLContext.
2) As said before, no, this will not work if the class of your UDF is unknown at the executors - in that case you need to serialize the class as well; the point is, it's beneficial to have a single directory for UDF meta data.

niketanpansare:
Ok. I will cleanup the UDF invocation in the next commit :+1:

PR #351, opened 2017-01-19
[SYSTEMML-540] [SYSTEMML-515] Allow an expression for sparsity

niketanpansare:
- This PR also improves the performance of dropout

### DIFF
--- a/scripts/staging/SystemML-NN/nn/layers/dropout.dml
+++ b/scripts/staging/SystemML-NN/nn/layers/dropout.dml
@@ -42,7 +42,16 @@ forward = function(matrix[double] X, double p, int seed)
    *  - out: Ouptuts, of same shape as X.
    *  - mask: Dropout mask used to compute the output.
    */
-  mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=p, seed=seed)
+  # Normally, we might use something like
+  #    `mask = rand(rows=nrow(X), cols=ncol(X), min=0, max=1, seed=seed) <= p`
+  # to create a dropout mask.  Fortunately, SystemML has a `sparsity` parameter on
+  # the `rand` function that allows use to create a mask directly.
+  if (seed == -1) {
+  	mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=p)
+  }
+  else {
+  	mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=p, seed=seed)
+  }
   out = X * mask / p
 }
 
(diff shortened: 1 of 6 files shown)


mboehm7:
Please also fix it for MR - you can simply use instruction patching as done for rows and cols.

scripts/staging/SystemML-NN/nn/layers/dropout.dml
```diff
@@ -42,10 +42,7 @@ forward = function(matrix[double] X, double p, int seed)
    *  - out: Ouptuts, of same shape as X.
    *  - mask: Dropout mask used to compute the output.
    */
-  if (seed == -1) {
-    seed = as.integer(floor(as.scalar(rand(rows=1, cols=1, min=1, max=100000))))
-  }
```
dusenberrymw: These three lines need to be added back so that a user can indicate that the dropout layer should use a random seed.

scripts/staging/SystemML-NN/nn/layers/dropout.dml
```diff
@@ -42,10 +42,7 @@ forward = function(matrix[double] X, double p, int seed)
...
-  mask = rand(rows=nrow(X), cols=ncol(X), min=0, max=1, seed=seed) <= p
```
dusenberrymw: Can we add a comment here explaining how we are creating the mask with the `sparsity` flag, since that is not a common construct?  I.e. add the following comment:

```
# Normally, we might use something like
#    `mask = rand(rows=nrow(X), cols=ncol(X), min=0, max=1, seed=seed) <= p`
# to create a dropout mask.  Fortunately, SystemML has a `sparsity` parameter on
# the `rand` function that allows use to create a mask directly.
mask = rand(rows=nrow(X), cols=ncol(X), min=1, max=1, sparsity=p, seed=seed)
```

src/main/java/org/apache/sysml/parser/DataExpression.java
```diff
@@ -1163,9 +1165,10 @@ else if (dataParam instanceof StringIdentifier) {
 			}
 			
 			//parameters w/o support for variable inputs (requires double/int or string constants)
-			if (!(getVarParam(RAND_SPARSITY) instanceof DoubleIdentifier || getVarParam(RAND_SPARSITY) instanceof IntIdentifier)) {
-				raiseValidateError("for Rand statement " + RAND_SPARSITY + " has incorrect value type", conditional);
-			}
+//			if(!(getVarParam(RAND_SPARSITY) instanceof DoubleIdentifier || getVarParam(RAND_SPARSITY) instanceof IntIdentifier
+//					|| getVarParam(RAND_SPARSITY) instanceof DataIdentifier)) {
+//				raiseValidateError("for Rand statement " + RAND_SPARSITY + " has incorrect value type", conditional);
+//			}
```
dusenberrymw: Could we just remove this code?

niketanpansare:
@mboehm7 @dusenberrymw Incorporated your comments, please review it.

@dusenberrymw Instead of generating random seed in DML, we can let rand chose a random seed instead by not providing it an argument :)

niketanpansare:
The [integration tests](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1030/) passed as well.

PR #352, opened 2017-01-23
[SYSTEMML-1192] Delete rmvar instructions via MLContext

deroneriksson:
Remove all rmvar instructions so that variables are maintained in the symbol
table, which is useful in an interactive environment such as the Spark Shell.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
@@ -116,6 +116,12 @@ public class MLContext {
 	 */
 	private ProjectInfo projectInfo = null;
 
+	/**
+	 * Whether or not all values should be maintained in the symbol table
+	 * after execution.
+	 */
+	private boolean maintainSymbolTable = false;
+
 	private List<String> scriptHistoryStrings = new ArrayList<String>();
 	private Map<String, Script> scripts = new LinkedHashMap<String, Script>();
 
@@ -273,6 +279,7 @@ public class MLContext {
 		scriptExecutor.setStatistics(statistics);
 		scriptExecutor.setStatisticsMaxHeavyHitters(statisticsMaxHeavyHitters);
 		scriptExecutor.setInit(scriptHistoryStrings.isEmpty());
+		scriptExecutor.setMaintainSymbolTable(maintainSymbolTable);
 		return execute(script, scriptExecutor);
 	}
 
@@ -357,6 +364,30 @@ public class MLContext {
 	 */
 	public void setExplain(boolean explain) {
 		this.explain = explain;
+	}
+
+
+	/**
+	 * Obtain whether or not all values should be maintained in the symbol table
+	 * after execution.
+	 * 
+	 * @return {@code true} if all values should be maintained in the symbol
+	 *         table, {@code false} otherwise
+	 */
+	public boolean isMaintainSymbolTable() {
+		return maintainSymbolTable;
+	}
+
+	/**
+	 * Set whether or not all values should be maintained in the symbol table
+	 * after execution.
+	 * 
+	 * @param maintainSymbolTable
+	 *            {@code true} if all values should be maintained in the symbol
+	 *            table, {@code false} otherwise
+	 */
+	public void setMaintainSymbolTable(boolean maintainSymbolTable) {
+		this.maintainSymbolTable = maintainSymbolTable;
 	}
 
 	/**
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -23,6 +23,7 @@ import java.io.FileNotFoundException;
 import java.net.URL;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


deroneriksson:
Example:
```
scala> val ml = new MLContext(sc)

Welcome to Apache SystemML!

ml: org.apache.sysml.api.mlcontext.MLContext = org.apache.sysml.api.mlcontext.MLContext@2dec57c5

scala> val script = dml("A = matrix('1 2 3 4', rows=2, cols=2); print(toString(A));")
script: org.apache.sysml.api.mlcontext.Script =
Inputs:
None

Outputs:
None

scala> ml.execute(script)
1.000 2.000
3.000 4.000

res0: org.apache.sysml.api.mlcontext.MLResults =
None
scala> script.displaySymbolTable
res1: String =
"Symbol Table:
  [1] (String) _Var3: 1.000 2.000
3.000 4.000

  [2] (String) _Var2: 1.000 2.000
3.000 4.000

  [3] (Matrix) _mVar1: MatrixObject: scratch_space//_p61451_192.168.1.103//_t0/temp1_1, [2 x 2, nnz=4, blocks (1000 x 1000)], binaryblock, dirty
"

**Comment by deroneriksson** [comment]
@mboehm7 Do you think this is OK in its current state (default on), or do you think this should by default be off (due to potential performance/memory issues), but have a MLContext flag that can be set to turn it on? Note that Script does have a clearSymbolTable() method.

**Comment by nakul02** [comment]
@deroneriksson - as we discussed offline, maybe this is useful for debugging.
Maybe we could disable `rmvar` via a flag (from the command line options) to aid in debugging, 
as opposed to disabling it by default.

**Comment by deroneriksson** [comment]
@nakul02 That sounds like a good idea. It can be quite useful for debugging but I was a bit concerned about performance and memory requirements. I'll update to by default use the existing behavior.

**Comment by deroneriksson** [comment]
Updated to by default use existing behavior (only maintain symbol table values for outputs). Added ability to turn on maintaining symbol table values.

Example:
```
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 2.1.0
      /_/
         
Using Scala version 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60)

scala> import org.apache.sysml.api.mlcontext.ScriptFactory._

scala> val ml = new MLContext(sc)

Welcome to Apache SystemML!

ml: org.apache.sysml.api.mlcontext.MLContext = org.apache.sysml.api.mlcontext.MLContext@f1e1d66

scala> val script = dml("A = matrix('1 2 3 4', rows=2, cols=2); print(toString(A));")
script: org.apache.sysml.api.mlcontext.Script =
Inputs:
None

Outputs:
None

scala> ml.execute(script)
1.000 2.000
3.000 4.000

res0: org.apache.sysml.api.mlcontext.MLResults =
None

scala> script.displaySymbolTable
res1: String =
"Symbol Table:
None
"

scala> ml.setMaintainSymbolTable(true)

scala> ml.execute(script)
1.000 2.000
3.000 4.000

res3: org.apache.sysml.api.mlcontext.MLResults =
None

scala> script.displaySymbolTable
res4: String =
"Symbol Table:
  [1] (String) _Var6: 1.000 2.000
3.000 4.000

  [2] (String) _Var5: 1.000 2.000
3.000 4.000

  [3] (Matrix) _mVar4: MatrixObject: scratch_space//_p63600_192.168.1.103//_t0/temp2_2, [2 x 2, nnz=4, blocks (1000 x 1000)], binaryblock, dirty
"
```

nakul02:
@niketanpansare - thoughts?

niketanpansare:
Conceptually, it feels a bit odd that 'script' has 'symbol table', shouldn't it be part of MLResults or MLContext ?

deroneriksson:
When the new MLContext API was created, the symbol table for a DML script was moved to the Script object, thus allowing multiple scripts to each have their own symbol table. The Script object is what is 'executed' by ScriptExecutor via the new MLContext and the resulting execution has an associated symbol table of values. Conceptually, I agree about MLResults, since an MLResults is associated with the results of an executed script, and MLResults is indeed passed a copy of the symbol table after execution.

This PR simply optionally allows the option of keeping symbol table values around after execution (by default it uses the old behavior). This can be useful when debugging (which I am currently doing via some UI tools).

niketanpansare:
LGTM and also +1 for adding any debugging feature :)

bertholdreinwald:
Thanks, Deron. I guess that this feature is useful for debugging scenarios. However, removing all rmvars is too coarse. 

What you want to accomplish here is to carry vars from 1 DML invocation to the next, which is useful. However, I think that a user needs to specify which vars to hold on to as only the user knows the larger flow in this interactive scenario. Honestly, I thought that the JMLC API has already a method to "pin" vars. Will try to find it tomorrow. Holding on too many vars will make subsequent memory estimates and optimizations challenging. 

Thoughts?

mboehm7:
Overall, there can indeed be memory leaks and performance issues but this really depends on the used API. For MLContext, which runs with caching enabled, unnecessary variables are simply spilled out to disk; variables below 4KB are not subject to caching but it would require tens of thousands of intermediates before this becomes a problem. For JMLC, caching is disabled, and hence we cannot apply it there.

However, I agree with Berthold that removing all rmvar instructions is anyway too coarse-grained. The majority of rmvar instructions refers to intermediates of a DAG which are never bound to logical variables names and hence not accessible for a user. To create a whitelist of variables you want to keep, you can simply iterate over all DAGs and collect the names of transient_write root nodes.

deroneriksson:
@bertholdreinwald @mboehm7 
This is meant as a debugging tool for developers, who can actually care about the intermediate variables that get created. SystemML often does lots of hidden magic, and being able to see the intermediates and their values can shed light on some of that. Anyway, if you would prefer not to have this feature in master, I can keep it in my own branch so that I can use it.

deroneriksson:
I'm going to go ahead and merge this optional troubleshooting feature since it does not affect the existing default behavior and needs to manually be turned on for debugging purposes. If it causes issues for users, I can remove it from master.

PR #355, opened 2017-01-24
[SYSTEMML-1160] [WIP] Enable prefetching of batches via for loop iterator

niketanpansare:
@mboehm7 @bertholdreinwald @dusenberrymw  To avoid wasting time, I have created this PR to allow committers to raise their concerns about the integration of this approach before we spend a week or two in this direction, rather than when the work is complete. 

The high-level proposal is to extend our for loop to iterate over the rows of matrix as follows:
```python
for( [Xb, yb] in [X, y], batch_size=64) {
  ....
}
```

This construct allows us to perform low-level optimization such as prefetching of blocks without requiring the user to do so via explicit udf `next_batch(X, batch_size=64)`. 

To support this construct without overhauling the engine, we support at max two iterator variables. This allows us to use our existing [IterablePredicate](https://github.com/niketanpansare/incubator-systemml/blob/d94c8799b13682c53bdacdbb876559c4c251332d/src/main/java/org/apache/sysml/parser/IterablePredicate.java) construct. In addition, we have extended [ForProgramBlock](https://github.com/niketanpansare/incubator-systemml/blob/d94c8799b13682c53bdacdbb876559c4c251332d/src/main/java/org/apache/sysml/runtime/controlprogram/ForProgramBlock.java#L132) to support iterating over MatrixObject rather than IntObject.

The alternative to this approach are as follows:
1. The user specifies the prefetch budget (let's say 1G or 10%) and we reserve that for SP/CP `next_batch(X, batch_size=64)`.
2. Refactor breast cancer script to fetch batches in two step: First we extract a big_mini_batch which will likely be a SP Indexing instruction that generates a MatrixBlock which can fit in CP. Then, in a for loop, we perform CP indexing on it.

I will hold off on making progress in this PR until we all agree on this approach.

**PLEASE DONOT MERGE THIS PR YET**

### DIFF
--- a/src/main/java/org/apache/sysml/hops/rewrite/RewriteForLoopVectorization.java
+++ b/src/main/java/org/apache/sysml/hops/rewrite/RewriteForLoopVectorization.java
@@ -64,25 +64,27 @@ public class RewriteForLoopVectorization extends StatementBlockRewriteRule
 		if( sb instanceof ForStatementBlock )
 		{
 			ForStatementBlock fsb = (ForStatementBlock) sb;
-			ForStatement fs = (ForStatement) fsb.getStatement(0);
-			Hop from = fsb.getFromHops();
-			Hop to = fsb.getToHops();
-			Hop incr = fsb.getIncrementHops();
-			String iterVar = fsb.getIterPredicate().getIterVar().getName();
-			
-			if( fs.getBody()!=null && fs.getBody().size()==1 ) //single child block
-			{
-				StatementBlock csb = (StatementBlock) fs.getBody().get(0);
-				if( !(   csb instanceof WhileStatementBlock  //last level block
-					  || csb instanceof IfStatementBlock 
-					  || csb instanceof ForStatementBlock ) )
+			if(fsb.getIterPredicate().getIterVar().size() == 1) {
+				ForStatement fs = (ForStatement) fsb.getStatement(0);
+				Hop from = fsb.getFromHops();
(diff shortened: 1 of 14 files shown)


mboehm7:
Thanks for asking but I'm against the proposed language feature because it's R-incompatible and it can be easily done automatically. 

This scenario is a classic workload that benefits from data partitioning. So far, we do automatic column/row partitioning in PARFOR but should extend this to block partitioning (there should be an existing JIRA) and to FOR loops as well. Apart from data partitioning, and as a first step, the blocking should be relatively easy to integrate via a new rewrite.

niketanpansare:
Thanks @mboehm7 for your reply. I will close this PR in favor of the approach you mentioned :)

dusenberrymw:
@mboehm7 Thanks for weighing in!  Will we be able to interleave computation and data fetching with that approach?

mboehm7:
@niketanpansare no need to immediately close it - let's have the discussion here.

@dusenberrymw With the full data partitioning, we could actually making the batch slicing a no_op after spending some time for initial partitioning, and hence there would be no need to interleave fetching. With blocking, there is still the slice operation at roughly 2GB/s, so roughly 50ms for a batch of 64*200K, which should be negligible compared to the remaining computation.

niketanpansare:
Sure, let's use this thread for discussion. I closed the PR as it introduces iterator based for loop and had not started any work on prefetching.

PR #356, opened 2017-01-24
[SYSTEMML-1155] Use artifact versions from parent pom

deroneriksson:
Use Apache parent pom artifact versions for maven-dependency-plugin, maven-jar-plugin, maven-clean-plugin, and maven-antrun-plugin.

Did not switch versions for maven-surefire-plugin and maven-failsafe-plugin since switching these with existing plugin configurations resulted in slower test suite run times.


gweidner:
Thank you Deron for updating the pom and for checking tests.
LGTM

PR #357, opened 2017-01-24
[SYSTEMML-1190] Cleanup Scala UDF invocation

niketanpansare:
Since this functionality is not urgent and @mboehm7 has some concerns on the existing approach, I will revert back the UDF invocation and close [SYSTEMML-1190](https://issues.apache.org/jira/browse/SYSTEMML-1190) as won't fix.

@mboehm7 @fschueler @bertholdreinwald


mboehm7:
thanks Niketan - I do appreciate it. Could we keep as a requirement for future work on integrating Scala UDFs that there should be unique directory of functions (and their meta data) at optimization/runtime level, no matter through which API we're coming? I really believe this will make our lives much easier.

niketanpansare:
Sure, @fschueler can you please update your design document to add the requirement ?

fschueler:
I updated the document. I see that your approach is still in our documentation and probably should be removed for now.

niketanpansare:
I will push the gh branch today.

bertholdreinwald:
thanks, Felix. Were do you keep the document?

fschueler:
Hi Berthold,

this is the document that I mentioned in the email thread on the Scala 
DSL/API Design. I am trying to summarize all of the approaches that 
exist for SystemML in Scala. The goal here should be to align them (also 
in respect to Python).
Here is the link again, please comment/suggest/edit the doc if you see 
something missing (everyone should be able to comment/edit): 
https://docs.google.com/document/d/1ntdfmSBB83zdcWZl2iv4yMsrA3ZHDM-mzX5lMfFpfeo/edit

-Felix

Am 26.01.2017 00:00 schrieb bertholdreinwald:

PR #360, opened 2017-01-25
[SYSTEMML-1194] Replace SQLContext with SparkSession

deroneriksson:
SQLContext has been deprecated in Spark 2 and has been replaced by SparkSession.

### DIFF
--- a/src/main/java/org/apache/sysml/api/MLOutput.java
+++ b/src/main/java/org/apache/sysml/api/MLOutput.java
@@ -26,7 +26,7 @@ import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
-import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.types.StructType;
 import org.apache.sysml.runtime.DMLRuntimeException;
 import org.apache.sysml.runtime.controlprogram.context.SparkExecutionContext;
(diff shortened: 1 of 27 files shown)


fschueler:
I think you forgot the `MLContext.getMLMatrix(...)` method?

deroneriksson:
Interestingly, it looks like the SQLContext constructors are deprecated but the class itself is not, so Eclipse did not flag the SQLContext reference as deprecated. I'll go ahead and make all those substitutions.

Also, I believe I will need to also make some Python updates, since the python code also uses SQLContext and apparently there is also some Scala code that uses it.

src/main/java/org/apache/sysml/api/MLOutput.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SQLContext;
```
dusenberrymw: Can you remove this import?

deroneriksson:
All main tests passed here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1108/

Python tests had the same results as the code on master:
```
test_matrix_agg_fn.py 15 ran
test_matrix_binary_op.py 29 ran, 2 failed (convertMBtoPy4JDenseArr - Sparse to dense conversion is not yet implemented)
test_mlcontext.py 9 ran
test_mllearn_df.py 3 ran, 1 failed (score test_linear_regression_sk2)
test_mllearn_numpy.py 5 ran, 1 failed (score test_linear_regression)
```

mboehm7:
LGTM - as a nice to have, you might want to create a util method for SparkSession creation as it's used in quite a number of places.

Furthermore, while looking over the changes, I've seen two orthogonal issues:
* The call sc.getConf().set("spark.memory.offHeap.enabled", "false"); seems ineffective because usually all configurations need to be set before the spark context is created.
* There are a number of misleading method names that refer to DataFrames but return DataSets

deroneriksson:
Thank you for reviewing @mboehm7! I'm going to go ahead and merge as it is for now.

PR #361, opened 2017-01-25
[SYSTEMML-1191] Create RC builds from branch

asurve:
I have experimental changes on my own test repository.

https://github.com/asurve/arvind-sysml/tree/branch-0.12

@lresende Can you review changes on my repository and this code change?
   I have created branch-0.12.0 SP3 on branch-0.12 to make sure these changes are good enough.

  I could test only release:prepare, not other changes like release:publish.

I have included @ckadner changes specific to passphrase as well, as I was trying them. I know Luciano has already merged those changes to master branch.

### DIFF
--- a/dev/release/release-build.sh
+++ b/dev/release/release-build.sh
@@ -61,6 +61,9 @@ release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --develo
 release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --developmentVersion="0.12.0-SNAPSHOT" --releaseRc="rc1" --tag="v0.11.0-incubating-rc1"
 release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --developmentVersion="0.12.0-SNAPSHOT" --releaseRc="rc1" --tag="v0.11.0-incubating-rc1"  --gitCommitHash="a874b73" --dryRun
 
+# Create 0.12 RC2 builds from branch-0.12 
+./release-build.sh --release-prepare --releaseVersion="0.12.0-incubating" --developmentVersion="0.12.1-incubating-SNAPSHOT" --releaseRc="rc2" --tag="v0.12.0-incubating-rc2" --gitCommitHash="branch-0.12"
+
 release-build.sh --release-publish --gitCommitHash="a874b73"
 release-build.sh --release-publish --gitTag="v0.11.0-incubating-rc1"
 
@@ -141,12 +144,12 @@ while [ "${1+defined}" ]; do
 done
 
 
-for env in GPG_PASSPHRASE; do
-  if [ -z "${!env}" ]; then
-    echo "ERROR: $env must be set to run this script"
-    exit_with_usage
-  fi
-done
+if [[ -z "$GPG_PASSPHRASE" ]]; then
+    echo 'The environment variable GPG_PASSPHRASE is not set. Enter the passphrase to'
+    echo 'unlock the GPG signing key that will be used to sign the release!'
+    echo
+    stty -echo && printf "GPG passphrase: " && read GPG_PASSPHRASE && printf '\n' && stty echo
+fi
 
 if [[ "$RELEASE_PREPARE" == "true" && -z "$RELEASE_VERSION" ]]; then
     echo "ERROR: --releaseVersion must be passed as an argument to run this script"
@@ -223,7 +226,7 @@ echo "  "
 function checkout_code {
     # Checkout code
     rm -rf $RELEASE_WORK_DIR
-    mkdir $RELEASE_WORK_DIR
+    mkdir -p $RELEASE_WORK_DIR
     cd $RELEASE_WORK_DIR
     git clone https://git-wip-us.apache.org/repos/asf/incubator-systemml.git
     cd incubator-systemml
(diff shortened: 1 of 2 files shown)


dev/release/release-build.sh
```diff
@@ -225,9 +233,21 @@ function checkout_code {
     rm -rf $RELEASE_WORK_DIR
     mkdir $RELEASE_WORK_DIR
     cd $RELEASE_WORK_DIR
-    git clone https://git-wip-us.apache.org/repos/asf/incubator-systemml.git
+    if [[ "$GIT_BRANCH" ]]; then
+        git clone -b $GIT_BRANCH https://git-wip-us.apache.org/repos/asf/incubator-systemml.git
```
ckadner: since you are doing the `git checkout $GIT_BRANCH` below, you should not need to do a `clone -b $GIT_BRANCH ...` here

pom.xml
```diff
@@ -25,7 +25,7 @@
 		<version>18</version>
 	</parent>
 	<groupId>org.apache.systemml</groupId>
-	<version>0.13.0-incubating-SNAPSHOT</version>
+	<version>0.12.1-incubating-SNAPSHOT</version>
```
ckadner: I think this pom file change should not be part of this commit

dev/release/release-build.sh
asurve: Just "git checkout branch name" be sufficient to create local branch, pull changes from branch locally?

ckadner: one or the other is sufficient

but, none of your changes should be required if you invoke the script with the `--GIT_REF="branch-name"` parameter

ckadner:
@asurve It seems that no changes to the `realease-build.sh` script are required. You can simply specify the branch when invoking the script with `--gitCommitHash="branch-name"`

I just did a `--dryRun` to verify it on my local clone

dev/release/release-build.sh
asurve: Dry run works to checkout from a commit hash, but maven command fails in non dryRun.

asurve:
@ckadner  Please see my earlier response.
I tried --dryRun. --dryRun works fine because it can create a branch locally based on commit hash user specifies. But when you run without --dryRun it fails as branch created locally does not accept any changes.
Do you see any issue with current approach added in this PR?

I don't think we need to find alternative solution. I want to make sure Release Candidate we will be creating with this solution  is right and does not introduce any other issue.

ckadner:
@asurve  when invoking the script with `--gitCommitHash="branch-name"` paramter you need to use an existing remote branch, the release-build script is not creating a new local branch

I just ran the unchanged script with  `--gitCommitHash="branch-name"` on Apache Bahir and accidentally missed to use the `--dryRun` flag. The release script finished without any error and created the release tag.

I would advocate to not introduce additional redundant if-then-else code blocks with potential for more errors.

asurve:
Does --gitComitHash parameter suppose to provide commit hash, not branch name? 

I run script with --dryRun and without --dryRun using --gitComitHash="Branch Hash", and had issue.

ckadner:
you would run it with `--gitComitHash="branch-0.12"` in order to create a new release tag on branch `branch-0.12`

asurve:
I understand that. That is another confusion we may introduce for user who may try this script.
I would prefer not to continue with additional experiments, if there is no issue with solution I would like to get going with build.

ckadner:
The problem I see with your changes is that you introduce additional complexity which is not needed to accomplish what can already be done.

The parameter you propose to introduce `--gitBranch` conflicts with the existing parameter `--gitCommitHash` which already has some conflicts with usage of the `--gitTag` parameter when running the `release-publish` goal.

Git uses commit hashes, branch names, tags as pointers to revisions. Most git commands accept any of them interchangeably.

You should get some feedback from @lresende as well. But if you go ahead with your changes then you need to add some additional parameter validation to eliminate conflicts (even more if-then-else code)

asurve:
Thanks for comments.

dev/release/release-build.sh
```diff
@@ -61,6 +61,9 @@ release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --develo
 release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --developmentVersion="0.12.0-SNAPSHOT" --releaseRc="rc1" --tag="v0.11.0-incubating-rc1"
 release-build.sh --release-prepare --releaseVersion="0.11.0-incubating" --developmentVersion="0.12.0-SNAPSHOT" --releaseRc="rc1" --tag="v0.11.0-incubating-rc1"  --gitCommitHash="a874b73" --dryRun
 
+# Create 0.12 RC2 builds from branch-0.12
```
ckadner: since this is inside a `cat << EOF` section this comment will be printed to the console as part of the scripts help text

ckadner: I would add to the explanation of `--gitCommitHash` in the _OPTIONS_ passage above
from:
`--gitCommitHash - Release tag or commit to build from, default master HEAD`
to:
`--gitCommitHash - Release tag, branch name, or commit to build from, default master HEAD`

dusenberrymw:
Ping.  What is the plan here?

PR #362, opened 2017-01-25
[SYSTEMML-1198] Ensure jar included in Python release artifact

gweidner:
Added condition in pre_setup.py to match different jar name for release build vs snapshot build.

### DIFF
--- a/src/main/python/pre_setup.py
+++ b/src/main/python/pre_setup.py
@@ -30,6 +30,6 @@ if os.path.exists(java_dir_full_path):
 os.mkdir(java_dir_full_path)
 root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.getcwd())))
 for file in os.listdir(os.path.join(root_dir, 'target')):
-    if fnmatch.fnmatch(file, 'systemml-*-incubating-SNAPSHOT.jar'):
+    if fnmatch.fnmatch(file, 'systemml-*-incubating-SNAPSHOT.jar') or fnmatch.fnmatch(file, 'systemml-*-incubating.jar'):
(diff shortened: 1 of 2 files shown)


gweidner:
A distribution build was tested two ways:
1) Normal development/test build where jar name has -SNAPSHOT.  Verified python artifact contained -SNAPSHOT jar.
2) Simulated release build, i.e., removed SNAPSHOT in pom version name.  Verified python artifact contained release jar (i.e., jar file without snapshot in name.

gweidner:
@dusenberrymw @niketanpansare

gweidner:
Here's link to RC1 artifact which had the issue found during validation testing:
https://dist.apache.org/repos/dist/dev/incubator/systemml/0.12.0-incubating-rc1/systemml-0.12.0-incubating-python.tgz

gweidner:
It looks like classloader.py might also need similar change:
Line 58:             if fnmatch.fnmatch(file, 'systemml-*-incubating-SNAPSHOT.jar'):

niketanpansare:
Thanks @gweidner :+1:

gweidner:
Thank you @dusenberrymw and @niketanpansare for reviewing.  I'm attaching screen captures from Python Notebook where I've uninstalled/re-installed the previous/updated artifact and restarted kernel.

notebook_matrix

notebook_logistic

gweidner:
For reference in case I've mis-interpreted anything, here are examples of error messages observed before making both changes:

notebook_directory_error

notebook_loader_error

bertholdreinwald:
Hah. I noticed that too by analyzing the different artifacts on apache site and maven site. Sent out an email to dev list prior to reading that thread.

PR #364, opened 2017-01-26
[SYSTEMML-527] Write function description parameter

deroneriksson:
Add description parameter to write function to allow writing description
to metadata. Additionally set author to be user name if available and add
a created timestamp.

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/spark/WriteSPInstruction.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/spark/WriteSPInstruction.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Random;
 
-import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.mapred.SequenceFileOutputFormat;
 import org.apache.spark.Accumulator;
@@ -56,6 +55,7 @@ public class WriteSPInstruction extends SPInstruction
 	private CPOperand input1 = null; 
 	private CPOperand input2 = null;
 	private CPOperand input3 = null;
+	private CPOperand input4 = null;
 	private FileFormatProperties formatProperties;
 	
 	//scalars might occur for transform
@@ -104,12 +104,14 @@ public class WriteSPInstruction extends SPInstruction
 			
 			boolean isInputMB = Boolean.parseBoolean(parts[7]);
 			inst.setInputMatrixBlock(isInputMB);
+
+			CPOperand in4 = new CPOperand(parts[8]);
+			inst.input4 = in4;
 		} else {
 			FileFormatProperties ffp = new FileFormatProperties();
-			String description = parts[4];
-			if (StringUtils.isNotBlank(description)) {
-				ffp.setDescription(description);
-			}
+
+			CPOperand in4 = new CPOperand(parts[4]);
+			inst.input4 = in4;
 			inst.setFormatProperties(ffp);
 		}
 		return inst;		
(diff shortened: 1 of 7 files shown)


deroneriksson:
This can be used for labelling where a matrix came from (for instance, "description": "I was generated by SystemML's Univariate Statistics"

Example writes:
```
A = matrix("1 2 3 4", rows=2, cols=2);
write(A, "results/A.csv", format="csv");
write(A, "results/Adesc.csv", format="csv", description="description 1");
write(A, "results/B.text", format="text");
write(A, "results/Bdesc.text", format="text", description="description 2");
write(A, "results/C.mm", format="mm");
write(A, "results/Cdesc.mm", format="mm", description="description 3");
write(A, "results/D.bin", format="binary");
write(A, "results/Ddesc.bin", format="binary", description="description 4");

X = 5;
Xdesc = "description " + X;
write(A, "results/Xdesc.csv", format="csv", description=Xdesc);

#This erroneously sets the description to "Ydesc"
Ydesc = "description 6";
if (1 == 1) {
    Ydesc = "this should be the description"
}
write(A, "results/Ydesc.csv", format="csv", description=Ydesc);
```

This gives:
A.csv.mtd
```
{
    "data_type": "matrix",
    "value_type": "double",
    "rows": 2,
    "cols": 2,
    "nnz": 4,
    "format": "csv",
    "header": false,
    "sep": ",",
    "author": "deroneriksson",
    "created": "2017-01-26 15:14:41 PST"
}
```

Adesc.csv.mtd:
```
{
    "data_type": "matrix",
    "value_type": "double",
    "rows": 2,
    "cols": 2,
    "nnz": 4,
    "format": "csv",
    "header": false,
    "sep": ",",
    "description": "description 1",
    "author": "deroneriksson",
    "created": "2017-01-26 15:14:41 PST"
}
```
Etc..

Note that all writes work as expected except the last one.

Here is the second to last:
Xdesc.csv.mtd:
```
{
...
    "description": "description 5",
    "author": "deroneriksson",
    "created": "2017-01-26 15:14:41 PST"
}
```

Here is the last (incorrect) one, which is outputting the description as the variable name rather than the variable value:
Ydesc.csv.mtd:
```
{
...
    "description": "Ydesc",
    "author": "deroneriksson",
    "created": "2017-01-26 15:14:41 PST"
}
```

@mboehm7 @niketanpansare Am I doing something wrong that is causing the incorrect behavior in the last example?

deroneriksson:
BTW, all tests passed here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/251/

src/main/java/org/apache/sysml/runtime/instructions/spark/WriteSPInstruction.java
```diff
@@ -103,6 +104,13 @@ public static WriteSPInstruction parseInstruction ( String str )
 			
 			boolean isInputMB = Boolean.parseBoolean(parts[7]);
 			inst.setInputMatrixBlock(isInputMB);
+		} else {
+			FileFormatProperties ffp = new FileFormatProperties();
+			String description = parts[4];
```
niketanpansare: Since the description can be a variable, at parse time you should do parse part[4] as CPOperand. Then in process instruction, you should check if the passed value is a variable or a literal. For example:
```bash
String fname = ec.getScalarInput(input2.getName(), ValueType.STRING, input2.isLiteral()).getStringValue();
```

deroneriksson: Thank you!

deroneriksson:
@niketanpansare @mboehm7 @dusenberrymw 
This now allows variables thanks to @niketanpansare feedback. Does this look good?

niketanpansare:
Please note that description cannot be an expression and always has to be a literal

deroneriksson:
Thank you for reviewing and for your help @niketanpansare!

BTW I also tried writes in a for loop and this is also handled correctly.
```
A = matrix("1 2 3 4", rows=2, cols=2);
for (i in 1:3) {
  n = A + 1;
  write(n, "results/myresult-" + i + ".csv", format="csv", description="mydescription " + i);
}
```

PR #366, opened 2017-01-27
[SYSTEMML-1208] Mimic main website styles in project documentation

deroneriksson:
Update project documentation styles to be similar to main website styles.

### DIFF
--- a/docs/_layouts/global.html
+++ b/docs/_layouts/global.html
@@ -56,8 +56,8 @@
                                 <li><b>Language Guides:</b></li>
                                 <li><a href="dml-language-reference.html">DML Language Reference</a></li>
                                 <li><a href="beginners-guide-to-dml-and-pydml.html">Beginner's Guide to DML and PyDML</a></li>
-                                <li><a href="beginners-guide-python.html">Beginner's Guide for Python users</a></li>
-                                <li><a href="python-reference.html">Reference Guide for Python users</a></li>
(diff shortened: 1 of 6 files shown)


deroneriksson:
Preview here: http://deroneriksson.github.io/incubator-systemml/

Checked with Firefox, Chrome, and Safari on OS X.

fschueler:
I love that it is integrated, but the orange font-color is making me dizzy! ;)

deroneriksson:
Maybe I could tone down the orange. I darkened the main text font color since it's nice to have a darker color when reading lots of documentation.

deroneriksson:
@fschueler What do you think about dimming the link orange from `#FF5003` to `#FF7032`?

fschueler:
I was looking at the MLContext section and because there are so many links it just made it a big orange wall. I actually like the current green/turquise/whatever color it is.

But I guess that's a matter of taste here ;)

deroneriksson:
I like both. I think toning down the orange is good since it blends in more now when pages have lots of links. It might be good to not use bold/strong font weights with the orange too (we have a lot on the Overview page) since the orange is already such a strong color.

deroneriksson:
@fschueler @gweidner 
Do you think this is ready to merge?

gweidner:
Thank you @deroneriksson !
LGTM

PR #369, opened 2017-02-02
[SYSTEMML-1224] Migrate Vector and LabeledPoint classes from mllib to ml

deroneriksson:
Migrate:
mllib.linalg.DenseVector to ml.linalg.DenseVector.
mllib.linalg.Vector to ml.linalg.Vector.
mllib.linalg.Vectors to ml.linalg.Vectors.
mllib.linalg.VectorUDT to ml.linalg.VectorUDT.
mllib.regression.LabeledPoint to ml.feature.LabeledPoint.


deroneriksson:
This addresses issue seen with:
```
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_mllearn_df.py
```
that gives:
```
java.lang.ClassCastException: org.apache.spark.ml.linalg.SparseVector cannot be cast to org.apache.spark.mllib.linalg.Vector
	at org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtils.countNnz(RDDConverterUtils.java:314)
```

All tests passed here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/256/

cc @iyounus

asurve:
stringDataFrameToVectorDataFrame() from RDDConverterUtilsExt has been removed due to unavailability of Vectors.parse() function.

@niketanpansare Can you look at the code change and address needed change if function is needed.

Data Generation needed for Performance testing is being blocked. Once this get fixed there are other issues to be addressed as well and urgent. Please look at this change sooner.

deroneriksson:
RDDConverterUtilsExt is an internal SystemML class (not an API class) with 0 references in the project. It should not be called by external users.

fschueler:
We also have to change our documentation accordingly. Currently the [LinarRegression Example](https://apache.github.io/incubator-systemml/spark-mlcontext-programming-guide#generate-linear-regression-test-data) uses the old version. Also: The `LinearDataGenerator` only generates "old" `mllib.regression.LabeledPoint` while the `LinearRegression.fit()` function doesn't take those as input.

fschueler:
`RDDConverterUtilsExt` is documented and might make it seem like it should be a user-facing API. We might want to change the docs there too if we don't want people to use it.

deroneriksson:
@fschueler With the new MLContext API, all conversions can either happen automatically or via api.mlcontext.MLContextConversionUtil, thus under the API package. If RDDConverterUtils/RDDConverterUtilsExt methods are needed that are not included in MLContextConversionUtil, please add wrappers to these to these calls to MLContextConversionUtil so that we can avoid tightly coupling user code to our internal code.

deroneriksson:
@fschueler Additionally RDDConverterUtilsExt has this javadoc, which suggests that these conversion utils are not guaranteed to be thoroughly tested and that they will move. This is one more good reason to move any needed converters to MLContextConversionUtil and to use this class in our documentation.
```
/**
 * NOTE: These are experimental converter utils. Once thoroughly tested, they
 * can be moved to RDDConverterUtils.
 */
```

fschueler:
Okay, maybe we can remove `RDDConverterUtilsExt` from the docs then.

deroneriksson:
I believe @iyounus needs this PR since the class cast exceptions are preventing deeper project Spark 2 testing.

cc @frreiss

iyounus:
We need this PR because some of the python tests don't run at all. With this fix the tests do run but one test fails. It would be better if this PR gets merged in so that I can investigate the failing python test.

niketanpansare:
Can you please look at 'def _java2py(sc, obj):' and similar py2java method in MLContext ?

deroneriksson:
@niketanpansare What about those methods?

niketanpansare:
I was addressing @iyounus comment about investing the failing python test. My suspicion is that these methods are probably the cause of the failures as they invoke corresponding [MLLib methods](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mlcontext.py#L78).

deroneriksson:
All tests passed: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1069/console

@niketanpansare Does this PR look good to you? I deleted stringDataFrameToVectorDataFrame() since Vectors.parse() doesn't exist for Vectors in the ml package and this method is not used in the project. If this method is needed by someone, perhaps it can be migrated to Spark 2 in another PR?

asurve:
I am ok to delete this method for now, but make sure to have Jira opened to get it added in new MLcontext API after 0.13 release.

deroneriksson:
Ran following python tests w/ this PR:
```
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_matrix_agg_fn.py
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_matrix_binary_op.py
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_mlcontext.py
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_mllearn_df.py
spark-submit --driver-class-path $SYSTEMML_HOME/SystemML.jar test_mllearn_numpy.py
```

These gave following results:
```
15 (All passed)
29 (All passed)
9 (All passed)
3 (1 failure - scoring issue)
5 (1 failure - scoring issue)
```

Before this PR, many ClassCastExceptions occur. This PR allows @iyounus to investigate the deeper scoring issue.

dusenberrymw:
As part of this, can we explore updating the use of `pyspark.mllib.common` in the Python MLContext with the newer [`pyspark.ml.common`](https://github.com/apache/spark/blob/master/python/pyspark/ml/common.py)?  We would need to change the [import](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mlcontext.py#L30) and usages ([1](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mlcontext.py#L78), [2](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mlcontext.py#L94)), and then test everything.

deroneriksson:
@dusenberrymw Perhaps a Python developer or data scientist can take on this issue in a separate JIRA issue and PR?

deroneriksson:
Can I get an L?
Can I get a G?
Can I get a T?
Can I get an M?
What's that spell? LGTM!
Anyone?
@niketanpansare @dusenberrymw @iyounus @asurve @fschueler

deroneriksson:
@dusenberrymw Thank you!!!

I'll create a JIRA for the Python change and also for migrating the deleted method that may or may not be used by anyone.

PR #370, opened 2017-02-02
[SYSTEMML-1222] MLContext input long to int cast

deroneriksson:
Convert MLContext long input to int since long not a supported value type.
Fix input value type error message to display input name rather than value.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -247,7 +247,7 @@ public final class MLContextUtil {
 			}
 		}
 		if (!supported) {
-			throw new MLContextException("Input name (\"" + value + "\") value type not supported: " + o.getClass());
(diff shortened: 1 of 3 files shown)


deroneriksson:
Can validate with:
```
val ml = new MLContext(sc)
val scr = dml("A=matrix('1 2 3 4', rows=2, cols=2);write(A, 'A.csv', format='csv');")
ml.execute(scr)
val scr2 = dml("A=read('A.csv', cols_in_block=X); print(toString(A));");
val X : Long = 5
scr2.in("X", X).out("A")
ml.execute(scr2).getMatrix("A").toDF().show()
```

cc @dusenberrymw

dusenberrymw:
LGTM.  Thanks, @deroneriksson.

deroneriksson:
@dusenberrymw Thanks for reviewing!

deroneriksson:
BTW, while preparing to merge, I noticed that the following code isn't Java 6 compatible:
```
value = (int) (long) value;
```

So, I am updating to the following which is also Java 6 compatible for backwards compatibility:
```
Long lng = (Long) value;
value = lng.intValue();
```

PR #371, opened 2017-02-02
[SYSTEMML-1181] Change sqlContext to spark in MLContext docs

fschueler:
The variable `sqlContext` is not available by default in the spark shell anymore, instead `spark` should be used to create `DataFrames`. 

Where methods expect an instance of `SqlContext`, arguments are replaced with `spark.sqlContext`.


deroneriksson:
LGTM. Thank you @fschueler!

PR #372, opened 2017-02-02
[SYSTEMML-1227] Cleanup pyc files

deroneriksson:
Delete *.pyc files under src/main/python when 'mvn clean' is run.
Add *.pyc to .gitignore.


deroneriksson:
Thank you for reviewing @dusenberrymw!

PR #373, opened 2017-02-03
[SYSTEMML-1220] dmlFromResource in Python MLContext

deroneriksson:
Add dmlFromResource/pydmlFromResource capability to Python MLContext.

### DIFF
--- a/src/main/python/systemml/mlcontext.py
+++ b/src/main/python/systemml/mlcontext.py
@@ -19,7 +19,7 @@
 #
 #-------------------------------------------------------------
 
-__all__ = ['MLResults', 'MLContext', 'Script', 'dml', 'pydml', '_java2py', 'Matrix']
+__all__ = ['MLResults', 'MLContext', 'Script', 'dml', 'pydml', 'dmlFromResource', 'pydmlFromResource', '_java2py', 'Matrix']
 
 import os
 
@@ -52,6 +52,24 @@ def dml(scriptString):
         raise ValueError("scriptString should be a string, got %s" % type(scriptString))
     return Script(scriptString, scriptType="dml")
 
+def dmlFromResource(resourcePath):
+    """
+    Create a dml script object based on a resource path.
+
+    Parameters
+    ----------
+    resourcePath: string
+        Path to a dml script on the classpath.
+
+    Returns
+    -------
+    script: Script instance
+        Instance of a script object.
+    """
+    if not isinstance(resourcePath, str):
+        raise ValueError("resourcePath should be a string, got %s" % type(resourcePath))
+    return Script(resourcePath, scriptType="dml", isResource=True)
+
 
 def pydml(scriptString):
     """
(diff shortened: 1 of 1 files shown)


deroneriksson:
Can test with code similar to:
```
mvn clean
pip3 install -e src/main/python
mvn clean package
PYSPARK_PYTHON=python3 pyspark --driver-class-path target/SystemML.jar
from systemml import MLContext, dmlFromResource
ml = MLContext(sc)
script = dmlFromResource("/scripts/algorithms/Univar-Stats.dml")
ml.execute(script)
```
(in this above example, inputs and outputs are hardcoded into script)

deroneriksson:
cc  @bertholdreinwald @dusenberrymw

deroneriksson:
@dusenberrymw @bertholdreinwald 
I think a larger issue is whether "algorithm customizability" is one of the main benefits of SystemML that we are promoting. Algorithm wrappers and pulling scripts from the jar basically are saying we're taking a static rather than dynamic approach.

mboehm7:
well, besides (1) custom analytics, there is also the argument of (2) adaptation to changing data and cluster characteristics. For me, referencing existing dml scripts is a must have feature (that should preferably be consistent across all APIs and language bindings).

deroneriksson:
Since our main algorithms are packaged in the jar, something like dmlFromResource does offer a simple way of accessing the script without needing to do something such as unpack the jar file or copy/paste the script from the file system or a URL. It's a great way to easily reference and run an algorithm and in no way prevents algorithm customization. If I would like to quickly demonstrate running an algorithm from the spark shell, most likely I would use dmlFromUrl (reference a script on github) or dmlFromReference (reference a script in the jar), an dmlFromResource would use the specific script packaged with the specific jar, so it's probably the better technique to use. So I would be in favor of adding dmlFromResource to the Python API.

If this PR looks good or if the Python needs any updates, please let me know.

@mboehm7 @bertholdreinwald @dusenberrymw

deroneriksson:
Since no one is giving this PR a LGTM, I am going to close this PR. If you would like this functionality, please contact me and I can re-open it.

@mboehm7 @bertholdreinwald @dusenberrymw

bertholdreinwald:
sorry for the late response. My opinion would be to have dmlFromResource, because
1. provides easy access to DML script in JAR, where ever the JAR is in a cloud environment.
2. consistency with Scala API.
3. having variety of where your DML comes from.
4. dmlFromUrl requires another hop.

I look at it as a progression of DML usage ... people may start with (1) DML as strings, (2) that will become wildly soon as the size of the script grows ... then it goes in a file. Where do i put it if i am in a hosted env. If I put it in git, then it will 1st go in staging area, and I can use it. (3) Once blessed, then a wrapper needs to be developed. 

All that takes time, and blocks experimentation with the script.

bertholdreinwald:
LGTM.

deroneriksson:
Thank you @bertholdreinwald. I agree.

PR #374, opened 2017-02-03
[SYSTEMML-951] Remove unnecessary reblock of binary blocks

niketanpansare:
Currently, we always force the reblocking of binary blocks into squared 1K x 1K. Though allowing 1K x 1K block size by default is a good decision, we should allow expert users to reblock the binary block to a non-squared blocks or blocks of different sizes (without changing the default block size). In this case, if SystemML reads the data in binary block format, we should respect the original blocking.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/rewrite/RewriteBlockSizeAndReblock.java
+++ b/src/main/java/org/apache/sysml/hops/rewrite/RewriteBlockSizeAndReblock.java
@@ -57,10 +57,28 @@ public class RewriteBlockSizeAndReblock extends HopRewriteRule
 		
 		//perform reblock and blocksize rewrite
 		for( Hop h : roots ) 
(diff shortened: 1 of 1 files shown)


mboehm7:
Rather than simply disabling any binary block reblocks, it would be better to expose both brlen and bclen in our configuration file and respective APIs in order to avoid confusion. If the input blocksizes match the current configuration, there is no reblock; otherwise, we'll bring the input into the requested blocking scheme (i.e., we should stick to the current semantics).   

Making the change towards rectangular blocks would mostly entail (1) changing the configuration management and APIs as the compiler/runtime already handles rectangular blocksizes, and (2) testing this at scale with good coverage (e.g., with our perf testsuite) as I would expect a number of related issues showing up.

niketanpansare:
Closing this PR for now. We can revisit this after discussion on https://github.com/apache/incubator-systemml/pull/389 and after experiments with rectangular blocks

PR #375, opened 2017-02-03
[SYSTEMML-1228] Fix SparkListener to work with Spark 2.1

fschueler:
This PR sets the Spark version in the pom.xml to 2.1 and updates the SparkListener to work with Spark 2.1.

The `makeOperationGraph` method now takes a parameter for the number of nodes retained which I set to Integer.MAX_LEVEL, which should retain all nodes. I wasn't sure what the best number should be here...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
 	<properties>
 		<hadoop.version>2.4.1</hadoop.version>
 		<antlr.version>4.5.3</antlr.version>
-		<spark.version>2.0.2</spark.version>
+		<spark.version>2.1.0</spark.version>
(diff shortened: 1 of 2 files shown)


niketanpansare:
@fschueler we can either go with this approach or cleanup the monitoring utils as in https://github.com/apache/incubator-systemml/pull/376

fschueler:
I don't know enough about the monitoring utils to decide that, that's why I went with the minimal fix ;)

niketanpansare:
Makes sense. Do you mind closing this PR in favor of [cleanup](https://github.com/apache/incubator-systemml/pull/376)  ?

PR #377, opened 2017-02-04
[SYSTEMML-1181] Remove old Spark MLContext API documentation

fschueler:
This removes documentation for the old MLContext API since its usage will be deprecated. With versioned documentation is should remain part of the documentation of releases that still support it. For the next release we should not document it to discourage its usage.

We should also write new documentation for using SystemML with Zeppelin and Jupyter notebooks.


deroneriksson:
LGTM. Thank you @fschueler. This keeps the Spark MLContext Programming Guide focused on the new API.

PR #378, opened 2017-02-05
[SYSTEMML-1233] Deprecate old MLContext API

deroneriksson:
Deprecate old MLContext API, to be removed in SystemML 1.0.
Ignore references to deprecated MLContext classes in project.


deroneriksson:
Thank you for reviewing @fschueler !

BTW, all tests pass except random unrelated FullReblockTest. https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1080/

PR #379, opened 2017-02-05
[SYSTEMML-1232] Migrate stringDataFrameToVectorDataFrame to ml Vector

deroneriksson:
Restore and migrate RDDConverterUtilsExt.stringDataFrameToVectorDataFrame
method from mllib Vector class to ml Vector class. Use NumericParser since
ml.linalg.Vectors.parse() does not exist.

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
@@ -28,6 +28,7 @@ import java.util.Iterator;
 
 import org.apache.hadoop.io.Text;
 import org.apache.spark.SparkContext;
+import org.apache.spark.SparkException;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
@@ -407,7 +408,9 @@ public class RDDConverterUtilsExt
 				ArrayList<Object> fieldsArr = new ArrayList<Object>();
 				for (int i = 0; i < oldRow.length(); i++) {
 					Object ci = oldRow.get(i);
-					if (ci instanceof String) {
+					if (ci == null) {
+						fieldsArr.add(null);
+					} else if (ci instanceof String) {
 						String cis = (String) ci;
 						StringBuffer sb = new StringBuffer(cis.trim());
 						for (int nid = 0; i < 2; i++) { // remove two level
@@ -421,10 +424,14 @@ public class RDDConverterUtilsExt
 						// have the replace code
 						String ncis = "[" + sb.toString().replaceAll(" *, *", ",") + "]";
 
-						// [ ] will always result in double array
-						double[] doubles = (double[]) NumericParser.parse(ncis);
-						Vector dense = Vectors.dense(doubles);
-						fieldsArr.add(dense);
+						try {
+							// ncis [ ] will always result in double array return type
+							double[] doubles = (double[]) NumericParser.parse(ncis);
+							Vector dense = Vectors.dense(doubles);
+							fieldsArr.add(dense);
+						} catch (Exception e) { // can't catch SparkException here in Java apparently
+							throw new DMLRuntimeException("Error converting to double array. " + e.getMessage(), e);
+						}
 
 					} else {
 						throw new DMLRuntimeException("Only String is supported");
--- a/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
@@ -21,14 +21,12 @@ package org.apache.sysml.runtime.instructions.spark.utils;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Scanner;
 
 import org.apache.hadoop.io.Text;
-import org.apache.spark.Accumulator;
 import org.apache.spark.SparkContext;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
@@ -40,6 +38,7 @@ import org.apache.spark.ml.linalg.VectorUDT;
 import org.apache.spark.ml.linalg.Vectors;
 import org.apache.spark.mllib.linalg.distributed.CoordinateMatrix;
 import org.apache.spark.mllib.linalg.distributed.MatrixEntry;
+import org.apache.spark.mllib.util.NumericParser;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.RowFactory;
@@ -47,15 +46,7 @@ import org.apache.spark.sql.SQLContext;
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
-
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-
-import scala.Tuple2;
-
 import org.apache.sysml.runtime.DMLRuntimeException;
-import org.apache.sysml.runtime.instructions.spark.functions.ConvertMatrixBlockToIJVLines;
-import org.apache.sysml.runtime.io.IOUtilFunctions;
 import org.apache.sysml.runtime.matrix.MatrixCharacteristics;
 import org.apache.sysml.runtime.matrix.data.MatrixBlock;
 import org.apache.sysml.runtime.matrix.data.MatrixCell;
@@ -63,7 +54,8 @@ import org.apache.sysml.runtime.matrix.data.MatrixIndexes;
 import org.apache.sysml.runtime.matrix.mapred.IndexedMatrixValue;
 import org.apache.sysml.runtime.matrix.mapred.ReblockBuffer;
 import org.apache.sysml.runtime.util.FastStringTokenizer;
-import org.apache.sysml.runtime.util.UtilFunctions;
+
+import scala.Tuple2;
 
 /**
  * NOTE: These are experimental converter utils. Once thoroughly tested, they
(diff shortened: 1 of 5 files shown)


deroneriksson:
Can test with code such as:
```
static class StringToRow implements Function<String, Row> {
	private static final long serialVersionUID = 3945939649355731805L;

	@Override
	public Row call(String str) throws Exception {
		return RowFactory.create(str);
	}
}
	
@Test
public void testStuff() throws DMLRuntimeException {
	List<String> list = new ArrayList<String>();
	list.add("((1.2, 4.3, 3.4))");
	list.add("(1.2, 3.4, 2.2)");
	list.add("[[1.2,34.3, 1.2, 1.2]]");
	list.add("[1.2, 3.4]");
	JavaRDD<String> javaRddString = sc.parallelize(list);
	JavaRDD<Row> javaRddRow = javaRddString.map(new StringToRow());
	SQLContext sqlContext = new SQLContext(sc);
	List<StructField> fields = new ArrayList<StructField>();
	fields.add(DataTypes.createStructField("C1", DataTypes.StringType, true));
	StructType schema = DataTypes.createStructType(fields);
	Dataset<Row> inDF = sqlContext.createDataFrame(javaRddRow, schema);
	Dataset<Row> outDF = RDDConverterUtilsExt.stringDataFrameToVectorDataFrame(sqlContext, inDF);
		
	List<Row> outputList = outDF.collectAsList();
	for (Row row : outputList) {
		System.out.println("ROW:" + row);
	}
}
```

deroneriksson:
All tests passed: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1081/console

frreiss:
@deroneriksson can you please add a unit test for this method? We need to prevent future breakage of code in other projects that calls `stringDataFrameToVectorDataFrame()`.

deroneriksson:
@frreiss 
Sure, I can add a basic unit test based on that example I included in this PR.

Since RDDConverterUtilsExt javadocs describe it as "NOTE: These are experimental converter utils. Once thoroughly tested, they can be moved to RDDConverterUtils", I really hope users will not directly reference RDDConverterUtilsExt. However, some of our older documentation did point to this non-API class so it is understandable that it may being used.

deroneriksson:
@frreiss For the new MLContext API, the org.apache.sysml.api.mlcontext.MLContextConversionUtil class was created to hopefully serve to buffer users from SystemML internal conversion functionality.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -362,4 +354,68 @@ private void flushBufferToList( ReblockBuffer rbuff,  ArrayList<Tuple2<MatrixInd
...
+						for (int nid = 0; i < 2; i++) { // remove two level
+														// nesting
+							if ((sb.charAt(0) == '(' && sb.charAt(sb.length() - 1) == ')')
+									|| (sb.charAt(0) == '[' && sb.charAt(sb.length() - 1) == ']')) {
+								sb.deleteCharAt(0);
+								sb.setLength(sb.length() - 1);
+							}
+						}
+						// have the replace code
+						String ncis = "[" + sb.toString().replaceAll(" *, *", ",") + "]";
```
frreiss: This regex substitution is going to be expensive. Have you benchmarked the performance of this conversion function and verified that it is acceptably fast?

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -362,4 +354,68 @@ private void flushBufferToList( ReblockBuffer rbuff,  ArrayList<Tuple2<MatrixInd
...
+
+						// [ ] will always result in double array
+						double[] doubles = (double[]) NumericParser.parse(ncis);
+						Vector dense = Vectors.dense(doubles);
+						fieldsArr.add(dense);
+
+					} else {
+						throw new DMLRuntimeException("Only String is supported");
```
frreiss: You should probably handle null values instead of erroring out when a field is null here.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -362,4 +354,68 @@ private void flushBufferToList( ReblockBuffer rbuff,  ArrayList<Tuple2<MatrixInd
...
+									|| (sb.charAt(0) == '[' && sb.charAt(sb.length() - 1) == ']')) {
+								sb.deleteCharAt(0);
+								sb.setLength(sb.length() - 1);
+							}
+						}
+						// have the replace code
+						String ncis = "[" + sb.toString().replaceAll(" *, *", ",") + "]";
+
+						// [ ] will always result in double array
+						double[] doubles = (double[]) NumericParser.parse(ncis);
```
frreiss: You should detect parse errors here and emit a usable error message when an input string doesn't properly parse.

deroneriksson:
Here is the original method:
```
	public static Dataset<Row> stringDataFrameToVectorDataFrame(SQLContext sqlContext, Dataset<Row> inputDF)
			throws DMLRuntimeException {

		StructField[] oldSchema = inputDF.schema().fields();
		//create the new schema
		StructField[] newSchema = new StructField[oldSchema.length];
		for(int i = 0; i < oldSchema.length; i++) {
			String colName = oldSchema[i].name();
			newSchema[i] = DataTypes.createStructField(colName, new VectorUDT(), true);
		}

		//converter
		class StringToVector implements Function<Tuple2<Row, Long>, Row> {
			private static final long serialVersionUID = -4733816995375745659L;
			@Override
			public Row call(Tuple2<Row, Long> arg0) throws Exception {
				Row oldRow = arg0._1;
				int oldNumCols = oldRow.length();
				if (oldNumCols > 1) {
					throw new DMLRuntimeException("The row must have at most one column");
				}

				// parse the various strings. i.e
				// ((1.2,4.3, 3.4))  or (1.2, 3.4, 2.2) or (1.2 3.4)
				// [[1.2,34.3, 1.2, 1.2]] or [1.2, 3.4] or [1.3 1.2]
				Object [] fields = new Object[oldNumCols];
				ArrayList<Object> fieldsArr = new ArrayList<Object>();
				for (int i = 0; i < oldRow.length(); i++) {
					Object ci=oldRow.get(i);
					if (ci instanceof String) {
						String cis = (String)ci;
						StringBuffer sb = new StringBuffer(cis.trim());
						for (int nid=0; i < 2; i++) { //remove two level nesting
							if ((sb.charAt(0) == '(' && sb.charAt(sb.length() - 1) == ')') ||
									(sb.charAt(0) == '[' && sb.charAt(sb.length() - 1) == ']')
									) {
								sb.deleteCharAt(0);
								sb.setLength(sb.length() - 1);
							}
						}
						//have the replace code
						String ncis = "[" + sb.toString().replaceAll(" *, *", ",") + "]";
						Vector v = Vectors.parse(ncis);
						fieldsArr.add(v);
					} else {
						throw new DMLRuntimeException("Only String is supported");
					}
				}
				Row row = RowFactory.create(fieldsArr.toArray());
				return row;
			}
		}

		//output DF
		JavaRDD<Row> newRows = inputDF.rdd().toJavaRDD().zipWithIndex().map(new StringToVector());
		// DataFrame outDF = sqlContext.createDataFrame(newRows, new StructType(newSchema)); //TODO investigate why it doesn't work
		Dataset<Row> outDF = sqlContext.createDataFrame(newRows.rdd(),
				DataTypes.createStructType(newSchema));

		return outDF;
	}
```

deroneriksson:
The extent of my changes were converting:
```
Vector v = Vectors.parse(ncis);
fieldsArr.add(v);
```
to
```
// [ ] will always result in double array
double[] doubles = (double[]) NumericParser.parse(ncis);
Vector dense = Vectors.dense(doubles);
fieldsArr.add(dense);
```

deroneriksson:
My goal here was not to improve the method but to restore the original behavior but using the ml.linalg.Vector import rather than mllib.linalg.Vector. I purposefully was avoiding additional changes to this method since I wanted to minimize potential impact to any user using this functionality.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
deroneriksson: This is the original implementation. I have not changed it. I would not use string concatenation here either since java strings are immutable and this is fairly wasteful.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -362,4 +354,68 @@ private void flushBufferToList( ReblockBuffer rbuff,  ArrayList<Tuple2<MatrixInd
...
+
+		// converter
+		class StringToVector implements Function<Tuple2<Row, Long>, Row> {
+			private static final long serialVersionUID = -4733816995375745659L;
+
+			@Override
+			public Row call(Tuple2<Row, Long> arg0) throws Exception {
+				Row oldRow = arg0._1;
+				int oldNumCols = oldRow.length();
+				if (oldNumCols > 1) {
```
deroneriksson: This is the original implementation.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -362,4 +354,68 @@ private void flushBufferToList( ReblockBuffer rbuff,  ArrayList<Tuple2<MatrixInd
 			ret.addAll(SparkUtils.fromIndexedMatrixBlock(rettmp));
 		}
 	}
+
+	public static Dataset<Row> stringDataFrameToVectorDataFrame(SQLContext sqlContext, Dataset<Row> inputDF)
+			throws DMLRuntimeException {
+
+		StructField[] oldSchema = inputDF.schema().fields();
```
deroneriksson: This is the original implementation.

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
aloknsingh: The idea is that since Vectors type of data is not exported in R they appear as the one String. So only one string is supported as the "[1, 3.4, 3, 123]" or varous sep and format (see the regexp)

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
aloknsingh: Hi @deroneriksson @frreiss ,
 Yes this won't be faster but it will solve the use cases and also this conversion will happen in parallel for each of partition of the DF and cost will be amortised . If there is the better solution we should be open but after some thinking I found that regexp would be more right. But we can only handle the special case i.e "[1, 132.34, 324,13]" .If this becomes issue too much for performance

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
aloknsingh: @frreiss Yes, I agree, in the orig version , we used the Vectors.parse which suppose to handle it. Also sometimes for example in hive , we do allow the bad records and it just ignore it. So what do you think we should follow? should we accept the bad record with the warn and move forward or error out even for one bad record?

src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
aloknsingh: @frreiss nice cache, yes this is the bug actually it is not checking for the null it is checking if it is the instance of string

deroneriksson: What is the behavior you would like? do you want a `if (ci == null)` to throw an error?

deroneriksson:
Added javadocs, null check, test cases, catch parse exception, etc. I'm going to merge this. Please let me know if any other modification are needed. Also feel free to submit PRs for any modifications needed.

deroneriksson:
BTW all tests passed here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1101/

PR #382, opened 2017-02-07
[SYSTEMML-1237] Add input float support to MLContext

deroneriksson:
Implicitly cast float to double in in(String, Object, Metadata) method
of Script class.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/Script.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/Script.java
@@ -324,6 +324,10 @@ public class Script {
 			// convert Long to Integer since Long not a supported value type
 			Long lng = (Long) value;
 			value = lng.intValue();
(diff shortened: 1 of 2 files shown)


deroneriksson:
Can test in spark shell with code such as:
```
$ spark-shell --executor-memory 4G --driver-memory 4G --driver-class-path=target/SystemML.jar

val ml = new MLContext(sc)
val scr = dml("c = a + b;")
val a:Float = 1
val b:Float = 2
scr.in("a", a).in("b", b).out("c")
print(ml.execute(scr).getDouble("c"))
```

niketanpansare:
LGTM :+1:

deroneriksson:
Thanks for reviewing @niketanpansare!

deroneriksson:
All tests passed here: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1084/

PR #391, opened 2017-02-14
[SYSTEMML-1259] Replace append with cbind for matrices

deroneriksson:
According to DML Language Ref:
"NOTE: append() has been replaced by cbind(), so its use is discouraged."


deroneriksson:
I think we should remove it from the matrix documentation. It can be used with strings, so it should be kept in that section of the documentation.

deroneriksson:
All tests passed: https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1113/

deroneriksson:
I removed append() from the matrix docs. I'll merge.

PR #395, opened 2017-02-16
[SYSTEMML-1279] Decrease numCols to prevent spark codegen issue

fschueler:
This should work around getting an exception from spark codegen.

### DIFF
--- a/docs/spark-mlcontext-programming-guide.md
+++ b/docs/spark-mlcontext-programming-guide.md
@@ -124,7 +124,7 @@ None
 
 ## DataFrame Example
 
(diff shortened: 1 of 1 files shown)


deroneriksson:
LGTM. The underlying issue isn't resolved but this does prevent the current warning which can be disconcerting to a new user. Thanks @fschueler!

deroneriksson:
Would you might updating line 127 (For demonstration purposes, we’ll use Spark to create a DataFrame called df of random doubles from 0 to 1 consisting of 10,000 rows and 1,000 columns.)?

deroneriksson:
And "For performance, we’ll specify metadata indicating that the matrix has 10,000 rows and 1,000 columns."

fschueler:
I updated the two lines - they didn't fit my previous search pattern ;)

deroneriksson:
Thanks @fschueler!

PR #396, opened 2017-02-16
[SYSTEMML-1280] Restore and deprecate SQLContext methods

deroneriksson:
Restore Java SQLContext method signatures (migrated to SparkSession by
SYSTEMML-1194) in case any users are using the old SystemML methods and
are unable to use SparkSessions.
Applies to old MLContext, MLMatrix, MLOutput, FrameRDDConverterUtils,
RDDConverterUtils, and RDDConverterUtilsExt.


bdwyer2:
Thanks!

PR #397, opened 2017-02-17
[SYSTEMML-1277] MLContext implicitly convert mllib Vector to ml Vector

deroneriksson:
Implicitly convert dataframe mllib.Vector columns to ml.Vector columns in
MLContext API.


deroneriksson:
cc @dusenberrymw

deroneriksson:
Great. Thanks for reviewing @dusenberrymw. Hopefully it works as well on the real-world data.

PR #399, opened 2017-02-17
[SYSTEMML-1246] Use correct jar name in sparkDML.sh of -bin artifact

gweidner:
Added resource filter to use correct jar name in sparkDML.sh of -bin artifact.  The pattern {project.artifactId}-${project.version} was used to handle both snapshot and release builds.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,6 +282,27 @@
 							<outputDirectory>${basedir}/target/lib/hadoop/bin</outputDirectory>
 						</configuration>
 					</execution>
+					
+					<execution>
+						<id>copy-resources-filtered</id>
+						<phase>compile</phase>
+						<goals>
+							<goal>copy-resources</goal>
+						</goals>
+						<configuration>
+							<overwrite>true</overwrite>
+							<outputDirectory>${basedir}/target/scripts</outputDirectory>
+							<resources>
+								<resource>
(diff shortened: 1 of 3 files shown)


gweidner:
Before this patch, the sparkDML.sh inside -bin referenced SystemML.jar.  For other artifacts where sparkDML.sh is bundled, e.g., SystemML.jar and systemml-0.13.0-incubating-SNAPSHOT.zip|tgz which includes SystemML.jar, the existing reference was correct and unchanged.  Note before this patch the systemml-0.13.0-incubating-SNAPSHOT.jar also contained the same sparkDML.sh as SystemML.jar so that value remains inconsistent but will need to be investigated in a separate JIRA.  This PR is targeted for the primary binary artifact referenced in download page.

deroneriksson:
LGTM. I built the distribution profile and verified systemml-0.13.0-incubating-SNAPSHOT-bin/scripts/sparkDML.sh contains the updated values for DEFAULT_SPARK_HOME.

In the future when there is more time, we might want to use maven to perform filtering (replacement) of substitution variables when copying the resources (search pom for "copy-project-info-for-python" and inspect "project_info.py" file) so that the values are dynamically pulled and substituted from the pom.xml file so that these sparkDML.sh version values will automatically be updated when the artifacts are built.

gweidner:

PR #401, opened 2017-02-20
[SYSTEMML-1266] Replace README.txt In Release Package

gweidner:
Replaced contents with readme.md as requested/directed.

### DIFF
--- a/src/main/standalone/README.txt
+++ b/src/main/standalone/README.txt
@@ -1,65 +1,138 @@
--------------------------------------------------------------------------------
-Apache SystemML (incubating)
--------------------------------------------------------------------------------
+﻿Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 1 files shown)


gweidner:
@asurve

deroneriksson:
LGTM. Thanks @gweidner!

When there is more time, we should consider changing src/main/standalone/* to src/main/resources/bin/*, or move it to src/main/resources/standalone/*. We can do string substitution variables with maven filtering to avoid hardcoding version numbers, since we can pull the version number from the pom.

gweidner:
Thank you @deroneriksson for reviewing!  That's a good idea regarding resource filtering.  I'll merge this PR but keep the JIRA open for that.

deroneriksson:
Cool! Thanks @gweidner

PR #405, opened 2017-02-24
[SYSTEMML-1347] SparkSession in Java/Scala MLContext API

deroneriksson:
Add MLContext constructor for SparkSession.
In MLContext, store SparkSession reference instead of SparkContext.
Remove unused monitoring parameter in MLContext.
Simplifications in MLContextUtil and MLContextConversionUtil.
Method for creating SparkSession in SparkExecutionContext.
Update tests for SparkSession.

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/context/SparkExecutionContext.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/context/SparkExecutionContext.java
@@ -33,11 +33,14 @@ import org.apache.spark.SparkConf;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.SparkSession.Builder;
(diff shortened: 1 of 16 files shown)


src/main/java/org/apache/sysml/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -260,12 +263,36 @@ public static SparkConf createSystemMLSparkConf() {
...
-	
+
+	/**
+	 * Create a SystemML-preferred Spark Session.
+	 * 
+	 * @param appName the application name
+	 * @param master the master value (ie, "local", etc)
+	 * @return
+	 */
+	public static SparkSession createSystemMLSparkSession(String appName, String master) {
```
dusenberrymw: Could we move this function into the `test` package?  Looks great for encapsulating a good default spark session for our test suites, but I'm afraid it could cause confusion later if it's left inside the runtime codebase.

deroneriksson:
Verified all tests still pass (https://sparktc.ibmcloud.com/jenkins/job/SystemML-OnDemand/277/console). I'll merge.

PR #407, opened 2017-02-27
[MINOR] Boolean hashcode for Java previous to 8

deroneriksson:
Boolean.hashCode(value) was added in Java 8. Can update to Boolean.valueOf(value).hashCode() for Java 7 (and 6).


deroneriksson:
cc @mboehm7

mboehm7:
LGTM - thanks for the fix @deroneriksson

PR #408, opened 2017-02-27
[SYSTEMML-1354] Move minimum recommended Spark version to pom.xml

deroneriksson:
Move MLContext minimum recommended Spark version to pom.xml to simplify project
maintenance. For SystemML jar, put value in manifest and read from there.
If no jar file present (development environment), read value from pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
 		<hadoop.version>2.6.0</hadoop.version>
 		<antlr.version>4.5.3</antlr.version>
 		<spark.version>2.1.0</spark.version>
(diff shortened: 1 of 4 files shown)


deroneriksson:
@dusenberrymw Sounds good. I'll make the update. If we find that a  `minimum.recommended.spark.version` is useful in the future (for instance, it could be `2.0.2` when `spark.version` is `2.1.0`), it is easy to modify the pom.xml to support it.

deroneriksson:
Thank you for the feedback @dusenberrymw. I will merge the updates.

PR #409, opened 2017-02-28
[SYSTEMML-1353] Initial Scala DSL implementation

fschueler:
This is the code for the initial Scala DSL implementation. Since it's huge, I split it up into logical commits that could potentially reviewed by different people. The changes to the `pom.xml` file include the maven snapshot repository for the Emma dependency and the Scala-reflect library.

In the MLContext API I replaced the `Matrix.java` with the `Matrix.scala`. I have not added all the transformations since I would like to discuss which transformations should really be exposed.

As for the DSL implementation I split it into two parts:

1) API - the user API for working with Matrices and built-in functions. 
2) Compiler - the compiler infrastructure to convert Scala ASTs to DML

Both include a set of tests but could probably be extended. The DSL still has some limitations which should be sufficiently described in the documentation that I added. 

I think before declaring it a non-experimental feature there should be some serious testing with actual usecases. This might be interesting for @dusenberrymw, @iyounus and others.


nakul02:
The code is extensively and very well documented!
Thank you!!!

deroneriksson:
@fschueler @frreiss @bertholdreinwald 
Hi, do we have any Scala experts who would be able to volunteer to review this PR?

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+          val body = indent(parts(1))
+
+          s"""
+             |for ($idx in $range) {
+             |$body
+             |}
+            """.stripMargin.trim
+        }
+
+        val escape = (str: String) => str
```
frreiss: You should escape backslashes first, not last. Otherwise you'll replace newline with `\\n` instead of `\n`

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+              /* matches apply methods with one argument */
+              case (Some(tgt), ((arg :: Nil) :: Nil)) if isApply(method) => {
+                val module = tgt(env)
+
+                if (module == "Vector") { // Vector.apply(Array(...))
+                  printConstructor(method, argss, env, true)
+                }
+
+                else {
+                  s"${tgt(env)}${printMethod(" ", method, " ")}${arg(env)}"
```
frreiss: Is this logic correct for every object the user might call `apply` on, or should there be some additional error checking code here?

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+
+                // binary operators
+                else {
+                  method.name.decodedName.toTermName match {
+                    case u.TermName("to")    => s"${tgt(env)} + 1:${arg(env)} + 1"
+                    case u.TermName("until") => s"${tgt(env)} + 1:${arg(env)}"
+                    case u.TermName("%") => s"($module %% ${args(0)})" // modulo in dml is %%
+                    case u.TermName("&&") => s"($module & ${args(0)})" // && in dml is &
+                    case u.TermName("||") => s"($module | ${args(0)})" // || in dml is |
+                    case _ => s"($module ${method.name.decodedName} ${args(0)})"
```
frreiss: Is this default cause guaranteed not to generate nonsensical DML, or should there be some additional error-checking code here?

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+                  "builtin"
+                }
+
+                  // apply as indexing - convert Scala 0-based to DML 1-based
+                else {
+                  val rows = args(0) // rows
+                  val cols = args(1) // columns
+
+                  (rows, cols) match {
+                    case (":::", c) if c.contains(":")  => s"$module[,$c]"
```
frreiss: I would have added parentheses around `$c` and `$r` in the format strings below, just in case. For example, `$c + 1` --> `($c) + 1`

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+                    case (":::", c)                     => s"$module[,$c + 1]"
+                    case (r, ":::") if r.contains(":")  => s"$module[$r,]"
+                    case (r, ":::")                     => s"$module[$r + 1,]"
+                    case (r, c) if  r.contains(":") &&
+                                    c.contains(":")     => s"$module[$r,$c]"
```
frreiss: Shouldn't this be `[$r + 1, $c + 1]`?

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+              case (Some(tgt), (x :: xs) :: Nil) if isUpdate(method) => {
+                val module = tgt(env)
+
+                // update on matrix objects (left indexing): A[r, c] = v === A.update(r, c, v)
+                val rows = args(0) // rows
+                val cols = args(1) // columns
+                val value = args(2) // value to update with
+
+                (rows, cols) match {
+                  case (":::", c) if c.contains(":")  => s"$module[,$c] = $value"
```
frreiss: Duplicate of logic in the case statement about 10 lines earlier. Consider breaking this logic out into a separate method.

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+                  printConstructor(method, argss, env, true)
+                }
+
+                else if (module == "package") {
+                  // builtin
+                  printBuiltin(tgt, method, argss, env)
+                }
+
+                else {
+                  "case (Some(tgt), (x :: xs) :: Nil)"
```
frreiss: This doesn't look like valid DML. Should there be an exception thrown here?

src/main/scala/org/apache/sysml/compiler/lang/source/DML.scala
```diff
@@ -0,0 +1,473 @@
...
+
+              // matches functions that are not defined in a module or class (udfs)
+              case (None, _) => {
+                val name = method.name.decodedName
+                val argString = args.mkString(", ")
+                s"$name($argString)"
+              }
+
+              case _ =>
+                abort(s"Unsupported function call! Calling the method $method is not supported!")
```
frreiss: Is it possible to add some location information about the original line of Scala code to this error?

src/main/scala/org/apache/sysml/compiler/macros/RewriteMacros.scala
```diff
@@ -0,0 +1,218 @@
...
+
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox
+
+
+class RewriteMacros(val c: blackbox.Context) extends MacroCompiler with DML {
+  import DML._
+  import u._
+
+  override val rootPkg = "org.apache.sysml.api"
```
frreiss: My copy of IntelliJ complains that this val doesn't override anything.

src/main/scala/org/apache/sysml/compiler/macros/RewriteMacros.scala
```diff
@@ -0,0 +1,218 @@
...
+        if (printDML) {
+          println("=" * 80)
+          println((" " * 26) + "RUNNING GENERATED DML SCRIPT")
+          println("=" * 80)
+          println(${formatted})
+          println("=" * 80)
+          println("Input parameters:" +  List(..${inParams}).mkString(", "))
+        }
+
+        val script = dml($dmlString).in(inputs).out(..${outParams})
```
frreiss: Maybe break this line out as a method of the superclass so that users can call it for debugging purposes.

src/test/java/org/apache/sysml/test/integration/functions/mlcontext/GNMFTest.java
```diff
@@ -67,12 +67,12 @@
...
 
 @RunWith(value = Parameterized.class)
-public class GNMFTest extends AutomatedTestBase 
+public class GNMFTest extends AutomatedTestBase
 {
 	private final static String TEST_DIR = "applications/gnmf/";
 	private final static String TEST_NAME = "GNMF";
 	private final static String TEST_CLASS_DIR = TEST_DIR + GNMFTest.class.getSimpleName() + "/";
-	
+
```
frreiss: Whitespace changes throughout this file might cause merge conflicts.

src/test/scala/org/apache/sysml/api/BaseAPISpec.scala
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.apache.sysml.api
+
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.{FreeSpec, Matchers}
+
+/**
+  * Specification for the SystemML API
```
frreiss: Please add a link to the appropriate parts of the scalatest documentation for the benefit of Java developers who are not familiar with `FreeSpec`.

src/test/scala/org/apache/sysml/api/linalg/ReplTest.scala
```diff
@@ -0,0 +1,149 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class ReplTest extends FreeSpec with Matchers {
+
+  private def scalaOptionError(msg: String): Unit = {
+    Console.err.println(msg)
+  }
+
+  val outputDir = Files.createTempDirectory("systemml_repl_test")
+  val jars = "" ///home/felix/repos/incubator-systemml/target/SystemML.jar"
```
frreiss: Hard-coded path should be relative to project root.

fschueler:
Thanks for reviewing @frreiss ! I will address your comments, rebase, and update the PR!

deroneriksson:
ping @fschueler @frreiss @asurve 
It sounds like the SystemML 1.0.0 release is rapidly approaching. Will it be possible to make the updates and merge for the 1.0.0 release?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1744/

fschueler:
I rebased and started looking into it a while ago - let me see what I can do over the weekend. What's the release plan for 1.0?

asurve:
Thanks Felix for looking into this one.
We were planning to do release candidate build this week. Based on response so far, build may be delayed by two weeks, so you have two more weeks to get verified changes in.Hope that helps.
-Arvind
Arvind Surve | Spark Technology Center  | http://www.spark.tc/


On Thursday, July 13, 2017, 1:50:30 AM PDT, Felix Schüler <notifications@github.com> wrote:


I rebased and started looking into it a while ago - let me see what I can do over the weekend. What's the release plan for 1.0?

j143:
Hi Felix, If you like, I will take up the conflicts and `MLContext` related changeset and will address Fred's comments.
cc @fschueler

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2280/

mboehm7:
Thanks but closing as outdated.

PR #410, opened 2017-02-28
[SYSTEMML-1282] Remove avro from bin and standalone jar artifacts

deroneriksson:
Remove avro libraries from -bin.tgz, -bin.zip, and -standalone.jar
artifacts to simplify project build dependency management and
reduce artifact size. Update corresponding LICENSE files.

### DIFF
--- a/src/assembly/bin.xml
+++ b/src/assembly/bin.xml
@@ -114,7 +114,6 @@
 		<dependencySet>
 			<includes>
 				<include>*:${artifactId}*</include>
-				<include>*:avro*</include>
 				<include>*:commons-cli*</include>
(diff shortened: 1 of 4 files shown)


deroneriksson:
Note that it can be difficult to track down whether dependencies such as these are required or not. If these dependencies are indeed needed, most likely NoClassDefFoundErrors will occur under certain situations.

deroneriksson:
I am going to go ahead and merge this PR since it can be difficult to hit the correct situations where dependencies such as these are needed. This way, the standalone jar and bin artifacts can be tested for a significant amount of time before the 1.0.0 release to see if these artifacts require these dependencies.

PR #411, opened 2017-03-01
[SYSTEMML-259] Function with no return value not require lvalue

deroneriksson:
If a user-defined function does not return a value, don't require
that the function is assigned to a variable since there is nothing
to assign. Add corresponding MLContext tests.


nakul02:
This program passes and gives no error, even though x is assigned nothing:

```
hello=function(){
    print('no return value')
}
x = hello()
```

Could you please check for this condition, i.e. check for the number of arguments in the function definition and match them to where they're being called?

deroneriksson:
Added warning messages if a function call does not return a value but is assigned to a variable or variables. I think a warning at this stage is better than an error since before this PR, a LHS value was a requirement when calling a function even if it did not return a value.

deroneriksson:
Thank you for reviewing @dusenberrymw and @nakul02! I'll merge.

PR #412, opened 2017-03-01
[SYSTEMML-1359][WIP] Gpu extra instrumentation

nakul02:
This is an initial pass at this. I will try to improve the interface of the functions so that the performance measurement code will not be all over the place. Doing something simple like the decorator pattern in python to measure the time a function or in scala with their call-by-name would have been the more elegant thing, but alas...

@niketanpansare, @bertholdreinwald


nakul02:
Example output:

```
Total elapsed time:		1816.851 sec.
Total compilation time:		1.819 sec.
Total execution time:		1815.032 sec.
Number of compiled Spark inst:	25.
CUDA/CuLibraries init time:	0.000/0.000 sec.
Number of executed GPU inst:	2877263.
GPU mem tx time  (alloc/dealloc/toDev/fromDev):	253.660/7.258/21.617/1.331 sec.
GPU mem tx count (alloc/dealloc/toDev/fromDev/evict):	2952022/2877268/1/37384/15/1.
GPU conversion time  (sparseConv/sp2dense/dense2sp):	4.761/0.005/0.183 sec.
GPU conversion count (sparseConv/sp2dense/dense2sp):	1/4/1.
Cache hits (Mem, WB, FS, HDFS):	112131/0/0/1.
Cache writes (WB, FS, HDFS):	37379/0/8.
Cache times (ACQr/m, RLS, EXP):	2.204/0.017/0.926/0.115 sec.
HOP DAGs recompiled (PRED, SB):	0/37367.
HOP DAGs recompile time:	54.006 sec.
Spark ctx create time (lazy):	0.912 sec.
Spark trans times (par,bc,col):	0.000/0.000/0.000 secs.
Total JIT compile time:		23.418 sec.
Total JVM GC count:		30.
Total JVM GC time:		47.37 sec.
-- 1) 	gpu_+ 	329.886 sec 	597865	ddgeaml[2.409s,149468], mmck[16.088s,298928], ao[300.328s,597865], H2D[0.000s,1], msk[6.114s,149469]
-- 2) 	gpu_* 	320.583 sec 	597873	ddgeaml[7.088s,448401], ao[290.769s,597873], msk[0.000s,4], H2D[0.002s,7], s2d[0.005s,4], mmck[14.217s,149468]
-- 3) 	gpu_ba+* 	292.030 sec 	411038	Mdmdm[7.580s,411037], ao[218.231s,411037], a[218.578s,411037], H2D[21.375s,37372], f[0.000s,1], dtl[0.003s,1], Msao[0.090s,1], d2s[0.180s,1], Msmsm[0.639s,1]
-- 4) 	gpu_- 	207.446 sec 	336305	msk[12.538s,298936], mmck[0.378s,2], ddgeaml[0.615s,37367], ao[191.496s,336305]
-- 5) 	gpu_-* 	157.722 sec 	298936	D2D[3.717s,298936], daxpy[3.967s,298936], ao[145.747s,298936]
-- 6) 	gpu_/ 	100.870 sec 	149472	ao[85.645s,149472], ddgeaml[0.000s,3], mmck[14.387s,149469]
-- 7) 	gpu_r' 	95.198 sec 	149468	ddgeaml[2.707s,149468], ao[88.895s,149468]
-- 8) 	gpu_exp 	88.521 sec 	149468	ao[79.928s,149468], expk[7.774s,149468]
-- 9) 	gpu_^2 	85.970 sec 	149469	ao[76.051s,149469], msk[9.048s,149469]
-- 10) 	gpu_uasqk+ 	35.690 sec 	37367	rallk[3.347s,74734], f[0.166s,74734], a[30.626s,74734], D2H[0.988s,37367], msk[1.336s,37367]
-- 11) 	rmvar 	10.888 sec 	4446694
-- 12) 	rexpand 	8.573 sec 	1
-- 13) 	gpu_uack+ 	6.087 sec 	1	ao[0.001s,1], rcolk[4.667s,1], H2D[1.419s,1]
-- 14) 	rangeReIndex 	4.714 sec 	37367
-- 15) 	gpu_uacsqk+ 	4.673 sec 	1	ao[0.000s,1], f[0.001s,1], a[0.002s,1], msk[0.067s,1], rcolk[4.602s,1]
-- 16) 	createvar 	3.462 sec 	2877279
-- 17) 	rsort 	1.778 sec 	1
-- 18) 	rand 	0.641 sec 	12
-- 19) 	print 	0.633 sec 	37368
-- 20) 	+ 	0.403 sec 	373669
-- 21) 	cpvar 	0.121 sec 	597890
-- 22) 	write 	0.116 sec 	8
-- 23) 	/ 	0.056 sec 	37367
-- 24) 	* 	0.055 sec 	37368
-- 25) 	nrow 	0.054 sec 	37367
-- 26) 	== 	0.042 sec 	112101
-- 27) 	assignvar 	0.033 sec 	112116
-- 28) 	%% 	0.028 sec 	74734
-- 29) 	&& 	0.017 sec 	37367
-- 30) 	< 	0.011 sec 	37368
-- 31) 	- 	0.010 sec 	37367
-- 32) 	> 	0.010 sec 	37367
-- 33) 	sqrt 	0.007 sec 	1
Statistics collection overhead: 5.339211545 sec
```

nakul02:
The meanings of all the weird looking symbols is specified [here](https://github.com/nakul02/incubator-systemml/blob/c0a6e1fb39781bd11230efeb13218166b4f877fb/src/main/java/org/apache/sysml/runtime/instructions/gpu/GPUInstruction.java#L38)

mboehm7:
it's certainly useful for local debugging but I'd recommend not to push this to master because these stats are supposed to be a clean summary.

nakul02:
I was thinking or providing an option to stats to print "advanced stats", thoughts?

mboehm7:
1) In general, I'm open to discuss "multiple levels" for the stats tool but that would require a holistic (i.e., system-wide) design. In any case, I really would like to retain concise summary statistics because a dump of every measurement quickly gets overwhelming, rendering it counter-productive. How about you create a separate class GPUStatistics similar to CacheStatistics for now and use it just for local debugging. Once, we settle on the right abstraction for the user-facing stats output we could simply embed this output. Furthermore, please double check the existing GPU summary statistics - the time/count outputs are supposed to be aligned with the remaining output.

2) Please, note that your current approach of measuring the overhead of statistics collection returns misleading numbers. The two hash table inserts are by far cheaper than getting the time with System.nanoTime(). Of course if depends on the hardware platform, but it can be even micro seconds. Your best estimate of getting the measurement overhead is to run your script twice - with and without stats - and compare the total execution times.

nakul02:
Thanks for the comments @mboehm7.

**_W.R.T 1._**
In general, I think we can both agree that more granular performance counters is a good thing to track down performance problems.

For the current statistics that are printed out, most of it can only be understood by someone familiar with the opcodes used for our instructions and what cache acquire and release mean. Same for most other things reported by statistics. 
This leads to me believe that the statistics aren't meant for a so-called **"end-user"**.

Keeping this in mind, I feel that with this PR, I need to care about the format of statistics only for developers of SystemML. In which case, I should ask the developers of SystemML what format they would like to see for these additional timers. Can you suggest how I should print these additional timers?

CacheStatistics provides specific counters and timers. The code I added to Statistics.java is really minimal. It adds [two](https://github.com/nakul02/incubator-systemml/blob/1b13587a5e5ca4e8f3bbe9670f8f9b3dda00b336/src/main/java/org/apache/sysml/utils/Statistics.java#L109-L110) HashMaps and a little bit of code to "[maintain](https://github.com/nakul02/incubator-systemml/blob/1b13587a5e5ca4e8f3bbe9670f8f9b3dda00b336/src/main/java/org/apache/sysml/utils/Statistics.java#L553-L597)" them. Also, it makes logical sense to have them in close proximity, code-wise to the [heavy-hitters maps](https://github.com/nakul02/incubator-systemml/blob/1b13587a5e5ca4e8f3bbe9670f8f9b3dda00b336/src/main/java/org/apache/sysml/utils/Statistics.java#L107-L108).

Also, could you explain, what you mean when you say "local debugging"? Do you mean, only for the purposes of my current development?


**W.R.T. 2**
"The two hash table inserts are by far cheaper than getting the time with System.nanoTime()." - This runs counter-intuitive to common sense. Could you please provide a reference?

mboehm7:
ad 1) yes, I mean for your current and future development. In fact, there are many users, not just developers, who use this stats tool. So, if we expose statistics through -stats, we have to hold us to a much higher standard than for dev profiling only.
 
ad 2) https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks

nakul02:
**WRT 1)**
Ok, I was thinking of doing this:
- Introduce a GPUStatistics class 
    - move all GPU related statistics there
- Introduce a "-devstats" option and print these advanced statistics only when that is enabled.
    - we could follow this with a key value pair, for instance, "GPU=true", "DNN=true" and so on, if you think that is a good idea, otherwise just print all the advanced stats.

**WRT 2)**
This is surprising, but I read this article: https://shipilev.net/blog/2014/nanotrusting-nanotime/ and he provides performance considerations when using `System.nanoTime()`. Basically, each call can take between `15-30ns` on Linux platforms, and upto `15μs` on a particularly bad case on Windows Server. The `get` on a HashMap [takes](https://dzone.com/articles/how-tinylog-12-will-simplify-your-logging) about `8-10ns` for the size of maps we are working with.
You are right, so then the overhead of statistics collection does return misleading numbers, I shall remove them.

mboehm7:
ad 1) What we usually do is to integrate such integration with certain log levels, usually DEBUG or TRACE depending on the granularity of outputs. This allows people to easily control the information of specific packages from outside of SytemML in a consistent manner. Furthermore, some components have local debug flags; if enabled they set their own output to DEBUG level. For example, please have a look at org.apache.sysml.hops.rewrite.ProgramRewriter.

nakul02:
Unfortunately, these detailed statistics do not semantically fall into the same category as either the `DEBUG` or `TRACE` log levels. Given that trace and debug are going to spit out a massive amount of output, I would like to treat these statistics (and others like it) separately.
These are the log4j logging levels, `ALL < TRACE < DEBUG < INFO < WARN < ERROR < FATAL < OFF`. Unfortunately, there are no other levels for me to fit the advanced statistics into.

In ProgramRewriter, it turns out that the so-called "local debug flag" is a [compile time constant](https://github.com/nakul02/incubator-systemml/blob/082a743dbb366b807067b5bdf1579ecdd3ced3ab/src/main/java/org/apache/sysml/hops/rewrite/ProgramRewriter.java#L57). This also seems to be the case for [LibMatrixDNN](https://github.com/nakul02/incubator-systemml/blob/9c31c2ca7624c61357894ec463223c507165d1d8/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java#L53).

Is adding a `-devstats` a bad idea?

nakul02:
@mboehm7 - I've updated the PR and refactored all GPU statistics to `GPUStatistics.java`.
Also the `-devstats` option has been introduced.
@niketanpansare, @bertholdreinwald - does this seem like a good idea?

mboehm7:
yes, these LDEBUG flags are compile-time to allow for quick testing in your local dev environment. However, the log levels can also be easily configured from outside by modifying the log4j.properties file in your classpath. Note that this configuration allows a fine granularity (e.g., different log levels for different packages, so you would simply set DEBUG for the gpu.context package).

I personally don't think it's a good idea to expose this dump of statistics as shown above through a command line option. Instead I would recommend to think about concise summary statistics (to be integrated in -stats) and keep the remaining detailed profiling internal.  However, I'll not stand in your way if you guys want to go ahead with this.

niketanpansare:
@nakul02  I think adding `devstats` is useful for performance debugging. The key question is how to expose it to the user. I would be in favor of treating devstats as additional statistics printed after statistics, rather than modifying the existing statistics output. Regarding how to enable it, my preference is in the following order:
1. Turned on by a boolean flag in GPUStatistics. Drawback: Requires source code change while performance debugging. 
2. Turned on by -stats and if LOG level is debug. We then can print additional devstats after Statistics.display() with LOG.debug. Drawback: Other debug outputs. Advantage: No additional flag.
3. `-devstats` or via Configuration property.

niketanpansare:
If you choose option 1, I am OK with the current output as it is not exposed to the regular user. Also, I would like to put 'Configuration property' option after option 1 as it is how we enable WIP features.

niketanpansare:
@nakul02 As I said earlier, I like the idea of having `devstats` for performance debugging. My worry is that what this PR attempts to introduce is just `gpu-devstats` (or at best `gpu+dnn-devstats`). If you (or other developers) are interested in generalizing this concept to other components, I am all up for a single `devstats` flag. Since there are no "free lunches", introducing an additional command-line flag (especially if intended only for development for only one component) becomes one more thing an end user has to be educated about ... also, adding a command-line flag has to be accompanied with supporting that feature via other APIs (such as MLContext) as well. If not used by anyone, it eventually becomes one more thing that is forgotten :)

Configuration property has less impact in that regards, as the assumption is that most end-users will use default configuration and only the expert users will attempt to modify the configuration. 

At high-level, we need to ask ourselves two questions:
1. Is the additional info we are providing summarized in statistics (in case of LibDNN it is) ?
2. Will anyone except the engine developer care about the additional info ?

If the answer is yes and no respectively, it is best to fall back to a boolean flag option. If unsure about question 2, the answer almost always is no :)

nakul02:
@niketanpansare - I am very much in favor of either a `-devstats` or `-gpudevstats` and/or a configuration property. In "generalizing this concept to other components", we can either wait until we have everything properly instrumented and then enable `-devstats` or add in the instrumentation incrementally, one component at a time. That is really up to the stakeholders of the project. I for one am very much for the incremental way. If its the everything-instrumented way is preferred, we can move it to a system config property only for now.

I'll have to disagree with your high-level questions and their answers, specifically - "Will anyone except the engine developer care about the additional info".We have a `-debug` [option](https://github.com/nakul02/incubator-systemml/blob/bbefe96b263f697eb3f4e0297379840930356c0d/src/main/java/org/apache/sysml/api/DMLScript.java#L293-L293). This is a flag clearly intended for the developers, much more so than a `-devstats` option. Yet it is user-facing. When you say that an end-user needs to be educated about it, what we do is print it out with the help message in usage. If i've haven't misunderstood you, that is all the education we'd need to do. 
Other counter-examples of something only the engine developer would care about is the different kinds of `-explain` (`hops`, `runtime`, `recompile_hops` & `recompile_runtime`). In my opinion, if a end-user is smart enough to understand these various kinds of "explain" dumps, an extra set of numbers printed by `-devstats` or its ilk should be a no-brainer. :)

In any case, I also understand that the `-explain` and `-debug` might be legacy options that have never been revisited under the set of high level questions that you pose and might just be technical debt that we need to clear up.
All I wanted to point out was that there is precedent, and adding a `-devstats` option is nothing extremely out of the ordinary.

niketanpansare:
@nakul02 I think you misunderstood what `-debug` flag refers to. It enables a GNU-style debugger that DML writer can use for stepping through their DML script (though I have not played with it for about a year :( ).

nakul02:
@mboehm7, @niketanpansare - i've removed the `-devstats` option and made the extra statistics available through a property in the config file.

niketanpansare:
LGTM. Thanks @nakul02 :+1:

nakul02:
Thanks @niketanpansare, [tests passed](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1190/), I shall merge.

PR #413, opened 2017-03-04
[SYSTEMML-1370] More robust conversion of numpy array to matrixblock by passing multiple blocks

niketanpansare:
Here is the code to test this functionality:

```python
from systemml import MLContext, dml, convertToMatrixBlock
import pandas as pd
nr = 46900
X_pd = pd.DataFrame(range(1,
(nr*784)+1,1),dtype=float).values.reshape(nr,784)
# Sending 8MB blocks at a time   ... succeeds
convertToMatrixBlock(sc, X_pd)
# Sending entire block at a time ... fails
convertToMatrixBlock(sc, X_pd, maxSizeBlockInMB=100000)
```

Drawback of multiple blocks approach: 
- We create two copies of the data (one for smaller blocks passed through Py4J and other for final output).

Advantage:
- More robust. Can handle large number of rows.

TODO: 
- Process multiple blocks in parallel. This requires investigation of whether to use PySpark's parallelism or create threads (and likely sockets) on the driver.
- Use squared blocks for extremely wide matrices.

@bertholdreinwald @dusenberrymw @iyounus

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
@@ -165,6 +165,37 @@ public class RDDConverterUtilsExt
 		return convertPy4JArrayToMB(data, (int) rlen, (int) clen, isSparse);
 	}
 	
+	public static MatrixBlock mergeRowBlocks(ArrayList<MatrixBlock> mb, int numRowsPerBlock, int rlen, int clen, boolean isSparse) throws DMLRuntimeException {
+		return mergeRowBlocks(mb, (long)numRowsPerBlock, (long)rlen, (long)clen, isSparse);
+	}
+	
+	/**
+	 * This creates a MatrixBlock from list of row blocks
+	 * 
+	 * @param mb list of row blocks
+	 * @param numRowsPerBlock number of rows per block
+	 * @param rlen number of rows
+	 * @param clen number of columns
+	 * @param isSparse is the output matrix in sparse format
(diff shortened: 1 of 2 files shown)


dusenberrymw:
Would it be better to just have our [`_py2java ` function](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mlcontext.py#L121) convert Pandas DataFrames to PySpark DataFrames using Spark's existing logic (i.e. `spark.createDataFrame(my_pandas_df)`), and then pass those into the engine?  That way, we wouldn't have to worry about memory limits, and we could reduce all this redundant logic and simply reuse our internal DataFrame conversion code.  We could do the same for 2D NumPy arrays.

niketanpansare:
I have not benchmarked the `numpy -> pandas_df -> spark df (which internally will have parallelize) ( in our systemml algo ---> reblock/collect -> convert to mb if matrix fit in cp` as opposed to `numpy -> mb` or `numpy -> blocked numpy -> blocked mb -> mb`. But, I would suspect the performance of latter would be much better. You can think former as passing `csv` data (with additional parallelize/collect calls) and latter as passing `blocked binary data`.

niketanpansare:
I would like to merge this PR as it makes the current approach more robust. We can explore alternative approaches in different PR. If you have objections, please let me know.

bertholdreinwald:
Having this direct, fast pass conversion from numpy.ndarray or Pandas DF to matrixblock is definitely performance beneficial compared to even only from numpy.ndarray to Spark data frames.

Especially considering use cases of apps/tools having small-ish/medium-ish data loaded into a numpy structure, and want to quickly hand it over to SystemML.

LGTM.

PR #414, opened 2017-03-05
[WIP][SYSTEMML-298] Print matrices without as.scalar or toString

deroneriksson:
This is a work-in-progress to potentially simplify DML for beginners by removing the need when printing to cast single cells using as.scalar (`print(as.scalar(A[1,1])`) or matrices using toString (`print(toString(A)`). This would allow DML such as `print(A[1,1])` and `print(A)`.

This PR handles the following situations:
```
A = matrix("1 2 3 4 5 6", rows=2, cols=3);
print(as.scalar(A[1,1]));
print(A[1,2]);
print("one two " + A[1,3]);
print(A[2,1] + " five six");
print("four " + A[2,2] + " six");
print(A[1,1] + " two " + A[1,3]);
B = A[2,3];
print(B);
print("five " + B);
print(B + " seven");
print("five " + B + " seven");
C=A[2,1];
print(C+' five '+B);
print(A);
print("$$$" + A);
print(A + "!!!");
print("@@@" + A + "%%%");
D=matrix("1 2 3 4", rows=2, cols=2);
print(D + "***" + A);
print(A[1:2,2:3]);
print("^^^" + A[1:2,2:3]);
print(A[1:2,2:3] + "###");
print("[[[" + A[1:1,1:3] + "]]]");
```

This PR does not handle all potential situations, such as `print(A+B)` where A and B are matrices. However, it is potentially a nice starting point for integrating some of these features or at least for discussion.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/rewrite/RewriteAlgebraicSimplificationStatic.java
+++ b/src/main/java/org/apache/sysml/hops/rewrite/RewriteAlgebraicSimplificationStatic.java
@@ -258,8 +258,10 @@ public class RewriteAlgebraicSimplificationStatic extends HopRewriteRule
 			Hop left = bop.getInput().get(0);
 			Hop right = bop.getInput().get(1);
 			//X/1 or X*1 -> X 
-			if(    left.getDataType()==DataType.MATRIX 
-				&& right instanceof LiteralOp && ((LiteralOp)right).getDoubleValue()==1.0 )
+			if(left.getDataType()==DataType.MATRIX 
+				&& right instanceof LiteralOp
+				&& (right.getValueType() != ValueType.STRING)
+				&& ((LiteralOp)right).getDoubleValue()==1.0 )
 			{
 				if( bop.getOp()==OpOp2.DIV || bop.getOp()==OpOp2.MULT )
 				{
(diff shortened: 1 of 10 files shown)


mboehm7:
thanks for pushing this forward @deroneriksson. Before we discuss the details of the PR, you might want to consider an alternative via a simple rewrite. You could introduce a new rewrite (subclass of HopRewriteRule) and simply inject an as.scalar cast (for 1x1 matrices) or toString (for nxm matrices) whenever you encounter an u(print) with a matrix input. This would be very clean as the entire logic would reside in this rewrite without the need to touch the runtime.

mboehm7:
just to add to my previous comment: to enable something like "@@@" + A + "%%%", we would also need to consider b(+) with matrix and scalar-string inputs.

deroneriksson:
@mboehm7 Thank you for the great feedback. I suspected that injecting as.scalar or toString in the proper locations would be a cleaner solution, but I wasn't sure about the right mechanism to do so, so I implemented the messier solution in this PR. I'll take a look at creating a rewrite by subclassing HopRewriteRule, since a simpler solutions sounds like a great idea.

deroneriksson:
I'm going to close this PR since it sounds like it took the wrong approach to printing matrices.

PR #418, opened 2017-03-07
[SYSTEMML-485] Rename unusual file extensions

deroneriksson:
Rename unusual file extensions (.Rt -> .R, .dmlt -> .dml, .data -> .csv, etc.) to help standardize file naming in project.


mboehm7:
Could you please delete the L2SVM files? The dmlt and Rt extensions are generally used to indicate temporary files - for algorithm testcases, we do not generate them, so this must be a relict from a long time ago. 

That reminds me, down the road, we should remove the entire "old codepath" from our testsuite. There are still various tests that use something like $$op1$$ + $$op2$$, replace these with input arguments, write the temporary dmlt script, and finally execute the script. This stems from a time when we did not have any input arguments (>5 years). Maybe we find a good way to automatically replace these tests.

deroneriksson:
Sounds good. I deleted the L2SVMTest.Rt and L2SVMTest.dmlt and also added a needed update to WriteCSVTest.dml.

mboehm7:
LGTM - thanks @deroneriksson

PR #426, opened 2017-03-13
[SYSTEMML-1396] Lazy cudafree

nakul02:
Lazily frees cuda memory. Reuses memory already allocation on the GPU as much as possible.

Caveat - It is possible to run into the issue of internal/external memory fragmentation.
cudaMemGetInfo returns the total available free memory. That reported memory may not be contiguous to satisfy a request for an allocation though. No successful experiments have run into this issue yet, but a bug that I had introduced in my initial coding showed me this issue.

Also, with 2 use cases that I ran - @prithvirajsen's  [Autoencoder](https://github.com/apache/incubator-systemml/pull/384) and @dusenberrymw's lennet script with some small sample data i was provided, showed between a **`5x-10x`** speedup.

The PR also adds some advanced timers and adds certain bugfixes.
@niketanpansare - can you please verify if [this](https://github.com/apache/incubator-systemml/commit/02ae4ed667849ee006b1320b451e89daa9c7cb16) bug fix is legit?

ping @bertholdreinwald


niketanpansare:
LGTM. This is an important performance update. We can revisit the fragmentation later. Thanks @nakul02  :+1:

nakul02:
Thanks @niketanpansare, will merge

PR #430, opened 2017-03-14
[SYSTEMML-1400] Include pandas in required python packages

gweidner:
Updated setup.py to also include pandas for pip install so 'import systemml' is successful since dependency in converters.py.


gweidner:
Note a minimum version was not included in the requirement specifier.  Observed version 0.19.2 installed in RHEL 7.2 after this change.

deroneriksson:
LGTM. Thank you @gweidner

gweidner:

deroneriksson:
Thank you for fixing this @gweidner. I have run into the same issue.

PR #432, opened 2017-03-17
[SYSTEMML-1403] Fixed the bug in GPU relu_maxpooling and also cleaned up LibMatrixCUDA code.

niketanpansare:
@nakul02


nakul02:
I've suggested a tiny change via [this](https://github.com/niketanpansare/incubator-systemml/pull/10) PR. It fixes the bug we were running into with `reluMaxpooling` in a slightly better way.

Other than that LGTM.

PR #433, opened 2017-03-17
[SYSTEMML-1411] Add CPU and GPU bias multiply operation

niketanpansare:
This operation is similar to bias_add, except that it performs element-wise multiplication rather than addition. It avoids unnecessary multiplication of ones and reshape. This pattern is common in deep learning layers such as batch normalization:
```python
out = matrix(..., rows=N, cols=C*H*W)
gamma = matrix(... rows=C, cols=1)
ones = matrix(1, rows=1, cols=H*W)
out_1 = out*matrix(gamma %*% ones, rows=1, cols=C*H*W)
OR 
out_1 = bias_multiply(out, gamma)
```

@nakul02 @dusenberrymw @bertholdreinwald


nakul02:
LGTM 👍

dusenberrymw:
Yeah this is good.  In general though, this and `bias_add` really show that we need to seriously consider adding tensors.  We don't need to support full out tensor multiplication (matrix multiplication is just fine), but it would be super helpful for these kinds of tasks.

niketanpansare:
I agree. tensor support will be extremely useful.

nakul02:
I agree too. General tensors would be a useful data structure to have at the DML level. We will probably be limited by the number of elements supported in a tensor because of the limitations of Java (integer addressable memory) in standalone mode, but we should be able to get around that in spark mode.

PR #435, opened 2017-03-18
[SYSTEMML-1103] Use Apache Commons CLI to parse command line arguments in DMLScript

nakul02:
Use of Apache Commons CLI to parse arguments passed to DMLScript.
This gives us the flexibility that Commons CLI has, to include a "long options name", specify a fixed number of arguments and so forth.
Please note that no additional dependency is added in the pom file.

@deroneriksson - do you think this is a good idea? could you try playing with it a little?
Ping - @niketanpansare, @mboehm7


deroneriksson:
@nakul02 Using commons cli to parse the command line arguments seems like a great idea to me. I'm going to be away from a computer until Tuesday or Wednesday but I'd be happy to try it out then. Please note that the javadoc comments need some updating (https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1240/console).

As an aside to @mboehm7 and @dusenberrymw regarding our discussion at https://issues.apache.org/jira/browse/SYSTEMML-1406, the 'files changed' view at https://github.com/apache/incubator-systemml/pull/435/files#diff-3739eb2a584f6704c8e77098d7d39bea is an example of why standardizing to spaces rather than tabs can improve readability on github. We could always use a predefined style such as https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml

dusenberrymw:
@deroneriksson, @mboehm7 Yes, this is a fantastic example of why spaces would be preferable.  The Google Java style is great too with its 2-space indentation and 4-space continuation indentation.

nakul02:
This PR keeps existing CLI options the same. The only change that is bought about is the way `-config` is specified.
Previously, to specify a config file, you'd say `-config=SystemML-config.xml`. This PR changes it to 
`-config SystemML-config.xml`. (Note the `=` instead of the ` `).

Scripts & tests have been updated to reflect this. 
If any of you have existing scripts on your machines that specify the old `-config=` format, please change it after the PR is merged in.

Additionally, now that Apache Commons CLI is used, the arguments can be specified in any order. The dml file need not be the first and the `nvargs`/`args` need not be the last. 

@deroneriksson, @asurve & @dusenberrymw - Could you please review this?

## What's wrong with the current options passing mechanism?
I also think that we should standardize how we pass in CLI options to SystemML. Right now we have the [`Ant`](https://ant.apache.org/manual/running.html) way of passing arguments, i.e. `-<option> <value1> <value2>` or some weird variation of it.
`-f` is a short option name, there is not corresponding long option name. Something like `--file`
`-config=` is a weird way of accepting arguments, when all the other options have a space as a character to separate the option and the value.

Typically with GNU like long options, you'd have a single dash and a single character short option and a double-dash and a longer option. This would be nice to have. 
We can discuss this here. If no one is really affected by it or has an opinion one way or another, could this please be merged in after review?

Ping @bertholdreinwald, @niketanpansare, @frreiss

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -102,66 +109,58 @@
...
+		public DMLScript.RUNTIME_PLATFORM execMode = OptimizerUtils.getDefaultExecutionMode();	// Execution mode standalone, MR, Spark or a hybrid
+		public boolean              gpu           = false;						// Whether to use the GPU
+		public boolean              forceGPU      = false;						// Whether to ignore memory & estimates and always use the GPU
+		public boolean              debug         = false;						// to go into debug mode to be able to step through a program
+		public ScriptType           scriptType    = ScriptType.DML;		// whether the script is a DML or PyDML script
+		public String               filePath 			= null;							// path to script
+		public String 							script				= null;							// the script itself
+		public boolean              help          = false;						// whether to print the usage option
+
+		public final static DMLOptions defaultOptions = new DMLOptions();
```
dusenberrymw: Can you fix the spacing in this section?  It would be better to just use spaces to align everything.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -102,66 +109,58 @@
...
-	public static boolean ENABLE_DEBUG_MODE = false; //default debug mode
-	public static boolean USE_LOCAL_SPARK_CONFIG = false; //set default local spark configuration - used for local testing
-	public static String DML_FILE_PATH_ANTLR_PARSER = null;
-	public static ExplainType EXPLAIN = ExplainType.NONE; //default explain
+	public static RUNTIME_PLATFORM 	rtplatform = DMLOptions.defaultOptions.execMode;							// the execution mode
+	public static boolean						STATISTICS = DMLOptions.defaultOptions.stats;									// whether to print statistics
+	public static int 							STATISTICS_COUNT = DMLOptions.defaultOptions.statsCount;			// statistics maximum heavy hitter count
+	public static boolean						ENABLE_DEBUG_MODE = DMLOptions.defaultOptions.debug; 					// debug mode
+	public static ExplainType 			EXPLAIN = DMLOptions.defaultOptions.explainType; 							// explain type
+	public static String						DML_FILE_PATH_ANTLR_PARSER = DMLOptions.defaultOptions.filePath;	// filename of dml/pydml script
```
dusenberrymw: Can you fix the spacing in this section?  It would be better to just use spaces to align everything.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -102,66 +109,58 @@
...
-	public static boolean FORCE_ACCELERATOR = false;
-	
-	// flag that indicates whether or not to suppress any prints to stdout
-	public static boolean _suppressPrint2Stdout = false;
-	
-	public static String _uuid = IDHandler.createDistributedUniqueID(); 
+	public static ScriptType 				SCRIPT_TYPE = DMLOptions.defaultOptions.scriptType;
+
+	public static boolean						USE_ACCELERATOR = DMLOptions.defaultOptions.gpu;
+	public static boolean 					FORCE_ACCELERATOR = DMLOptions.defaultOptions.forceGPU;
```
dusenberrymw: Can you fix the spacing in this section?  It would be better to just use spaces to align everything.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -230,127 +439,81 @@ public static void main(String[] args)
...
+			// String[] scriptArgs = null; //optional script arguments
+			// boolean namedScriptArgs = false;
+
+			STATISTICS 				= dmlOptions.stats;
+			STATISTICS_COUNT 	= dmlOptions.statsCount;
+			USE_ACCELERATOR 	= dmlOptions.gpu;
+			FORCE_ACCELERATOR = dmlOptions.forceGPU;
+			EXPLAIN 					= dmlOptions.explainType;
+			ENABLE_DEBUG_MODE = dmlOptions.debug;
+			SCRIPT_TYPE				= dmlOptions.scriptType;
```
dusenberrymw: Can you fix the spacing in this section?  It would be better to just use spaces to align everything.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -396,7 +568,7 @@ else if (args[i].startsWith("-args") || args[i].startsWith("-nvargs")) {
 				arg.equalsIgnoreCase("-stats") || 
 				arg.equalsIgnoreCase("-exec") ||
 				arg.equalsIgnoreCase("-debug") ||
-				arg.startsWith("-config="))
+				arg.startsWith("-config "))
```
dusenberrymw: I think there is an extra space here: `arg.startsWith("-config "))` --> `arg.startsWith("-config"))`

src/main/java/org/apache/sysml/api/MLContext.java
```diff
@@ -1399,26 +1388,29 @@ private void checkIfRegisteringInputAllowed() throws DMLRuntimeException {
...
-	private MLOutput compileAndExecuteScript(String dmlScriptFilePath, String [] args, boolean isNamedArgument, boolean isPyDML, String configFilePath) throws IOException, DMLException {
-		return compileAndExecuteScript(dmlScriptFilePath, args, true, isNamedArgument, isPyDML, configFilePath);
+	private MLOutput compileAndExecuteScript(String dmlScriptFilePath, String [] args, boolean isNamedArgument, ScriptType scriptType, String configFilePath) throws IOException, DMLException {
+		return compileAndExecuteScript(dmlScriptFilePath, args, true, isNamedArgument, scriptType, configFilePath);
 	}
-	
+
 	/**
-	 * All the execute() methods call this, which  after setting appropriate input/output variables
+	 * 	 * All the execute() methods call this, which  after setting appropriate input/output variables
```
dusenberrymw: Can you fix this?

nakul02:
Thanks @dusenberrymw, the tabs have been converted to spaces, changes in DMLScript.java look clean now.

nakul02:
Here is the usage string it prints out:

```
usage: systemml
 -args <argN>           parameterize DML script with positional
                        parameters, after -args flag, each argument must
                        be an unnamed-argument, where 1st value after
                        -args will replace $1 in DML script, 2nd value
                        will replace $2, etc. Either -args or -nvargs can
                        be used, not both.
 -clean                 cleanup all SystemML working directories (FS,
                        DFS), All other flags are ignored in this mode.
 -config <filename>     use a given configuration file  (default: use
                        parameter values in default SystemML-config.xml
                        config file; if <filename> is prefixed with hdfs
                        or gpfs it is read from DFS, otherwise from local
                        file system)
 -debug                 run in debug mode
 -exec <mode>           execution mode can be hadoop, singlenode,
                        hybrid[default], hybrid_spark, spark
 -explain <level>       explain plan levels can be hops, runtime[default],
                        recompile_hops, recompile_runtime
 -f <filename>          will be interpreted as a filename path (if
                        <filename> is prefixed with hdfs or gpfs it is
                        read from DFS, otherwise from local file system)
                        Either this option must be specified or -s
 -gpu <force>           use CUDA instructions when reasonable. If the
                        force option is set (by specifying -gpu force),
                        the conservative memory estimates are skipped and
                        GPU is used wherever possible
 -help                  show usage message
 -nvargs <key=value>    parameterize DML script with named parameters.
                        after -nvargs flag, each argument must be be
                        named-argument of form argName=argValue, where
                        value will replace $argName in DML script, argName
                        must be a valid DML variable name (start with
                        letter, contain only letters, numbers, or
                        underscores). Either -args or -nvargs can be used,
                        not both.
 -python                parses Python-like DML
 -s <script_contents>   A script to execute directly. Either this options
                        must be specified or -f
 -stats <count>         monitor and report caching/recompilation
                        statistics, default heavy hitter count is 10
```

niketanpansare:
@nakul02 I tested few scenarios changing the order of arguments, quoting args and all worked. Here are few scenarios that I didn't like:
Scenario 1: Incorrect handling of -foo
```bash
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args hi -stats -foo temp1.dml
Exception in thread "main" org.apache.sysml.api.DMLException: java.io.FileNotFoundException: oo (No such file or directory)
        at org.apache.sysml.api.DMLScript.executeScript(DMLScript.java:531)
        at org.apache.sysml.api.DMLScript.main(DMLScript.java:209)
```

Scenario 2: Quoting doesnot work
```bash
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args "hi hii" -stats -f temp1.dml
Doesnot work as expected (but works in current master).
```

Scenario 3: Better error messages for file name
```bash
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args "hi hii" -stats -blah temp1.dml
Invalid argument specified for -stats option, must be a valid integer
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args "hi hii" -blah temp1.dml
17/03/22 18:46:00 ERROR DMLScript: Failed to execute DML script.
org.apache.sysml.parser.LanguageException: DML script was not specified!
```
--------------------------------------
Something to think about (not too strong opinion on it):

```bash
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args a=hi -stats -f temp1.dml
Passes ... Would you prefer -args arguments to be quoted in this scenario "a=hi" ?
```

As an FYI, incorrect usage of nvargs works as expected:
```bash
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -nvargs hi -stats -f temp1.dml
Invalid argument specified for -nvargs option, must be a list of space separated K=V pairs, where K is a valid name of a variable in the DML/PyDML program
```

nakul02:
Thanks for the various scenarios @niketanpansare 👍 

 ### For scenario 1 
Unfortunately, commons CLI provides an optional attached short options, much like you'd say `gcc -O2` or `gcc -O3`. Thus it treats `-foo` as `-f` followed by `oo`. 
I am not really sure what the solution is to this problem.

 ### For scenario 2
I checked on mac and linux. The behavior in this scenario is the same when the `bin/systemml` script is used.

#### temp1.dml
```r
print($1)
```

#### On Master Branch
```bash
➜  incubator-systemml git:(master) ✗ bin/systemml temp1.dml -args "hi hiii"
...
hi
...
```

#### On this branch
```bash
➜  incubator-systemml git:(SYSTEMML-1103_use_commons_cli) ✗ bin/systemml temp1.dml -args "hi hiii"
...
hi
...
```

However, when I run it using the java command, specifying all the arguments manually, it seems to work for me, both for CP mode and Spark:

```bash
➜  incubator-systemml git:(SYSTEMML-1103_use_commons_cli) ✗ java -classpath "conf/*:target/lib/*:target/systemml-1.0.0-incubating-SNAPSHOT.jar" org.apache.sysml.api.DMLScript -f temp1.dml -args "hi hiii"
log4j:WARN No appenders could be found for logger (org.apache.hadoop.util.Shell).
log4j:WARN Please initialize the log4j system properly.
hi hiii
```


```bash
➜  incubator-systemml git:(SYSTEMML-1103_use_commons_cli) ✗ $SPARK_HOME/bin/spark-submit --master local target/SystemML.jar -f temp1.dml -args "hi hiii"
log4j:WARN No appenders could be found for logger (org.apache.hadoop.util.Shell).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
hi hiii
```

The error might happen because of the use of ours spark or standalone scripts. 
Could you please check again and let me know if you still get the error? 
Please use the latest version of this PR.




### For scenario 3

I do agree with this example:
```
~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args "hi hii" -stats -blah temp1.dml
Invalid argument specified for -stats option, must be a valid integer
```
The error message could have been better. 
I do have an explanation, not a solution to this however. Here it is in case you are interested:
When `-stats` is parsed, the parser expects 0 or 1 arguments. The parser then comes across `-blah`. Now it can't find `-blah` or any prefix substring of `blah` to be a valid option, therefore it assumes it to be an option to `-stats` and stores it as a string. Later this option is tried to be parsed as an integer, when that fails, this error message is printed out.
There is a more obvious bug that I fixed. It is mandatory to specify either of `-clean`, `-help`, `-f` or `-s`. If none of those are specified, an error should be printed. I accidentally called `isRequired` instead of `setRequired`. Anyways, that is fixed now.

This is what gets printed out:
```
Missing required option: [-help shows usage message, -s specified script string to execute directly, -f specifies dml/pydml file to execute; path can be local/hdfs/gpfs (prefixed with appropriate URI), -clean cleans up all SystemML working directories (FS, DFS); all other flags are ignored in this mode. 
```


### Regarding  `-args` accepting `a=b` as a single argument and not spitting out an error. 
In my opinion, which to be clear is just MY opinion, `a=b` is a valid string. If a user wants to get this in, we should in no way design the system to disallow it. However, if other stakeholders of the project feel that this is not needed, I'll change it now or we can change it later.

nakul02:
I've since changed the usage string to be more concise, here it is:

```usage: systemml
 -args <argN>           specifies positional parameters; first value will
                        replace $1 in DML program; $2 will replace 2nd and
                        so on
 -clean                 cleans up all SystemML working directories (FS,
                        DFS); all other flags are ignored in this mode.
 -config <filename>     uses a given configuration file (can be on
                        local/hdfs/gpfs; default values in
                        SystemML-config.xml
 -debug                 runs in debug mode; default off
 -exec <mode>           sets execution mode; can be 'hadoop' /
                        'singlenode' / 'hybrid'[default] / 'hybrid_spark'
                        / 'spark'
 -explain <level>       explains plan levels; can be 'hops' /
                        'runtime'[default] / 'recompile_hops' /
                        'recompile_runtime'
 -f <filename>          specifies dml/pydml file to execute; path can be
                        local/hdfs/gpfs (prefixed with appropriate URI)
 -gpu <force>           uses CUDA instructions when reasonable; set
                        <force> option to skip conservative memory
                        estimates and use GPU wherever possible; default
                        off
 -help                  shows usage message
 -nvargs <key=value>    parameterizes DML script with named parameters of
                        the form <key=value>; <key> should be a valid
                        identifier in DML/PyDML
 -python                parses Python-like DML
 -s <script_contents>   specified script string to execute directly
 -stats <count>         monitors and reports caching/recompilation
                        statistics; heavy hitter <count> is 10 unless
                        overridden; default off
```

niketanpansare:
@nakul02 LGTM, but I would prefer that you allow other committers to test this PR. Now, the quotes works when called via spark-submit:
```bash
$ ~/spark-2.1.0-bin-hadoop2.7/bin/spark-submit SystemML.jar -args " world1 world2" -stats -f temp1.dml
hello world1 world2
$ cat temp1.dml
print("hello" + $1)
```

If the quotes with space don't work, I would recommend that you either fix the standalone script or open a JIRA for quotes bug.

@bertholdreinwald can you please review this PR ?

nakul02:
Passed all [tests](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1275/)

nakul02:
Thanks @dusenberrymw, I'll merge

PR #434, opened 2017-03-17
[SYSTEMML-1408] Add padding parameters to max-pooling layers

dusenberrymw:
This adds padding parameters to the max-pooling layers, along with the
associated tests.  Also, there are some general code formatting updates.

PR #437, opened 2017-03-22
[SYSTEMML-1428] Fixed maxpooling functions for padding > 0

niketanpansare:
@bertholdreinwald @dusenberrymw

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
+++ b/src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
@@ -559,10 +559,10 @@ public class LibMatrixDNN {
 		Iterator<IJV> iter = input.sparseBlock.getIterator(n, n+1);
 		int [] tensorIndexes = new int[3];
 		
-		int start_index_h = params.start_indexes_h[p];
-		int end_index_h = params.end_indexes_h[p];
-		int start_index_w = params.start_indexes_w[q];
-		int end_index_w = params.end_indexes_w[q];
+		int start_index_h = Math.max(params.start_indexes_h[p], 0);
+		int end_index_h = Math.min(params.end_indexes_h[p], params.H);
+		int start_index_w = Math.max(params.start_indexes_w[q], 0);
+		int end_index_w = Math.min(params.end_indexes_w[q], params.W);
 		
 		int maxIndex = -1; 
 		double maxVal = -Double.MAX_VALUE;
@@ -592,10 +592,10 @@ public class LibMatrixDNN {
 	}
 	
 	private static int getMaxIndex(int p, int q, int inputOffset, double [] inputArray, ConvolutionParameters params) {
-		int start_index_h = params.start_indexes_h[p];
-		int end_index_h = params.end_indexes_h[p];
-		int start_index_w = params.start_indexes_w[q];
-		int end_index_w = params.end_indexes_w[q];
+		int start_index_h = Math.max(params.start_indexes_h[p], 0);
+		int end_index_h = Math.min(params.end_indexes_h[p], params.H);
+		int start_index_w = Math.max(params.start_indexes_w[q], 0);
+		int end_index_w = Math.min(params.end_indexes_w[q], params.W);
 		
 		int maxIndex = -1; 
 		double maxVal = -Double.MAX_VALUE;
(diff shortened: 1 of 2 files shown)


dusenberrymw:
@niketanpansare Unfortunately, I'm still experiencing test failures with this PR using the tests added in #434.  Can you grab that version of the `nn` library and run the tests to observe the issues?

niketanpansare:
@dusenberrymw interesting, I compared this PR with the results returned by CuDNN. Here is the code that I used for testing: https://github.com/niketanpansare/systemml-cudnn-test ... In particular, can you please check if the tests satisfy the assumption "do not treat pad as zero" (see https://github.com/apache/incubator-systemml/pull/437/files#diff-ec9eaa0bc47052d8d2acbba33fbd8f89R570)  which is consistent with CuDNN ?

dusenberrymw:
Indeed, I was not aware of that assumption, but I see that is true within other implementations as well, albeit the docs aren't always correct :).  I'll make a few updates to my code & tests and then see where we are at.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
```diff
@@ -433,16 +433,14 @@ private static void fillIndexesArray(ConvolutionParameters params) {
 		params.start_indexes_w = new int[params.Q];
 		params.end_indexes_w = new int[params.Q];
 		for (int p = 0; p < params.P; p++) {
-			int start_index_h = p * params.stride_h - params.pad_h;
-			final int end_index_h = Math.min(start_index_h + params.R, params.H);
-			start_index_h = Math.max(start_index_h, 0);
+			int start_index_h = p * params.stride_h - params.pad_h; // Math.max(p * params.stride_h - params.pad_h, 0);
+			int end_index_h = start_index_h + params.R; // Math.min(start_index_h + params.R, params.H);
```
dusenberrymw: Can you remove the commented out code before merging?

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
```diff
@@ -433,16 +433,14 @@ private static void fillIndexesArray(ConvolutionParameters params) {
...
 			params.start_indexes_h[p] = start_index_h;
 			params.end_indexes_h[p] = end_index_h;
 		}
 		for (int q = 0; q < params.Q; q++) {
-			int start_index_w = Math.max(q * params.stride_w - params.pad_w, 0);
-			int end_index_w = Math.min(start_index_w + params.S, params.W);
-			start_index_w = Math.max(start_index_w, 0);
+			int start_index_w =  q * params.stride_w - params.pad_w; // Math.max(q * params.stride_w - params.pad_w, 0);
+			int end_index_w = start_index_w + params.S; // Math.min(start_index_w + params.S, params.W);
```
dusenberrymw: Can you remove the commented out code before merging?

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
```diff
@@ -899,7 +909,7 @@ public static void maxpooling(MatrixBlock input, MatrixBlock outputBlock, Convol
 		//post-processing: maintain nnz
 		outputBlock.recomputeNonZeros();
 	}
-
+
```
dusenberrymw: nitpick: can you remove this whitespace?

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDNN.java
```diff
@@ -919,8 +929,11 @@ private static void doPooling(int n, ConvolutionParameters params) throws DMLRun
 				final int inOffset1 = inOffset + c*HW;
 				for (int p = 0; p < params.P; p++) {
 					for (int q = 0; q < params.Q; q++, out_index++) {
-						for (int h = params.start_indexes_h[p]; h < params.end_indexes_h[p]; h++) {
-							for (int w = params.start_indexes_w[q]; w < params.end_indexes_w[q]; w++) {
+						// Note: We do not treat pad as zero and hence we don't do:  
+						// outputArray[out_index] = Math.max(outputArray[out_index], 0) 
+						// if params.start_indexes_h[p] < 0 || params.end_indexes_h[p] >= params.H || params.start_indexes_w[q] < 0 || params.end_indexes_w[q] >= params.W
```
dusenberrymw: This comment is good to leave for now.

PR #440, opened 2017-03-27
[SYSTEMML-1103] use Apache Commons CLI

nakul02:
This PR is being re-opened since the previous one was committed onto master and needed to be undone as per suggestions by @mboehm7 (Conversation - [here](https://github.com/apache/incubator-systemml/commit/346d1c01ad94c5b8178b8c9baf7d38e0867805da#commitcomment-21510083))

Follow the conversation history in #435


deroneriksson:
Hi @nakul02. Thank you for your continued efforts on this important PR!

Couple comments:
1) In CLIOptionsParserTest, in Eclipse I see 10 warnings about "The value of the local variable ... is not used". Could you either not create these local variable or suppress the warnings?

2) Could .api.ScriptType be moved back to .api.mlcontext.ScriptType? The reasoning for this is so that this enum can be used via spark-shell (or a notebook) via a single import, which imports all the primary classes/enums needed for the MLContext API
(ie, `import org.apache.sysml.api.mlcontext._`)?

nakul02:
@deroneriksson - fixed

deroneriksson:
LGTM. Worked in all sanity checks I tried (in eclipse, standalone, args, nvargs, etc).

nakul02:
Thanks @deroneriksson

dusenberrymw:
Let's merge this as soon as the 0.14 rc1 is tagged.

src/main/java/org/apache/sysml/api/jmlc/Connection.java
```diff
@@ -63,6 +54,15 @@
...
 import org.apache.wink.json4j.JSONObject;
 
+import java.io.BufferedReader;
+import java.io.Closeable;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.HashMap;
+import java.util.Map;
```
mboehm7: Please, fix the ordering of imports - why did you replace the existing sorted imports with unsorted lists - there are multiple files with similar problems.

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -216,7 +234,200 @@ public static void main(String[] args)
...
+		if (line.hasOption("exec")){
+			String execMode = line.getOptionValue("exec");
+			if (execMode != null){
+				if (execMode.equalsIgnoreCase("hadoop")) dmlOptions.execMode = RUNTIME_PLATFORM.HADOOP;
+				else if (execMode.equalsIgnoreCase("singlenode")) dmlOptions.execMode = RUNTIME_PLATFORM.SINGLE_NODE;
+				else if (execMode.equalsIgnoreCase("hybrid")) dmlOptions.execMode = RUNTIME_PLATFORM.HYBRID;
+				else if (execMode.equalsIgnoreCase("hybrid_spark")) dmlOptions.execMode = RUNTIME_PLATFORM.HYBRID_SPARK;
+				else if (execMode.equalsIgnoreCase("spark")) dmlOptions.execMode = RUNTIME_PLATFORM.SPARK;
+				else throw new org.apache.commons.cli.ParseException("Invalid argument specified for -exec option, must be one of [hadoop, singlenode, hybrid, hybrid_spark, spark]");
+			}
```
mboehm7: Use ExplainType.valueOf() instead, potentially surrounded by error handling, instead of this redundant string matching

src/main/java/org/apache/sysml/api/DMLScript.java
```diff
@@ -216,7 +234,200 @@ public static void main(String[] args)
...
+		if (line.hasOption("f")){
+			dmlOptions.filePath = line.getOptionValue("f");
+		}
+
+		if (line.hasOption("s")){
+			dmlOptions.script = line.getOptionValue("s");
+		}
+
+		// Positional arguments map is created as ("$1", "a"), ("$2", 123), ....
+		if (line.hasOption("args")){
```
mboehm7: Please simplify to for(int k=1, k<=argValues.length)

nakul02: I am not sure how'd you'd put this in a loop. Can you please clarify exactly how I would do this?

src/main/java/org/apache/sysml/api/DMLScript.java
nakul02: As much as possible, I am trying to do what the current master does. Using valueOf will not give us "equalsIgnoreCase" functionality (AFAIK). I will prefer to keep this the way it is.

src/main/java/org/apache/sysml/api/DMLScript.java
mboehm7: no the comment was just misplaced - it refers to the loop below that starts with for (String str : argValues) - I would simplify these 7 lines of code to 3.

src/main/java/org/apache/sysml/api/jmlc/Connection.java
nakul02: Will do. I guess, we are conforming to eclipse style imports. This is the intellij style imports.
I followed [this](http://stackoverflow.com/questions/14716283/is-it-possible-for-intellij-to-organize-imports-the-same-way-as-in-eclipse) guide. 
We should quickly decide on a code style that we can all conform to and have all our IDEs use this style file.

src/main/java/org/apache/sysml/api/DMLScript.java
mboehm7: same thing here the comment refers to the code below - ExplainType, which uses valueOf in master

mboehm7: sorry, should use Explain.parseExplainType

nakul02: I've removed Explain.parseExplainType. The reason for this is that parsing should not be used anywhere other than through the CLI. For the tests, for example, I will have it so that it sends in an instance of `DMLOptions` instead of a string array of `args` with all the systemml options.

In any case, my preference is towards this.

deroneriksson:
@nakul02 I'm highly in favor of merging this now. It's up to 23 commits and has 3 LGTMs ( @dusenberrymw @niketanpansare and myself) and has been thoroughly reviewed by the community.

nakul02:
Thanks @deroneriksson, I shall merge.

nakul02:
Actually, on second thought, I'll wait till 0.14 is voted on and passed and then merge this in.

nakul02:
Since 0.14 is on a separate branch I am planning on merging this.
@dusenberrymw, @deroneriksson, @niketanpansare - is that ok to do?

niketanpansare:
Can you please resolve the conflicts ? I would like to test it before giving LGTM.

nakul02:
@niketanpansare - on it

nakul02:
Passed [tests ](https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1353/)
@niketanpansare  - could you please test this out, I will merge when I have your ok, thanks.

niketanpansare:
Ran few sanity tests to validate this PR. LGTM Thanks @nakul02 :+1:

nakul02:
Great, I shall merge. Thanks for the review and feedback @deroneriksson, @dusenberrymw, @niketanpansare and @mboehm7.

PR #447, opened 2017-04-01
SystemML-NN Updates

dusenberrymw:
This PR contains a group of commits that are mostly performs a cleanup and reorganization of the `nn` library in preparation for one more large API update, and then graduation from the `staging` directory for the 1.0 release.

Commit a67f07a *does* make a breaking API change by updating the convolution and max pooling layer names to include "2d".  This way, we can add 1D and 3D variants in the future.  @niketanpansare You will need to update Caffe2DML with `conv*.dml` -> `conv2d*.dml`, and `max_pool*.dml` -> `max_pool2d*.dml`.


niketanpansare:
LGTM. Thanks @dusenberrymw ... I will make necessary changes in Caffe2DML :+1:

PR #452, opened 2017-04-05
[SYSTEMML-1462] Extract breast cancer preprocessing code into a package

dusenberrymw:
Extract breast cancer preprocessing code into a Python package + script.

This commit extracts the breast cancer project preprocessing functions
into a new `breastcancer` Python package, along with a new `preprocess.py`
script.  The notebook has all functions removed, leaving only the settings
and execution of the preprocessing using the new package, and the new script
is essentially a copy of that for use with `spark-submit`.


deroneriksson:
LGTM. Thank you @dusenberrymw.

PR #456, opened 2017-04-10
[SYSTEMML-1510] Add Pygments license in source distribution

asurve:
Updated license text for Pygments in source license file.

@deroneriksson Can you please review change?

Note:
    - I don't have Pygments version to specify, I don't think it matters.
    - License specifies to refer to AUTHORS file from Pygments source code, not sure if it needs to be added anywhere.


deroneriksson:
I believe this is correct. Thank you @asurve.

LGTM.

PR #459, opened 2017-04-12
[SYSTEMML-1516] Improve output size calculation in conv2d & max_pool2d

dusenberrymw:
The calculation of the output sizes (`Hout`, `Wout`) for conv2d and
max_pool2d layers rely on a flooring operation to only allow for valid
operations.  Previously, we simply relied on the `as.integer` call to
implicitly perform this flooring operation during truncation of the
double to an integer.  While this gave the correct results, it was less
readable than it should be, and relied on truncation behavior rather
than explicit mathematical behavior.  This commit simply adds an
explicit `floor` operation to each calculation.


deroneriksson:
LGTM. Thank you @dusenberrymw!

PR #460, opened 2017-04-16
[SYSTEMML-1529] Improve accuracy in test_svm of test_mllearn_numpy.py

gweidner:
Adjusted tolerance parameter in test_svm to prevent _occasional_ failure due to accuracy not meeting test criteria of greater than 95%.

### DIFF
--- a/src/main/python/tests/test_mllearn_numpy.py
+++ b/src/main/python/tests/test_mllearn_numpy.py
@@ -149,7 +149,7 @@ class TestMLLearn(unittest.TestCase):
         y_train = y_digits[:int(.9 * n_samples)]
         X_test = X_digits[int(.9 * n_samples):]
         y_test = y_digits[int(.9 * n_samples):]
-        svm = SVM(sparkSession, is_multi_class=True, tol=0.001)
+        svm = SVM(sparkSession, is_multi_class=True, tol=0.0001)
         mllearn_predicted = svm.fit(X_train, y_train).predict(X_test)
         from sklearn import linear_model, svm
         clf = svm.LinearSVC()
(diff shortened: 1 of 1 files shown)


gweidner:
Instrumentation displaying the accuracy showed the score above 95% most of the time.  Occasionally the result was at or below the test threshold of 0.95 resulting in a test failure.  For example, one failure was exactly 0.95, another slightly below at 0.94444.

First tried intercept 2 by using normalize=True and keeping fit_intercept default value of false.  However, this was invalid based on resulting program exception:
`
Caused by: org.apache.sysml.runtime.DMLScriptException: Stopping due to invalid argument: Currently supported intercept options are 0 and 1
        at org.apache.sysml.runtime.instructions.cp.ScalarBuiltinCPInstruction.processInstruction(ScalarBuiltinCPInstruction.java:64)
`

Next tried disabling regularization with C=float("inf") but observed a failure where accuracy as low as 0.92777.

Then changed tolerance to 0.001 based on default value from scripts/algorithms/m-svm.dml and test/scripts/applications/m-svm/m-svm.pydml.  Accuracy was observed between 0.977 and 0.994.  Note the default parameters in python/systemml/mllearn/estimators.py were kept as-is.

mboehm7:
LGTM - thanks @gweidner. As a nice to have could you try tol=0.0001 as that's the default of sklearn.LinearSVC. Unfortunately, the documentation does not tell if this is a target or per iteration tolerance but if it is - similar to our algorithm - a target tolerance then I'd like to use consistent parameters for comparison.

gweidner:
Thank you for reviewing Matthias - much appreciated!  Good point regarding tolerance - let me retry with tol=0.0001.  Before this PR I had only used the default value in estimators.py of 0.000001.

gweidner:
I ran a series of experiments at `tol=0.0001` where I ran test_svm only and always printed the accuracy.  The results looked good as twice an accuracy of 1.0 was reported.  Also repeated the full test_mllearn_numpy group with test_svm at `tol=0.0001` using updated commit.  Since all 7 tests passed as well, merging to master.

PR #461, opened 2017-04-17
[SYSTEMML-1532] Python launch script for standalone

nakul02:
This python script to launch standalone systemml has been tested on
- Windows 10 command line
- Windows 10 with Cygwin
- Mac OS Sierra
- Linux

Note the addition of "cygwin". We currently have not script that does this. It did require figuring out a couple of quirks with path separators and such because of the mixed environment that cygwin provides.
The behavior of this script is very similar to the bash script `bin/systemml` in so far as I thought to be reasonable. This is only an addition, the original `bin/systemml` and `bin/systemml.bat` scripts remain intact.

This script is only for developers, (I think)? 

Ping @deroneriksson, @dusenberrymw, 
Pinging windows users - @niketanpansare, @gweidner


niketanpansare:
I would recommend not using the name `systemml.py` as our python api has a module with the same name. 

So as to know the reasoning behind this PR, is it fair to assume that having one single python file for all environment help simplify developer testing ?

nakul02:
@niketanpansare - yes, a single python file for all environments, i think, is a good thing.
Also, it adds support for cygwin (which is my preferred setup when on windows).
Do you have suggestions for names?

niketanpansare:
something like run_systemml or launch_systemml ?

nakul02:
@niketanpansare  - Changed the name to `systemml-standalone.py`.
There will be a corresponding `systemml-spark-submit.py`, `systemml-spark-shell.py` and `systemml-pyspark-shell.py` (if those make sense).

bin/systemml-standalone.py
```diff
@@ -0,0 +1,200 @@
...
+            if l and not(l.startswith('#')) and '=' in l:
+                k, v = l.split('=', 1)
+                k = k.strip()
+                v = v.strip()
+                if len(v) > 0:
+                    quoted = v[0] == v[len(v) - 1] and v[0] in ['"', "'"]
+
+                # strip quotes
+                if quoted:
+                    v = v[1:-1]
```
dusenberrymw: I would just condense this down to:
```
if len(v) > 0 and v[0] == v[len(v) - 1] and v[0] in ['"', "'"]:
  # strip quotes
  v = v[1:-1]
```

bin/systemml-standalone.py
```diff
@@ -0,0 +1,200 @@
...
+    env_vars = parse_env_file(systemml_env_path)
+    os.environ.update(env_vars)
+
+
+# Invoke the jar with options and arguments
+cmd = ['java', systemml_default_java_opts, 'org.apache.sysml.api.DMLScript', '-f', script_file, '-exec singlenode', '-config', systemml_config_path] + sys.argv[2:]
+# For debugging
+# print(' '.join(cmd))
+
+return_code = os.system(' '.join(cmd))
```
dusenberrymw: I think the use of the newer `subprocess.call` function is encouraged now (actually, there is a `subprocess.run` function, but it is for >= 3.5): https://docs.python.org/3.6/library/subprocess.html#replacing-os-system

nakul02: I was having trouble with environment variables when using `subprocess.call` which is why i switched to `os.system`.

niketanpansare:
>> systemml-standalone.py. There will be a corresponding systemml-spark-submit.py, systemml-spark-shell.py and systemml-pyspark-shell.py

If possible, I would recommend creating a single script and handling different scenarios within it using parameters. Likely I am missing some context; can you explain why we would need systemml-spark-submit.py, systemml-spark-shell.py and systemml-pyspark-shell.py ? Would it not simply invoke spark-submit with --jars or --driver-class-path ?

>>  I think it will be useful to have a single script for all environments, but we should make a decision and deprecate the other scripts if we prefer this one.

+1.

nakul02:
We can discuss whether one script with Args is better or more. Personally I prefer having more to clearly separate the different entry points into SystemML. This is much like Spark. A separate script for pyspark, spark-submit, etc.

niketanpansare:
Multiple files/entry points has higher chance of inconsistencies as evident from our APIs. Since this PR doesnot add other files and since this is intended for developers, I am OK with this PR :+1:  ... When we plan to make it available to general audience, I would recommend discussing it on our mailing list.

As a side note, I understand other committers might have different views on this, but I am not fan of having wrapper invocation scripts over standard programs (such as java, spark-submit, pyspark, spark-shell). I would prefer better documentation of usage rather than hiding them in these scripts. For example: we simply ask users to invoke SystemML using
`java -cp target/*:SystemML.jar -f ...` or `spark-submit SystemML.jar -f ...` or `pyspark --driver-class-path SystemML.jar --jars SystemML.jar -f ..`.

nakul02:
Thank you for the `ok` on this PR 👍 

Well, `spark` does have different scripts. And they hide the messy invocation behind scripts. I would suggest that we should learn from it what we can.
I suspect what will probably end up happening is that people will create scripts on their own if we don't. 

I am a fan of documentation. 🥇

mboehm7:
well, I kind of agree with both sides here: currently, I use my custom sparkDML2.sh scripts to simplify the invocation through spark_submit. I don't use sparkDML.sh for the reason Niketan brought up: we want to feed our flags unchanged into such as script, while sparkDML requires to modify dash to dash dash (for consistency with shell script conventions) and some flags cannot be passed at all.

bertholdreinwald:
how about extending sparkDML.sh ?

nakul02:
@bertholdreinwald - that is the next step. Either I or the GSoC student can work on it (assuming the project goes through). I was thinking of redoing the sparkDML as a python script.

nakul02:
@niketanpansare, @dusenberrymw, does this look good to you?
Can I merge?

niketanpansare:
LGTM :+1:

nakul02:
I shall merge, thank you

PR #463, opened 2017-04-18
[SYSTEMML-259] Function with no return value fix

deroneriksson:
Add additional target null checks so that if a function does not return a
value and is not inlined, it does not require an lvalue.
Add MLContext tests with forced function calls.


deroneriksson:
Thank you for reviewing @dusenberrymw!

PR #465, opened 2017-04-20
[SYSTEMML-1549] Cox.dml - return S & T in usable format

bdwyer2:
Returns `S` & `T` as a matrix instead of string.

### DIFF
--- a/scripts/algorithms/Cox.dml
+++ b/scripts/algorithms/Cox.dml
@@ -388,41 +388,59 @@ CI_l = b - se_b * z_alpha_2;
 CI_r = b - se_b + z_alpha_2;
 
 ######## SOME STATISTICS AND TESTS
+S = matrix(0, 6, 1);
+T = matrix(0, 3, 3);
+
 # no. of records
 S_str = "no. of records " + N;
+S[1, 1] = N;
 
 # no.of events
 S_str = append (S_str, "no. of events " + sum (E));
+S[2, 1] = sum (E);
 
 # log-likelihood
 loglik = -o;
 S_str = append (S_str, "loglik " + loglik + " ");
+S[3, 1] = loglik;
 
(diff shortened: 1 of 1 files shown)


dusenberrymw:
@bdwyer2 I agree that this is definitely useful.  I need to think about whether there is a way to do it without affecting the regular usage of the existing scripts.  Or perhaps we can just update it with the upcoming 1.0 release.

aloknsingh:
@dusenberrymw @bdwyer2 The above will not break the existing script. I believe it is returning extra matrixes so it should be safe

deroneriksson:
@dusenberrymw I don't see any tests in the project that test Cox.dml (including scripts/perftest/runAll.sh, where there is a comment where the test should be), so I don't think this change would break anything.

Is line 213 ok (`write (S_str, fileS, format = fmtO)`)?

dusenberrymw:
Is this comment still applicable to the matrices (vs. the original string format)?
```
# Two log files containing a summary of some statistics of the fitted model:
 # 1- File S with the following format 
 #	- line 1: no. of observations
 #	- line 2: no. of events
 #   - line 3: log-likelihood 
 #	- line 4: AIC
 #	- line 5: Rsquare (Cox & Snell)
 #	- line 6: max possible Rsquare
 # 2- File T with the following format
 #	- line 1: Likelihood ratio test statistic, degree of freedom, P-value
 #	- line 2: Wald test statistic, degree of freedom, P-value
 #	- line 3: Score (log-rank) test statistic, degree of freedom, P-value
```

deroneriksson:
LGTM. Thank you @bdwyer2! I'll merge.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1482/

PR #466, opened 2017-04-20
[SYSTEMML-692] Separated Caffe2dml into a separate jar

niketanpansare:
@mboehm7 @bertholdreinwald @dusenberrymw @deroneriksson @asurve  This PR separates Caffe2DML-related files into a separate jar. Let's use this PR to discuss whether:
1. **Current Master:** Caffe2DML should be integrated in main SystemML jar. Pros: One jar for all. Cons: +1 MB.
2. **This PR:** Caffe2DML should be released as a separate jar. Pros: -1MB. Cons: _One more artifact to release_.
3. Caffe2DML should be a separate project. I don't like this option as Caffe2DML is an API of SystemML, not a standalone tool.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,6 +255,27 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
 				<executions>
+				  <execution>
+		            <goals><goal>jar</goal></goals>
+		            <phase>package</phase>
+		            <configuration>
+		              <archive>
+						<index>true</index>
+						<manifestEntries>
+							<Build-Timestamp>${maven.build.timestamp}</Build-Timestamp>
+							<Main-Class>org.apache.sysml.api.DMLScript</Main-Class>
+						</manifestEntries>
+					  </archive>
+					  <classifier>lite</classifier>
+		              <excludes>
+		                <exclude>**/caffe/*</exclude>
+		                <exclude>**/org/tensorflow/*</exclude>
+		                <exclude>**/org/tensorflow/framework/*</exclude>
+		                <exclude>**/org/tensorflow/util/*</exclude>
+		                <exclude>**/org/apache/sysml/api/dl/*</exclude>
+		              </excludes>
+		            </configuration>
+		          </execution>
 		          <execution>
 		            <id>caffe2dml</id>
 		            <goals><goal>jar</goal></goals>
@@ -277,22 +298,6 @@
 		            </configuration>
 		          </execution>
 		        </executions>
-		        <configuration>
-	              <archive>
-					<index>true</index>
-					<manifestEntries>
-						<Build-Timestamp>${maven.build.timestamp}</Build-Timestamp>
-						<Main-Class>org.apache.sysml.api.DMLScript</Main-Class>
-					</manifestEntries>
-				  </archive>
-	              <excludes>
-	                <exclude>**/caffe/*</exclude>
-	                <exclude>**/org/tensorflow/*</exclude>
-	                <exclude>**/org/tensorflow/framework/*</exclude>
-	                <exclude>**/org/tensorflow/util/*</exclude>
-	                <exclude>**/org/apache/sysml/api/dl/*</exclude>
-	              </excludes>
-	            </configuration>
 			</plugin>
 
 			<plugin>
@@ -612,6 +617,7 @@
 						<phase>package</phase>
 						<configuration>
 							<target name="copy and rename JAR">
+								<move file="${project.build.directory}/${project.artifactId}-${project.version}-lite.jar" tofile="${project.build.directory}/${project.artifactId}-${project.version}.jar" />
 								<copy file="${project.build.directory}/${project.artifactId}-${project.version}.jar" tofile="${project.build.directory}/SystemML.jar" />
 							</target>
 						</configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -254,15 +254,45 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
-				<configuration>
-					<archive>
+				<executions>
+		          <execution>
+		            <id>caffe2dml</id>
+		            <goals><goal>jar</goal></goals>
+		            <phase>package</phase>
+		            <configuration>
+		              <archive>
 						<index>true</index>
 						<manifestEntries>
 							<Build-Timestamp>${maven.build.timestamp}</Build-Timestamp>
 							<Main-Class>org.apache.sysml.api.DMLScript</Main-Class>
 						</manifestEntries>
-					</archive>
-				</configuration>
+					  </archive>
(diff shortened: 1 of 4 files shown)


mboehm7:
I'm in favor of (2) - in addition to the size of the jar, it also affects build time, and it removes the tensorflow package from our jar which might confuse people to think SystemML has a tensorflow dependency.

bertholdreinwald:
For certain deployment scenarios, people ask for a "as small as possible" footprint SystemML jar file as they only want to use some limited, core capability of SystemML, e.g. do not want to use caffe proto, or MR/Spark backend, just fast in-memory Java processing. These scenarios give the 1st option a strong -1.

niketanpansare:
Thanks Berthold and Matthias for your feedback. I guess we are all in agreement that bundling Caffe2DML separately is the way to go.

Nakul, Mike, Arvind, Deron: Can you please provide your feedback as this PR introduces another artifact in our release ?

asurve:
Today we have single jar file for deployment. By having multiple artifacts/jar files, how deployment will be handled? 
In the cluster environment do we expect user to configure every node with additional artifact/jar file?

deroneriksson:
First off, WRT build time, I really don't think build time is an issue. Adding a few seconds to a project build seems trivial if this adds useful functionality that will help users. If anything, splitting to a separate jar built in the same project might increase build time compared to packaging this in the same jar. However, I only think build time is an issue if it did something like add a minute or more to the build, and even then I would think that is fine if it adds critical functionality to the project.

Having a separate jar created by the project's pom adds additional complexity to our build. This makes the pom even more difficult to maintain. So with regards to build maintenance, (1) is better than (2). That said, I'm a little bit frustrated by the 75 java warnings that get introduced to the project from protobuf-generated classes, but I'm not sure what can be done there. Personally I like to have 0 java warnings in a project. That way, I can easily see in my IDE if I have introduced new java warnings, whereas if there is a big pool of java warnings already, it gets harder to spot the ones that I might be introducing.

Also, if there is an additional artifact (2), please be aware that it should have the required LICENSE, NOTICE, and DISCLAIMER files and the LICENSE needs to accurately reflect the contents of the jar. Also, tasks such as artifact validation and integrating the additional artifact into the release process need to be considered.

niketanpansare:
I have renamed `systeml-*-caffe2dml.jar` to `systeml-*-extra.jar` . We can use it to include BLAS dependency as well.

>> Today we have single jar file for deployment. By having multiple artifacts/jar files, how deployment will be handled? In the cluster environment do we expect user to configure every node with additional artifact/jar file?

The user has to add the additional jar into `--jars` command, no need to configure every node.

>> First off, WRT build time, I really don't think build time is an issue.  Adding a few seconds to a project build seems trivial if this adds useful functionality that will help users.

I agree.

>> That said, I'm a little bit frustrated by the 75 java warnings that get introduced to the project from protobuf-generated classes, but I'm not sure what can be done there.

I too don't like to warning. This is a known issue in protobuf with Java 8: https://github.com/google/protobuf/issues/304

>> Also, if there is an additional artifact (2), please be aware that it should have the required LICENSE, NOTICE, and DISCLAIMER files and the LICENSE needs to accurately reflect the contents of the jar.

Thanks @deroneriksson  I have added META-INF of the current jar into the `extra` jar. Can someone help me with the LICENSE for the `extra` jar ?

dusenberrymw:
I prefer option (1).  At this point, adding an additional jar artifact just complicates the project further, and will lead to confusion on how to actually use SystemML for Caffe-based deep learning using the Caffe2DML API.  Additionally, when we add Keras support as well, this will likely share much of the infrastructure from Caffe2DML, and I don't want to have to create yet another "Keras2DML" jar.

As for the lightweight deployment, that certainly is a viable use-case.  However, I would prefer to have an optional "lightweight" profile for that scenario that simply builds a true minimal version of the project, rather than have the main jar be somewhat small.  This would not be released as a download for our releases, but would be available as an optional maven build, just like our current Java-only standalone build profile.

deroneriksson:
I agree with @dusenberrymw about (1). I like his idea of having a lightweight jar rather than limiting the main jar. Also if truly lightweight is a goal, then I think the hadoop dependency requirements should be removed from JMLC execution. I haven't tried it lately, but I believe JMLC execution requires at minimum hadoop-auth (67K), hadoop-common (3.4M), hadoop-mapreduce-client-core (1.5M), and possibly guava (1.6M) as a transitive dependency of these.

mboehm7:
Just to clarify the build time issue: people like me build SystemML 10s to 100s of times a day; from my perspective our build time is borderline right now - I find myself increasingly often switching tasks while the build is running which distracts the dev workflow. So not including this tool into the core build process would help keeping the build time low.

And I absolutely agree with regard to the introduced warnings. Hopefully, excluding caffe2dml from our build, will also resolve this issue.

mboehm7:
@niketanpansare isn't there also a fourth option that removes the tensorflow classes and only adds the core dml generation? This would kind of combine all advantages.

niketanpansare:
>>  isn't there also a fourth option that removes the tensorflow classes and only adds the core dml generation? This would kind of combine all advantages.

The warnings are coming from files generated by caffe.proto, not tensorflow's proto files. The fourth option you mentioned would involve:
- Create a separate repository that will be managed by some committer X (not Apache !!!).
- This repo will compile the java classes from caffe.proto and tensorflow.proto
- X will go through necessary open-source related process and then host it on either central maven on private maven repo on github.
- SystemML will then have an explicit dependency on a privately managed project.

mboehm7:
no, I was just wondering why we need these tensorflow classes at all - at the end of the day it's just about parsing some input specification and generating dml. I'm still in favor of having caffe2dml in the SystemML repository but it would be great to smooth out the build issues.

deroneriksson:
Regarding the Java warnings in Eclipse for the generated source files, I just tried the solution posted here and it removed the warnings for the generated source files: http://stackoverflow.com/questions/1127920/how-to-suppress-java-warnings-for-specific-directories-or-files-such-as-generate

Given that I can suppress these warnings, I am happier with the (1) solution. :-)

niketanpansare:
>>  I'm still in favor of having caffe2dml in the SystemML repository but it would be great to smooth out the build issues.
>> Given that I can suppress these warnings

Great !! Once we agree to a solution, Deron can resolve the warning in a separate PR.

>> I am happier with the (1) solution.
>>  However, I would prefer to have an optional "lightweight" profile for that scenario that simply builds a true minimal version of the project, rather than have the main jar be somewhat small.

If we all agree to solution (1), I am happy to create a "lightweight" profile.

>> why we need these tensorflow classes

TensorFlow classes are required for [tensorboard integration with Caffe2DML](https://github.com/apache/incubator-systemml/blob/master/src/main/python/systemml/mllearn/estimators.py#L615). In future, this will allow us to plot histograms, network, etc.

capture

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1427/

PR #467, opened 2017-04-21
[SYSTEMML-1553] Add Caffe and Tensorflow License

niketanpansare:
@asurve @bertholdreinwald @deroneriksson can you please review this PR ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -854,6 +854,10 @@
 								<exclude>src/main/python/systemml.egg-info/**</exclude>
 								<!-- Compiled ptx file from nvcc -->
 								<exclude>src/main/cpp/kernels/SystemML.ptx</exclude>
+								<!-- Proto files -->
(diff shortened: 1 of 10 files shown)


deroneriksson:
Hi @niketanpansare, I believe the full Apache license doesn't need to be included in its own section in src/main/python/LICENSE. It can be added to section that mentions Apache Wink.

In addition, the caffe and tensorflow files appear to be added to the main jar. If this is so, these need to be mentioned in:
src/assembly/bin/LICENSE
src/assembly/distrib/LICENSE
src/assembly/inmemory/LICENSE
src/assembly/jar/LICENSE
src/assembly/standalone-jar/LICENSE

deroneriksson:
BTW this is why I think we should remove "distrib". We have way too many LICENSE files to manage right now. :-)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1429/

PR #468, opened 2017-04-21
[SYSTEMML-1554] IPA Scalar Transient Read Replacement

dusenberrymw:
Currently, during IPA we collect all variables (scalars & matrices)
eligible for propagation across blocks (i.e. not updated in block), and
then propagate the only the matrix sizes across the blocks. It seems
plausible that we could also replace all eligible scalar transient reads
with literals based on the variables that have already been collected.
The benefit is that many ops will be able to determine their respective
output sizes during regular compilation, instead of having to wait until
dynamic recompilation, and thus we can reduce the pressure on dynamic
recompilation.

Are there drawbacks to this approach?  The use case is that I was seeing a large number of memory warnings while training a convolutional net due to the sizes being unknown during regular compilation, yet the engine only having CP versions of the ops.  Additionally, I was running into actual heap space OOM errors for situations that should not run out of memory, and thus I started exploring.

I've attached an example script and the explain plan (hops & runtime) w/ and w/o the IPA scalar replacement to the associated JIRA issue.

cc @mboehm7

### DIFF
--- a/src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
+++ b/src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
@@ -467,13 +467,12 @@ public class InterProceduralAnalysis
 	 *
 	 * @param sb  DML statement blocks.
 	 * @param fcand  Function candidates.
-	 * @param callVars  Calling program's map of variables eligible for
-	 *                     propagation.
+	 * @param callVars  Map of variables eligible for propagation.
 	 * @param fcandSafeNNZ  Function candidate safe non-zeros.
 	 * @param unaryFcands  Unary function candidates.
 	 * @param fnStack  Function stack to determine current scope.
-	 * @throws HopsException
-	 * @throws ParseException
+	 * @throws HopsException  If a HopsException occurs.
+	 * @throws ParseException  If a ParseException occurs.
 	 */
 	private void propagateStatisticsAcrossBlock( StatementBlock sb, Map<String, Integer> fcand, LocalVariableMap callVars, Map<String, Set<Long>> fcandSafeNNZ, Set<String> unaryFcands, Set<String> fnStack )
 		throws HopsException, ParseException
@@ -566,10 +565,15 @@ public class InterProceduralAnalysis
 	 *
 	 * This replaces scalar reads and typecasts thereof with literals.
 	 *
+	 * Ultimately, this leads to improvements because the size
+	 * expression evaluation over DAGs with scalar symbol table entries
+	 * (which is also applied during IPA) is limited to supported
+	 * operations, whereas literal replacement is a brute force method
+	 * that applies to all (including future) operations.
+	 *
 	 * @param roots  List of HOPs.
-	 * @param vars  Calling program's map of variables eligible for
-	 *                 propagation.
-	 * @throws HopsException
+	 * @param vars  Map of variables eligible for propagation.
+	 * @throws HopsException  If a HopsException occurs.
 	 */
 	private void propagateScalarsAcrossDAG(ArrayList<Hop> roots, LocalVariableMap vars)
 		throws HopsException
@@ -608,9 +612,9 @@ public class InterProceduralAnalysis
 	/**
 	 * Propagate matrix sizes across DAGs.
 	 *
-	 * @param roots  List of HOPs.
+	 * @param roots  List of HOP DAG root nodes.
 	 * @param vars  Map of variables eligible for propagation.
-	 * @throws HopsException
+	 * @throws HopsException  If a HopsException occurs.
 	 */
 	private void propagateStatisticsAcrossDAG( ArrayList<Hop> roots, LocalVariableMap vars )
 		throws HopsException
@@ -643,15 +647,15 @@ public class InterProceduralAnalysis
 	 * block.
 	 *
 	 * @param prog  The DML program.
-	 * @param roots List of HOPs in this DAG for propagation.
+	 * @param roots List of HOP DAG root notes for propagation.
 	 * @param fcand  Function candidates.
 	 * @param callVars  Calling program's map of variables eligible for
 	 *                     propagation.
 	 * @param fcandSafeNNZ  Function candidate safe non-zeros.
 	 * @param unaryFcands  Unary function candidates.
 	 * @param fnStack  Function stack to determine current scope.
-	 * @throws HopsException
-	 * @throws ParseException
+	 * @throws HopsException  If a HopsException occurs.
+	 * @throws ParseException  If a ParseException occurs.
 	 */
 	private void propagateStatisticsIntoFunctions(DMLProgram prog, ArrayList<Hop> roots, Map<String, Integer> fcand, LocalVariableMap callVars, Map<String, Set<Long>> fcandSafeNNZ, Set<String> unaryFcands, Set<String> fnStack )
 			throws HopsException, ParseException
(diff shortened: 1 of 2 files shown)


mboehm7:
thanks @dusenberrymw I gave it a try on our ARIMA application testcase which historically was challenging for scalar propagation. Unfortunately, it failed due to an - probably unrelated - issue in replaceLiteralFullUnaryAggregate. Once I resolved this, I'll play around with it a bit more. 

Once this PR is in, we should also think about the problem of propagating scalars into functions if functions are called once or with consistent scalar inputs.

dusenberrymw:
@mboehm7 Great, interested to see what else is needed for this to be generally applicable.  Also, definitely +1 for propagating scalars into functions.  In particular, we should allow for the case of functions for which any subset of the inputs are consistent scalars.  I.e., a function may have an unknown matrix size as an input, but then have several other scalar inputs that are always consistent.

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -543,7 +546,27 @@ else if (sb instanceof ForStatementBlock) //incl parfor
...
+	 * @param roots  List of HOPs.
+	 * @param vars  Map of variables eligible for propagation.
+	 * @throws HopsException
+	 */
+	private void propagateScalarsAcrossDAG(ArrayList<Hop> roots, LocalVariableMap vars)
+		throws HopsException
+	{
+		for (Hop hop : roots) {
+			try {
+				Recompiler.rReplaceLiterals(hop, vars);
```
mboehm7: We now expose more fine-grained control over literal replacement - please use rReplaceLiterals(hop, vars, true), which restricts literal replacements to scalars only, disallowing replacing small matrix operations as it would fail for temporary matrix object that we use for size propagation.

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -543,7 +546,27 @@ else if (sb instanceof ForStatementBlock) //incl parfor
...
+			} catch (Exception ex) {
+				throw new HopsException("Failed to replace transient reads w/ literals.", ex);
```
mboehm7: Maybe use a slightly more general exception message as literal replacement affects not only transient reads, but also value types casts on top of them.

dusenberrymw:
Thanks, @mboehm7.  Looks like it is failing a test still -- `org.apache.sysml.test.integration.functions.misc.DataTypeChangeTest#testDataTypeChangeValidate4c`.  Looking into it, it fails due to trying to cast a Matrix to a Scalar object.  At a deeper level, it looks like the propagated variable map is holding onto the "matrix" `X`, rather than dropping it as it should, since X is turned into a scalar by the call `X = foo(X)`.  Interestingly, the FunctionOp for the `foo` function is marked as having an `Unknown` datatype and valuetype.  That to me seems to be  big issue, but I'm not sure exactly where that is failing.  Thoughts?  Overall, this seems like a bug that was just hidden before, rather than being newly introduced.

mboehm7:
yes this is almost certainly a bug - originally we allowed data type changes in conditional control flow (e.g., if branch assigns a scalar and else branch a matrix), in which case we assign UNKNOWN for subsequent references. However, I modified this years ago because SystemML could not compile valid instructions for these scenarios unless we extend the recompiler to actually update the data type and block sizes there.

dusenberrymw:
By the way, aside from that test, everything else passed.

dusenberrymw:
cc @mboehm7 Can you review this fix for the datatype conversion issue?  I'm also waiting for the full testing with Jenkins.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1437/

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -460,8 +460,22 @@ private void determineFunctionCandidatesNNZPropagation(Map<String, FunctionOp> f
...
-	
-	private void propagateStatisticsAcrossBlock( StatementBlock sb, Map<String, Integer> fcand, LocalVariableMap callVars, Map<String, Set<Long>> fcandSafeNNZ, Set<String> unaryFcands, Set<String> fnStack ) 
+
+	/**
+	 * Perform intra-procedural analysis (IPA) by propagating statistics
+	 * across statement blocks.
+	 *
+	 * @param sb  DML statement blocks.
+	 * @param fcand  Function candidates.
+	 * @param callVars  Calling program's map of variables eligible for
```
mboehm7: Not necessarily called by a program

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -460,8 +460,22 @@ private void determineFunctionCandidatesNNZPropagation(Map<String, FunctionOp> f
...
+	 *                     propagation.
+	 * @param fcandSafeNNZ  Function candidate safe non-zeros.
+	 * @param unaryFcands  Unary function candidates.
+	 * @param fnStack  Function stack to determine current scope.
+	 * @throws HopsException
+	 * @throws ParseException
```
mboehm7: I know any description for the exceptions is absolutely redundant but probably better than a javadoc warning

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -543,7 +560,28 @@ else if (sb instanceof ForStatementBlock) //incl parfor
...
+
+	/**
+	 * Propagate scalar values across DAGs.
+	 *
+	 * This replaces scalar reads and typecasts thereof with literals.
+	 *
+	 * @param roots  List of HOPs.
+	 * @param vars  Calling program's map of variables eligible for
+	 *                 propagation.
+	 * @throws HopsException
```
mboehm7: for all following instances, the comment from above applies

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -543,7 +560,28 @@ else if (sb instanceof ForStatementBlock) //incl parfor
...
+	 */
+	private void propagateScalarsAcrossDAG(ArrayList<Hop> roots, LocalVariableMap vars)
+		throws HopsException
+	{
+		for (Hop hop : roots) {
+			try {
+				Recompiler.rReplaceLiterals(hop, vars, true);
```
mboehm7: I was thinking about the need for this and I guess it leads to improvements because the size expression evaluation over dags with scalar symbol table entries (which is also applied during IPA) is limited in supported operations whereas literal replacement and constant folding is a brute force method which applies to all (including future) operations because we simply compile runtime instructions during constant folding. Bottom line: it's good to have it in IPA.

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -566,8 +604,15 @@ private void propagateStatisticsAcrossPredicateDAG( Hop root, LocalVariableMap v
 			throw new HopsException("Failed to update Hop DAG statistics.", ex);
 		}
 	}
-	
-	private void propagateStatisticsAcrossDAG( ArrayList<Hop> roots, LocalVariableMap vars ) 
+
+	/**
+	 * Propagate matrix sizes across DAGs.
+	 *
+	 * @param roots  List of HOPs.
```
mboehm7: List of HOP DAG root nodes

dusenberrymw:
Thanks, @mboehm7.  I'll update the docs and merge.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1442/

PR #471, opened 2017-04-26
[SYSTEMML-591] Initial attempt to consolidate replicated compilation chain

niketanpansare:
This PR only consolidates MLContext and DMLScript's execute method. But JMLC and old MLContext are not touched in this PR; latter is going to be removed soon. This addresses several inconsistencies/duplications:
- `Statistics.startRunTimer()`. In subsequent PR, we should address `Statistics.startCompileTimer();`
- Printing Statistics via Log4J.
- Recently added boolean flags such as systemml.stats.extraDNN, systemml.stats.extraGPU, etc.
- Spoof was not enabled via MLcontext.
- GPU creation logic and error message

@mboehm7 @deroneriksson @nakul02  can you please review this PR ?

### DIFF
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -88,11 +88,8 @@ import org.apache.sysml.runtime.controlprogram.context.SparkExecutionContext;
 import org.apache.sysml.runtime.controlprogram.parfor.ProgramConverter;
 import org.apache.sysml.runtime.controlprogram.parfor.stat.InfrastructureAnalyzer;
 import org.apache.sysml.runtime.controlprogram.parfor.util.IDHandler;
-import org.apache.sysml.runtime.instructions.gpu.context.GPUContext;
-import org.apache.sysml.runtime.instructions.gpu.context.GPUContextPool;
 import org.apache.sysml.runtime.io.IOUtilFunctions;
 import org.apache.sysml.runtime.matrix.CleanupMR;
-import org.apache.sysml.runtime.matrix.data.LibMatrixDNN;
 import org.apache.sysml.runtime.matrix.mapred.MRConfigurationNames;
 import org.apache.sysml.runtime.matrix.mapred.MRJobConfiguration;
 import org.apache.sysml.runtime.util.LocalFileUtils;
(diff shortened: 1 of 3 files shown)


akchinSTC:
Build successful

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1381/

mboehm7:
thanks for working on this; couple of high-level comments:
1) the up-to-date issue is https://issues.apache.org/jira/browse/SYSTEMML-1325
2) the purpose of this issue is to consolidate the compilation chain from script to runtime program to create consistency and fix a couple of existing bugs, not just the actual execute 
3) I would recommend to remove the old mlcontext first in order to avoid unnecessary updates
4) further, please start from DMLScript's compilation chain as this is the only feature-complete compilation chain and thus, the ground truth.

mboehm7:
Furthermore, it would be great to design a configurable compilation chain first, implement it as a new class, and subsequently replace all duplicated compilation chains (for that the new chain need to be sufficiently flexible to allow a one-line invocation from mlcontext/jmlc/dmlscript/debug)

niketanpansare:
>> I would recommend to remove the old mlcontext first in order to avoid unnecessary updates

I am not sure the right protocol here. Should this be discussed on the mailing list before removing an API ? If this is not required as the old MLContext has been deprecated for a while, I can remove the old MLContext as part of this PR. Else, I suggest that it should be addressed in a separate PR. 

>> the purpose of this issue is to consolidate the compilation chain from script to runtime program to create consistency and fix a couple of existing bugs, not just the actual execute
>> further, please start from DMLScript's compilation chain as this is the only feature-complete compilation chain and thus, the ground truth.
>> Furthermore, it would be great to design a configurable compilation chain first, implement it as a new class, and subsequently replace all duplicated compilation chains (for that the new chain need to be sufficiently flexible to allow a one-line invocation from mlcontext/jmlc/dmlscript/debug)

I agree, this PR doesnot resolve the JIRA. In fact, this PR is deliberately scoped to only address the execute() part of the compilation chain for two reasons:
- The inconsistencies in execute() part could results in significant bugs when used via MLContext.
- I do not have free cycles (at least for next few weeks) to consolidate the entire compilation chain.

If you prefer, I can rename the PR to SYSTEMML-1047  and SYSTEMML-1506 and we can merge it in.  I am also OK with closing this PR if you or someone else wants to redesign the entire chain; in which case, this PR will just get in the way :)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1430/

PR #473, opened 2017-04-26
[SYSTEMML-1511] Tab completion for scripts using MLContext

deroneriksson:
Generate MLContext classes at build time to allow tab completion
and function execution from the Spark Shell for existing scripts in
the main jar file. The generated script classes are subclasses of Script.
The results of functions that return multiple values are encapsulated in
output classes. This avoids the generic type erasure issue associated with
returning results as tuples.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -632,9 +632,10 @@
 			<plugin>
 				<groupId>org.codehaus.mojo</groupId>
 				<artifactId>exec-maven-plugin</artifactId>
-				<version>1.5.0</version>
+				<version>1.6.0</version>
 				<executions>
 					<execution>
+						<id>install-dependency-for-integration-tests</id>
 						<phase>prepare-package</phase>
 						<goals>
 							<goal>java</goal>
@@ -644,6 +645,20 @@
 							<arguments>
 								<argument>true</argument> <!-- <argument>${skipTests}</argument> -->
 								<argument>${enableGPU}</argument>
+							</arguments>
+						</configuration>
+					</execution>
+					<execution>
+						<id>generate-classes-for-mlcontext</id>
+						<phase>prepare-package</phase>
+						<goals>
+							<goal>java</goal>
+						</goals>
+						<configuration>
+							<mainClass>org.apache.sysml.utils.GenerateClassesForMLContext</mainClass>
+							<arguments>
+								<arguments>scripts</arguments>
+								<arguments>target/classes</arguments>
 							</arguments>
 						</configuration>
 					</execution>
@@ -1043,7 +1058,7 @@
 					<plugin>
 						<groupId>org.codehaus.mojo</groupId>
 						<artifactId>exec-maven-plugin</artifactId>
-						<version>1.5.0</version>
+						<version>1.6.0</version>
 						<executions>
 							<execution>
 								<id>generate-python-package-1</id>
--- a/src/main/java/org/apache/sysml/api/DMLScript.java
+++ b/src/main/java/org/apache/sysml/api/DMLScript.java
@@ -178,6 +178,10 @@ public class DMLScript
 	public static boolean _suppressPrint2Stdout = false;  // flag that indicates whether or not to suppress any prints to stdout
 	public static boolean USE_LOCAL_SPARK_CONFIG = false; //set default local spark configuration - used for local testing
 	public static boolean _activeAM = false;
+	/**
+	 * If true, allow DMLProgram to be generated while not halting due to validation errors/warnings
+	 */
+	public static boolean VALIDATOR_IGNORE_ISSUES = false;
 
 	public static String _uuid = IDHandler.createDistributedUniqueID();
 	private static final Log LOG = LogFactory.getLog(DMLScript.class.getName());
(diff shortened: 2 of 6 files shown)


deroneriksson:
Generate data, train, and predict example using nn library in Spark Shell:
```
val ml = new MLContext(spark)
val mnist_lenet = ml.nn.examples.Mnist_lenet
val data = mnist_lenet.generate_dummy_data
val t = mnist_lenet.train(data.x,data.y,data.x,data.y,data.c,data.hin,data.win,1)
mnist_lenet.predict(data.x,data.c,data.hin,data.win,t.w1,t.b1,t.w2,t.b2,t.w3,t.b3,t.w4,t.b4)
```

Examples of function display:
```
// display function contents to end of comments
mnist_lenet.generate_dummy_data_
// display full function contents
println(mnist_lenet.generate_dummy_data__)
```

Example using a scripts/algorithms/A.dml file containing:
```
add = function(matrix[double] m, int num) return (matrix[double] n) { n = m + num }
sub = function(matrix[double] m, int num) return (matrix[double] n) { n = m - num }
mult = function(matrix[double] m, int num) return (matrix[double] n) { n = m * num }
```
Spark Shell:
```
val x = Array(Array(1.0, 2.0), Array(3.0, 4.0))
val a = ml.algorithms.A
for (i <- 1 to 3) { a.add(x,i).toDF.show }
for (i <- 1 to 3) { a.sub(a.mult(a.add(x,i),i),i).toDF.show }
```

akchinSTC:
Build successful

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1390/

nakul02:
This is awesome work @deroneriksson !
Could you please add a readme with some examples to try out?

deroneriksson:
Thanks for the great review @dusenberrymw!

1) For the mappings, a script is being mapped to a Java class, which is uppercased, while functions are being mapped to methods on the Java class, which by convention are lowercased. However, this is a Java convention and is not required. If allowing uppercased methods seems like it feels right from a usability perspective for data scientists, we can definitely do that easily.
2) I was attempting to create Scala-style minimalism, but something like _docs/__docs/_source/__source is definitely very easy to do and certainly clearer.
3) Currently the contents for each class are being pulled from the jar file. The file system is not overriding the content of the jar file. 1) and 2) are trivial but 3) is a bit more work but certainly achievable. We should be able to make script and method contents dynamic but the function signatures probably need to stay static due to java classloading issues. It might be good to do 3) as a separate PR.

dusenberrymw:
For (1), in particular, I was thinking it would be nice to follow the case of the function *parameters* and return variables defined in the DML.  So for `val dummy = clf.generate_dummy_data()`, I would expect to be able to use `dummy.X`, `dummy.Y`, `dummy.Hin`, etc., just as defined in the DML file.

For (2), I agree that the minimalism is great for the current solution.  Let's get some feedback from others as to whether minimalism or verbosity would be better here, as I'm not even completely sure one way or another.  For myself, I understand the current syntax, but I'm just not sure about others.  If we do go with the more verbose way, I think two underscores would be needed, as many of the DML functions will have underscores in the name, and thus a single underscore may just look like another DML function.

For (3) yeah the ultimate ideal would be realtime, dynamic generation of everything.  Of course that looks pretty hard, so I think keeping the function signatures static is fine.  Allowing for dynamic function bodies would definitely be extremely useful and needed.

deroneriksson:
Updated 1) and 2). The following now works:
```
val ml = new MLContext(spark)
val clf = ml.nn.examples.Mnist_lenet
val dummy = clf.generate_dummy_data
val dummyVal = clf.generate_dummy_data
val params = clf.train(dummy.X, dummy.Y, dummyVal.X, dummyVal.Y, dummy.C, dummy.Hin, dummy.Win, 1)
val probs = clf.predict(dummy.X, dummy.C, dummy.Hin, dummy.Win, params.W1, params.b1, params.W2, params.b2, params.W3, params.b3, params.W4, params.b4)
val perf = clf.eval(probs, dummy.Y)

clf.eval__docs
clf.eval__source
```

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1434/

dusenberrymw:
@deroneriksson Awesome, that LGTM.  Let's merge this, and then follow-up with (3) in a separate PR.  Also, for (3), I noticed that `"scripts/"` is being prepended to the script paths that are generated, so that causes those files to always be sourced from the JAR.  Interestingly, since files like `mnist_lenet.dml` import `nn/../..`, I believe the `nn` files would be loaded dynamically for invocations of functions from `mnist_lenet.dml`.

deroneriksson:
Adding dynamic loading of DML script content sounds like it could be very useful during DML development. Doing this in a separate PR sounds great.

I'm about to be away from any computers until Sunday. Feel free to merge this if you think it would be useful during the next few days. Otherwise, I will merge most likely on Sunday or Monday. I'm a little apprehensive to merge and then all of a sudden not be available in case there are any issues.

PR #477, opened 2017-04-30
[SYSTEMML-540] Additional tests to compare the accuracy of different convolution related operators with CuDNN

niketanpansare:
Though our JUnit tests have enough coverage, these scripts are more exhaustive as it compares the results under different settings:
- Sparsity
- With/without BLAS
- With/Without GPU

@dusenberrymw can you please review this PR ?

### DIFF
--- a/scripts/nn/test/compare_backends/gen_conv2d.dml
+++ b/scripts/nn/test/compare_backends/gen_conv2d.dml
@@ -1,0 +1,25 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+X = rand(rows=$N, cols=$C*$H*$W, sparsity=$sp, min=-0.5, max=1)
+w = rand(rows=$K, cols=$C*$R*$S, sparsity=$sp, min=-0.5, max=1)
+write(X, "input.mtx", format="binary")
+write(w, "filter.mtx", format="binary")
--- a/scripts/nn/test/compare_backends/gen_conv2d.dml
+++ b/scripts/nn/test/compare_backends/gen_conv2d.dml
@@ -20,6 +20,6 @@
 #-------------------------------------------------------------
 
 X = rand(rows=$N, cols=$C*$H*$W, sparsity=$sp, min=-0.5, max=1)
-w = rand(rows=$K, cols=$C*$R*$S, sparsity=$sp, min=-0.5, max=1)
+w = rand(rows=$F, cols=$C*$Hf*$Wf, sparsity=$sp, min=-0.5, max=1)
 write(X, "input.mtx", format="binary")
 write(w, "filter.mtx", format="binary")
--- a/scripts/nn/test/compare_backends/gen_conv2d_bwd_data.dml
+++ b/scripts/nn/test/compare_backends/gen_conv2d_bwd_data.dml
@@ -19,10 +19,11 @@
 #
 #-------------------------------------------------------------
 
-P = floor(($H + 2 * $pad - $R) / $stride) + 1
-Q = floor(($W + 2 * $pad - $S) / $stride) + 1
+Hout = floor(($H + 2 * $pad - $Hf) / $stride) + 1
+Wout = floor(($W + 2 * $pad - $Wf) / $stride) + 1
 
-w = rand(rows=$K, cols=$C*$R*$S, sparsity=$sp, min=-0.5, max=1)
-dout = rand(rows=$N, cols=$K*P*Q, sparsity=$sp, min=-0.5, max=1)
+w = rand(rows=$F, cols=$C*$Hf*$Wf, sparsity=$sp, min=-0.5, max=1)
+dout = rand(rows=$N, cols=$F*Hout*Wout, sparsity=$sp, min=-0.5, max=1)
+
 write(w, "filter.mtx", format="binary")
 write(dout, "dout.mtx", format="binary")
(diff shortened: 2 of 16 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1431/

scripts/nn/test/compare_backends/test_conv2d.sh
```diff
@@ -0,0 +1,67 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+export SPARK_HOME=~/spark-2.1.0-bin-hadoop2.7
```
dusenberrymw: I think it would be good to remove this and just assume the user will have the Spark `bin` folder in the path.  That way, this will be able to be run by anyone, rather than forcing another dev to have this specific Spark directory available.  Same thing for the other shell scripts.

scripts/nn/test/compare_backends/test_conv2d.sh
```diff
@@ -0,0 +1,67 @@
...
+N=5
+C=3
+H=28
+W=28
+K=32 
+R=3 
+S=3
+for sparsity in 0.1 0.2 0.5 0.6 0.9
+do
+	$SPARK_HOME/bin/spark-submit SystemML.jar -f gen_conv2d.dml -nvargs sp=$sparsity N=$N C=$C H=$H W=$W K=$K R=$R S=$S
```
dusenberrymw: Based on the above comment, let's update this to `spark-submit ...` so that others can easily use it.  Assumes that the user has placed the Spark `bin` folder on the path, which is reasonable.  Same thing for the other shell scripts.

scripts/nn/test/compare_backends/test_conv2d.sh
```diff
@@ -0,0 +1,67 @@
...
+	for stride in 1 2 3
+	do
+		for pad in 0 1 2
+		do
+			$SPARK_HOME/bin/spark-submit SystemML.jar -f test_conv2d.dml -nvargs stride=$stride pad=$pad out=out_cp.csv N=$N C=$C H=$H W=$W K=$K R=$R S=$S
+			$SPARK_HOME/bin/spark-submit --jars $jars SystemML.jar -f test_conv2d.dml -stats -gpu force  -nvargs stride=$stride pad=$pad out=out_gpu.csv N=$N C=$C H=$H W=$W K=$K R=$R S=$S
+			$SPARK_HOME/bin/spark-submit SystemML.jar -f compare.dml -args out_cp.csv out_gpu.csv "conv2d:stride="$stride",pad="$pad",sparsity="$sparsity
```
dusenberrymw: Likewise, let's update these to `spark-submit ...`.  Same thing for the other shell scripts.

scripts/nn/test/compare_backends/run_tests.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
dusenberrymw: Can you update this to `#!/usr/bin/env bash` to use the user's preferred bash, rather than a hardcoded system version?  Also applies to the other shell scripts.

scripts/nn/test/compare_backends/gen_conv2d.dml
```diff
@@ -0,0 +1,25 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+X = rand(rows=$N, cols=$C*$H*$W, sparsity=$sp, min=-0.5, max=1)
+w = rand(rows=$K, cols=$C*$R*$S, sparsity=$sp, min=-0.5, max=1)
```
dusenberrymw: Can we update the filter size variables here and elsewhere to match those used elsewhere in the library?  Specifically, `K` -> `F`, `R` -> `Hf`, and `S` -> `Wf`.  I think this would really help for others in the future who come to run these tests and reason about them.

scripts/nn/test/compare_backends/test_conv2d_bwd_filter.sh
```diff
@@ -0,0 +1,67 @@
...
+for sparsity in  0.1 0.2 0.5 0.6 0.9
+do
+	for stride in 1 2 3
+	do
+		for pad in 0 1 2
+		do
+			$SPARK_HOME/bin/spark-submit SystemML.jar -f gen_conv2d_bwd_filter.dml -nvargs sp=$sparsity N=$N C=$C H=$H W=$W K=$K R=$R S=$S stride=$stride pad=$pad
+			$SPARK_HOME/bin/spark-submit SystemML.jar -f test_conv2d_bwd_filter.dml -nvargs stride=$stride pad=$pad out=out_cp.csv N=$N C=$C H=$H W=$W K=$K R=$R S=$S
+			$SPARK_HOME/bin/spark-submit --jars $jars SystemML.jar -f test_conv2d_bwd_filter.dml -stats -gpu force -nvargs stride=$stride pad=$pad out=out_gpu.csv N=$N C=$C H=$H W=$W K=$K R=$R S=$S
+			$SPARK_HOME/bin/spark-submit SystemML.jar -f compare.dml -args out_cp.csv out_gpu.csv "conv2d_backward_filter:stride="$stride",pad="$pad",sparsity="$sparsity
```
dusenberrymw: Can you add a few quick comments in this section that basically say that you're (1) generating data, (2) running a test in CPU mode, (3) running the same test in a GPU mode, and (4) comparing the CPU vs GPU results to make sure they are the same?  Same thing for the other shell scripts.

scripts/nn/test/compare_backends/gen_conv2d_bwd_data.dml
```diff
@@ -0,0 +1,28 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+P = floor(($H + 2 * $pad - $R) / $stride) + 1
+Q = floor(($W + 2 * $pad - $S) / $stride) + 1
```
dusenberrymw: Similar to updating the variable names for filter sizes, can we also update the output size variables to be the same as the rest of the library?  Specifically, `P` -> `Hout`, and `Q` -> `Wout`?  Might also help to update `H` -> `Hin`, and `W` -> `Win`.  Once again, by maintaining consistency, I think this will help others in the future using and reasoning about these tests.  This would apply to the rest of the files too.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1443/

PR #481, opened 2017-05-03
[SYSTEMML-1572] Enable native BLAS on remote executors

niketanpansare:
@mboehm7 can you please review this PR ?

@bertholdreinwald


mboehm7:
I'll try to have a look next week. In the meantime, you might want to revisit the integration approach - the large number of files you had to modify is already an indicator that this is probably not a good approach. 

It should be fairly simply: create a single static call which dispatches between matrix block (which should not call libnative at all) and libnative, initialize once, and call it for each matrix multiplication.

niketanpansare:
@mboehm7 This approach allows the matrix multiplication on remote executor to respect the configuration flag `native.blas`. Since the configuration is not sent to the executor, not sure how single static call will enable that. But, we can discuss this next week after you have had chance to look at it :)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1435/

bertholdreinwald:
Not working as despite    <native.blas>true</native.blas> in SystemML-config.xml, another env var disables it.

17/05/03 09:07:33 WARN utils.NativeHelper: Using internal Java BLAS as native BLAS support is disabled by the environment variable 'SYSTEMML_BLAS='.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1438/

mboehm7:
Thanks but closing as outdated (and already working).

PR #484, opened 2017-05-06
[SYSTEMML-1561] Improve constant folding during compilation

dusenberrymw:
In our `nn` library, our convolution and pooling layers have to pass
around the spatial dimensions (height and width) of the images that are
stretched out into rows of the input/output matrices.  These output
dimensions are computed within the forward functions of the above
layers as small scalar equations.  From a mathematical standpoint,
these sizes can be determined at compile time, and it is nice to have
these size equations in DML (v.s. hiding them inside the engine within
built-in functions).  However, we do not currently evaluate these
expressions during compilation, and thus we are left with unknown sizes
even during recompilation.  This naturally leads to max memory estimates
and thus often leads to unnecessary distributed runtime ops rather than
simple CP ones.

Essentially, this commit fixes that issue by rerunning static rewrites
\+ IPA again immediately after the initial IPA pass as a kind of "second
chance".  By doing this, we're able to apply this constant folding
rewrite for this scenario.  This make sense because during the initial
static rewrite pass, we can't apply constant folding to the `Hout`,
`Wout`, etc. DAGs due to the leaf nodes being scalar transient reads.
After IPA with the new scalar replacement, these DAGs will consist
entirely of operations on literal leaf nodes, and thus will be eligible
for constant folding.  Then, after that second pass of static rewrites,
we can benefit from IPA again by being able to now perform scalar
replacement for functions/other DAGs that consume the `Hout`, `Wout`,
etc. DAGs, which are now literals.


dusenberrymw:
cc @mboehm7 Can you review this?
cc @niketanpansare This will help to speed up our deep learning scripts and prevent unknown sizes from being propagated.  I'm seeing hundreds of seconds of improvement on ur MNIST LeNet example due to a drop in the number of compiled & executed Spark instructions.  Can you also review this?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1444/

dusenberrymw:
ping @mboehm7

mboehm7:
sorry - I didn't get the chance to try it out yet, but I'll look into in the next days.

src/main/java/org/apache/sysml/hops/OptimizerUtils.java
```diff
@@ -157,6 +157,18 @@
...
+	 * Enables an additional "second chance" pass of static rewrites + IPA after the initial pass of
+	 * IPA.  Without this, there are many situations in which sizes will remain unknown even after
+	 * recompilation, thus leading to distributed ops.  With the second chance enabled, sizes in
+	 * these situations can be determined.  For example, the alternation of constant folding
+	 * (static rewrite) and scalar replacement (IPA) can allow for size propagation without dynamic
+	 * rewrites or recompilation due to replacement of scalars with literals during IPA, which
+	 * enables constant folding of sub-DAGs of literals during static rewrites, which in turn allows
+	 * for scalar propagation during IPA.
+	 */
+	public static boolean ALLOW_IPA_SECOND_CHANCE = true;
```
mboehm7: please modify OptimizerUtils.constructCompilerConfig to set this constant according to the chosen optimization level

src/main/java/org/apache/sysml/parser/DMLTranslator.java
```diff
@@ -267,10 +266,20 @@ public void rewriteHopsDAG(DMLProgram dmlp)
...
+			if (OptimizerUtils.ALLOW_IPA_SECOND_CHANCE) {
+				// SECOND CHANCE:
+				// Rerun static rewrites + IPA to allow for further improvements, such as making use
+				// of constant folding (static rewrite) after scalar -> literal replacement (IPA),
+				// and then further scalar -> literal replacement (IPA).
+				rewriter.rewriteProgramHopDAGs(dmlp);
+				resetHopsDAGVisitStatus(dmlp);
+				ipa.analyzeProgram(dmlp);
+				resetHopsDAGVisitStatus(dmlp);
+			}
```
mboehm7: I would recommend to move this into InterProceduralAnalysis.analyzeProgram in order to reuse the created function call graph and other invariant information. In case you want to leaf it as is, please remove the TODO from InterProceduralAnalysis, line 218

src/main/java/org/apache/sysml/hops/OptimizerUtils.java
dusenberrymw: Which optimization levels should it be enabled for?

mboehm7: I would recommend >=2 (which is what your code currently implicitly does because it is conditioned on inter procedural analysis) but it would be good to make it explicit

dusenberrymw: Done.  I set the flag to false for levels < 2, as is done for the existing flags.

src/main/java/org/apache/sysml/parser/DMLTranslator.java
dusenberrymw: Well I think it might be cleaner to leave it up at this level because other wise I'd have to pass in the `rewriter` into `InterProceduralAnalysis.analyzeProgram`, or create another one, both of which would just get messy.  I removed the TODO message.  Does that work?

dusenberrymw:
@mboehm7 Can this be merged in now?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1484/

dusenberrymw:
Jenkins tests passed, will merge.

PR #487, opened 2017-05-08
[SYSTEMML-1589] Check for the empty block case

niketanpansare:
@mboehm7  can you please review this ?


mboehm7:
LGTM - thanks @niketanpansare ; Btw, the unrelated change from isEmptyBlock() (i.e., isEmptyBlock(true)) to isEmpty (i.e., isEmptyBlock(false)) is the preferred way because it avoids unnecessary scans of allocated dense or sparse inputs with nnz=0.

niketanpansare:
Thanks @mboehm7  ... I will merge :)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1450/

PR #490, opened 2017-05-09
Added conv2d transpose layer along with associated tests

prithvirajsen:
@dusenberrymw @niketanpansare @bertholdreinwald 
Can you please review this PR?

Besides including the forward, backward functions for conv2d transpose, this PR also includes tests (for the forward function and the gradient checks). 

Also, I've included a weight initializer for bilinear interpolation. This is useful if someone wants to upsample using this layer.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1454/

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+   *  - C: Number of input channels (dimensionality of depth).
+   *  - Hin: Input height.
+   *  - Win: Input width.
+   *  - Hf: Filter height.
+   *  - Wf: Filter width.
+   *  - strideh: Stride over height.
+   *  - stridew: Stride over width.
+   *  - padh: Padding for top and bottom sides.
+   *  - padw: Padding for left and right sides.
+   *  - out_padh: extra padding for top side. This should
```
niketanpansare: @prithvirajsen can you please explain how to compute `out_padh` from Caffe spec ?

Here is an example Caffe layer:
```
layer {
  name: "upconv_u3d_u2a"
  type: "Deconvolution"
  bottom: "u3d"
  top: "u2a"
  convolution_param {
    num_output: 256
    bias_term: false
    pad: 1
    kernel_size: 4
    group: 256
    stride: 2
    weight_filler {
      type: "bilinear"
    }
  }
}
``` 

and here is the caffe's documentation: http://caffe.berkeleyvision.org/tutorial/layers/deconvolution.html

It is quite likely that caffe might be using some default for this.

prithvirajsen: I don't see any options for Caffe's Deconvolution layer to express the extra padding args. Perhaps set them to 0 in Caffe2DML?

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+  factor_up = ceil(K / 2)
+  center = (2 * factor_up - factor_up %% 2 - 1) / 2 / factor_up
+  vect = 1 - abs(seq(0, K-1) / factor_up - center)
+  weights = matrix(vect %*% t(vect), rows=1, cols=K*K)
+
+  if(C > 1){
+    repl_weights = matrix(1, rows=C, cols=1) %*% cbind(weights, matrix(0, rows=1, cols=C*K*K))
+    repl_weights_row = matrix(repl_weights, rows=1, cols=C*(C+1)*K^2)
+    W = matrix(repl_weights_row[1,1:(C*K)^2], rows=C, cols=C*K^2)
+  }else W = weights
```
dusenberrymw: Can we replace this if/else statement and just replicate W to shape `(C, C*K*K)` always?  If `C=1`, the optimizer will just remove those statements, so no harm.  This way, this function will be a bit clearer.  I think the following would work (and the comments help):

```
...
channels = matrix(1, rows=C, cols=1)  # for replicating K*K kernel for each channel
weights = matrix(vect %*% t(vect), rows=1, cols=K*K)  # create K*K kernel
weights = matrix(channels %*% weights, rows=1, cols=C*K*K)  # replicate K*K kernel for each input channel
weights = channels %*% weights  #  output same number of channels; shape (C, C*K*K)
```

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+   * via bilinear interpolation, bias is set to 0.
+   * 
+   * Inputs:
+   *  - C: Number of input channels (dimensionality of depth).
+   *  - K: Kernel size (upsampling requires a square filter
+   *      of size K X K).
+   *
+   * Outputs:
+   *  - W: Weights, of shape (F, C*Hf*Wf).
+   *  - b: Biases, of shape (F, 1).
```
dusenberrymw: Can you update the docs here with the correct output sizes?

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+ */
+ 
+forward = function(matrix[double] X, matrix[double] W, matrix[double] b,
+                   int C, int Hin, int Win, int Hf, int Wf,
+		           int strideh, int stridew, int padh, int padw,
+		           int out_padh, int out_padw)
+    return (matrix[double] out, int Hout, int Wout){
+  /*
+   * Computes the forward pass for a 2D spatial transpose convolutional 
+   * layer with F filters.  The input data has N examples, each represented
```
dusenberrymw: nitpick: can you move "represented" to the next line?  I've tried to keep all docs <= 72 characters in length according to other style guides.

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+/*
+ * 2D Transpose convolutional layer.
+ *
+ * Utilizes built-in convolution operators for higher performance.
+ */
+ 
+forward = function(matrix[double] X, matrix[double] W, matrix[double] b,
+                   int C, int Hin, int Win, int Hf, int Wf,
+		           int strideh, int stridew, int padh, int padw,
+		           int out_padh, int out_padw)
```
dusenberrymw: Looks like a few tabs were accidentally inserted here.  Can you update them each to 2 spaces?

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+   *  - out: Outputs, of shape (N, F*Hout*Wout).
+   *  - Hout: Output height.
+   *  - Wout: Output width.
+   */
+  N = nrow(X)
+  F = nrow(W)
+  Hout = strideh * (Hin-1) - 2*padh + Hf + out_padh
+  Wout = stridew * (Win-1) - 2*padw + Wf + out_padw
+
+  out = conv2d_backward_data(W, X, stride=[strideh,stridew], padding=[padh,padw],
```
dusenberrymw: Can you add a comment line or two here that explains to future viewers why we're able to use `conv2d_backward_data` to implement transpose convolution?  Of course, we know why it works, but a comment may help others in the future reason about what this layer is doing when they're building models.

scripts/nn/layers/conv2d_transpose.dml
```diff
@@ -0,0 +1,172 @@
...
+   *  - dW: Gradient wrt `W`, of shape (F, C*Hf*Wf).
+   *  - db: Gradient wrt `b`, of shape (F, 1).
+   */
+  N = nrow(X)
+  F = nrow(W)
+
+  dW = conv2d_backward_filter(dout, X, stride=[strideh,stridew], padding=[padh,padw],
+                              input_shape=[N,C,Hout,Wout], filter_shape=[F,C,Hf,Wf])
+
+  dX = conv2d(dout, W, input_shape=[N,C,Hout,Wout], filter_shape=[F,C,Hf,Wf],
```
dusenberrymw: Similar to the comment in `forward`, can you add a few comments here for how we're able to use `conv2d_backward_filter` (since we "hack" it with a swapped `dout` & `X`)  and `conv2d` to implement the backward pass of transpose convolution?  Again, we of course know why this is possible, but it may help others in the future.

scripts/nn/test/grad_check.dml
```diff
@@ -616,6 +617,114 @@ conv2d_simple = function() {
...
+  /*
+   * Gradient check for the 2D convolution transpose layer.
+   */
+  print("Grad checking the 2D convolution transpose layer with L2 loss.")
+
+  N = 1
+  C = 1
+  Hin = 3
+  Win = 3
+  F = 1
```
dusenberrymw: Can you increase `N`, `C`, and `F` to `2`?  That will help to ensure that there are no issues related to the shapes that may be masked when those are 1.

scripts/nn/test/grad_check.dml
```diff
@@ -616,6 +617,114 @@ conv2d_simple = function() {
...
+  stride = 2
+  pad = 1
+  out_pad = 1
+
+  X = rand(rows=N, cols=C*Hin*Win)
+
+  [W,b] = conv2d_transpose::init(F, C, Hf, Wf)
+  
+  [out, Hout, Wout] = 
+ 	conv2d_transpose::forward(X, W, b, C, Hin, Win, Hf, Wf, stride, stride, pad, pad, out_pad, out_pad)
```
dusenberrymw: Can you update the formatting for this function so that no lines, such as this one, are longer than 100 characters?  In addition to keep docs <=72, I've kept all lines <=100, based on other style guides.

scripts/nn/test/grad_check.dml
```diff
@@ -616,6 +617,114 @@ conv2d_simple = function() {
...
+      [outph, Hout, Wout] = 
+      	conv2d_transpose::forward(X, W, b, C, Hin, Win, Hf, Wf, stride, stride, pad, pad, out_pad, out_pad)
+
+	  lossph = l2_loss::forward(outph, y)
+
+	  X[i,j] = old
+
+	  dX_num = (lossph-lossmh) / (2*h)
+
+	  rel_error = test_util::check_rel_grad_error(as.scalar(dX[i,j]), dX_num, lossph, lossmh)
```
dusenberrymw: Looks like some tabs were accidentally added in this for loop.  Can you replace them each with 2 spaces?  Also, can you please remove the extra new lines?

Same thing for the for loops below.

scripts/nn/layers/conv2d_transpose.dml
prithvirajsen: Thanks for all the comments. I fixed all of them barring the one about the 'if' stmt. I've added more comments to explain what a channel-wise independent multi-channel upsampling filter looks like and why the 'if' stmt is needed. Please take a look when you get a chance.

prithvirajsen: Actually, I take that back. The 'if' condition can be commented out. The lines within the 'if' block will work even in the case of C=1. It just results in redundant computation (I don't think the optimizer is smart enough to recognize this). On the other hand, I don't think this bit of redundant computation will ever result in a bottleneck.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1457/

PR #491, opened 2017-05-11
[SYSTEMML-1598] Output statistics via MLContext to stdout

deroneriksson:
If ml.setStatistics(true), output statistics to standard output
rather than to log4j.

### DIFF
--- a/src/main/java/org/apache/sysml/api/ScriptExecutorUtils.java
+++ b/src/main/java/org/apache/sysml/api/ScriptExecutorUtils.java
@@ -16,10 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.apache.sysml.api;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import org.apache.sysml.api.mlcontext.MLContext;
+import org.apache.sysml.api.mlcontext.MLContextException;
+import org.apache.sysml.api.mlcontext.ScriptExecutor;
 import org.apache.sysml.conf.DMLConfig;
 import org.apache.sysml.hops.codegen.SpoofCompiler;
 import org.apache.sysml.runtime.DMLRuntimeException;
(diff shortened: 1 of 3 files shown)


deroneriksson:
Example:
```
spark-shell --executor-memory 4G --driver-memory 4G --driver-class-path=target/SystemML.jar

val ml = new MLContext(spark)
val mnist_lenet = ml.nn.examples.Mnist_lenet
ml.setStatistics(true)
val data = mnist_lenet.generate_dummy_data
ml.setStatistics(false)
val data = mnist_lenet.generate_dummy_data
```

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1458/

mboehm7:
This touches upon a broader question, i.e., should all explain/stats output go to stdout or the logs? Just to give the history: We had deployments, where SystemML was called in a separate process and writing to stdout would clog up the stdout queue if nobody is reading from it. Hence all output except print/stop goes to the logs. That's also the reason why our scripts have options to write any debug output to text files instead of printing it. However, nowadays such deployments are rather uncommon and we could reconsider this in general.

deroneriksson:
When working in an interactive environment such as Spark Shell, outputting to stdout I feel is highly preferrable. I'm less certain about other environments.

As an illustration, the above example generates:
```
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 2.1.0
      /_/
         
Using Scala version 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60)

scala> val ml = new MLContext(spark)

Welcome to Apache SystemML!

ml: org.apache.sysml.api.mlcontext.MLContext = org.apache.sysml.api.mlcontext.MLContext@4930213b

scala> val mnist_lenet = ml.nn.examples.Mnist_lenet
mnist_lenet: org.apache.sysml.scripts.nn.examples.Mnist_lenet =
Inputs:
None

Outputs:
None

scala> ml.setStatistics(true)

scala> val data = mnist_lenet.generate_dummy_data
SystemML Statistics:
Total elapsed time:		0.115 sec.
Total compilation time:		0.000 sec.
Total execution time:		0.115 sec.
Number of compiled Spark inst:	0.
Cache hits (Mem, WB, FS, HDFS):	3/0/0/0.
Cache writes (WB, FS, HDFS):	5/0/0.
Cache times (ACQr/m, RLS, EXP):	0.000/0.000/0.001/0.000 sec.
Spark ctx create time (lazy):	0.000 sec.
Spark trans times (par,bc,col):	0.000/0.000/0.000 secs.
Total JIT compile time:		0.067 sec.
Total JVM GC count:		15.
Total JVM GC time:		0.815 sec.
-- 1) 	rand 	0.103 sec 	2	
-- 2) 	round 	0.007 sec 	1	
-- 3) 	ctableexpand 	0.004 sec 	1	
-- 4) 	createvar 	0.001 sec 	5	
-- 5) 	seq 	0.000 sec 	1	
-- 6) 	rmvar 	0.000 sec 	5	
-- 7) 	cpvar 	0.000 sec 	2	
-- 8) 	assignvar 	0.000 sec 	3	

data: org.apache.sysml.scripts.nn.examples.mnist_lenet.Generate_dummy_data_output =
X (Matrix): MatrixObject: scratch_space//_p60985_192.168.1.104//_t0/temp0_0, [1024 x 784, nnz=802816, blocks (1000 x 1000)], binaryblock, dirty
Y (Matrix): MatrixObject: scratch_space//_p60985_192.168.1.104//_t0/temp4_4, [1024 x 10, nnz=1024, blocks (1000 x 1000)], binaryblock, dirty
C (long): 1
Hin (long): 28
Win (long): 28

scala> ml.setStatistics(false)

scala> val data = mnist_lenet.generate_dummy_data
data: org.apache.sysml.scripts.nn.examples.mnist_lenet.Generate_dummy_data_output =
X (Matrix): MatrixObject: scratch_space//_p60985_192.168.1.104//_t0/temp5_5, [1024 x 784, nnz=802816, blocks (1000 x 1000)], binaryblock, dirty
Y (Matrix): MatrixObject: scratch_space//_p60985_192.168.1.104//_t0/temp9_9, [1024 x 10, nnz=1024, blocks (1000 x 1000)], binaryblock, dirty
C (long): 1
Hin (long): 28
Win (long): 28

scala>
```

We could even potentially make it configurable whether the output goes to stdout and/or logging.

nakul02:
Even `explain` via `MLContext` [is printed to](https://github.com/apache/incubator-systemml/blob/master/src/main/java/org/apache/sysml/api/mlcontext/ScriptExecutor.java#L188)`System.out`. 

Would it make sense to provide an extra method - `ml.showStatistics()` - that would print out the statistics to stdout and by default it could print to log4j?

deroneriksson:
Until a few days ago, statistics also printed to System.out. This PR is just reverting to that behavior.

deroneriksson:
BTW @mboehm7 brings up a good point about statistics/explain and how they should be handled. This probably will be an issue to address in the consolidated compilation chain, where statistics/explain can potentially be configurable based on how SystemML is being run or configurable by user preference.

deroneriksson:
@nakul02 If we make it configurable via MLContext, I feel strongly that the default should be stdout.

nakul02:
@deroneriksson - that makes sense. The user specified the `-stats` and `-explain` flag or use the `set{Statistics,Explain}` flags knowing full well what he/she wants to see printed to screen. 
I agree, the defaults should be stdout.
Also, since @mboehm7 said that the legacy deployment scenarios are uncommon, defaulting to stdout is probably ok.

deroneriksson:
Cool, I will go ahead and merge.

BTW, I like the idea of potentially making the output of stats/explain configurable in the near future.

deroneriksson:
Actually, it looks like there were test failures on Jenkins. I will investigate.

dusenberrymw:
Yeah stdout should be the default, but we should definitely think about a simple configuration, such as an `ml.setOutputType(...)`.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1461/

PR #492, opened 2017-05-11
[SYSTEMML-1479] Integrated DeConvolution and Concat layer in Caffe2DML

niketanpansare:
@prithvirajsen @dusenberrymw can you please review this PR ?

@bertholdreinwald End-to-end integration of a network using them (for example: Unet) is still remaining.

### DIFF
--- a/src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
+++ b/src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
@@ -156,14 +156,133 @@ class Data(val param:LayerParameter, val id:Int, val net:CaffeNetwork, val numCh
 class BatchNorm(val param:LayerParameter, val id:Int, val net:CaffeNetwork) extends CaffeLayer with HasWeight with HasBias {
   // val scale =  
   override def sourceFileName = "batch_norm2d"
+  /*
+   * Initialize the parameters of this layer.
+   *
+   * Note: This is just a convenience function, and parameters
+   * may be initialized manually if needed.
+   *
+   * Inputs:
+   *  - C: Number of input channels (dimensionality of input depth).
+   *
+   * Outputs:
+   *  - gamma: Scale parameters, of shape (C, 1).
+   *  - beta: Shift parameters, of shape (C, 1).
+   *  - ema_mean: Exponential moving average of the mean, of
+   *      shape (C, 1).
+   *  - ema_var: Exponential moving average of the variance, of
+   *      shape (C, 1).
+   */
   override def init(dmlScript:StringBuilder) = invokeInit(dmlScript, List[String](gamma, beta, ema_mean, ema_var), numChannels)
   var update_mean_var = true
+  /*
+   * Computes the forward pass for a 2D (spatial) batch normalization
+   * layer.  The input data has N examples, each represented as a 3D
+   * volume unrolled into a single vector.
+   *
+   * A spatial batch normalization layer uses the per-channel sample
+   * mean and per-channel uncorrected sample variance during training
+   * to normalize each channel of the input data.  Additionally, it
+   * introduces learnable parameters (gamma, beta) to control the
+   * amount of normalization.
+   *
+   *   `y = ((x-mean) / sqrt(var+eps)) * gamma + beta`
+   *
+   * This implementation maintains exponential moving averages of the
+   * mean and variance during training for use during testing.
+   *
+   * Reference:
+   *  - Batch Normalization: Accelerating Deep Network Training by
+   *    Reducing Internal Covariate Shift, S. Ioffe & C. Szegedy, 2015
+   *    - https://arxiv.org/abs/1502.03167
+   *
+   * Inputs:
+   *  - X: Inputs, of shape (N, C*Hin*Win).
+   *  - gamma: Scale parameters, of shape (C, 1).
+   *  - beta: Shift parameters, of shape (C, 1).
+   *  - C: Number of input channels (dimensionality of input depth).
+   *  - Hin: Input height.
+   *  - Win: Input width.
+   *  - mode: 'train' or 'test' to indicate if the model is currently
+   *      being trained or tested.  During training, the current batch
+   *      mean and variance will be used to normalize the inputs, while
+   *      during testing, the exponential average of the mean and
+   *      variance over all previous batches will be used.
+   *  - ema_mean: Exponential moving average of the mean, of
+   *      shape (C, 1).
+   *  - ema_var: Exponential moving average of the variance, of
+   *      shape (C, 1).
+   *  - mu: Momentum value for moving averages.
+   *      Typical values are in the range of [0.9, 0.999].
+   *  - epsilon: Smoothing term to avoid divide by zero errors.
+   *      Typical values are in the range of [1e-5, 1e-3].
+   *
+   * Outputs:
+   *  - out: Outputs, of shape (N, C*Hin*Win).
+   *  - ema_mean_upd: Updated exponential moving average of the mean,
+   *      of shape (C, 1).
+   *  - ema_var_upd: Updated exponential moving average of the variance,
+   *      of shape (C, 1).
+   *  - cache_mean: Cache of the batch mean, of shape (C, 1).
+   *      Note: This is used for performance during training.
+   *  - cache_var: Cache of the batch variance, of shape (C, 1).
+   *      Note: This is used for performance during training.
+   *  - cache_norm: Cache of the normalized inputs, of
+   *      shape (C, N*Hin*Win). Note: This is used for performance
+   *      during training.
+   */
   def forward(dmlScript: StringBuilder, isPrediction: Boolean): Unit = {
     val mode = if(isPrediction) "\"test\"" else "\"train\""
     invokeForward(dmlScript, List[String](out, withSuffix(ema_mean), withSuffix(ema_var), withSuffix(cache_mean), withSuffix(cache_var), withSuffix(cache_norm)), 
         X, gamma, beta, numChannels, Hin, Win, mode, ema_mean, ema_var,  ma_fraction, eps)  
   }
-  
+  /*
+   * Computes the backward pass for a 2D (spatial) batch normalization
+   * layer.
+   *
+   * Inputs:
+   *  - dout: Gradient wrt `out` from upstream, of shape (N, C*Hin*Win).
+   *  - out: Outputs from the forward pass, of shape (N, C*Hin*Win).
+   *  - ema_mean_upd: Updated exponential moving average of the mean
+   *      from the forward pass, of shape (C, 1).
+   *  - ema_var_upd: Updated exponential moving average of the variance
+   *      from the forward pass, of shape (C, 1).
+   *  - cache_mean: Cache of the batch mean from the forward pass, of
+   *      shape (C, 1).  Note: This is used for performance during
+   *      training.
+   *  - cache_var: Cache of the batch variance from the forward pass,
+   *      of shape (C, 1).  Note: This is used for performance during
+   *      training.
+   *  - cache_norm: Cache of the normalized inputs from the forward
+   *      pass, of shape (C, N*Hin*Win).  Note: This is used for
+   *      performance during training.
+   *  - X: Input data matrix to the forward pass, of
+   *      shape (N, C*Hin*Win).
+   *  - gamma: Scale parameters, of shape (C, 1).
+   *  - beta: Shift parameters, of shape (C, 1).
+   *  - C: Number of input channels (dimensionality of input depth).
+   *  - Hin: Input height.
+   *  - Win: Input width.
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -212,15 +344,131 @@ class Elementwise(val param:LayerParameter, val id:Int, val net:CaffeNetwork) ex
...
+   *
+   * Inputs:
+   *  - n_i * c_i * h * w for each input blob i from 1 to K.
+   *
+   * Outputs:
+   *  - out: Outputs, of shape 
+   *    - if axis = 0: (n_1 + n_2 + ... + n_K) * c_1 * h * w, and all input c_i should be the same.
+   *    - if axis = 1: n_1 * (c_1 + c_2 + ... + c_K) * h * w, and all input n_i should be the same.
+   */
+  override def forward(dmlScript: StringBuilder, isPrediction: Boolean): Unit = {
```
dusenberrymw: We should just add a new `nn/layers/concat.dml` layer, rather than embed it in this Scala code.  That way, we can test it for correctness using the same grad checks the rest of the `nn` library uses, and can use it in other contexts.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -249,53 +497,217 @@ class SoftmaxWithLoss(val param:LayerParameter, val id:Int, val net:CaffeNetwork
 }
 
 class ReLU(val param:LayerParameter, val id:Int, val net:CaffeNetwork) extends CaffeLayer {
+  // TODO: Leaky ReLU: negative_slope [default 0]: specifies whether to leak the negative part by multiplying it with the slope value rather than setting it to 0.
```
dusenberrymw: Yeah, we should add this as another layer.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -249,53 +497,217 @@ class SoftmaxWithLoss(val param:LayerParameter, val id:Int, val net:CaffeNetwork
...
+   *  - padw: Padding for left and right sides.
+   *      A typical value is 0.
+   *
+   * Outputs:
+   *  - out: Outputs, of shape (N, C*Hout*Wout).
+   *  - Hout: Output height.
+   *  - Wout: Output width.
+   */
   override def forward(dmlScript:StringBuilder, isPrediction:Boolean) = 
     invokeForward(dmlScript, List[String](out, "ignoreHout_"+id, "ignoreWout_"+id),
```
dusenberrymw: By the way, with the new IPA scalar replacement + constant folding updates I've made, it's no longer necessary to compute `Hout` and `Wout` within Caffe2DML.  Once the constant folding PR is merged, they will be computed and replaced with literals during the initial compilation, thus no performance hit.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -304,54 +716,292 @@ class MaxPooling(val param:LayerParameter, val id:Int, val net:CaffeNetwork) ext
...
+    invokeBackward(dmlScript, outSuffix, List[String]("dOut" + id, dWeight, dBias), 
+        dout, Hout, Wout, X, weight, bias, numChannels, Hin, Win, kernel_h, kernel_w, stride_h, stride_w, pad_h, pad_w)
+  // n * c_o * h_o * w_o, where h_o = (h_i + 2 * pad_h - kernel_h) / stride_h + 1 and w_o likewise.
+  override def outputShape = ( numChannels, Hout, Wout )
+  // -------------------------------------------------
+  def numChannels = bottomLayerOutputShape._1
+  def Hin = bottomLayerOutputShape._2
+  def Win = bottomLayerOutputShape._3
+  // Hout = strideh * (Hin-1) - 2*padh + Hf + out_padh
+  def Hout:String =  try {
```
dusenberrymw: Once the constant folding PR is in, you can just use the `Hout` and `Wout` returned from `forward`.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -304,54 +716,292 @@ class MaxPooling(val param:LayerParameter, val id:Int, val net:CaffeNetwork) ext
...
+  // pad (or pad_h and pad_w) [default 0]: specifies the number of pixels to (implicitly) add to each side of the input
+  def pad_h =   if(convParam.hasPadH) convParam.getPadH.toString 
+                   else if(convParam.getPadCount > 0)  convParam.getPad(0).toString 
+                   else "0"
+  def pad_w =   if(convParam.hasPadW) convParam.getPadW.toString 
+                   else if(convParam.getPadCount > 0)  convParam.getPad(0).toString 
+                   else "0"
+}
+
+class DeConvolution(val param:LayerParameter, val id:Int, val net:CaffeNetwork) extends CaffeLayer with HasWeight with HasBias {
```
dusenberrymw: What do you think about updating this to, say, `Convolution2DTranspose`?  I know that Caffe used "deconvolution", but it would be nice to not perpetuate that incorrect naming.  Also, Caffe2 now uses "ConvTranspose" instead of "Deconvolution" --> https://caffe2.ai/docs/operators-catalogue.html#convtranspose.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -99,8 +99,20 @@ trait CaffeLayer extends BaseDMLGenerator {
...
+  // Assumption: the first variable of resultVariables is always dX
   def invokeBackward(dmlScript:StringBuilder, outSuffix:String, resultVariables:List[String],  arguments:String*):Unit = {
-    invoke(dmlScript, sourceFileName + "::", resultVariables.map(_ + outSuffix), "backward", arguments.toList)
+    invoke(dmlScript, sourceFileName + "::", resultVariables.map(_ + outSuffix), "backward", arguments.toList, false)
+    val bottomLayerIDs = net.getBottomLayers(param.getName).map(l => net.getCaffeLayer(l).id)
+    dmlScript.append("; ")
+    bottomLayerIDs.map(bottomLayerID => dmlScript.append( dX(bottomLayerID) + outSuffix + " = " + resultVariables(0) + outSuffix + "; "))
+    dmlScript.append("\n")
+  }
+  def assignDoutToDX(dmlScript:StringBuilder, outSuffix:String):Unit = {
```
dusenberrymw: What is this used for?

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -190,8 +321,9 @@ class Scale(val param:LayerParameter, val id:Int, val net:CaffeNetwork) extends
   if(!param.getScaleParam.getBiasTerm) throw new LanguageException("Add \"scale_param { bias_term: true }\" to the layer " + param.getName)
   override def sourceFileName = null
   override def init(dmlScript: StringBuilder): Unit = {}
+  // TODO: Generalize this !!
   def forward(dmlScript: StringBuilder, isPrediction: Boolean): Unit = assign(dmlScript, out, X)
-  override def backward(dmlScript: StringBuilder, outSuffix:String): Unit = assign(dmlScript, dX + outSuffix, dout)
+  override def backward(dmlScript: StringBuilder, outSuffix:String): Unit = assignDoutToDX(dmlScript, outSuffix)
```
dusenberrymw: Shouldn't we either just throw errors for Caffe2DML layers that aren't implemented yet?  Otherwise, I'm afraid we'll accidentally mask issues.  This layer can be mapped to the [`scale_shift2d`](https://github.com/apache/incubator-systemml/blob/master/scripts/nn/layers/scale_shift2d.dml) layer.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
```diff
@@ -144,16 +156,135 @@ class Data(val param:LayerParameter, val id:Int, val net:CaffeNetwork, val numCh
...
+   *  - C: Number of input channels (dimensionality of input depth).
+   *
+   * Outputs:
+   *  - gamma: Scale parameters, of shape (C, 1).
+   *  - beta: Shift parameters, of shape (C, 1).
+   *  - ema_mean: Exponential moving average of the mean, of
+   *      shape (C, 1).
+   *  - ema_var: Exponential moving average of the variance, of
+   *      shape (C, 1).
+   */
```
dusenberrymw: Thanks for adding the docs everywhere.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1460/

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: I thought of adding concat.dml, but it is difficult to generalize it because we donot support arbitrary number of arguments to a UDF.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Hout and Wout also needs to be computed for reading .caffemodel file at compile time.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
dusenberrymw: How about we add it for 2 arguments, and then Caffe2DML can just call it repeatedly.  Similar to what is being done now, except nest calls to `concat::forward`.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Btw, does Hout  and Wout cause issue with parfor based approach ?

dusenberrymw: No, because with parfor distrib sgd, the layers are called (forward & backward) within the body of the parfor loop, and only the gradients are aggregated.  Regardless, Hout & Wout will be the same for all parallel instances of the parfor.  Did you have any other possible issues in mind?

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: The backward might be inefficient in that case :(

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Since the output shape logic is not too complex, I would really prefer to keep it until we are satisfied with the performance of distributed training with and without Caffe2DML.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: I am OK with either naming scheme as long as we are consistent.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Sure, and Caffe2DML can redirect to the appropriate layer.

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Yes, I intend to support that in different PR. Just don't want to make this PR too large :)

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: Updated the documentation :)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1464/

src/main/scala/org/apache/sysml/api/dl/CaffeLayer.scala
niketanpansare: To map this layer to [scale_shift2d.dml](https://github.com/apache/incubator-systemml/blob/master/scripts/nn/layers/scale_shift2d.dml), we will have to remove following code from batch_norm2d.dml:
https://github.com/apache/incubator-systemml/blob/master/scripts/nn/layers/batch_norm2d.dml#L118-L119

dusenberrymw: No we don't.  Just leave `gamma` and `beta` in their initialized forms (1s and 0s, respectively), and don't update them when you receive their gradients from `batch_norm2d::backward`.  That way, those lines will simply remain identity ops.

PR #493, opened 2017-05-11
[SYSTEMML-1470] Reset MLContext stats between runs

deroneriksson:
Update ScriptExecutor to reset stats between executions if needed.
Update resetGlobalFlags to set DMLScript.STATISTICS_COUNT using
DMLOptions.defaultOptions.statsCount rather than hardcoded value.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/ScriptExecutor.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/ScriptExecutor.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.api.ScriptExecutorUtils;
+import org.apache.sysml.api.DMLScript.DMLOptions;
 import org.apache.sysml.api.jmlc.JMLCUtils;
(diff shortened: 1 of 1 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1462/

src/main/java/org/apache/sysml/api/mlcontext/ScriptExecutor.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.api.ScriptExecutorUtils;
+import org.apache.sysml.api.DMLScript.DMLOptions;
```
nakul02: Because of this weird dependency on DMLScript, should we consider moving DMLOptions into its own class?
Weirdly, I had thought of doing that, but since the Options are closely tied with parsing CLI, it didn't seem to make sense back then. But looking at this `import` statement, I am beginning to reconsider.

deroneriksson: I would say that DMLOptions is so important to SystemML that it should probably be moved to its own class.

deroneriksson:
Thank you for reviewing @nakul02!

PR #497, opened 2017-05-13
[SYSTEMML-1596] Set runtime platform via MLContext

deroneriksson:
Update MLContext and ScriptExecutor to allow for convenient setting
of the runtime platform. This is useful when debugging and developing.

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
@@ -31,7 +31,6 @@ import org.apache.spark.SparkContext;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.sql.SparkSession;
 import org.apache.sysml.api.DMLScript;
-import org.apache.sysml.api.DMLScript.RUNTIME_PLATFORM;
 import org.apache.sysml.api.MLContextProxy;
 import org.apache.sysml.api.jmlc.JMLCUtils;
 import org.apache.sysml.conf.ConfigurationManager;
@@ -117,6 +116,12 @@ public class MLContext {
 	private ExplainLevel explainLevel = null;
 
 	/**
+	 * The runtime platform on which to execute. By default, MLContext runs on
+	 * {@code ExecutionType.DRIVER_AND_SPARK}.
+	 */
+	private ExecutionType executionType = ExecutionType.DRIVER_AND_SPARK;
+
+	/**
 	 * Whether or not all values should be maintained in the symbol table
 	 * after execution.
 	 */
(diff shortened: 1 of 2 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1467/

dusenberrymw:
Nice.  For `DRIVER` mode, will we be able to setup MLContext to work without needing Spark?  I.e. could we simply use SystemML as a single Scala/Java program via MLContext?

deroneriksson:
Currently a lot of the primary classes in the API (MLContext, MLContextUtil, MLContextConversionUtil, Matrix, MLResults, etc) utilize Spark classes. We could probably do something such as create parent classes without the Spark class references to allow MLContext usage without Spark, although of course it could quite possibly be more difficult than that.

deroneriksson:
@dusenberrymw Do you think this should be merged before the 1.0.0 release? In communication with @feihugis, it seems like setting the 'runtime platform'/'execution type' can be useful, and having it built into the API can avoid the need for code such as `org.apache.sysml.api.DMLScript.rtplatform = org.apache.sysml.api.DMLScript.RUNTIME_PLATFORM.SPARK`.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1745/

dusenberrymw:
Yeah we should definitely merge this functionality.  Is everyone okay with the updated names of `DRIVER` instead of `SINGLE_NODE`, and `DRIVER_AND_SPARK` instead of `HYBRID_SPARK`?  To me they are clearer, but it does introduce inconsistencies.  cc @nakul02

deroneriksson:
BTW the MLContext API naming differences vs the 'internal' global rtplatform in DMLScript are:
DMLScript.RUNTIME_PLATFORM.SINGLE_NODE -> ExecutionType.DRIVER
DMLScript.RUNTIME_PLATFORM.SPARK -> ExecutionType.SPARK
DMLScript.RUNTIME_PLATFORM.HADOOP -> ExecutionType.HADOOP
DMLScript.RUNTIME_PLATFORM.HYBRID_SPARK -> ExecutionType.DRIVER_AND_SPARK
DMLScript.RUNTIME_PLATFORM.HYBRID -> ExecutionType.DRIVER_AND_HADOOP

The reasoning for this:
SINGLE_NODE is confusing since SystemML can run in CP on a single node, Spark can run on a single node, and Hadoop can run on a single node.
HYBRID is CP+Hadoop, whereas HYBRID_SPARK is CP+Spark. The name 'HYBRID' is not descriptive of the Hadoop part of CP+Hadoop.

I am not 100% happy with the term DRIVER but I think it is more descriptive than SINGLE_NODE.

nakul02:
+1 to the renaming suggestions.
If possible, we could have the `internal` and `external` names be the same. Appropriate changes in the launch scripts & @krishnakalyan3's performance test scripts would be needed too.

deroneriksson:
Thank you for reviewing @dusenberrymw and @nakul02. I'll merge.

I think standardizing the internal/external names would be great, especially if we could address the lack of clarity regarding the 1) SINGLE_NODE and 2) HYBRID vs HYBRID_SPARK internal representations. I would also be very happy with names such as:
  SINGLE_NODE -> CP
  HYBRID -> HYBRID_HADOOP

mboehm7:
What is the reasoning behind these inconsistent and misleading execution modes? Singlenode means CP, hybrid means CP and MR, and hybrid_spark means CP and SPARK - note that CP instructions are not necessarily executed in the driver (e.g., in the parfor optimizer decides for remote execution, then CP instructions are executed in the executor processes and map/reduce tasks respectively).

deroneriksson:
@mboehm7 Statements such as "hybrid means CP and MR, and hybrid_spark means CP and SPARK" point out the confusing, inconsistent naming in DMLScript.RUNTIME_PLATFORM. Using a name such as "hybrid" instead of "hybrid_hadoop" or "hybrid_mapreduce" obfuscates meaning for our users and just does not make sense to me.

mboehm7:
globally renaming hybrid to hybrid_mr as well as hadoop to mr is fine, I just don't think that "DRIVER" is a good choice because it attaches a location, which is not necessarily true with parfor for containerized jmlc deployments.

mboehm7:
btw, the name hybrid is not too far fetched, because we switch automatically from hybrid_mr to hybrid_spark when we recognize that we're running in a spark context (at least through spark_submit).

mboehm7:
so what's the conclusion here: revert or create global consistency?

bertholdreinwald:
The original intent of these modes was to specify what kind of runtime instructions can be generated. It was primarily internally. Later with different SystemML deployment modes, the same “flag” was used for the user to be able to specify the deployment mode.

From user’s perspective, there are 2 or 3 modes depending on how you count: no use of a data parallel platform, i.e. single node, or use of data parallel platform, i.e. MR, or Spark. Yes, you can run MR and Spark on single node, but from SystemML perspective it is still using a data parallel platform.

If people don’t like to use “CP”, or “DRIVER”, or “SINGLE_NODE”, may be “LOCAL” is a better choice. I think using “HYBRID …” is very descriptive and important for SystemML.

There should be 3 modes exposed to the user:
(1) LOCAL
(2) HYBRID_MR
(3) HYBRID_SPARK

Internally, we should not get rid of SPARK and MR modes as they are being used for experimentation, debugging, and developing forcing all ops to SPARPK or MR. But they are useless to the user.

deroneriksson:
In my opinion, "SPARK" and "HADOOP" are not useless for the user. One of the first things I typically say to a user running into an issue in "HYBRID_SPARK" mode is to try running in "SPARK" mode, which typically reveals various issues that are useful when debugging. Usually in order to set this in the past (via MLContext in a notebook) I have had to instruct the user to perform a hack by setting DMLScript.rtplatform = DMLScript.RUNTIME_PLATFORM.SPARK.

PR #502, opened 2017-05-15
[SYSTEMML-1600] Add version info in MLContext

krishnakalyan3:
Append SystemML version number to MLContext welcome message. 
Tested local in my local system.

ping @deroneriksson 

```
Welcome to Apache SystemML!
Version 1.0.0-incubating-SNAPSHOT
ml: org.apache.sysml.api.mlcontext.MLContext = org.apache.sysml.api.mlcontext.MLContext@1755aee6
```

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -975,10 +975,11 @@ public final class MLContextUtil {
 	 */
 	public static String welcomeMessage() {
         StringBuilder sb = new StringBuilder();
-        sb.append("\nWelcome to Apache SystemML!\n Version ");
+        sb.append("\nWelcome to Apache SystemML!\n");
         try {
             ProjectInfo info = ProjectInfo.getProjectInfo();
             if (info.version() != null) {
+            	sb.append("Version ");
                 sb.append(info.version());
             }
         } catch (MLContextException e) {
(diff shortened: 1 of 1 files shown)


akchinSTC:

src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
```diff
@@ -974,8 +974,15 @@ public static String displayOutputs(Set<String> outputNames, LocalVariableMap sy
 	 * @return the SystemML welcome message
 	 */
 	public static String welcomeMessage() {
-		StringBuilder sb = new StringBuilder();
-		sb.append("\nWelcome to Apache SystemML!\n");
+        StringBuilder sb = new StringBuilder();
+        sb.append("\nWelcome to Apache SystemML!\n Version ");
```
deroneriksson: Hi @krishnakalyan3. Thank you for working on this!

Would you mind moving `" Version "` to within the `if (info.version() != null) { }` block? Otherwise, if someone is developing using MLContext in an IDE (where no SystemML jar file is available), they see output such as the following in the console:
```
Welcome to Apache SystemML!
 Version 
```
rather than just
```
Welcome to Apache SystemML!
```
Thanks!

krishnakalyan3: @deroneriksson Thank you for the review. I have made the necessary changes

PR #509, opened 2017-05-23
[SYSTEMML-540] Support both FP32 and FP64 MKL DNN kernels for convolution and matrix multiplication

niketanpansare:
Using randomly generated 64 images of size 256x256 with sparsity 0.9 and 32 filter of size 5x5. Stride = [1,1] and pad=[1,1].

A. With channels = 32
Time in seconds for 10 conv2d calls using MKL DNN:
native.blas.datatype=float: 10.765 seconds
native.blas.datatype=double: 71.118 seconds

Time in seconds for 10 conv2d calls using native im2col and OpenBLAS gemm:
native.blas.datatype=float: 21.963 seconds
native.blas.datatype=double: 34.881 seconds

B. With channels = 3
Time in seconds for 10 conv2d calls using MKL DNN:
native.blas.datatype=float: 5.144 seconds
native.blas.datatype=double: 12.599 seconds

Time in seconds for 10 conv2d calls using native im2col and OpenBLAS gemm:
native.blas.datatype=float: 7.918 seconds
native.blas.datatype=double: 8.688 seconds

@nakul02 @mboehm7 @bertholdreinwald @dusenberrymw


mboehm7:
How does this fit into SystemML's execution model, which tries to ensure equivalent results (modulo rounding issues) across backends?

niketanpansare:
By default, we always use double precision which guarantees equivalent results. The user has to explicitly turn on single precision and hence is aware of loss of precision.

mboehm7:
which would be ok, iff all backends and operations would support single precision accuracy.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1485/

niketanpansare:
Sorry, I disagree with the iff clause especially for an optional feature that is not turned on by default. If it helps, I plan to support lower precision for GPU backend once I get some free cycles.

mboehm7:
See a partial integration for a subset of operations like that creates more problems than it solves. In the presence of rewrites and operator selection, it is not even clear to a user, at script level, which operations are affected. In addition, there are issues like converting inputs from double to float, which memory requirements are not capture by the memory estimates of the compiler. Finally, my major concern is that we only exploit low-hanging fruits like calling existing library functions without addressing the actual hard part of a full compiler and runtime integration for all operations. Supporting different matrix value types is great but requires a proper design and consistent integration.

niketanpansare:
I second the idea of supporting different matrix types. Given the short runway and stiff competition, we should at least attempt to give users the flexibility to address temporary shortcomings until we come up with a full integration. This also helps validate the major effort required for full integration of a given feature.

Again, we should not stop optional features (turned off by default), meant for advanced users that provides significant performance benefits to a large class of deep learning problems (the field in which SystemML is not yet competitive). I am OK with flagging these features with necessary disclaimers in the documentation :)

src/main/cpp/libmatrixmult.h
```diff
@@ -51,12 +52,19 @@
...
 void matmult(double* m1Ptr, double* m2Ptr, double* retPtr, int m1rlen,
              int m1clen, int m2clen, int numThreads);
-             
-void tsmm(double* m1Ptr, double* retPtr, int m1rlen, int m1clen, bool isLeftTranspose,  int numThreads);
-             
+void matmult(float* m1Ptr, float* m2Ptr, float* retPtr, int m1rlen, int m1clen,
+             int m2clen, int numThreads);
+
+void tsmm(double* m1Ptr, double* retPtr, int m1rlen, int m1clen,
+          bool isLeftTranspose, int numThreads);
```
nakul02: This is dangerous territory - overloading a function with `double` and `float` parameters. You are better of naming them differently or using c++ templates.

nakul02:
Having a different value for matrices seems like a better idea than this, as a holistic feature.
In the face of stiff competition, maybe quickly adding features, and consequently, technical debt is not a good idea.

However, if the benefits of this are large and made obvious by experiments on end-to-end DL use cases, I am ok with it. 
If this PR ends up getting merged, this need to either be implemented to completion or be yanked out completely. I am absolutely against having incomplete features for so-called advanced users and let future contributors deal with the technical debt.

@niketanpansare - do you have any experimental results that you can post here?

dusenberrymw:
Interesting to see that the im2col + OpenBLAS combo is actually faster for doubles than MKL-DNN.  Just shows that the focus is on single precision. 👍 

As for how to move forward, we know that we will need to support single precision (or even lower), especially on GPUs (modern models may not fit with double precision).  So, the conversation here needs to be on what are the concrete steps that would be needed for adding this to the engine in a timely manner.  As @niketanpansare has noted, we are up against some serious competition, so we really need to be focused.

One possibility here is to have an internal staged rollout of single precision in the engine, but externally present the option of either single or double precision to the user, without any specification of which ops/backends are affected.

niketanpansare:
>> do you have any experimental results (end-to-end DL) that you can post here?

Here are the statistics of one of the parfor worker using in Mike's final prediction demo at strata:
```bash
-- 1) conv2d_bias_add 71633.031 sec 1201029
-- 2) ba+* 27327.518 sec 4832432
-- 3) + 27052.159 sec 2811118
-- 4) * 11866.432 sec 1201023
-- 5) / 11417.598 sec 1246365
-- 6) - 11406.372 sec 1269053
-- 7) sqrt 10791.259 sec 1202002
-- 8) sel+ 7915.427 sec 1065033
-- 9) rmvar 2576.081 sec 20220609
-- 10) leftIndex 1010.286 sec 45338
```

Setup:
- Total number of nodes: 9
- Each node has Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (48 vcores) and 100GB executor memory. The data is stored on HDFS is parquet format.
- Python script used in the current setup: https://github.com/niketanpansare/bc_weights/blob/master/scoring_rowwisecropping.py
- DML script called by the above Python script: https://github.com/niketanpansare/bc_weights/blob/master/resnet_prediction_parfor_rowwisecropping.dml
- We use im2col based implementation (which is also what Caffe, TensorFlow and other frameworks use) for conv2d_bias_add. We use MKL's dgemm for matrix multiplication in this operator. 

You can compare `conv2d_bias_add (71633.031 sec)`  to `openblas native.blas.datatype=double (34.881 seconds)` with greater than or equal to 32 channels. With `mkl dnn native.blas.datatype=float (10.765 seconds)`, this number should come down by 3.25x. You should see similar performance gains for `ba+*`.

>> I am absolutely against having incomplete features for so-called advanced users and let future contributors deal with the technical debt.

I think there has been a misunderstanding. The PR itself is complete: all it does is utilizes BLAS+MKL DNN more efficiently. @nakul02  Please let me know what specific part is bothering you.

Challenges with supporting multiple types:
- Multi-month project with no volunteer to take it on.
- API complexity: Either introduce generic types/algebra in DML or write `k` copies of nn layers. Highlighting Mike's slack comment: `max(0, X)` explodes, `max(0.0, X)` works just fine, and `max(X, 0)` works just fine as well.
- Implementation complexity: `n` times more operators to guarantee numerical stability (for example: if we we are concerned with precision difference between sgemm and dgemm, then imagine the operation: float * double + int with kahan summation). 

Not supporting single-precision implies even a fair reviewer will always see a performance difference of 2-3x on CP (probably on GPU too). This difference is significant enough to shoo away deep learning users, where a typical training takes months.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1488/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1490/

niketanpansare:
Ok, since the general consensus is against this approach, I will close this PR. Please feel free to take over the task of supporting lower precision on CP. Thanks.

PR #511, opened 2017-05-23
[SYSTEMML-1303] Remove deprecated old MLContext API

deroneriksson:
Remove deprecated old MLContext API, which is scheduled to be removed in version 1.0.0.


mboehm7:
LGTM - thanks @deroneriksson; minor comments: (1) I would recommend not to modify the formatting of apache licenses because they are simply copies of each other, and (2) now that the old mlcontext is removed, we should be able to consolidate the test packages into functions/mlcontext.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1487/

deroneriksson:
@mboehm7 @dusenberrymw 
WRT 1) I think it might be a good idea to remove the trailing whitespace in all the Apache license headers since git tends to treat trailing whitespace as an error, 2) I agree that consolidating mlcontext and functions/mlcontext sounds like a good idea

mboehm7:
I don't have a strong opinion here, but I would leave the formatting of the licenses and empty lines  as they are (tabs/spaces aligned with previous line) because this is the default in eclipse and it keeps the history clean.

deroneriksson:
Although I have used the same Eclipse defaults for about 15 years, I just recently switched to removing trailing whitespace. Git/GitHub tend to flag trailing whitespace as an error and developers such as @dusenberrymw have convinced me that removing trailing whitespace tends to be a good practice to follow. I updated my Eclipse using:
```
Preferences --> Java --> Editor --> Save Actions --> Configure
Code Organizing --> Remove trailing whitespace (and All lines)
```

PR #513, opened 2017-05-24
[SYSTEMML-1625] GPU Unit Tests

nakul02:
Ping @bertholdreinwald, @niketanpansare


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1491/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1493/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1494/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1496/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1499/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1500/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1512/

pom.xml
```diff
@@ -426,7 +426,7 @@
 
 					<!-- Third argument prevents Java from popping up lots of windows on
 						MacOS -->
```
nakul02: @deroneriksson, @frreiss, @gweidner, @dusenberrymw - I've increased the maximum memory needed by the JVM from 2G to 8G. Is this ok?

src/main/java/org/apache/sysml/runtime/instructions/GPUInstructionParser.java
```diff
@@ -42,81 +42,79 @@
...
-		String2GPUInstructionType.put( "^2"   , GPUINSTRUCTION_TYPE.ArithmeticBinary); //special ^ case
-		String2GPUInstructionType.put( "*2"   , GPUINSTRUCTION_TYPE.ArithmeticBinary); //special * case
-		String2GPUInstructionType.put( "-nz"  , GPUINSTRUCTION_TYPE.ArithmeticBinary); //special - case
-		String2GPUInstructionType.put( "+*"   , GPUINSTRUCTION_TYPE.ArithmeticBinary);
-		String2GPUInstructionType.put( "-*"   , GPUINSTRUCTION_TYPE.ArithmeticBinary);
+		String2GPUInstructionType.put( "+",    GPUINSTRUCTION_TYPE.ArithmeticBinary);
+		String2GPUInstructionType.put( "-",    GPUINSTRUCTION_TYPE.ArithmeticBinary);
+		String2GPUInstructionType.put( "*",    GPUINSTRUCTION_TYPE.ArithmeticBinary);
+		String2GPUInstructionType.put( "/",    GPUINSTRUCTION_TYPE.ArithmeticBinary);
+		//String2GPUInstructionType.put( "%%",   GPUINSTRUCTION_TYPE.ArithmeticBinary);
```
nakul02: @niketanpansare - I've commented out the `%%` and `%/%` ops because they were mentioned here, but an implementation was not to be found? Is this correct or have I overlooked something important?

nakul02:
@bertholdreinwald , @niketanpansare  - could you please review/leave comments?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1515/

src/main/java/org/apache/sysml/runtime/instructions/GPUInstructionParser.java
niketanpansare: That is correct

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -2348,14 +2346,21 @@ else if(op.fn instanceof Multiply || op.fn instanceof And) {
...
+                // If the IEEE 754 standard defines otherwise, this logic needs
+                // to be re-enabled and the Java computation logic for divide by zero
+                // needs to be revisited
+                //else if(op.fn instanceof Divide && isSparseAndEmpty(gCtx, in)) {
+                //	setOutputToConstant(ec, gCtx, instName, Double.NaN, outputName);
+                //}
+                //else if(op.fn instanceof Divide) {
+                //	//For division, IEEE 754 defines x/0.0 as INFINITY and 0.0/0.0 as NaN.
+                //	compareAndSet(ec, gCtx, instName, in, outputName, 0.0, 1e-6, Double.NaN, Double.POSITIVE_INFINITY, Double.POSITIVE_INFINITY);
+                //}
```
niketanpansare: Are the integration test cases validating this passing ? I would recommend taking that into consideration.

pom.xml
mboehm7: no, please leave it at 2G as there are a couple of tests are designed for 2G which would otherwise fail with an unexpected number of compiled/executed distributed operations.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
mboehm7: SystemML generally has limited support for INF and NaN as even sparse linear algebra is technically incorrect in the presence of NaNs. However, for existing values, we need to ensure that the results are equivalent to Java.

mboehm7:
@nakul02 could please help me understand how you disable these tests in environments without GPUs?

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDatagen.java
```diff
@@ -510,14 +511,19 @@ private static void genRandomNumbers(boolean invokedFromCP, int rl, int ru, int
 
 		// Setup Pseudo Random Number Generator for cell values based on 'pdf'.
 		if (valuePRNG == null) {
-			if ( rgen._pdf.equalsIgnoreCase(RAND_PDF_UNIFORM)) 
+			switch (rgen._pdf) {
+			case UNIFORM:
```
mboehm7: please globally fix the formatting of such switch statement

src/test/java/org/apache/sysml/test/gpu/BinaryOpTests.java
```diff
@@ -0,0 +1,73 @@
...
+	@Test
+	public void testSolve() {
+		// Test Ax = b
+		// Dimensions of A = m * n
+		// Dimensions of x = n * 1
+		// Dimensions of b = m * 1
+
+		String scriptStr = "x = solve(A, b)";
+		double sparsity = 1.0; // Only dense matrices supported by "solve"
+		final int[] sides = { 32, 33, 128, 256, 513, 2049 };
```
mboehm7: Why not use parameterized junit tests for that? it would have the advantage of being able to re-run individual instances.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDatagen.java
nakul02: How specifically do you mean?

nakul02:
@mboehm7 - GPU tests cases are disabled by default. 
To enable them, one would have to say 
`mvn verify -PgpuTests`

pom.xml
nakul02: Which ones? I could go and fix them.

src/test/java/org/apache/sysml/test/gpu/BinaryOpTests.java
nakul02: I have made them parameterixable when needed. Take a look at the other test cases in other classes.

mboehm7:
ok that sounds good - and for the nightly builds we would enable them explicitly?

src/test/java/org/apache/sysml/test/gpu/BinaryOpTests.java
mboehm7: well, whenever there is no need to explicitly kick them off via dedicated functions, we still use 	@Parameters to specify such parameters in a way that allows us to reexecute individual parameters on failures - see for example, the GLM application test.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDatagen.java
mboehm7: please use indentation for the case statements.

pom.xml
mboehm7: just look at the failed builds - I guess SparsityRecompileTest is a good example. However, note that we cannot modify these tests because people need to be able to runs the entire testsuite offline on some older laptop as well.

nakul02:
That's correct

pom.xml
nakul02: I can work on either fixing the tests or fixing these newly introduced tests. Either ways supporting "older laptops" with presumably very little memory sounds infeasible. Even apache spark recommends more than 2GB of memory on the driver.

nakul02: Just out of curiosity, are you aware of anyone me trying to run the systemml tests on an "older laptop"?

mboehm7: Well my own dev environment are two laptops with 32GB and 16GB respectively, but I like to have headroom for active development while running the tests. Honestly, I don't see a good reason why we should unnecessarily increase this requirement. Btw, even the Spark default driver memory setting is 1GB.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+    private final List<Double> sparsitylst = Arrays.asList(1.0);    // Only test for dense
+    */
+
+	private final List<Integer> Nlst = Arrays.asList(128, 64);
+	private final List<Integer> Clst = Arrays.asList(30, 3);
+	private final List<Integer> Hlst = Arrays.asList(256, 64);
+	private final List<Integer> Wlst = Arrays.asList(256, 64);
+	private final List<Integer> Klst = Arrays.asList(30, 20);
+	private final List<Integer> Rlst = Arrays.asList(128, 4);
+	private final List<Integer> Slst = Arrays.asList(128, 4);
```
dusenberrymw: Can you update the sizes lists for `R` and `S` to include `3`?  Maybe just replace 4 with 3.  3x3 filters are super common, so it would be good to have explicit tests for that size.  Also, 1x1 would be good too, since that is used to alter the number of channels only, and I could imagine the engine having an issue for that size.  128 is good too since it's possible to formulate an affine layer (i.e. "fully-connected layer", "dense layer", etc.) as a convolution with filter spatial sizes equal to the input spatial sizes.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+	private final List<Integer> strideHeightLst = Arrays.asList(9, 3);
+	private final List<Integer> strideWidthLst = Arrays.asList(9, 3);
```
dusenberrymw: Can you include stride size `1` here, possibly just replacing 3?  Stride 1 is super common, so it would be good to have explicit tests.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+
+														long doutSize = N * K * P * Q * 8l;
+														if (doutSize > MAX_OP_SIZE) // dout/output size
+															continue;
+
+														double imageSizeInMB = imageSize / (1024.0 * 1024.0);
+														double filterSizeInMB = filterSize / (1024.0 * 1024.0);
+														double doutSizeInMB = doutSize / (1024.0 * 1024.0);
+														System.out
+																.format("conv2d_backward_filter, image[%d,%d,%d,%d](%.1fMB), filter[%d,%d,%d,%d](%.1f), dout[%d,%d,%d,%d](%.1fMB), stride[%d,%d], padding[%d,%d]",
```
dusenberrymw: I think this should be `conv2d ...`.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+													for (double sparsity : sparsitylst) {
+
+														// Make sure ops fit in GPU memory and within constraints of cudnn
+														long imageSize = N * C * H * W * 8l;
+														if (imageSize > MAX_OP_SIZE)  // image size
+															continue;
+														long filterSize = K * C * R * S * 8l;
+														if (filterSize > MAX_OP_SIZE)  // filter size
+															continue;
+														// filter is smaller than image + padding
```
dusenberrymw: The filter could also be equal to the image+padding size.  This is used in some cases if we want to reformulate an affine layer as a convolution, which is nice if we want to do surgery on a model with an affine layer to convert it to a fully-convolutional model so that it can be applied to any image size, rather than a fixed image size as it would with an affine layer present.

dusenberrymw: Also, it would be nice to include this scenario in one of the test cases above.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+							for (long R : Rlst) {
+								for (long S : Slst) {
+									for (long strideH : strideHeightLst) {
+										for (long strideW : strideWidthLst) {
+											for (long padH : padHeightLst) {
+												for (long padW : padWidthLst) {
+													for (double sparsity : sparsitylst) {
+
+														// filter is smaller than image + padding
+														if (R > (H + padH) || S > (W + padW))
```
dusenberrymw: Same as above, the filter could be equal in size to that of the image + padding.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
```
dusenberrymw: Same as above, the filter could be equal in size to that of the image + padding.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+						for (long R : Rlst) {
+							for (long S : Slst) {
+								for (long strideH : strideHeightLst) {
+									for (long strideW : strideWidthLst) {
+										for (long padH : padHeightLst) {
+											for (long padW : padWidthLst) {
+												for (double sparsity : sparsitylst) {
+
+													// pool is smaller than image + padding
+													if (R > (H + padH) || S > (W + padW))
```
dusenberrymw: Similar to conv2d, the max pool window could be equal in size to the image + pad.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
```
dusenberrymw: Same as above, the max pool window could be equal in size to the image + pad.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
```diff
@@ -0,0 +1,490 @@
...
+														inputs.put("padH", padH);
+														inputs.put("padW", padW);
+														inputs.put("image", image);
+														inputs.put("filter", filter);
+														List<Object> outCPU = runOnCPU(spark, scriptStr, inputs,
+																Arrays.asList("O"));
+														List<Object> outGPU = runOnGPU(spark, scriptStr, inputs,
+																Arrays.asList("O"));
+														assertHeavyHitterPresent("gpu_conv2d");
+														assertEqualObjects(outCPU.get(0), outGPU.get(0));
```
dusenberrymw: Is this testing absolute error difference or a relative error difference?  I would recommend using a relative error or otherwise the magnitude of the values being compared could mask an issue.  For example, if the two values are ~1e5, the difference between them ends up being ~1e6, and we use a difference threshold of 1e5, then the test would pass, but the difference is actually relatively huge, and likely indicative of an issue.

pom.xml
deroneriksson: Normally the test suite run by maven utilizes 4 processes running concurrently, with each one utilizing 2GB, so computers with 16GB RAM (such as my laptop) can run the test suite. If you need more memory than 2GB per process, perhaps you can make it configurable.

mboehm7: after thinking about this a bit more - could it be that the 8GB are only required to cover an existing memory leak (maybe in mlcontext or spark internals)? Even without these new tests, our testsuite already has an issue of memory leaks when run in a single JVM process but I was not able to really resolve them so far. These issues don't show up in our nightly builds because, as @deroneriksson mentioned, maven uses an individual process per suite.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixDatagen.java
nakul02: I tried reformatting using the style format that @deroneriksson checked in (#515). It prefers this style - i.e. `case` at the same indentation level as `switch`.

mboehm7: hmm, thanks for pointing this out; apparently I missed this PR somehow - then we need to have a broader discussion regarding the style template in general. Regarding the switch statement it's probably safe to use indentation here because it significantly increases readability.

nakul02: I am going to leave this the way the now committed style format requires me to. If the future discussion about the style template produces a different outcome, I shall change it to that.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
nakul02: good catch, thanks

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
nakul02: good catch, thanks

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
nakul02: The loop is skipped if `filter width` or `filter height` is strictly greater than `image+padding`.
Is this sufficient?

nakul02:
@mboehm7 , @deroneriksson -  I've reverted back to using 2g for the verify phase. 
@dusenberrymw - incorporated your suggestions.

deroneriksson:
LGTM.

src/test/java/org/apache/sysml/test/gpu/NeuralNetworkOpTests.java
dusenberrymw: Ah, yeah I just read the comment and quickly skimmed the actual code here.  The code is fine, can you just update the comment?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1520/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1522/

nakul02:
Tests passed, With LGTMs from @dusenberrymw, @deroneriksson & @niketanpansare, I shall merge.
Thank you for reviewing @niketanpansare, @mboehm7, @deroneriksson & @dusenberrymw

PR #516, opened 2017-05-26
[SYSTEMML-1380] Kmeans isY and verb parameters should be boolean

krishnakalyan3:
In the Kmeans.dml script, the 'isY' and 'verb' input parameters are integers. However, they are basically on/off switches, so replacing the integer types with boolean types could potentially be a little clearer to users.

I tested this on my local system.

### DIFF
--- a/docs/algorithms-clustering.md
+++ b/docs/algorithms-clustering.md
@@ -123,10 +123,10 @@ apart is a "false negative" etc.
                                     maxi=[int]
                                     tol=[double]
                                     samp=[int]
-                                    isY=[int]
+                                    isY=[boolean]
                                     Y=[file]
                                     fmt=[format]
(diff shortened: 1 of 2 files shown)


akchinSTC:

krishnakalyan3:
cc @deroneriksson

deroneriksson:
@iyounus Are you aware of any code invoking Kmeans.dml in the project (such as performance suite) where any updates are needed for these parameters? I looked at runKmeans.sh and I don't see any other updates needed.

krishnakalyan3:
@deroneriksson thank you for the review. I am sorry for missing out conditional if updates.

deroneriksson:
LGTM. Thank you @krishnakalyan3!

PR #521, opened 2017-05-30
[SYSTEMML-445] [WIP] Prepare GPU backend to support arbitrary datatype

niketanpansare:
@nakul02 Does this design work for you ?

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/CustomSystemMLKernels.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/CustomSystemMLKernels.java
@@ -1,0 +1,159 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sysml.runtime.instructions.gpu.context;
+
+import static jcuda.driver.JCudaDriver.cuLaunchKernel;
+import static jcuda.driver.JCudaDriver.cuModuleGetFunction;
+import static jcuda.driver.JCudaDriver.cuModuleLoadDataEx;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+
+import org.apache.sysml.runtime.DMLRuntimeException;
+import org.apache.sysml.runtime.io.IOUtilFunctions;
+
+import jcuda.Pointer;
+import jcuda.driver.CUfunction;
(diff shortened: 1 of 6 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1509/

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/CustomSystemMLKernels.java
```diff
@@ -0,0 +1,159 @@
...
+
+import jcuda.Pointer;
+import jcuda.driver.CUfunction;
+import jcuda.driver.CUmodule;
+import jcuda.driver.CUresult;
+
+/**
+ * Utility class that allows LibMatrixCUDA as well as GPUObject to invoke custom CUDA kernels.
+ * 
+ * The utility org.apache.sysml.runtime.instructions.gpu.context.JCudaKernels simplifies the launching of the kernels.
```
nakul02: Update comment to point to CustomSystemMLKernel

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -3332,7 +3333,7 @@ public static void solve(ExecutionContext ec, GPUContext gCtx, String instName,
 	@SuppressWarnings("unused")
 	private static void debugPrintMatrix(Pointer in, int rlen, int clen){
 		double[] data = new double[rlen * clen];
-		cudaMemcpy(Pointer.to(data), in, rlen*clen*Sizeof.DOUBLE, cudaMemcpyDeviceToHost);
+		cudaMemcpy(Pointer.to(data), in, rlen*clen*handle.sizeOfDatatype(), cudaMemcpyDeviceToHost);
```
nakul02: I am sure these won't work right off the bat. 
You'd first need to cast them to a float array on the host machine, then copy them to the GPU.
(Also I understand this is a WIP, and there is more work to be done. In light of that, please ignore this comment if you had already planned to do what I suggested).

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/JCudaKernels.java
```diff
@@ -18,142 +18,43 @@
...
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.HashMap;
-
-import org.apache.sysml.runtime.DMLRuntimeException;
-import org.apache.sysml.runtime.io.IOUtilFunctions;
+// This indirection prepares SystemML's GPU backend to allow
+// for arbitrary datatype
+public abstract class JCudaKernels {
+	private static JCudaKernels singletonObj = new DoublePrecisionKernels();
```
nakul02: I'd prefer no singleton objects, specially when you get it from an abstract class. I prefer factories for this sort of stuff. Granted that it adds more cruft. 
An alternative way to do this is to add the `sizeOfDatatype()` and `cudnnDataType()` methods to jcudakernels, or more aptly to `GPUContext` (since all cublas, cusparse, cusolve, cudnn library calls need to have that data type, not just the custom kernels). In the GPUContext, you could control which data type to select via an enumeration. 
I recommend an enumeration here, because (a) the types are limited - float, double, short... (b) what you plan to do with this information is very limited (`cudnnDataType` and `sizeOfDatatype`).

Also, the decision to make it single vs double precision should not be embedded this deep down in the code. 
it should be controlled by a systemml property (from the SystemML-config.xml file) or via a command line parameter.

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/JCudaKernels.java
```diff
@@ -18,142 +18,43 @@
...
+	public static JCudaKernels handle() {
+		return singletonObj;
+	}
+	public abstract int cudnnDataType();
```
nakul02: Documentation here, please

PR #522, opened 2017-05-31
[SYSTEMML-1236] Move MLContextProxy out of api package

deroneriksson:
MLContextProxy is used internally by SystemML but not my end users, so move
this class to a non-API package.
Remove some unnecessary casting since only 1 MLContext class now exists.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1516/

PR #523, opened 2017-05-31
[SYSTEMML-1562] Remove history from MLContext

deroneriksson:
Remove Script history from MLContext for multiple reasons:
1) Reduce number of object references
2) In environments such as Spark Shell, a user typically already
   has references to the Script objects
3) History implementation can lead to confusing behavior since the
   LinkedHashMap values can point to the same Script object, which
   can change over time.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1517/

PR #530, opened 2017-06-07
[Doc] Change PCA scale value

krishnakalyan3:
Currently in the matrix factorisation docs the value of scale is 10. 
This value can be either `0` or `1`.


akchinSTC:

krishnakalyan3:
ping @deroneriksson @nakul02

nakul02:
LGTM :+1: 
I shall merge.

mboehm7:
the fix is fine; down the road, however, we should really consider to make scale, center, and projectdata boolean inputs in order to avoid such issues altogether.

PR #531, opened 2017-06-07
[SYSTEMML-947] Remove binary block classes from MLContext

deroneriksson:
Remove BinaryBlockMatrix and BinaryBlockFrame classes from
MLContext API and incorporate similar functionality into
Matrix and Frame classes.


deroneriksson:
@mboehm7 Please let me know what you think about this implementation.

src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
```diff
@@ -146,12 +207,38 @@ public MatrixObject toMatrixObject() {
...
+	 */
+	public MatrixBlock toMatrixBlock() {
+		if (matrixMetadata == null) {
+			throw new MLContextException("Matrix metadata required to convert binary blocks to a MatrixBlock.");
+		}
+		if (binaryBlocks != null) {
+			return MLContextConversionUtil.binaryBlocksToMatrixBlock(binaryBlocks, matrixMetadata);
+		} else if (matrixObject != null) {
+			toBinaryBlocks();
+			return MLContextConversionUtil.binaryBlocksToMatrixBlock(binaryBlocks, matrixMetadata);
```
mboehm7: Could you use toBinaryBlocks() instead of binaryBlocks here? It would have the same effect but would make it more explicit without knowing that toBinaryBlocks also modifies the state.

src/main/java/org/apache/sysml/api/mlcontext/Matrix.java
```diff
@@ -160,11 +247,45 @@ public BinaryBlockMatrix toBinaryBlockMatrix() {
...
 	@Override
 	public String toString() {
-		return matrixObject.toString();
+		if (matrixObject != null) {
+			return matrixObject.toString();
+		} else if (matrixMetadata != null) {
+			return matrixMetadata.toString();
+		} else {
+			return super.toString();
+		}
```
mboehm7: nice to have comment: could we make this easier to read via something like "return (matrixObject !=null) ? matrixObject.toString() : (matrixMetadata !=null ) ? matrixMetadata.toString() : super.toString();"?

src/main/scala/org/apache/sysml/api/ml/BaseSystemMLClassifier.scala
```diff
@@ -225,7 +226,7 @@ trait BaseSystemMLClassifierModel extends BaseSystemMLEstimatorModel {
     // ml.setExplainLevel(ExplainLevel.RECOMPILE_RUNTIME)
     val modelPredict = ml.execute(script._1.in(script._2, X, new MatrixMetadata(X.getNumRows, X.getNumColumns, X.getNonZeros)))
     val ret = PredictionUtils.computePredictedClassLabelsFromProbability(modelPredict, isSingleNode, sc, probVar)
-              .getBinaryBlockMatrix("Prediction").getMatrixBlock
+              .getMatrix("Prediction").toMatrixBlock
```
mboehm7: down the road, it's probably a good idea to modify the APIs of these pipelines to return the model as Matrix instead of as MatrixBlock in order to avoid OOMs in case of huge models.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1537/

deroneriksson:
@mboehm7 Thank you for reviewing. I made the requested updates although I didn't change to the nested ternary operators since I personally have more difficulty reading them than the existing if/else statements.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1538/

PR #534, opened 2017-06-08
[SYSTEMML-1662] Extend Hop Validator

dhutchis:
This PR adds additional checks to the Hop Dag Validator.

* Check that each Hop has the correct number of inputs. Adds a `checkArity()` method to Hop and implements it in every Hop.
* Matrix data type Hops must have a Double value type
* Visit flag must be properly reset via `Hop.resetVisitStatus()`. Uses an external Set to track which Hops have been seen and compares this to the visit flag of each Hop.
* Add related Javadoc 

There is a CHECK flag in `ProgramRewriter` that enables the validator after applying rewrites. It is currently set to true.

There are several tests that fail with this validator in place. We should create additional issues for the problems that the validator uncovers.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -115,7 +115,12 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 		//compute unknown dims and nnz
 		refreshSizeInformation();
 	}
-	
+
+	@Override
+	public int getArity() {
+		return 2;
+	}
+
 	public void setHasLeftPMInput(boolean flag) {
 		_hasLeftPMInput = flag;
 	}
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -118,8 +118,7 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 
 	@Override
 	public void checkArity() throws HopsException {
-		int sz = _input.size();
-		HopsException.check(sz == 2, this, "should have arity 2 but has arity %d", sz);
+		HopsException.check(_input.size() == 2, this, "should have arity 2 but has arity %d", _input.size());
 	}
 
 	public void setHasLeftPMInput(boolean flag) {
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -117,8 +117,9 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 	}
 
 	@Override
-	public int getArity() {
-		return 2;
+	public void checkArity() throws HopsException {
+		int sz = _input.size();
+		HopsException.check(sz == 2, this, "should have arity 2 but has arity %d", sz);
 	}
 
 	public void setHasLeftPMInput(boolean flag) {
--- a/src/main/java/org/apache/sysml/hops/DataOp.java
+++ b/src/main/java/org/apache/sysml/hops/DataOp.java
@@ -182,7 +182,12 @@ public class DataOp extends Hop
 		if (dop == DataOpTypes.TRANSIENTWRITE)
 			setInputFormatType(FileFormatTypes.BINARY);
 	}
-	
+
+	@Override
+	public int getArity() {
+		return -1;
+	}
+
 	public DataOpTypes getDataOpType()
 	{
 		return _dataop;
--- a/src/main/java/org/apache/sysml/hops/DataOp.java
+++ b/src/main/java/org/apache/sysml/hops/DataOp.java
@@ -34,7 +34,10 @@ import org.apache.sysml.runtime.controlprogram.caching.MatrixObject.UpdateType;
 import org.apache.sysml.runtime.matrix.MatrixCharacteristics;
 import org.apache.sysml.runtime.util.LocalFileUtils;
 
-
+/**
+ *  A DataOp can be either a persistent read/write or transient read/write - writes will always have at least one input,
+ *  but all types can have parameters (e.g., for csv literals of delimiter, header, etc).
+ */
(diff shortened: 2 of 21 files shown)


akchinSTC:

deroneriksson:
Test this Jenkins

dhutchis:
Just to be clear, this PR creates many test failures as a result of enabling the Validator

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1551/

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -115,7 +115,13 @@ public AggBinaryOp(String l, DataType dt, ValueType vt, OpOp2 innOp,
 		//compute unknown dims and nnz
 		refreshSizeInformation();
 	}
-	
+
+	@Override
+	public void checkArity() throws HopsException {
+		int sz = _input.size();
+		HopsException.check(sz == 2, this, "should have arity 2 but has arity %d", sz);
```
mboehm7: Could you please substitute sz directly (whenever feasible) to increase readability?

src/main/java/org/apache/sysml/hops/DataOp.java
```diff
@@ -182,7 +185,29 @@ public DataOp(String l, DataType dt, ValueType vt,
...
+	@Override
+	public void checkArity() throws HopsException {
+		int sz = _input.size();
+		int pz = _paramIndexMap.size();
+		switch (_dataop) {
+		case PERSISTENTREAD:
+		case TRANSIENTREAD:
+			HopsException.check(sz == pz, this,
+					"in %s mode has %d inputs and %d parameters",
+					_dataop, sz, pz);
```
mboehm7: please use _dataop.name() instead of the implicit toString.

src/main/java/org/apache/sysml/hops/DataOp.java
```diff
@@ -182,7 +185,29 @@ public DataOp(String l, DataType dt, ValueType vt,
...
+			break;
+		case PERSISTENTWRITE:
+		case TRANSIENTWRITE:
+		case FUNCTIONOUTPUT:
+			HopsException.check(sz == pz + 1, this,
+					"in %s mode has %d inputs and %d parameters (expect 1 more input for write mode)",
```
mboehm7: Please don't call it mode but operator type

src/main/java/org/apache/sysml/hops/Hop.java
```diff
@@ -77,8 +77,8 @@
 	protected long _nnz = -1;
 	protected UpdateType _updateType = UpdateType.COPY;
 
-	protected ArrayList<Hop> _parent = new ArrayList<Hop>();
-	protected ArrayList<Hop> _input = new ArrayList<Hop>();
+	protected ArrayList<Hop> _parent = new ArrayList<>(5);
+	protected ArrayList<Hop> _input = new ArrayList<>(5);
```
mboehm7: Please leave the original specification for now and don't force the initial capacity - the array header anyway dominates here, so allocating an initial array of 5 instead of 10 elements makes no difference.

src/main/java/org/apache/sysml/hops/HopsException.java
```diff
@@ -45,4 +45,14 @@ public HopsException(String message, Throwable cause) {
         super(message, cause);
     }
 
+    /** If the condition fails, print the message formatted with objects. */
```
mboehm7: If you add javadoc, please specify the parameters and conditions, as otherwise this will throw javadoc warnings/errors.

src/main/java/org/apache/sysml/hops/HopsException.java
```diff
@@ -45,4 +45,14 @@ public HopsException(String message, Throwable cause) {
...
+    public static void check(boolean condition, String message, Object... objects) throws HopsException {
+        if (!condition)
+            throw new HopsException(String.format(message, objects));
+    }
+    /** If the condition fails, print the Op and its Id, along with the message formatted with objects. */
```
mboehm7: same as above

src/main/java/org/apache/sysml/hops/rewrite/HopDagValidator.java
```diff
@@ -28,85 +30,98 @@
...
-		throws HopsException
-	{
-		if(hop.isVisited())
-			return;
+	private static void rValidateHop(final Hop hop, final ValidatorState state) throws HopsException {
+		final long id = hop.getHopID();
+
+		final boolean seen = !state.seen.add(id);
+		check(seen == hop.isVisited(), hop,
+				"seen previously is %b but hop visited previously is %b", seen, !seen);
```
mboehm7: Please simplify this message - providing both inputs is not useful because 'seen' implies '!seen'.

src/main/java/org/apache/sysml/hops/rewrite/ProgramRewriter.java
```diff
@@ -55,7 +55,7 @@
 	
 	//internal local debug level
 	private static final boolean LDEBUG = false; 
-	private static final boolean CHECK = false;
+	private static final boolean CHECK = true;
```
mboehm7: DISABLE for the final patch!

dhutchis: Will do :)

deroneriksson:
Test this Jenkins

deroneriksson:
add to whitelist.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1559/

dhutchis:
@mboehm7 Try this set. Let me know if you would prefer a rebase.

mboehm7:
LGTM - thanks @dhutchis. This feature is awesome and already discovered quite a number of subtle bugs.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1564/

PR #540, opened 2017-06-10
[SYSTEMML-1663] Rewrite rule: element-wise multiply chains

dhutchis:
This PR adds the rule `RewriteEMult`. It operates on sub-dags of Hops composed of BinaryOps that do multiplication, re-ordering the multiplicands (by the commutative and associative properties of multiplication) and collapsing identical multiplicands into element-wise powers. For example, it simplifies `A * B * A` to `(A^2) * B`.  Here is a more general example: Hop tree

    ((((A * v) * c) * v) * c) * B

is rewritten to

    A * (B * ((v^2) * (c^2)))

Notice that the matrices were grouped together (they are actually ordered so that the sparsest matrix appears first, which later sparsity-driving fusion can take advantage of), then the vectors are grouped together (more precisely, row vectors then column vectors), and then the scalars. Whenever a multiplicand appears more than once, it is consolidated into a power, to the number of times it appears.

This rule interacts with a few other rules. I modified them as follows.

1. In `AggUnaryOp`, I added conditions to the `TernaryAggregate` detection method to act on expressions of the form `sum(A^2 * B)`. Since I was knee-deep in this method anyway, I also added a condition to detect `sum(A^3)`. These changes were necessary since prior to RewriteEMult, they would have been `sum(A * B * A)` or `sum(A * A * A)`, both of which are converted into TernaryAggregates which is a faster, fused operator (i.e., it is faster because it produces fewer intermediates).

2. In the dynamic algebraic rewrite `simplifyDotProductSum`, I added a condition to make it play nicely with the `TernaryAggregate` rewrite described above.

The following new tests verify these rewrites:

1. `RewriteEMultChainTest` tests rewriting the expression `2*X*3*Y*4*X`. The constants in the expression group together and are constant folded into 24.  The resulting expression is `Y*(X^2)*24`.
2. `ABATernaryAggregateTest` ensures that TernaryAggregate plays nicely with RewriteEMult.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1565/

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -490,33 +490,38 @@ private boolean isTernaryAggregateRewriteApplicable(ExecType et)
...
-							&& HopRewriteUtils.isEqualSize(input12.getInput().get(1), input1)	
-							&& HopRewriteUtils.isEqualSize(input11, input1);
+			if (input1.getParent().size() == 1
+					&& input1 instanceof BinaryOp) { //sum single consumer
+				BinaryOp binput1 = (BinaryOp)input1;
+
+				if (binput1.getOp() == OpOp2.POW
+						&& binput1.getInput().get(1) instanceof LiteralOp) {
+					LiteralOp lit = (LiteralOp)binput1.getInput().get(1);
+					ret = lit.getLongValue() == 3;
```
mboehm7: There is a HopRewriteUtil method to probe for literals of a given value.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -627,28 +632,68 @@ private boolean isUnaryAggregateOuterCPRewriteApplicable()
...
+			switch (b11.getOp()) {
+			case MULT: // A*B*C case
+				in1 = input11.getInput().get(0).constructLops();
+				in2 = input11.getInput().get(1).constructLops();
+				in3 = input12.constructLops();
+				handled = true;
+				break;
+			case POW: // A*A*B case
+				Hop b112 = b11.getInput().get(1);
+				if ( !(input12 instanceof BinaryOp && ((BinaryOp)input12).getOp()==OpOp2.MULT)
```
mboehm7: just for simplification, please use the existing HopRewriteUtil method for these checks.

src/main/java/org/apache/sysml/hops/OptimizerUtils.java
```diff
@@ -110,8 +110,13 @@
...
 	public static boolean ALLOW_CONSTANT_FOLDING = true;
 	
-	public static boolean ALLOW_ALGEBRAIC_SIMPLIFICATION = true; 
-	public static boolean ALLOW_OPERATOR_FUSION = true; 
+	public static boolean ALLOW_ALGEBRAIC_SIMPLIFICATION = true;
+	/**
+	 * Enables rewriting chains of element-wise multiplies that contain the same multiplicand more than once, as in
+	 * `A*B*A ==> (A^2)*B`.
+	 */
+	public static boolean ALLOW_EMULT_CHAIN_REWRITE = true;
```
mboehm7: Please don't introduce a new constant here but use the existing ALLOW_SUM_PRODUCT_REWRITES (which applies to a number of other rewrites and operators such as mmchain and tak+ as well)

src/main/java/org/apache/sysml/hops/rewrite/HopRewriteUtils.java
```diff
@@ -241,7 +241,24 @@ public static void addChildReference( Hop parent, Hop child, int pos ){
...
+	 * Replace an old Hop with a replacement Hop.
+	 * If the old Hop has no parents, then return the replacement.
+	 * Otherwise rewire each of the Hop's parents into the replacement and return the replacement.
+	 * @param old To be replaced
+	 * @param replacement The replacement
+	 * @return replacement
+	 */
+	public static Hop replaceHop(final Hop old, final Hop replacement) {
+		final ArrayList<Hop> rootParents = old.getParent();
+		if (rootParents.isEmpty())
```
mboehm7: Why was there a need to introduce this custom replacement method? The case of empty parents should never happen for these rewrites because there will always be a consumer such as u(print), d(twrite), d(pwrite), or fun; otherwise the expression would be dead code eliminated before.

src/main/java/org/apache/sysml/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
```diff
@@ -2050,7 +2050,14 @@ private Hop simplifyDotProductSum(Hop parent, Hop hi, int pos)
...
 					&& !HopRewriteUtils.isBinary(hi2.getInput().get(0), OpOp2.MULT)
-					&& !HopRewriteUtils.isBinary(hi2.getInput().get(1), OpOp2.MULT) )
+					&& !HopRewriteUtils.isBinary(hi2.getInput().get(1), OpOp2.MULT)
+					&& !(HopRewriteUtils.isBinary(hi2.getInput().get(0), OpOp2.POW)      // do not rewrite (A^2)*B
+						&& hi2.getInput().get(0).getInput().get(1) instanceof LiteralOp  // let tak+* handle it
+						&& ((LiteralOp)hi2.getInput().get(0).getInput().get(1)).getLongValue() == 2)
+					&& !(HopRewriteUtils.isBinary(hi2.getInput().get(1), OpOp2.POW)      // do not rewrite B*(A^2)
+						&& hi2.getInput().get(1).getInput().get(1) instanceof LiteralOp  // let tak+* handle it
+						&& ((LiteralOp)hi2.getInput().get(1).getInput().get(1)).getLongValue() == 2)
+					)
```
mboehm7: this is good but could please make these additional checks conditional on ALLOW_SUM_PRODUCT_REWRITES?

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+import java.util.TreeMap;
+
+import org.apache.sysml.hops.BinaryOp;
+import org.apache.sysml.hops.Hop;
+import org.apache.sysml.hops.HopsException;
+import org.apache.sysml.hops.LiteralOp;
+import org.apache.sysml.parser.Expression;
+
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
```
mboehm7: We removed the dependency to google guava a while ago because we didn't use it a lot and it frequently resulted in version conflicts regarding our different backends. I just double checked the code base, and apparently the new Caffe2DML tool reintroduced this dependency (cc @niketanpansare). Since we'll likely clean this up before our SystemML 1.0 release, it would be great if you could use default Java or commons collections.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+ * Prerequisite: RewriteCommonSubexpressionElimination must run before this rule.
+ *
+ * Rewrite a chain of element-wise multiply hops that contain identical elements.
+ * For example `(B * A) * B` is rewritten to `A * (B^2)` (or `(B^2) * A`), where `^` is element-wise power.
+ * The order of the multiplicands depends on their data types, dimentions (matrix or vector), and sparsity.
+ *
+ * Does not rewrite in the presence of foreign parents in the middle of the e-wise multiply chain,
+ * since foreign parents may rely on the individual results.
+ */
+public class RewriteEMult extends HopRewriteRule {
```
mboehm7: Could we please rename this to RewriteElementwiseMultChainOptimization (for consistency with our existing RewriteMatrixMultChainOptimization)?

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+			// the hop with more nnz is first; unknown nnz (-1) last
+			int c = Long.compare(o1.getNnz(), o2.getNnz());
+			if (c != 0) return c;
+			return Long.compare(o1.getHopID(), o2.getHopID());
+		}
+		private final int[] orderDataType;
+		{
+			Expression.DataType[] dtValues = Expression.DataType.values();
+			orderDataType = new int[dtValues.length];
+			for (int i = 0, valuesLength = dtValues.length; i < valuesLength; i++) {
```
mboehm7: This comparator is a little verbose - couldn't we simply create a tiny lookup table which would allow something like "rank = rankLookup[dt.ordinal()]"

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+	 * Check if a node has a parent that is not in the set of emults. Recursively check children who are also emults.
+	 * @param emults The set of BinaryOp element-wise multiply hops in the emult chain.
+	 * @param child An interior emult hop in the emult chain dag.
+	 * @return Whether this interior emult or any child emult has a foreign parent.
+	 */
+	private static boolean checkForeignParent(final Set<BinaryOp> emults, final BinaryOp child) {
+		final ArrayList<Hop> parents = child.getParent();
+		if (parents.size() > 1)
+			for (final Hop parent : parents)
+				//noinspection SuspiciousMethodCalls
```
mboehm7: What does this comment mean in this context?

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+		emults.add(root);
+
+		final ArrayList<Hop> inputs = root.getInput();
+		final Hop left = inputs.get(0), right = inputs.get(1);
+
+		if (isBinaryMult(left)) findEMultsAndLeaves((BinaryOp) left, emults, leaves);
+		else leaves.add(left);
+
+		if (isBinaryMult(right)) findEMultsAndLeaves((BinaryOp) right, emults, leaves);
+		else leaves.add(right);
```
mboehm7: formatting

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+	 * @param emults The set of BinaryOp element-wise multiply hops in the emult chain.
+	 * @param leaves The multiset of multiplicands in the emult chain.
+	 * @return If the multiset is worth optimizing.
+	 */
+	private static boolean isOptimizable(Set<BinaryOp> emults, final Multiset<Hop> leaves) {
+		// Old criterion: there should be at least one repeated leaf
+//		for (Multiset.Entry<Hop> hopEntry : leaves.entrySet()) {
+//			if (hopEntry.getCount() > 1)
+//				return true;
+//		}
```
mboehm7: Please remove commented code.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
```diff
@@ -0,0 +1,284 @@
...
+//		return false;
+		return emults.size() >= 2;
```
mboehm7: Why is it >= 2 and not >2. For two inputs, there is nothing to reorder, and the existing algebraic simplification rewrites will anyway canonicalize a A*A to A^2.

src/test/java/org/apache/sysml/test/integration/functions/misc/RewriteEMultChainTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.sysml.test.integration.TestConfiguration;
+import org.apache.sysml.test.utils.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Test whether `A*B*A` successfully rewrites to `(A^2)*B`.
+ */
+public class RewriteEMultChainTest extends AutomatedTestBase
+{
```
mboehm7: Please add this test to org.apache.sysml.test.integration.functions.misc.ZPackageSuite.

src/test/java/org/apache/sysml/test/integration/functions/misc/RewriteEMultChainTest.java
```diff
@@ -0,0 +1,127 @@
...
+	private static final String TEST_NAME1 = "RewriteEMultChainOpXYX";
+	private static final String TEST_DIR = "functions/misc/";
+	private static final String TEST_CLASS_DIR = TEST_DIR + RewriteEMultChainTest.class.getSimpleName() + "/";
+	
+	private static final int rows = 123;
+	private static final int cols = 321;
```
mboehm7: It would be good to add also some tests that involve scalars and vectors.

src/test/java/org/apache/sysml/test/integration/functions/ternary/ABATernaryAggregateTest.java
```diff
@@ -0,0 +1,417 @@
...
+import org.apache.sysml.utils.Statistics;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Similar to {@link TernaryAggregateTest} except that it tests `sum(A*B*A)`.
+ * Checks compatibility with {@link org.apache.sysml.hops.rewrite.RewriteEMult}.
+ */
+public class ABATernaryAggregateTest extends AutomatedTestBase
+{
```
mboehm7: Please add this test to org.apache.sysml.test.integration.functions.ternary.ZPackageSuite.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
mboehm7: cc @deroneriksson (regarding guava)

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
dhutchis: This is the number of emult hops. If there are 2 emults, then there are 3 inputs, and so we may do useful work

mboehm7: got it - thanks for clarifying.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
dhutchis: I left this in to show the (weaker) alternative I first considered - only rewriting expressions when at least one multiplicand occurs more than once. If we are agreed on rewriting more generally any emult chain with at least 2 emults (3 inputs), then I will delete it.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
dhutchis: `emults` is a set of BinaryOp. `parent` can be any Hop. Intellij warns when calling `contains()` with a different type than the type of the set; in this case Hop instead of BinaryOp. However the call is safe and so I include the statement to suppress the warning.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
dhutchis: Yes, I do have a small lookup table; see `orderDataType`. The lookup table is not sufficient because we need to distinguish between two Hops that are both matrices.  Two matrices should be checked to see if they are (row or column) vectors, and further to see if we have sparsity information on them, since it is useful to group vectors together as well as to lead the emult with the sparsest matrices.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
mboehm7: ok interesting - could you please use @SuppressWarnings instead.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
mboehm7: Actually I like the sorting a lot because this canonicalizes these expressions. Down the road, we will likely normalize more general expressions so, please go for the current option.

src/main/java/org/apache/sysml/hops/rewrite/HopRewriteUtils.java
dhutchis: This seems like a generally useful method. It correctly rewires an old Hop to a new one regardless of whether the old Hop has parents (i.e. is interior) or not (i.e. is a root). 

Though your logic is sound, why make developers think hard about whether they can safely assume a Hop could be a root or not when they could use one method that always works? Also, if dead code elimination didn't run for some reason, then the code without this method would become buggy, so that option seems less resilient.

It's a minor thing, but I think using a method like this to do rewiring would make life easier for developers.

src/main/java/org/apache/sysml/hops/rewrite/RewriteEMult.java
mboehm7: yes, you're right - I somehow read this as method; In that case, please move the member to the head of the class, and try to eliminate verbosity (for example, we could express the same thing in 6 lines of code via something like "orderedDataType[DataType.SCALAR.ordinal()] = 4;", etc)

src/main/java/org/apache/sysml/hops/rewrite/HopRewriteUtils.java
mboehm7: well, it's a nice to have comment but it's generally useful to use common primitives for all rewrites (which makes later refactoring much simpler). Maybe you can simply add this empty check to rewireAllParentChildReferences (which would also work for empty parents) in order to avoid the unnecessary ArrayList creation?

mboehm7:
btw, there are apparently a number of failed tests - I would recommend to start with tests like GLMDMLTest which currently fail with index out of bounds exceptions. 

Subsequently, we need to look at a couple of tests where the new rewrite destroys sparsity exploiting operators (e.g., WeightedUnaryMatrixMultTest, where W*(2*(U%*%t(V))) is reordered to something like 2*(W*(U%*%t(V)))) which creates an unnecessary sparse intermediate, while otherwise we would compile a single fused operator. We might want to consider a different ordering of matrices - (scalars) - vectors, where all scalars evaluated as a sub dag first.

dhutchis:
I added scalars to the test case. These sort as expected, though I did have the sort order reversed at first (it is so easy to get the wrong order on compare methods). Another rule splifies the scalars later.  The test is `2*X*3*Y*4*X` which rewrites to `Y*(X^2)*24`.

I also manually checked that vectors sort correctly. Unfortunately I couldn't figure out how to get R to element-wise multiply a matrix by a scalar, so I left this out. I also don't know how to check the output to see if the scalar rewriting (to 24) or vector re-ordering is successful, since these do not appear in the heavy hitter output.

dhutchis:
Ah okay, I see the overlap. Sure, let's go with the existing method.

dhutchis:
The tests should pass now. I had a visit status "hole in the middle" bug ~_~, now fixed

dhutchis:
Welp, I take that back. The test `ElementwiseAdditionMultiplicationTest.testTerm3` fails due to a numerical problem. This test computes `(A*B)*C`.  The rewrite re-orders this multiplication to `A*(B*C)`. Evidently this is enough of a difference to make the results not match. Look at these differences:

	(1,1) mismatch: expected 369.2084378683617, actual 369.2084378683618
	(3,3) mismatch: expected 656.166016794371, actual 656.1660167943709
	(2,3) mismatch: expected 42.394954799451014, actual 42.39495479945101
	(3,1) mismatch: expected 125.49664338031643, actual 125.49664338031641
	(1,2) mismatch: expected 222.4603650962903, actual 222.46036509629027
	(5,1) mismatch: expected 20.237240311264287, actual 20.237240311264284
	(1,4) mismatch: expected 183.04079237620041, actual 183.04079237620044
	(2,4) mismatch: expected 213.53422063724432, actual 213.53422063724435
	(5,3) mismatch: expected 31.27887759608941, actual 31.278877596089405
	(3,4) mismatch: expected 96.92834272428424, actual 96.92834272428425
	(5,2) mismatch: expected 61.48072685290718, actual 61.480726852907175

I say we change the test :)

We'll see if there are other problems, like the sparsity-order-exploiting case you mentioned.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1570/

mboehm7:
great - good progress. 

1) Regarding the check for scalar rewriting, you could run in forced spark execution mode and check for the number of execute spark instructions (scalar operations will never result in distributed operations). 

2) So I hope you already used your extended hop dag validator to find it. ;-)

3) Yes, you can change the test to use a tolerance of 1e-10, which can be passed as epsilon into compareResults.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1571/

src/main/java/org/apache/sysml/hops/rewrite/ProgramRewriter.java
```diff
@@ -96,6 +96,8 @@ public ProgramRewriter( boolean staticRewrites, boolean dynamicRewrites )
 			_dagRuleSet.add(     new RewriteRemoveUnnecessaryCasts()             );		
 			if( OptimizerUtils.ALLOW_COMMON_SUBEXPRESSION_ELIMINATION )
 				_dagRuleSet.add( new RewriteCommonSubexpressionElimination()     );
+			if ( OptimizerUtils.ALLOW_SUM_PRODUCT_REWRITES)
+				_dagRuleSet.add( new RewriteElementwiseMultChainOptimization()                              ); //dependency: cse
```
mboehm7: Formatting: please align the closing parenthesis.

dhutchis:
Great, here is the rebase. Thanks for timing this on the application scripts, @mboehm7

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1614/

deroneriksson:
@mboehm7
Please note that re-introduction of guava classes in the project such as in HopDagValidator and RewriteElementwiseMultChainOptimization means that the guava dependency needs to be handled in situations such as scoring using JMLC. The guava-11.0.2 jar is 1.6M, so this can potentially impact the memory footprint required to do JMLC.

cc @bertholdreinwald @frreiss

mboehm7:
@deroneriksson - as discussed in the comments above, I absolutely agree but did not want to unnecessarily stall the PR for this; I'll take of that in the context of testing this rewrite more thoroughly in the wild.

deroneriksson:
Great @mboehm7. I failed to see the comment WRT guava earlier. I see from the above comments that you noticed the guava dependency immediately. Good catch!

dhutchis:
I made a PR #549 that removes the added Guava dependencies. Sorry for the trouble.

deroneriksson:
@dhutchis Thanks! Usually I would highly encourage the use of a library like Guava in this situation. We just have a constraint to try minimize the dependencies/memory needed for the in-memory scoring via JMLC.

PR #541, opened 2017-06-13
[SYSTEMML-1686] Fix transpose conv2d filter shape & input size

dusenberrymw:
Currently, the transpose conv2d layer (`nn/layers/conv2d_tranpose.dml`)
has a bug in which the filters tensor `W` has an incorrect shape, and
the `conv2d_backward_data` op has an incorrect input shape argument.
This results in an exception when the number of input channels `C` is
not equal to the number of filters `F` (i.e. number of output channels).
Since the transpose conv2d op is the gradient of the conv2d op, the
filter tensor needs to have the shape `(C, F, Hf, Wf)` for `F` filters,
rather than `(F, C, Hf, Wf)`, in order to map from an input with `C`
channels to an output with `F` channels during the input data gradient
function (`conv2d_backward_data`) that is used in the forward pass.
Additionally, the `input_shape` argument for `conv2d_backward_data`
needs to be `(N, F, Hout, Wout)`, rather than `(N, C, Hout, Wout)` in
order to map from an input with `C` channels to an output with `F`
channels.  Our previous test cases did not catch this issue because
the tests used `C = F = 1`.

This commit fixes these issues, and improves the robustness of the
associated tests.


dusenberrymw:
cc @prithvirajsen, @niketanpansare, @bertholdreinwald

dusenberrymw:
@niketanpansare This will require a change in Caffe2DML for the shape of the filters `W`, unless you are using the `init` function already.  We still are using `F` filters, but the shape of `W` is now `(C, F*Hf*Wf)`.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1575/

prithvirajsen:
I'm having trouble understanding the organization of the numbers within a filter. For the simple case of F=2, C=1, Hf=2, Wf=2 such that the first filter is
`1 2
3 4`
and the second filter is
`5 6
7 8`
what should W look like?

dusenberrymw:
@prithvirajsen For that example, `W` would be a single row that looks like `1 2 3 4 5 6 7 8`.  The shape would be `(C, F*Hf*Wf) = (1, 2*Hf*Wf)`.  Essentially, we are just taking the normal `(F, C, Hf, Wf)` filter tensor, transposing it to `(C, F, Hf, Wf)` and then flattening the latter three dimensions as usual to `(C, F*Hf*Wf)`.

The motivation here is to think about the shape of `W` during a regular 2D conv function, i.e. `(F, C, Hf, Wf)`.  In the forward *and* backward pass of regular 2D conv, `W` maintains the same `(F, C, Hf, Wf)`shape.  Specifically, in the backward pass, the data gradient function `conv2d_backward_data` accepts the same `W` of shape `(F, C, Hf, Wf)` to map from gradients wrt the output of shape `(N, F, Hout, Wout)` to gradients wrt to the input of shape `(N, F, Hin, Win)`.

In a 2D transpose conv function, we use the 2D conv data gradient function as the forward function.  If we use the same size terminology as from regular 2D conv, we would be mapping from an input with `F` channels to an output with `C` channels (which seems backward, but is correct since the forward pass of 2D transpose conv is defined as the backward pass of 2D conv), and we would be using the same filter `W` of shape `(F, C, Hf, Wf)`.  Now, in order to make the terminology of the 2D transpose conv layer consistent with the rest of the library, we would need to swap the variable names `F` and `C` so that we are mapping from an input with `C` channels to an output with `F` channels.  Thus, we would use a filter `W` of shape `(C, F, Hf, Wf`).

dusenberrymw:
I just pushed a fix for a bug in the associated unit test, and in the process have uncovered an optimizer bug.  Essentially, for the [`conv2d_transpose` unit test](https://github.com/apache/systemml/pull/541/files#diff-69f1f9f08f0278c1419977edb4932fa1L112), the [`bias_add` function in `conv2d_transpose::forward`](https://github.com/apache/systemml/pull/541/files#diff-46627746109842440ab3f87b3e9b9197L85) is causing the number of rows in `out` to incorrectly increase from `2` to `3`.  If an `if (1==1) {}` cut is added just before the `bias_add` function, the issue does not occur.  Interestingly, the [gradient check for `conv2d_transpose`](https://github.com/dusenberrymw/systemml/blob/181585887fcf50ae7e9f7ec475a8596c63416916/scripts/nn/test/grad_check.dml#L620) does *not* have the same issue, despite also calling the same `conv2d_transpose::forward` function.

cc @mboehm7 Can you look into this?

cc @bertholdreinwald, @niketanpansare

mboehm7:
@niketanpansare do you wanna look into this bug described above (as it's likely related to your ConvolutionOp) - just let me know

niketanpansare:
Sure, will look into it.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1578/

prithvirajsen:
@dusenberrymw You answered my question but I am not sure I fully understood the representation.

For the following example W in FCHW format with F=C=Hf=Wf=2:
`20 21 22 23 24 25 26 27`
`30 31 32 33 34 35 36 37`
Is this the corresponding CFHW representation?
`20 21 22 23 30 31 32 33`
`24 25 26 27 34 35 36 37`

If yes, then I think the above transformation can be achieved with a simple DML function. Do you think that would be useful?

dusenberrymw:
Yeah`W` is a 4D tensor, and it is transformed from shape `(F, C, Hin, Win)` to `(C, F, Hin, Win)` by transposing the first two dimensions.

Here's some Python code for demonstration, using your example `W`.

1. Create `W` with shape `(F,C,Hin,Win)`:
```python
import numpy as np
F = C = Hin = Win = 2
w = np.array([20, 21, 22, 23, 24, 25, 26, 27, 30, 31, 32, 33, 34, 35, 36, 37])
w = w.reshape(F, C, Hin, Win)
print(w)
```
```
[[[[20 21]
   [22 23]]

  [[24 25]
   [26 27]]]


 [[[30 31]
   [32 33]]

  [[34 35]
   [36 37]]]]
```

2. Transpose to `(C, F, Hin, Win)`:
```python
w = w.transpose(1, 0, 2, 3)  # (F,C,Hin,Win) -> (C,F,Hin,Win)
print(w)
```
```
[[[[20 21]
   [22 23]]

  [[30 31]
   [32 33]]]


 [[[24 25]
   [26 27]]

  [[34 35]
   [36 37]]]]
```

3. Flatten to `(C, F*Hin*Win)` for SystemML:
```python
w = w.reshape(C, F*Hin*Win)
print(w)
```
```
[[20 21 22 23 30 31 32 33]
 [24 25 26 27 34 35 36 37]]
```

In general, having a DML `nn/util::transpose4d(dim1, dim2, dim3, dim4)` function would be useful for arbitrary dimension permutations of our (flattened) tensors, as is possible in NumPy.  If we had that function, then this layer could accept the normal filter tensor `W` of shape `(F, C*Hin*Win)` and then just transpose it to `(C, F*Hin*Win)` before usage.  For this layer, the drawback to using that would be having to perform the transposition during every `forward` & `backward` call.

TensorFlow follows the same approach of swapping the `F` & `C` dimensions for performance reasons (keep in mind that the spatial dimensions come first in TF).  [`conv2d`](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d) uses a filter tensor of shape `(Hf, Wf, C, F)`, and [`conv2d_transpose`](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose) use a filter tensor of shape `(Hf, Wf, F, C)`.

prithvirajsen:
I wouldn't want to transform W each time forward/backward is called either. The transformation should only be done once and then the transformed W should continue to be used.

If you're referring to a generic reshape that can perform arbitrary reshapes on 4D tensors, then I'm not sure if this can be achieved efficiently in DML (I think the corresponding builtin function would be very useful). A special reshape for FCHW -> CFHW is however, achievable in DML.

mboehm7:
ok quick update on the "optimizer issue" - it has nothing to do with the compiler or runtime; the root cause is that CONV2D_BACKWARD_DATA messes up the dimensions during parsing, BIAS_ADD propagates these dims, the nrow is replaced abd never reaches the compiler (which would actually correct these mistakes).

dusenberrymw:
Thanks, @mboehm7.  That fixed the issue.  Does anyone else have any remaining concerns regarding this PR?

prithvirajsen:
looks good to me.

niketanpansare:
Thanks @dusenberrymw LGTM :+1:

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1582/

PR #542, opened 2017-06-15
[SYSTEMML-1674][SYSTEMML-1675] Add new 2D depthwise convolution layers

dusenberrymw:
This adds (1) a new 2D depthwise convolution layer, and (2) a new 2D depthwise transpose convolution layer.

A depthwise convolution (1) applies a different set of M filters to each input channel separately, thus expanding each input channel to M output channels, and (2) concatenates the results into a single volume with C*M output channels.  This is in contrast to a regular 2D convolution, in which all of the filters would be applied to all of the input channels at once.

A depthwise transpose convolution (1) applies a different filter to each unique group of M input channels separately, thus condensing each group of M input channels to 1 output channel, and (2) concatenates the results into a single volume with C/M output channels.  This is in contrast to a regular 2D transpose convolution, in which all of the filters would be applied to all of the input channels at once.

These layers use for loop implementations for now, and I'll work on possible vectorized updates at a later date.

In addition to the new layers, this also adds the associated unit and gradient tests.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1583/

niketanpansare:
@dusenberrymw for integration with Caffe2DML, does this layer map to:
```
layer {
  type: "Deconvolution"
  convolution_param {
    num_output: 254
    bias_term: false
    pad: 1
    kernel_size: 4
    group: 254
    stride: 2
    weight_filler {
      type: "bilinear"
    }
  }
}
```

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1585/

dusenberrymw:
@niketanpansare Well for this particular PR, `conv2d_depthwise.dml` is the regular "Convolution", but I'm working on the related one for conv2d transpose ("deconvolution").  For the `groups` attribute, it will be a depthwise conv when `groups` is equal to the number of **input** channels, i.e. the `num_output` from the **previous** layer.  Also, I would think that Caffe will have an upper bound on the number of groups as the minimum of either the number of input channels or the number of filters.  Otherwise, it wouldn't be possible to divide the number of input channels by the number of groups if there are more groups than input channels.  It would be worth checking on that.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1609/

dusenberrymw:
cc @niketanpansare, @prithvirajsen, @bertholdreinwald

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1615/

prithvirajsen:
Looks good to me.

The `for` loops may add to running time (why are the `parfor`s commented out?). At some point (when the builtin functions for conv2d, conv2d_backaward_* are relatively stable), we should compare running times for the nn layer functions against other packages.

dusenberrymw:
Thanks, @prithvirajsen.  I agree that we should compare individual layers with other systems at some point in the near future.  As for the commented out `parfor` loops, I originally wanted to use them, but they first caused exceptions (fixed in master), and then they end up being slower.  The builtin `conv2d*` functions are internally multithreaded, but if they are run in the context of a local `parfor`, they are reduced to single-threaded ops, and the overall runtime slows down (~1.5x - 2x in my quick tests).  It's worth exploring more at some point, because for large numbers of channels >> the number of the examples, the `parfor` loops could possibly be faster.

PR #543, opened 2017-06-15
[SYSTEMML-540] Allow for `allreduce_parallel_batches` for multi-GPU training

niketanpansare:
- Introduced `allreduce_parallel_batches`  for multi-GPU training.
- Moved `train_algo` and `test_algo` from solver specification to Python API to conform with Caffe.
- Updated documentation for Caffe2DML.

@dusenberrymw @bertholdreinwald


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1586/

dusenberrymw:
👍  Awesome!

src/main/scala/org/apache/sysml/api/dl/Caffe2DML.scala
```diff
@@ -368,23 +483,22 @@ class Caffe2DML(val sc: SparkContext, val solverParam:Caffe.SolverParameter,
...
+    tabDMLScript.append("# Data structure to store gradients computed in parallel\n")
     net.getLayers.map(layer => net.getCaffeLayer(layer)).map(l => {
-      if(l.shouldUpdateWeight) assign(tabDMLScript, l.dWeight + "_agg", matrix("0", "parallel_batches", multiply(nrow(l.weight), ncol(l.weight))))
-      if(l.shouldUpdateBias) assign(tabDMLScript, l.dBias + "_agg", matrix("0", "parallel_batches", multiply(nrow(l.bias), ncol(l.bias)))) 
+      if(l.shouldUpdateWeight) assign(tabDMLScript, l.dWeight + "_agg", matrix("0", parallel_batches, multiply(nrow(l.weight), ncol(l.weight))))
+      if(l.shouldUpdateBias) assign(tabDMLScript, l.dBias + "_agg", matrix("0", parallel_batches, multiply(nrow(l.bias), ncol(l.bias)))) 
     })
   }
-  private def flattenAndStoreAggGradients_j():Unit = {
+  private def flattenGradients():Unit = {
```
dusenberrymw: There's still one subtle piece here that could be addressed for the allreduce parallel batches case.  In that scenario, we are effectively computing gradients for the larger group of mini-batches (i.e. `group_batch_size`), but we split it up into mini-batches of size `batchSize` and compute the gradients for each mini-batch in the group.  So, the gradient from **each** mini-batch in the group will be the average gradient over **`batch_size`** examples (since we are minimizing an average loss function).  When we flatten and aggregate them all together, we want the final gradients to be the average gradients over **`group_batch_size`** (that way, it is equivalent to an allreduce over a single batch of size `group_batch_size`).  If we simply flatten, store, and then add them all together, then the resulting gradient after aggregation will be `group_batch_size/batch_size` times larger than it should be, and thus the user's learning rate would need to be adjusted.  Therefore, we can re-weight the mini-batch gradients in the flatten stage so that when we add them together in the aggregation stage, the resulting gradients will be the average gradient over `group_batch_size`.  The benefit is that the learning rate will not need to be changed, regardless of whether we use allreduce or allreduce_parallel_batches, and thus the user will see equivalent mathematical convergence to an implementation that simply operated directly on a one batch of size `group_batch_size` at a time, while experiencing a systems performance speedup due to us cutting up the `group_batch_size` into smaller batches and running them in parallel.  We could imagine fixing the issue during the aggregation step instead of the flatten step, but since the final batch in the group could be of a smaller size (i.e. when `group_batch_size` isn't perfectly divisible by `batch_size`), performing the re-weighting during (or just before) the flatten step maintains correctness in all scenarios.

Effectively, this comes done to multiplying the flattened gradients (`matrix(l.dWeight, "1", multiply(nrow(l.weight), ncol(l.weight))` and equivalent for the biases) by a `weighting` factor during the flatten step, where `weighting = batch_size / group_batch_size`.  Equivalently, it might be cleaner to add a `reweightGradients(...)` function that is called just before `flattenGradients()`.

Here are the relevant lines from the distributed LeNet example: https://github.com/apache/systemml/blob/master/scripts/nn/examples/mnist_lenet_distrib_sgd.dml#L202-L204

Finally.... awesome work!  Really excited about the progress on the Caffe2DML project. 👍

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1603/

src/main/scala/org/apache/sysml/api/dl/Caffe2DML.scala
niketanpansare: Addressed it in https://github.com/apache/systemml/pull/527/files#diff-6381abbdd5a26b5c5274c00903555197R513

PR #546, opened 2017-06-16
[SYSTEMML-1701] fix need to use -force for gpu

nakul02:
@niketanpansare -
It turns out that I had removed the code to initialize the initial GPU memory budget as part of my large GPU backend redesign (My bad).
It has now been restored in a more appropriate place.

@bertholdreinwald -
It also turns out that Estimates for the GPU are the same as for CP operations. In so far as I can tell, this should work in most cases. I will have to look at the implementations carefully.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -49,6 +49,7 @@ import org.apache.sysml.parser.Expression.DataType;
 import org.apache.sysml.parser.Expression.ValueType;
 import org.apache.sysml.runtime.controlprogram.ParForProgramBlock.PDataPartitionFormat;
 import org.apache.sysml.runtime.controlprogram.context.SparkExecutionContext;
+import org.apache.sysml.runtime.instructions.gpu.context.GPUContextPool;
 import org.apache.sysml.runtime.matrix.MatrixCharacteristics;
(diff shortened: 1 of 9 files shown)


mboehm7:
the memory estimates will not work if sparse inputs are converted to dense for any GPU operation or if the requirements for intermediate memory are different.

mboehm7:
please also think about how to handle the case where the driver memory budget is smaller than the GPU device memory.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1590/

nakul02:
Thanks for the feedback @mboehm7 
I had not considered the case when the driver memory budget is smaller than the GPU device memory.

I shall work on the memory estimates in detail in another JIRA/PR. 
This one contains the fix for properly initializing the initial GPU memory budget.

niketanpansare:
>> please also think about how to handle the case where the driver memory budget is smaller than the GPU device memory.

Good point. We need to have a separate discussion on how to handle this case optimally (for example: partitioned mb in CP). For 1.0, we could use `min(CP memory budget, GPU memory)`.

>> the memory estimates will not work if sparse inputs are converted to dense for any GPU operation or if the requirements for intermediate memory are different.

Sure, we can handle this in a separate PR.

LGTM. Thanks @nakul02  👍

nakul02:
Thanks @niketanpansare, I shall merge.

PR #548, opened 2017-06-16
[SYSTEMML-1679] Add a new threshold utility function

feihugis:
This function accepts a matrix X and a threshold parameter thresh to get an indicator matrix with
values in {0, 1} depending on whether or not the values in X are above thresh. It can be used, for
example, to determine the predicted class in a binary classification problem given the output of
a sigmoid layer.

### DIFF
--- a/scripts/nn/test/test.dml
+++ b/scripts/nn/test/test.dml
@@ -605,3 +605,20 @@ tanh = function() {
   }
 }
 
+threshold = function() {
+    /*
+     * Test for threshold function.
+    */
+    print("Testing the threshold function.")
+
+    # Generate data
+    matrix = matrix("0.31 0.24 0.87
+
+                    0.45 0.66, 0.65
+
+                    0.24 0.91 0.13", rows=3, cols=3)
+    thresh = 0.5
+
+    # Get the indicator matrix
+    indicator_matrix = util::threshold(matrix, thresh)
+}
--- a/scripts/nn/test/test.dml
+++ b/scripts/nn/test/test.dml
@@ -612,13 +612,20 @@ threshold = function() {
     print("Testing the threshold function.")
 
     # Generate data
-    matrix = matrix("0.31 0.24 0.87
+    X = matrix("0.31 0.24 0.87
 
-                    0.45 0.66, 0.65
+                0.45 0.66 0.65
 
-                    0.24 0.91 0.13", rows=3, cols=3)
+                0.24 0.91 0.13", rows=3, cols=3)
     thresh = 0.5
+    target_matrix = matrix("0.0 0.0 1.0
+
+                            0.0 1.0 1.0
+
+                            0.0 1.0 0.0", rows=3, cols=3)
 
     # Get the indicator matrix
-    indicator_matrix = util::threshold(matrix, thresh)
+    indicator_matrix = util::threshold(X, thresh)
+
+    out = test_util::check_all_equal(indicator_matrix, target_matrix)
 }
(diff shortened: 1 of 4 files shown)


scripts/nn/util.dml
```diff
@@ -200,3 +200,19 @@ unpad_image = function(matrix[double] img_padded, int Hin, int Win, int padh, in
...
 }
 
+threshold = function(matrix[double] X, double thresh)
+    return (matrix[double] out) {
+  /*
+   * Computes an indicator matrix with values in (0, 1) depending on whether or not the values in X
+   * are above the input threshold
+   *
+   * Inputs:
+   *  - X: Input Matrix.
```
dusenberrymw: Just for consistency with the rest of the library can you update this to `X: Inputs, of shape (any, any).
`?

scripts/nn/util.dml
```diff
@@ -200,3 +200,19 @@ unpad_image = function(matrix[double] img_padded, int Hin, int Win, int padh, in
...
+   *  - thresh: Input threshold
+   *
+   * Outputs:
+   *  - out: Output indicator matrix
```
dusenberrymw: Just for consistency with the rest of the library can you update this to `out: Outputs, of same shape as `X`.`

scripts/nn/util.dml
```diff
@@ -200,3 +200,19 @@ unpad_image = function(matrix[double] img_padded, int Hin, int Win, int padh, in
   }
 }
 
+threshold = function(matrix[double] X, double thresh)
+    return (matrix[double] out) {
+  /*
+   * Computes an indicator matrix with values in (0, 1) depending on whether or not the values in X
```
dusenberrymw: Just for consistency with the rest of the library can you limit the line length of these docs to 72 characters?

scripts/nn/test/test.dml
```diff
@@ -605,3 +605,20 @@ tanh = function() {
...
+    # Generate data
+    matrix = matrix("0.31 0.24 0.87
+
+                    0.45 0.66, 0.65
+
+                    0.24 0.91 0.13", rows=3, cols=3)
+    thresh = 0.5
+
+    # Get the indicator matrix
+    indicator_matrix = util::threshold(matrix, thresh)
```
dusenberrymw: This is good so far, but we need to add a `target` matrix with the expected output for this example, and then compare `target` with `indicator_matrix` using the `util::check_all_equal` function (located in `nn/test/util.dml`) to test that this function is working correctly.  It can be helpful to purposely make it wrong temporarily to ensure that the testing code itself is working.

scripts/nn/util.dml
deroneriksson: @dusenberrymw Would it be possible to add some DML recommendations to http://apache.github.io/systemml/contributing-to-systemml.html#dml-code-format or to create a JIRA to update the "DML Code Format" section listing some recommendations such as 72 character line limits, since SystemML really needs more consistency in our DML? The nn library is great in this regard and it would be nice for the rest of our DML to eventually adopt the same great practices.

dusenberrymw: Yeah, that would be awesome.  I have a bunch of style guidelines that I've been implicitly following, but I haven't written them up.  2-space indentation, 4-space continuation lines, 72-character docstrings, the formatting of inputs & outputs in the docstring, 100-character code line-length, the function & return declaration formatting....

dusenberrymw:
Merged.  Thanks!

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1606/

PR #549, opened 2017-06-20
[SYSTEMML-1724] Remove Guava from compile-time dependencies

dhutchis:
Remove addition of Guava compile-time dependency from SYSTEMML-1663.
Replace with Java 8 and standard collections classes.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1618/

mboehm7:
Thanks @dhutchis for taking care of this. After debugging the failed test, it turned out that the hop comparator had an issue in one of the comparison cases, leading to "random" failures. I just delivered a fix for it but also temporarily disabled this rewrite because there are additional issues, where we would create invalid plans (e.g., with row and column vectors).

LGTM for this PR.

PR #550, opened 2017-06-20
[SYSTEMML-703] Prepare for GPU on Jenkins, disable expensive NN tests

nakul02:
@asurve - To run the GPU tests on the Jenkins machines, use this line:
```
mvn clean verify -PgpuTests
```

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -259,6 +259,18 @@ For examples, see the [Spark MLContext Programming Guide](http://apache.github.i
 Verify that the performance suite located at scripts/perftest/ executes on Spark and Hadoop. Testing should
 include 80MB, 800MB, 8GB, and 80GB data sizes.
 
+# Run NN Unit Tests for GPU
+
+<a href="#release-candidate-checklist">Up to Checklist</a>
+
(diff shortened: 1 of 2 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1620/

dusenberrymw:
Curious, aside from the `testMaxPoolBackward ` test, roughly how long do the other NN tests take?

nakul02:
The `conv2d*` tests did not finish, because of [JIRA 1650 ](https://issues.apache.org/jira/browse/SYSTEMML-1650)

dusenberrymw:
Ah yeah, that's right.  I was also thinking that we could force the [NNTest.java](https://github.com/apache/systemml/blob/master/src/test/java/org/apache/sysml/test/integration/scripts/nn/NNTest.java) tests to run on a GPU if available, since those won't take as long, but they will fail for that same reason.

nakul02:
Ping @asurve - let me know when you want me to merge this.

asurve:
LGTM

nakul02:
thanks @asurve, I shall merge

PR #552, opened 2017-06-21
[SYSTEMML-1728] Reshape util to convert tensors in NCHW to CNHW format

prithvirajsen:
Reshape util to convert tensors in NCHW to CNHW format.

@dusenberrymw can you please review this PR? 

cc @bertholdreinwald @niketanpansare


scripts/utils/tensor_utils.dml
```diff
@@ -0,0 +1,73 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+/*
+ * Reshape util for tensors in NCHW format.
+ * Transposes the 1st and 2nd dimensions.
+ */
+reshape_NCHW_to_CNHW = function(matrix[double] X, int C) return (matrix[double] out){
```
dusenberrymw: Can we rename this to `transpose_NCHW_to_CNHW`?  The reasoning is that NumPy has a [`reshape`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.reshape.h) and [`transpose`](https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.transpose.html) function.  The former just changes the shape of the tensor, without moving any of the elements.  This would be equivalent to our `matrix(X, rows=a, cols=b)` for the 2D case.  The latter function permutes the axes as desired, which would be equivalent to this new function you've added.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1624/

prithvirajsen:
@dusenberrymw 

Thanks for reviewing. I think I implemented all of your suggestions. Please take a look and let me know in case I missed something.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1627/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1628/

PR #553, opened 2017-06-22
[SYSTEMML-1713] correct memory estimates for gpu

nakul02:
Ping @mboehm7, @bertholdreinwald, @niketanpansare, @dusenberrymw

### DIFF
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -21,19 +21,19 @@ package org.apache.sysml.hops;
 
 import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.api.DMLScript.RUNTIME_PLATFORM;
+import org.apache.sysml.hops.Hop.MultiThreadedHop;
 import org.apache.sysml.hops.rewrite.HopRewriteUtils;
 import org.apache.sysml.lops.Aggregate;
 import org.apache.sysml.lops.Binary;
 import org.apache.sysml.lops.DataPartition;
 import org.apache.sysml.lops.Group;
-import org.apache.sysml.hops.Hop.MultiThreadedHop;
 import org.apache.sysml.lops.Lop;
 import org.apache.sysml.lops.LopProperties.ExecType;
 import org.apache.sysml.lops.LopsException;
 import org.apache.sysml.lops.MMCJ;
+import org.apache.sysml.lops.MMCJ.MMCJType;
 import org.apache.sysml.lops.MMRJ;
 import org.apache.sysml.lops.MMTSJ;
-import org.apache.sysml.lops.MMCJ.MMCJType;
 import org.apache.sysml.lops.MMTSJ.MMTSJType;
 import org.apache.sysml.lops.MMZip;
 import org.apache.sysml.lops.MapMult;
@@ -343,11 +343,48 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 	protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz )
 	{
 		double ret = 0;
-		
+
+		if (DMLScript.USE_ACCELERATOR) {
+			// In GPU Mode, intermediate memory is only needed in case of one of the matrix blocks is sparse
+			// When sparse block is converted to dense and a dense MM takes place, we need (dim1 * dim2)
+			// When dense block is converted to sparse and a sparse MM takes place, we need (dim1 * dim2 * 2)
+
+			Hop in1 = _input.get(0);
+			Hop in2 = _input.get(1);
+			double in1Sparsity = OptimizerUtils.getSparsity(in1.getDim1(), in1.getDim2(), in1.getNnz());
+			double in2Sparsity = OptimizerUtils.getSparsity(in2.getDim1(), in2.getDim2(), in2.getNnz());
+
+			boolean in1Sparse = in1Sparsity < MatrixBlock.SPARSITY_TURN_POINT;
+			boolean in2Sparse = in2Sparsity < MatrixBlock.SPARSITY_TURN_POINT;
+
+			boolean in1UltraSparse = in1Sparsity < MatrixBlock.ULTRA_SPARSITY_TURN_POINT;
+			boolean in2UltraSparse = in2Sparsity < MatrixBlock.ULTRA_SPARSITY_TURN_POINT;
+
+			// For Matmult X * Y, if X is sparse, Y is dense, X is converted to dense
+			// If X is ultrasparse, Y is converted to sparse
+			if (in1Sparse ^ in2Sparse) { // one sparse, one dense
+				if (in1Sparse) {
+					if (in1UltraSparse) {
+						ret += 2 * OptimizerUtils.estimateSizeExactSparsity(in2.getDim1(), in2.getDim2(), in2.getNnz());
+					} else {
+						ret += OptimizerUtils.estimateSizeExactSparsity(in1.getDim1(), in1.getDim2(), in1.getNnz());
+					}
+				} else if (in2Sparse) {
+					if (in2UltraSparse) {
+						ret += 2 * OptimizerUtils.estimateSizeExactSparsity(in1.getDim1(), in1.getDim2(), in1.getNnz());
+					} else {
+						ret += OptimizerUtils.estimateSizeExactSparsity(in2.getDim1(), in2.getDim2(), in2.getNnz());
+					}
+				}
+
+			}
+
+		}
+
 		//account for potential final dense-sparse transformation (worst-case sparse representation)
 		if( dim2 >= 2 ) //vectors always dense
-			ret = OptimizerUtils.estimateSizeExactSparsity(dim1, dim2, MatrixBlock.SPARSITY_TURN_POINT);
-		
+			ret += OptimizerUtils.estimateSizeExactSparsity(dim1, dim2, MatrixBlock.SPARSITY_TURN_POINT);
+
 		return ret;
 	}
 	
@@ -544,8 +581,8 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 		int k = OptimizerUtils.getConstrainedNumThreads(_maxNumThreads);
 		
 		ExecType et = ExecType.CP;
-		if(DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < GPUContextPool
-				.initialGPUMemBudget())) {
+		double maxBudgetForGPU = Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget());
+		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < maxBudgetForGPU)) {
 			et = ExecType.GPU;
 		}
 		
@@ -623,9 +660,9 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 		throws HopsException, LopsException
 	{	
 		Lop matmultCP = null;
-		
-		if(DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < GPUContextPool
-				.initialGPUMemBudget())) {
+
+		double maxBudgetForGPU = Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget());
+		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < maxBudgetForGPU)) {
 			Hop h1 = getInput().get(0);
 			Hop h2 = getInput().get(1);
 			Lop left; Lop right;
--- a/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggBinaryOp.java
@@ -581,8 +581,8 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 		int k = OptimizerUtils.getConstrainedNumThreads(_maxNumThreads);
 		
 		ExecType et = ExecType.CP;
-		double maxBudgetForGPU = Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget());
-		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < maxBudgetForGPU)) {
+		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR
+				|| getMemEstimate() < Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget()))) {
 			et = ExecType.GPU;
 		}
 		
@@ -661,8 +661,8 @@ public class AggBinaryOp extends Hop implements MultiThreadedHop
 	{	
 		Lop matmultCP = null;
 
-		double maxBudgetForGPU = Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget());
-		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < maxBudgetForGPU)) {
+		if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR
+				|| getMemEstimate() < Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget()))) {
 			Hop h1 = getInput().get(0);
 			Hop h2 = getInput().get(1);
 			Lop left; Lop right;
--- a/src/main/java/org/apache/sysml/hops/AggUnaryOp.java
+++ b/src/main/java/org/apache/sysml/hops/AggUnaryOp.java
@@ -149,8 +149,8 @@ public class AggUnaryOp extends Hop implements MultiThreadedHop
 				}				
 				else { //general case		
 					int k = OptimizerUtils.getConstrainedNumThreads(_maxNumThreads);
-					double maxBudgetForGPU = Math.min(GPUContextPool.initialGPUMemBudget(), OptimizerUtils.getLocalMemBudget());
-					if (DMLScript.USE_ACCELERATOR && (DMLScript.FORCE_ACCELERATOR || getMemEstimate() < maxBudgetForGPU)) {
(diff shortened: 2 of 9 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1625/

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -344,11 +344,30 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
 	protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz )
 	{
 		double ret = 0;
-		
+
+		if (DMLScript.USE_ACCELERATOR) {
+			// In GPU Mode, these are the intermediate estimates for matmult:
+			// Both dense or both sparse - no intermediate
```
mboehm7: please avoid redundant (1st line) and contradictory (2nd line) comments; we're in a method that only computes the intermediate memory requirements so saying "no intermediate" is confusing.

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -344,11 +344,30 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
...
+			// One sparse, one dense - (dim1 * dim2) OR (dim1 * dim2 + input in sparse format)
+			// In the worst case, an intermediate block other other than a potential temporary sparse-dense block
+			// can be created.
```
mboehm7: This comment is confusing as well - what is the difference between an intermediate and temporary block?

There is also a typo ("other other")

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -344,11 +344,30 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
...
+			Hop in1 = _input.get(0);
+			Hop in2 = _input.get(1);
+			boolean in1Sparse = MatrixBlock.evalSparseFormatInMemory(in1.getDim1(), in1.getDim2(), in1.getNnz());
+			boolean in2Sparse = MatrixBlock.evalSparseFormatInMemory(in2.getDim1(), in2.getDim2(), in2.getNnz());
```
mboehm7: Use the OptimizerUtils estimation functions in order to guard against unknown nnz - theses methods here assume known nnz and compute garbage otherwise. Note that you need to consider the worst-case though, meaning if you don't know the sparsity you need to consider the largest possible sparse matrix (see sparsity turn point).

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -344,11 +344,30 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
...
+
+			if ((!in1Sparse && in2Sparse) || (in2Sparse && !in2Sparse)) {
```
mboehm7: The second condition is wrong. You probably mean in1Sparse && !in2Sparse.

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
```diff
@@ -344,11 +344,30 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
...
+				long size1 = OptimizerUtils.estimateSize(in1.getDim1(), in1.getDim2());
+				long size2 = OptimizerUtils.estimateSize(in2.getDim1(), in2.getDim2());
+				ret += Math.max(size1, size2);
```
mboehm7: Why do you take the maximum of both, if you already know which one to convert?

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -329,8 +329,15 @@ public boolean allowsAllExecTypes()
...
-	{		
-		double sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz);
+	{
+		double sparsity = -1;
+		if (DMLScript.USE_ACCELERATOR) {
+			// The GPU version (for the time being) only does dense outputs
+			sparsity = 1.0;
+		} else {
+			sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz);
+		}
```
mboehm7: First, please use a single-line statements for such code (double sparsity=DMLScript.USE_ACCELERATOR ? 1 : OptimizerUtils.getSparsity(dim1, dim2, nnz)) - it's takes unnecessarily long to read this.

Second, and as a general note - it's fine to use internally dense outputs but these outputs need to be converted to sparse for any subsequent operations. Hence, the output memory estimate cannot be modified here because it is the input memory estimate of data-dependent operations (potentially executed in CP).

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -352,25 +359,41 @@ protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz
...
 			case VAR:
 				//worst-case correction LASTFOURROWS / LASTFOURCOLUMNS
-				if( _direction == Direction.Col ) //(potentially sparse)
+				if (DMLScript.USE_ACCELERATOR) {
+					// The GPU implementation only operates on dense data
+					// It allocated 2 dense blocks to help with these ops:
+					// it calculates the mean,
+					// subtracts the mean (row-wise or column-wise) from elements of the matrix into temp1
+					// it then squares temp1 to temp2
+					// it then divides all elements / each row / each column by the (size-1), (rowSize-1) or (colSize-1)
```
mboehm7: The comment is confusing - you state 2 dense blocks, but you needs means, temp1, and temp2.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -352,25 +359,41 @@ protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz
...
+					val = OptimizerUtils.estimateSize(dim1, dim2);
+					// additionally, in case or row or column wise aggregate unary ops,
+					// a temporary row or col vector is allocated
+					if (_direction == Direction.Col){
+						val += OptimizerUtils.estimateSize(dim1, 1);
```
mboehm7: This estimate seems incorrect - you talk about matrices in the size of the input here but you use (dim1 and dim2) which is the size of the output.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
```diff
@@ -352,25 +359,41 @@ protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz
...
+					} else if (_direction == Direction.Row){
+						val += OptimizerUtils.estimateSize(1, dim2);
+					}
+				} else if( _direction == Direction.Col ) { //(potentially sparse)
 					val = OptimizerUtils.estimateSizeExactSparsity(4, dim2, sparsity);
-				else if( _direction == Direction.Row ) //(always dense)
+				} else if( _direction == Direction.Row ) { //(always dense)
```
mboehm7: Please, refrain from introducing these unnecessary curly braces for very small branches.

src/main/java/org/apache/sysml/hops/TernaryOp.java
```diff
@@ -727,9 +727,15 @@ protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
...
 			case PLUS_MULT:
-			case MINUS_MULT:
-				sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz); 
+			case MINUS_MULT: {
+				if (DMLScript.USE_ACCELERATOR) {
+					// For the GPU, the input is converted to dense
+					sparsity = 1.0;
+				} else {
+					sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz);
+				}
```
mboehm7: Same as above for AggUnaryOp.

src/main/java/org/apache/sysml/hops/UnaryOp.java
```diff
@@ -548,8 +549,13 @@ public void computeMemEstimate(MemoTable memo)
...
 	protected double computeOutputMemEstimate( long dim1, long dim2, long nnz )
-	{		
-		double sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz);
+	{
+		double sparsity = -1;
+		if (DMLScript.USE_ACCELERATOR) {
+			sparsity = 1.0; // Output is always dense (for now) on the GPU
+		} else {
+			sparsity = OptimizerUtils.getSparsity(dim1, dim2, nnz);
+		}
```
mboehm7: same as above.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -1364,16 +1368,22 @@ private static void copyUpperToLowerTriangle(GPUContext gCtx, String instName, M
...
-	 * @param left                  Matrix A
-	 * @param right                 Matrix B
-	 * @param outputName            Name of the output matrix C (in code generated after LOP layer)
-	 * @param isLeftTransposed      op for A, transposed or not
-	 * @param isRightTransposed     op for B, tranposed or not
-	 * @return	output of matrix multiply
+	 * <p>
+	 * Memory Requirements -
+	 * Both dense - inputs, output, no intermediate
+	 * Both sparse - inputs, output, no intermediate
```
mboehm7: Are you sure that for all branches there are no intermediates? For example in our matrix multiplication library, we often use temporary memory for partial aggregates (w/o synchronization) in the number of threads.

src/main/java/org/apache/sysml/hops/UnaryOp.java
```diff
@@ -563,6 +569,10 @@ protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz
 			// getMemEstimate works for both cases of known dims and worst-case stats
 			ret = getInput().get(0).getMemEstimate() * 3; 
 		}
+
+		if (DMLScript.USE_ACCELERATOR) {
+			OptimizerUtils.estimateSize(dim1, dim2); // Intermediate memory required to convert sparse to dense
+		}
 		
 		return ret;
 	}
```
mboehm7: Is the ConvolutionOp missing? If cuDNN is called there internally it also needs to be reflected in the memory estimates.

src/main/java/org/apache/sysml/hops/BinaryOp.java
```diff
@@ -818,10 +819,24 @@ protected double computeIntermediateMemEstimate( long dim1, long dim2, long nnz
...
+			if (DMLScript.USE_ACCELERATOR) {
+				// Solve on the GPU takes an awful lot of intermediate space
+				// First the inputs are converted from row-major to column major
+				// Then a workspace and a temporary output (workSize, tauSize) are needed
+				long m = getInput().get(0).getDim1();
+				long n = getInput().get(0).getDim2();
+				long tauSize = OptimizerUtils.estimateSize(m, 1);
+				long workSize = OptimizerUtils.estimateSize(m, n);
+				long AtmpSize = OptimizerUtils.estimateSize(m, n);
+				long BtmpSize = OptimizerUtils.estimateSize(n, 1);
```
mboehm7: Hm, are all these intermediates required in our JVM process? If not, we need to think about separating CP and GPU memory requirements - this is probably useful anyway to make the estimates tighter but it might be more involved as it would change the conditions for selecting GPU instructions.

src/main/java/org/apache/sysml/hops/AggBinaryOp.java
nakul02: I do, will fix this.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
nakul02: I think I didn't make it clear in the comment, the output block is reused to put the (row/column wise) means into. 
I shall fix that in the comment.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
nakul02: Good catch, I realized this after the fact and went back to fix it in all the places, but missed this one.
Thanks. I will fix this.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
nakul02: I will have to disagree on this point. There does seem to be some sort of rules of thumb available online to always have curly braces, even when you only have a single statement.

src/main/java/org/apache/sysml/hops/BinaryOp.java
nakul02: These estimates are only for the GPU and will be calculated only when `USE_ACCELERATOR` is true.

src/main/java/org/apache/sysml/hops/TernaryOp.java
nakul02: Same as above for AggUnaryOp.

src/main/java/org/apache/sysml/hops/UnaryOp.java
nakul02: same as above.

src/main/java/org/apache/sysml/hops/UnaryOp.java
nakul02: I am not sure what you mean when you say "if cuDNN is called there internally".
`ConvolutionOp.java` is where cuDNN estimates are calculated.

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
nakul02: I haven't been able to find anything else.

nakul02:
"we should think about separating JVM and GPU memory requirements and budgets." - I very much agree. We should also put a plan in place to decouple other backends. It makes the adding and removing/disabling of backends straight-forward.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1633/

src/main/java/org/apache/sysml/hops/UnaryOp.java
mboehm7: Well, I just find it suspicious that the ConvolutionOp has intermediate memory requirements of zero although you just assigned memory intermediates to matrix multiplications. This is probably something that needs fixing in general (also for CP).

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
mboehm7: I would recommend to use the same output memory estimate (in order to ensure that the compilation of subsequent operations is not affected) and account for the dense block only in the intermediate requirements for GPUs.

mboehm7:
sure, but let's be more specific here: these memory estimates reflect singlenode operations; any distributed backends anyway have very different semantics and memory estimates and budgets. The interesting thing about GPUs is that the memory estimates will be very similar (once we have full support for dense/sparse) but the budget is different.

src/main/java/org/apache/sysml/hops/UnaryOp.java
nakul02: For matrix multiplication, intermediate memory is needed when one of the matrices is sparse and the other one dense. cuDNN is used only for dense matrices. If the input matrix is sparse, it is converted to dense.

nakul02:
That is correct, the memory budgets for GPU are different now because of the incomplete support for dense/sparse matrices.

src/main/java/org/apache/sysml/hops/AggUnaryOp.java
nakul02: Could you please explain what you mean by "compilation of subsequent operations is not affected"
If I revert my changes, SystemML will not correctly account for the fact that the output on the GPU will be dense.

mboehm7: I was referring to a scenario where the subsequent (data-dependent) operation cannot be compiled to GPU - in this case this wrong output estimate would negatively affect the subsequent operation (e.g., SPARK instead of CP). You have two options here: (1) either check the parent (for applicable GPU operations), or (2) leave it as proposed as we'll anyway clean this up once there is proper support for dense and sparse as well as separate memory budgets and estimates. I leave it up to you.

src/main/java/org/apache/sysml/hops/UnaryOp.java
mboehm7: This is ok for now but I'd like to get a reference for this. I could construct scenarios with small-ish output where partial aggregations are usually advantageous but this requires additional memory even for dense inputs. Anyway, if we do not have any details about these internals there is probably nothing we can do about it.

mboehm7:
that is fine - let me know once you think this PR is ready and I'll make another full pass. The separation in CP and GPU estimates and budget can be done in a separate PR.

nakul02:
Ping @mboehm7 - I am going to leave the proposed change as is for the case of when "subsequent (data-dependent) operation cannot be compiled to GPU".
Additionally, the situation when driver mem budget is less than GPU mem budget is taken care of.
Could you please review this PR (I think its ready).

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1695/

nakul02:
Ping @mboehm7

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1713/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1715/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1717/

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/CSRPointer.java
```diff
@@ -54,7 +54,7 @@
 
 	private static final Log LOG = LogFactory.getLog(CSRPointer.class.getName());
 
-	private static final double ULTRA_SPARSITY_TURN_POINT = 0.0004;
+	private static final double ULTRA_SPARSITY_TURN_POINT = 0.00004;
```
mboehm7: as a nice to have, I would prefer to simply reference MatrixBlock.ULTRA_SPARSITY_TURN_POINT for consistency.

nakul02:
Thanks @mboehm7 - I'll fix the merge conflicts and failing tests and merge.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1752/

nakul02:
Tests passed, I shall rebase & merge.

PR #559, opened 2017-06-30
[SYSTEMML-1744] JCuda jars in extra assembly jar

nakul02:
Ping @niketanpansare - can you please test, if possible, whether this works for a pip install?
Ping @deroneriksson, @gweidner  - can you please review this?

### DIFF
--- a/src/assembly/extra/LICENSE
+++ b/src/assembly/extra/LICENSE
@@ -467,22 +467,22 @@ The following compile-scope dependencies come under the MIT License
 
 JCuda (jcuda.org)
 
-org.jcuda.jcuda:0.8.0
-org.jcuda.jcublas:0.8.0
-org.jcuda.jcufft:0.8.0
-org.jcuda.jcusparse:0.8.0
-org.jcuda.jcusolver:0.8.0
-org.jcuda.jcurand:0.8.0
-org.jcuda.jnvgraph:0.8.0
-org.jcuda.jcudnn:0.8.0
-org.jcuda.jcuda-natives:0.8.0
-org.jcuda.jcublas-natives:0.8.0
-org.jcuda.jcufft-natives:0.8.0
-org.jcuda.jcusparse-natives:0.8.0
-org.jcuda.jcusolver-natives:0.8.0
-org.jcuda.jcurand-natives:0.8.0
-org.jcuda.jnvgraph-natives:0.8.0
-org.jcuda.jcudnn-natives:0.8.0
+org.jcuda:jcuda:0.8.0
+org.jcuda:jcublas:0.8.0
+org.jcuda:jcufft:0.8.0
+org.jcuda:jcusparse:0.8.0
+org.jcuda:jcusolver:0.8.0
+org.jcuda:jcurand:0.8.0
+org.jcuda:jnvgraph:0.8.0
+org.jcuda:jcudnn:0.8.0
+org.jcuda:jcuda-natives:0.8.0
+org.jcuda:jcublas-natives:0.8.0
+org.jcuda:jcufft-natives:0.8.0
+org.jcuda:jcusparse-natives:0.8.0
+org.jcuda:jcusolver-natives:0.8.0
+org.jcuda:jcurand-natives:0.8.0
+org.jcuda:jnvgraph-natives:0.8.0
+org.jcuda:jcudnn-natives:0.8.0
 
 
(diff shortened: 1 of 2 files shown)


nakul02:
Something to be aware of - This includes the platform specific cuda jars for which SystemML was built on. 

In an offline discussion with @niketanpansare, he said it was ok to have Linux jcuda native jars, because that would probably be the more widely used platform. For other platforms, we could provide instructions for their setup.

Another option would be to just include ALL or.jcuda jars for all platforms. This would blow up the `extra` jar size. Also, I'd need to test if that would work.

niketanpansare:
@nakul02 The jcuda jars are included likely as resource rather than `compile` by this PR.  Also, I don't see jcuda classes. Here is the content of `extra` jar:
```
lib/
lib/jcuda-0.8.0.jar
org/apache/sysml/api/dl/BaseDMLGenerator$$anonfun$invoke$2.class
org/apache/sysml/api/dl/Caffe2DML$$anonfun$org$apache$sysml$api$dl$Caffe2DML$$backward$1.class
org/apache/sysml/api/dl/Caffe2DMLModel$$anonfun$getPredictionScript$8.class
org/apache/sysml/api/dl/CaffeNetwork$.class
org/apache/sysml/api/dl/DMLGenerator$$anonfun$initWeights$1.class
org/apache/sysml/api/dl/HasWeight$class.class
org/apache/sysml/api/dl/IsLossLayer.class
org/apache/sysml/api/dl/LearningRatePolicy.class
org/apache/sysml/api/dl/Nesterov$.class
org/apache/sysml/api/dl/Nesterov.class
...
```

Hence,  the following python code throws an error `java.lang.ClassNotFoundException: jcuda.runtime.JCuda`:
```python
import numpy as np
from sklearn import datasets
from systemml.mllearn import LinearRegression
# Load the diabetes dataset
diabetes = datasets.load_diabetes()
# Use only one feature
diabetes_X = diabetes.data[:, np.newaxis, 2]
# Split the data into training/testing sets
X_train = diabetes_X[:-20]
X_test = diabetes_X[-20:]
# Split the targets into training/testing sets
y_train = diabetes.target[:-20]
y_test = diabetes.target[-20:]
# Create linear regression object
regr = LinearRegression(spark, fit_intercept=True, C=float("inf"), solver='direct-solve').setGPU(True).setForceGPU(True).setStatistics(True)
# Train the model using the training sets
regr.fit(X_train, y_train)
```

src/assembly/extra/LICENSE
```diff
@@ -460,3 +460,51 @@ Copyright 2017 The TensorFlow Authors.  All rights reserved.
...
    See the License for the specific language governing permissions and
    limitations under the License.
+
+===============================================================================
+
+The following compile-scope dependencies come under the MIT License
+
+JCuda (jcuda.org)
+
+org.jcuda.jcuda:0.8.0
```
deroneriksson: Please update these to reflect maven groupId:artifactId:version, so `org.jcuda.jcuda:0.8.0` is `org.jcuda:jcuda:0.8.0`, etc.

deroneriksson:
I see the following after `mvn clean package -P distribution`: 
image

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1700/

nakul02:
@niketanpansare - can you please check now?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1705/

niketanpansare:
LGTM. Thanks @nakul02 :+1:  ... I would recommend to wait until you get LGTM from @deroneriksson or @gweidner

nakul02:
Thanks @niketanpansare 
Will do.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1714/

deroneriksson:
LGTM. Thank you @nakul02.

For reference, on GitHub, the 0.8.0 tag does not include the MIT LICENSE file that exists on the master branch. However, based on the discussion at https://github.com/jcuda/jcuda-main/issues/18, 0.8.0 is indeed MIT licensed and the license file is at http://jcuda.org/License.txt.

nakul02:
Thanks @deroneriksson, I shall merge.

PR #566, opened 2017-07-12
[Minor] Minor bug fix in org.apache.sysml.hops.ipa.FunctionCallSizeInfo#toString()

feihugis:
Integer can not be converted to Long. This bug can be reproduced by setting the LogLevel to be `DEBUG` and running the MNIST example.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1732/

mboehm7:
thanks for catching this @feihugis - LGTM.

PR #567, opened 2017-07-12
[SYSTEMML-1663] Fix rewrite rule: element-wise multiply chains

dhutchis:
This PR adjusts the element-wise multiply chain rewrite rule. 
It now groups inputs into (1) unknowns/frames/objects, (2) non-vector matrices, (3) row vectors, and (4) column vectors and scalars. Each group is multiplied among all its inputs, in order of increasing nnz. Then, we multiply across groups as `(((other)*matrices)*row_vectors)*col_vectors_and_scalars`. The resulting order looks like the following:

<pre>
    (((unknown * object * frame) * ([least-nnz-matrix * matrix] * most-nnz-matrix))
     * ([least-nnz-row-vector * row-vector] * most-nnz-row-vector))
    * ([[scalars * least-nnz-col-vector] * col-vector] * most-nnz-col-vector)
</pre>

Adds a test that tests all these different kinds of inputs.

Moves the rewrite to dynamic rewrites. Does not rewrite if the top multiply's dims are unknown.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/rewrite/ProgramRewriter.java
+++ b/src/main/java/org/apache/sysml/hops/rewrite/ProgramRewriter.java
@@ -54,7 +54,7 @@ public class ProgramRewriter
 	private static final Log LOG = LogFactory.getLog(ProgramRewriter.class.getName());
 	
 	//internal local debug level
-	private static final boolean LDEBUG = false; 
+	private static final boolean LDEBUG = false;
 	private static final boolean CHECK = false;
 	
 	private ArrayList<HopRewriteRule> _dagRuleSet = null;
@@ -96,8 +96,6 @@ public class ProgramRewriter
 			_dagRuleSet.add(     new RewriteRemoveUnnecessaryCasts()             );		
 			if( OptimizerUtils.ALLOW_COMMON_SUBEXPRESSION_ELIMINATION )
 				_dagRuleSet.add( new RewriteCommonSubexpressionElimination()     );
-			//if ( OptimizerUtils.ALLOW_SUM_PRODUCT_REWRITES)
-			//	_dagRuleSet.add( new RewriteElementwiseMultChainOptimization()   ); //dependency: cse
 			if( OptimizerUtils.ALLOW_CONSTANT_FOLDING )
 				_dagRuleSet.add( new RewriteConstantFolding()                    ); //dependency: cse
 			if( OptimizerUtils.ALLOW_ALGEBRAIC_SIMPLIFICATION )
@@ -125,7 +123,9 @@ public class ProgramRewriter
 		// DYNAMIC REWRITES (which do require size information)
 		if( dynamicRewrites )
 		{
-			_dagRuleSet.add(     new RewriteMatrixMultChainOptimization()         ); //dependency: cse 
+			_dagRuleSet.add(     new RewriteMatrixMultChainOptimization()         ); //dependency: cse
+			if ( OptimizerUtils.ALLOW_SUM_PRODUCT_REWRITES)
+				_dagRuleSet.add( new RewriteElementwiseMultChainOptimization()    ); //dependency: cse
 			
 			if( OptimizerUtils.ALLOW_ALGEBRAIC_SIMPLIFICATION )
 			{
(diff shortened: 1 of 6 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1734/

dhutchis:
The failing test is the following: `R = W*(2*(U%*%t(V)));`.
The opcode `wumm` changed to `wdivmm` with an additional argument of -1, and with the "multiply by literal 2" appearing above `wdivmm`. Anyone knows what this means?

j143:
Seems this is the verdict.
```
23:59:53 Tests run: 26, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 539.071 sec <<< FAILURE! - in org.apache.sysml.test.integration.functions.quaternary.WeightedUnaryMatrixMultTest
02:24:21   WeightedUnaryMatrixMultTest.testWeightedUnaryMMMult2SparseCP:131->runWeightedUnaryMMTest:297 Missing opcode wumm
02:24:21   WeightedUnaryMatrixMultTest.testWeightedUnaryMMMult2SparseSP:173->runWeightedUnaryMMTest:301 Missing opcode sp_wumm
02:24:21   WeightedUnaryMatrixMultTest.testWeightedUnaryMMMult2DenseCP:111->runWeightedUnaryMMTest:297 Missing opcode wumm
02:24:21   WeightedUnaryMatrixMultTest.testWeightedUnaryMMMult2DenseSP:153->runWeightedUnaryMMTest:301 Missing opcode sp_wumm

dhutchis:
@j143 indeed, see the previous comment. Do you know what `wumm` and `wdivmm` are?  I believe that this happened because the `2*` scalar multiply moved around. I'd like to understand what the conditions for `wumm` are and why this is better than `wdivmm` so that I can try to work with it in the ordering of inputs to an element-wise multiply.

mboehm7:
here is an overview https://github.com/apache/systemml/blob/master/docs/devdocs/MatrixMultiplicationOperators.txt; sorry for the delay - I'll make a pass over this PR tonight.

deroneriksson:
Also, see the online version at http://apache.github.io/systemml/engine-dev-guide.html#complex-matrix-multiplication-operators and http://apache.github.io/systemml/engine-dev-guide.html#core-matrix-multiplication-primitives

dhutchis:
Okay, looks like the new element-wise rewrite order no longer meets pattern 2: `(W*sop(U%*%t(V),c))` where `sop` is `*` and `c` is `2`.
Instead it meets pattern 7: `W*(U%*%t(V))`.

This happens because we moved the `*2` to the outside, as in `( W*(U%*%t(V)) ) * 2`.
I guess this new pattern is a bit less efficient because the *2 is an extra operation after the wdivmm.
But if I were to try special casing it, it would be weird to stick the scalar in the middle of the two matrix multiplies in general. So I would have to look for exactly this pattern.
Or I could try modifying wumm to recognize if there is a "multiply by constant" outside the `W*(U%*%t(V))` which we could absorb into it.

j143:
@dhutchis modifying `wumm` to recognize if there is a "multiply by constant" outside the `W*(U%*%t(V))`, seems a best option.

dhutchis:
Got it!

mboehm7:
thanks @dhutchis - overall this looks already very good. 

Just one minor knit pick: could you please add a condition of nl.getParent().size()==1 to the new wumm rewrite to ensure we only merge this in if there are not multiple consumers, because the *2 is non-essential for the fused operator and we would redundantly compile a wdivmm in this case as well.

Regarding the modification of wumm, I would have actually recommended to disable the respective tests because even with wdivmm and *2 it would be sparsity-exploiting and these fused operators are going to be superseded by codegen anyway soon. However, since you've already done the work, we'll of course go with the modified rewrite. Thanks.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1749/

dhutchis:
Squashed into one commit and rebased onto master.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1759/

mboehm7:
thanks @dhutchis - this LGTM.

PR #569, opened 2017-07-12
[SYSTEMML-1763] Fix Explain countCompiledInstructions for CP

deroneriksson:
Increment the ExplainCounts numCPInst value in countCompiledInstructions
based on the value of the CP parameter.
Change countCompiledInstructions method return type to void.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1737/

PR #571, opened 2017-07-12
[SYSTEMML-1764] Fix cbind value in AppendGAlignedSP constructor

deroneriksson:
Set the _cbind field to the value of the cbind parameter in the AppendGAlignedSP
constructor.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1740/

PR #273, opened 2016-10-22
[WIP] Implementation of SVD

nakul02:
@niketanpansare, @mboehm7, @bertholdreinwald, @iyounus
This initial implementation is keeping in line with the discussion on the dev mailing list.
The CP 'svd' function is implemented and backed by Apache Commons Math.

PR #582, opened 2017-07-19
[SYSTEMML-1216][WIP] implementing singular value decomposition function - svd( )

j143:
svd

reference: wikipedia.

### DIFF
--- a/src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
+++ b/src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
@@ -181,22 +181,28 @@ public class BuiltinFunctionExpression extends DataIdentifier
 			// setup output properties
 			DataIdentifier svdOut1 = (DataIdentifier) getOutputs()[0];
 			DataIdentifier svdOut2 = (DataIdentifier) getOutputs()[1];
+			DataIdentifier svdOut3 = (DataIdentifier) getOutputs()[2];
 			
-			if ( inrows != incols ) {
-			       raiseValidateError("LU Decomposition can only be done on a square matrix. Input matrix is rectangular (rows=" + inrows + ", cols=" + incols + ")", conditional);
-			}
+			long inrows = getFirstExpr().getOutput().getDim1();
+			long incols = getFirstExpr().getOutput().getDim2();
 			
-			//Output1 - P
+			//Output1 - U (unitary matrix) 
 			svdOut1.setDataType(DataType.MATRIX);
 			svdOut1.setValueType(ValueType.DOUBLE);
 			svdOut1.setDimensions(inrows, inrows);
 			svdOut1.setBlockDimensions(getFirstExpr().getOutput().getRowsInBlock(), getFirstExpr().getOutput().getColumnsInBlock());
 			
-			//Output2 - Q
+			//Output2 - W (the rectangular diagonal matrix)
 			svdOut2.setDataType(DataType.MATRIX);
 			svdOut2.setValueType(ValueType.DOUBLE);
-			svdOut2.setDimensions(inrows, inrows);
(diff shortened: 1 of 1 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1778/

j143:
Hi! 
I have changed the file in the `parser/BuiltinFunctionExpression.java`, now where can I implement the problem logic.  Which files needs to changed, give me some help please.

j143:
cc @mboehm7 .

mboehm7:
Before implementing svd as a builtin function, I would recommend to investigate if it could be implemented efficiently in dml or not. If yes, then I would prefer having a dml bodied function because then all other compilation and runtime techniques automatically apply. If it turns out that the dml version has performance or scalability issues, we can discuss to add a builtin function instead. In this case, we would need to break this endeavor into two steps of local and distributed operations. If you need this svd builtin immediately, you can workaround via a java-based external function that calls some local implementation.  

Just as a background: Yes, there are builtin functions like solve, inverse, qr, lu, and eigen which only have local operations (by calling out to commons math). These functions were added for convenience but are typically only applied on small data. However, the missing distributed operations are unsatisfying because they do not adhere to the overall philosophy of SystemML. Moving forward, we would like to either add distributed operators for these operations or replace them by dml scripts which would automatically be compiled to distributed operations.

j143:
@mboehm7  thanks for looking into this. Like `qr`, `lu`, and `eigen` we try to implement a local `svd( )` function which is helpful for the development of scalable `svd( )` in dml as we have done for `qr`, `lu` 
 in this directory https://github.com/j143/systemml/tree/master/scripts/staging/scalable_linalg .

mboehm7:
I understand - are you saying you cannot implement a svd in dml efficiently?

j143:
I think yes. But, I am following the instructions in this [jira](https://issues.apache.org/jira/browse/SYSTEMML-1214). 
If you feel that we should implement scalable `svd( )` in dml w/o a local built-in, we can try that.

mboehm7:
thanks @j143 - yes, as mentioned above, I think we should first give a basic dml implementation as try and only resort to a builtin function if the dml implementation has fundamental issues in terms of performance or scalability.

nakul02:
Something else to keep in mind - there are GPU implementations for SVD, QR, Solve, etc available through cuSolver. These will be more efficient if svd were a buitkin function rather than a DML bodied function.  

@j143 - you can take a look at PR #273 also, it has an earlier attempt at this. 

Having a "local" svd is good. The distributed version can be built as a Builtin function or as a DML bodies function. The advantage of the DML bodied distributed version is that it can use the GPU implementation as it's building blocks.

j143:
Hi @mboehm7 !
1. May be we can consider one more thing here, Imran Younus has tried to implement the distributed svd() with the help of local svd( ) already in PR #273 , (which I believe is robust w/ acceptable accuracy, isn't it?) of the latest algorithm in the literature. 

2. It may be worthy of taking into account that too(?). Please give me, 20-25 hrs on studying the current literature that the distributed algorithm is using at its backbone, to understand whether the algorithm have any conflicts with distribution calculations at the matrix block level.

3. Will it helps us understand when implementing the other functions, internal or external implementation that perform well in our setting?. As @nakul02 pointed out, we try the above implementations for POV of GPU too. 

Please provide any thoughts.

nakul02:
@j143 - are you planning on following up on the implementation of `svd()` in a new PR?


As per your request on PR #273, I've rebased my svd branch against master - https://github.com/nakul02/systemml/tree/svd


Please note that as more commits are added to master, it may soon become stale. If you plan to use this to continue working on `svd()`, please clone my svd branch to your fork/clone of the systemml repo and rebase that.

PR #583, opened 2017-07-19
[MINOR] Refactoring - boolean constructors, returns, semicolons

deroneriksson:
1) Boolean constructor creates new Boolean instances, so follow best
practice of using Boolean.TRUE, Boolean.FALSE, and Boolean.valueOf()
rather than the Boolean constructor. See Boolean constructor javadocs
for more information.
2) Remove unnecessary return statements
3) Remove extraneous semicolons


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1779/

PR #584, opened 2017-07-19
[SYSTEMML-1774] Bugfix for illegal arguments in ConvolutionUtils

feihugis:
When using `Arrays.fill()` in `ConvolutionUtils.#binaryOperationInPlace` to fill the values of the 1-dimension destination array by `fromIndex` and `toIndex`, the column index `cix` of the sparse matrix block need to be added into `fromIndex` and `endIndex` first, but it is missed at `line 95` for `toIndex`.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1780/

j143:
It works, @feihugis !

deroneriksson:
Thank you @feihugis for the fix and @j143 for testing.
LGTM.

PR #588, opened 2017-07-21
[SYSTEMML-1703] Fix input data processing for Caffe VGG-19 model

asurve:
Fixed input data format to Caffe Model. This has been verified with VGG-19 models for various color and black/white image. 
For other models updated function can be used with appropriate parameters.

@niketanpansare Can you please review the change?

### DIFF
--- a/src/main/python/systemml/converters.py
+++ b/src/main/python/systemml/converters.py
@@ -237,7 +237,7 @@ def getModelMean(model_name):
 
 # Example usage: convertImageToNumPyArr(im, img_shape=(3, 224, 224), add_rotated_images=True, add_mirrored_images=True)
 # The above call returns a numpy array of shape (6, 50176) in NCHW format
-def convertImageToNumPyArr(image_file_name, img_shape=None, add_rotated_images=False, add_mirrored_images=False,
+def convertImageToNumPyArr(im, img_shape=None, add_rotated_images=False, add_mirrored_images=False,
     color_mode = 'RGB', mean=None):
 
     ## Input Parameters
@@ -257,7 +257,6 @@ def convertImageToNumPyArr(image_file_name, img_shape=None, add_rotated_images=F
         raise ValueError('Expected the number of channels to be either 1 or 3')
 
     from PIL import Image
-    im = Image.open(image_file_name)
 
     if size is not None:
         im = im.resize(size, Image.LANCZOS)
--- a/src/main/python/systemml/converters.py
+++ b/src/main/python/systemml/converters.py
@@ -25,6 +25,9 @@ import numpy as np
 import pandas as pd
 import os
 import math
+import skimage
+from skimage.transform import resize
+
 from pyspark.context import SparkContext
 from scipy.sparse import coo_matrix, spmatrix, csr_matrix
 from .classloader import *
@@ -39,7 +42,7 @@ def getNumCols(numPyArr):
 
 def get_pretty_str(key, value):
     return '\t"' + key + '": ' + str(value) + ',\n'
-        
+
 def save_tensor_csv(tensor, file_path, shouldTranspose):
     w = w.reshape(w.shape[0], -1)
     if shouldTranspose:
@@ -51,29 +54,29 @@ def save_tensor_csv(tensor, file_path, shouldTranspose):
         file.write(get_pretty_str('cols', w.shape[1]))
         file.write(get_pretty_str('nnz', np.count_nonzero(w)))
         file.write('\t"format": "csv",\n\t"description": {\n\t\t"author": "SystemML"\n\t}\n}\n')
-    
+
 def convert_caffemodel(sc, deploy_file, caffemodel_file, output_dir, format="binary", is_caffe_installed=False):
     """
-    Saves the weights and bias in the caffemodel file to output_dir in the specified format. 
+    Saves the weights and bias in the caffemodel file to output_dir in the specified format.
     This method does not requires caffe to be installed.
-    
+
     Parameters
     ----------
     sc: SparkContext
         SparkContext
-    
+
     deploy_file: string
         Path to the input network file
-        
+
     caffemodel_file: string
         Path to the input caffemodel file
-    
+
     output_dir: string
         Path to the output directory
-    
+
     format: string
         Format of the weights and bias (can be binary, csv or text)
-    
+
     is_caffe_installed: bool
         True if caffe is installed
     """
@@ -104,17 +107,17 @@ def convert_caffemodel(sc, deploy_file, caffemodel_file, output_dir, format="bin
         utilObj = sc._jvm.org.apache.sysml.api.dl.Utils()
         utilObj.saveCaffeModelFile(sc._jsc, deploy_file, caffemodel_file, output_dir, format)
 
-    
+
 def convert_lmdb_to_jpeg(lmdb_img_file, output_dir):
     """
     Saves the images in the lmdb file as jpeg in the output_dir. This method requires caffe to be installed along with lmdb and cv2 package.
     To install cv2 package, do `pip install opencv-python`.
-    
+
     Parameters
     ----------
     lmdb_img_file: string
         Path to the input lmdb file
-    
+
     output_dir: string
         Output directory for images (local filesystem)
     """
@@ -163,7 +166,7 @@ def _convertSPMatrixToMB(sc, src):
     buf3 = bytearray(col.tostring())
     createJavaObject(sc, 'dummy')
     return sc._jvm.org.apache.sysml.runtime.instructions.spark.utils.RDDConverterUtilsExt.convertSciPyCOOToMB(buf1, buf2, buf3, numRows, numCols, nnz)
-            
+
 def _convertDenseMatrixToMB(sc, src):
     numCols = getNumCols(src)
     numRows = src.shape[0]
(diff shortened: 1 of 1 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1790/

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,86 @@ def convertToNumPyArr(sc, mb):
...
+        flag for color format. True (default) loads as RGB while False
+        loads as intensity (if image is already grayscale).
+
+    Returns
+    -------
+    image : an image with type np.float32 in range [0, 1]
+        of size (H x W x 3) in RGB or
+        of size (H x W x 1) in grayscale.
+    """
+    import cv2
```
dusenberrymw: I would heavily recommend that we don't introduce a dependency on OpenCV.  Let's instead continue to use the PIL package (i.e. Pillow).  Pillow is much lighter weight.  `np.asarray(Image.open(filename)).astype(np.float64)` will return a (H,W,C) RGB array with double values in [0,255].

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,86 @@ def convertToNumPyArr(sc, mb):
...
+    image = skimage.img_as_float(cv2.imread(filename))
+    # cv2 reads in BGR format, so converting back to RGB format.
```
dusenberrymw: A nice side-effect is that Pillow (i.e. `PIL` package) reads in RGB format.

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,86 @@ def convertToNumPyArr(sc, mb):
...
+        resized_im = zoom(image, scale + (1,), order=interp_order)
+    return resized_im.astype(np.float32)
+
+
 # Example usage: convertImageToNumPyArr(im, img_shape=(3, 224, 224), add_rotated_images=True, add_mirrored_images=True)
 # The above call returns a numpy array of shape (6, 50176) in NCHW format
-def convertImageToNumPyArr(im, img_shape=None, add_rotated_images=False, add_mirrored_images=False):
-    from PIL import Image
+def convertImageToNumPyArr(image_file_name, img_shape=None, add_rotated_images=False, add_mirrored_images=False,
+    color=True, transpose=(2,0,1), color_mode = 'BGR', raw_scale=255, mean=[103.939, 116.779, 123.68]):
```
niketanpansare: @asurve I would not recommend using the VGG model as baseline for setting the default values. (for example: mean, color_mode).

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,86 @@ def convertToNumPyArr(sc, mb):
...
+        loads as intensity (if image is already grayscale).
+
+    Returns
+    -------
+    image : an image with type np.float32 in range [0, 1]
+        of size (H x W x 3) in RGB or
+        of size (H x W x 1) in grayscale.
+    """
+    import cv2
+    image = skimage.img_as_float(cv2.imread(filename))
```
niketanpansare: I understand the intent here is to be consistent with BVLC Caffe, but skimage is an additional dependency. Do you think it is necessary ?

src/main/python/systemml/converters.py
dusenberrymw: I don't think it's a good idea to include VGG-specific preprocessing into this function.  I would keep the `convertImageToNumPyArr` as a super lightweight function that just converts a `(H,W,C)` array into a `(1,C*H*W)` array, and have a separate preprocessing function that is a bit more general.  Especially the means -- those are dependent on a specific dataset for use by a specific model.

src/main/python/systemml/converters.py
asurve: I replaced skimage. Thanks

src/main/python/systemml/converters.py
asurve: done.

src/main/python/systemml/converters.py
asurve: yes, that was for short term until I address other models, but anyway changed it.

asurve: VGG model specific processing include color_mode ('BGR") and mean value.
Most of the other changes are needed for other models as well, I believe.
Once I see other models, I can revisit if we need to differentiate.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1798/

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,38 @@ def convertToNumPyArr(sc, mb):
...
+    except:
+        mean = None
+    return mean
+
+
 # Example usage: convertImageToNumPyArr(im, img_shape=(3, 224, 224), add_rotated_images=True, add_mirrored_images=True)
 # The above call returns a numpy array of shape (6, 50176) in NCHW format
-def convertImageToNumPyArr(im, img_shape=None, add_rotated_images=False, add_mirrored_images=False):
-    from PIL import Image
+def convertImageToNumPyArr(image_file_name, img_shape=None, add_rotated_images=False, add_mirrored_images=False,
```
niketanpansare: @asurve  I would not recommend changing image object to image file.  Consider this example: http://apache.github.io/systemml/beginners-guide-caffe2dml#how-to-prepare-a-directory-of-jpeg-images-for-training-with-caffe2dml-

src/main/python/systemml/converters.py
```diff
@@ -19,18 +19,21 @@
...
 import os
 import math
+
 from pyspark.context import SparkContext
 from scipy.sparse import coo_matrix, spmatrix, csr_matrix
 from .classloader import *
 
 SUPPORTED_TYPES = (np.ndarray, pd.DataFrame, spmatrix)
 
+MODEL_MEAN = {'VGG-19':[103.939, 116.779, 123.68]}
```
niketanpansare: should this be model mean or dataset mean ?

src/main/python/systemml/converters.py
asurve: ok, done

src/main/python/systemml/converters.py
asurve: Mean is calculated on training data for VGG-19 model trained on. While using VGG-19 model, this mean to be subtracted from input data.

src/main/python/systemml/converters.py
```diff
@@ -212,10 +215,38 @@ def convertToNumPyArr(sc, mb):
...
-    from PIL import Image
+def convertImageToNumPyArr(im, img_shape=None, add_rotated_images=False, add_mirrored_images=False,
+    color_mode = 'RGB', mean=None):
+
+    ## Input Parameters
+
+    # color_mode: In case of VGG models which expect image data in BGR format instead of RGB for other most models,
+    # color_mode parameter is used to process image data in BGR format.
+
+    # mean: mean value is used to subtract from input data from every pixel value. Default value specified is for VGG-19 model.
```
niketanpansare: can you please update the documentation to specify that by default we donot subtract mean ?

src/main/python/systemml/converters.py
niketanpansare: OK, that makes sense. Then, I would recommend calling it `getDatasetMean` and `DATASET_MEAN = {'ImageNet...SoAndSoYear':[103.939, 116.779, 123.68]}`

src/main/python/systemml/converters.py
```diff
@@ -224,24 +255,45 @@ def convertImageToNumPyArr(im, img_shape=None, add_rotated_images=False, add_mir
...
     if expected_mode is not im.mode:
         im = im.convert(expected_mode)
+
     def _im2NumPy(im):
         if expected_mode == 'L':
             return np.asarray(im.getdata()).reshape((1, -1))
         else:
-            # (H,W,C) --> (C,H,W) --> (1, C*H*W)
-            return np.asarray(im).transpose(2, 0, 1).reshape((1, -1))
+            im = (np.array(im).astype(np.float))
```
niketanpansare: is `astype(np.float)` required ?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1808/

src/main/python/systemml/converters.py
asurve: Means are defined in float values. Image values are in integer (0-255), without conversion precision will be lost.

src/main/python/systemml/converters.py
asurve: Updated parameter description.

src/main/python/systemml/converters.py
asurve: Yes, technically we can go with DataSet name, year date version etc. 
By doing so intuition to pick mean value for associated model gets lost. It becomes harder to DS to figure out which mean to pick up.
These mean values are associated with model, unless someone going to change dataset for same model and keep model name exactly same (which I am not sure).

At this point I have changed to DATASET_MEAN (though I don't like it) and name of that dataset as VGG_ILSVRC_19_2014.

niketanpansare: Thanks for the change, but I would not want you to be uncomfortable with it. The reason I suggested using dataset instead of model is because you might later want to create a transfer learning notebook with `vgg` model on `cats and dogs` dataset.

niketanpansare:
LGTM Thanks @asurve 👍

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1810/

asurve:
Thanks @niketanpansare

PR #597, opened 2017-07-27
[SYSTEMML-1813] Breast cancer preprocessing simplification and cleanup

dusenberrymw:
In anticipation of near-future algorithmic improvements to the
preprocessing to improve model training, this simplifies and cleans up
the preprocessing code as follows.

- Previously, we were processing all slides into one large saved
DataFrame, and then splitting that DataFrame into train and validation
DataFrames.  This commit simplifies this by splitting the slide numbers
into train and validation sets, and then processing those slides
separately.  This effectively skips the creation of the large DataFrame,
and removes the need to split that large DataFrame into train/val ones,
which should provide a large performance benefit.  The DataFrame `union`
method can be used to combine two DataFrames row-wise.
- Previously, we maintained a list of "broken" slides that were manually
removed.  This commit removes that manual list, and instead adds a
try/except filtering step to automatically remove problematic slides.
- This commit moves ad-hoc sampling code into a new `sample` function.
- This commit moves code to add row indices to a DataFrame into a new
`add_row_indices` function.

The benefit is that near-future algorithmic improvements to the
preprocessing code will be much easier to incorporate.

### DIFF
--- a/projects/breast_cancer/MachineLearning-Keras-ResNet50.ipynb
+++ b/projects/breast_cancer/MachineLearning-Keras-ResNet50.ipynb
@@ -10,7 +10,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
    "outputs": [],
    "source": [
     "%load_ext autoreload\n",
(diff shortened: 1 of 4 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1815/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1959/

PR #601, opened 2017-08-01
[SYSTEMML-1821] Improve the training process in mnist_lenet_distrib_sgd.dml

feihugis:
We optimize `mnist_lenet_distrib_sgd.dml` by removing the unnecessary intermediate matrixes. However, there comes an error `org.apache.hadoop.mapred.FileAlreadyExistsException` with it when running the distributed training of MNIST_LeNet example. The revisions in 'org.apache.sysml.runtime.controlprogram.ParForProgramBlock#exportMatricesToHDFS' can solve the error, but may not the best solution.

### DIFF
--- a/scripts/nn/examples/mnist_lenet_distrib_sgd.dml
+++ b/scripts/nn/examples/mnist_lenet_distrib_sgd.dml
@@ -120,8 +120,8 @@ train = function(matrix[double] X, matrix[double] Y,
       # the other groups.
       group_beg = ((g-1) * group_batch_size) %% N + 1
       group_end = min(N, group_beg + group_batch_size - 1)
-      X_group_batch = X[group_beg:group_end,]
-      y_group_batch = Y[group_beg:group_end,]
+      row_group = group_beg - group_beg + 1
+     
 
       # Data structure to store gradients computed in parallel
       dW1_agg = matrix(0, rows=parallel_batches, cols=nrow(W1)*ncol(W1))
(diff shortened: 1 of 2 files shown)


feihugis:
@mboehm7 @dusenberrymw Could you have a look at this PR, and give some suggestions?

mboehm7:
thanks @feihugis - just to clarify the modification of ParForProgramBlock is conceptually unnecessary because CachableData::export would only write if the matrix object is "dirty" (i.e., its in-memory representation differs from its hdfs representation), the write goes into a different file system, format, persistent write, or there are pending RDD operations. However, we recently changed the file scheme handling and the handling of pending RDD operations is non-trivial - so there might be internal bugs. Since this primitive is called from many different places, we should fix the internals and not try to patch it from outside. I don't have time tonight but can look into this tomorrow.

mboehm7:
ok, I just did a quick test through spark_submit on a previous setup (with updated mnist_lenet_distrib_sgd.dml) and this run did not show any export issues on parfor. Maybe this is an issue of missing scratch_space cleanup in MLContext as pointed out a while ago (https://www.mail-archive.com/dev@systemml.incubator.apache.org/msg01665.html). @deroneriksson maybe we can simply delete the scratch_space on MLContext initialization and close (unless we already do it and I just missed it), which would still allow to pass variables between script invocations.

However, the script later ran into indexing issues with the parfor body (index range overflows the matrix dimensions on slicing out batches). @feihugis could you please double check the modified indexing expressions with input data that is unaligned with the batch sizes?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1840/

scripts/nn/examples/mnist_lenet_distrib_sgd.dml
```diff
@@ -120,8 +120,8 @@ train = function(matrix[double] X, matrix[double] Y,
       # the other groups.
       group_beg = ((g-1) * group_batch_size) %% N + 1
       group_end = min(N, group_beg + group_batch_size - 1)
-      X_group_batch = X[group_beg:group_end,]
-      y_group_batch = Y[group_beg:group_end,]
+      row_group = group_beg - group_beg + 1
```
dusenberrymw: I think this should be `group_end - group_beg + 1`.

dusenberrymw: Also, just for quick readability, I think it might be helpful to rename `row_group` to maybe `nrow_group`.

feihugis: Oh, my bad. I will double check it and rename the variable.

scripts/nn/examples/mnist_lenet_distrib_sgd.dml
```diff
@@ -134,12 +134,15 @@ train = function(matrix[double] X, matrix[double] Y,
...
       # Run graph on each mini-batch in this group in parallel (ideally on multiple GPUs)
-      parfor (j in 1:parallel_batches) {
+      # mode=REMOTE_SPARK, opt=CONSTRAINED
+      parfor (j in 1:parallel_batches, mode=REMOTE_SPARK, opt=CONSTRAINED, resultmerge=REMOTE_SPARK) {
         # Get a mini-batch in this group
-        beg = ((j-1) * batch_size) %% nrow(X_group_batch) + 1
-        end = min(nrow(X_group_batch), beg + batch_size - 1)
-        X_batch = X_group_batch[beg:end,]
-        y_batch = y_group_batch[beg:end,]
+        batch_beg = group_beg + ((j-1) * batch_size) %% row_group + 1
```
dusenberrymw: Also, I don't think the `+ 1` is needed here since `group_beg` already contains it and we're collapsing the indexing expressions into one.

feihugis: Yes, I agree with you.

feihugis:
@mboehm7 Thanks for your help! The indexing expressions have been updated.  I'm guessing the reason you did not meet the export issues on parfor may be that the program exits before entering the export part due to the indexing issues. Could you test it again?

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1844/

dusenberrymw:
@feihugis By the way, I like the updated code in #609 that moves the reformulation to its own file.

PR #609, opened 2017-08-08
[SYSTEMML-1831] Improve the efficiency of matrix subsetting

feihugis:
For the `rangeReIndex` operation, it needs to read the whole input matrix into memory first and then do the subsetting in the memory. It is not efficient because it needs to read a lot of unnecessary data, and even will be out of memory if the size of input matrix exceeds the available memory.

This PR optimizes the data pipeline of matrix subsetting. It reads the keys in the Hadoop sequence file, and identify the keys overlapped with the input range index. Then the values specified by the identified keys will be read.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1455,6 +1455,11 @@
 			<version>1.9.5</version>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>org.scala-lang.modules</groupId>
+            <artifactId>scala-xml_2.11</artifactId>
+            <version>1.0.6</version>
+        </dependency>
 	</dependencies>
 
 </project>
--- a/scripts/nn/examples/mnist_lenet_distrib_sgd.dml
+++ b/scripts/nn/examples/mnist_lenet_distrib_sgd.dml
@@ -134,7 +134,7 @@ train = function(matrix[double] X, matrix[double] Y,
       db4_agg = matrix(0, rows=parallel_batches, cols=nrow(b4)*ncol(b4))
 
       # Run graph on each mini-batch in this group in parallel (ideally on multiple GPUs)
-      parfor (j in 1:parallel_batches) {
+      parfor (j in 1:parallel_batches, mode=REMOTE_SPARK, opt=CONSTRAINED, resultmerge=REMOTE_SPARK) {
         # Get a mini-batch in this group
         beg = ((j-1) * batch_size) %% nrow(X_group_batch) + 1
         end = min(nrow(X_group_batch), beg + batch_size - 1)
--- a/scripts/nn/examples/mnist_lenet_distrib_sgd_optimize.dml
+++ b/scripts/nn/examples/mnist_lenet_distrib_sgd_optimize.dml
@@ -1,0 +1,404 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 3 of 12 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1895/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1898/

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1902/

feihugis:
cc @mboehm7 @dusenberrymw @nakul02  Could you help review this PR? The scala code and revisions in the pom.xml file are just for running the test case. They will be removed after the review.

mboehm7:
Thanks for working on this @feihugis. I hate to say it after you've already done the work but I think that this change would actually be counter-productive for performance in most cases. For example, consider the scripts for bivariate or univariate statistics. Normally, each remote worker would read the dataset once and then slice out individual columns in-memory. In contrast, with this change each column indexing would effectively read and deserialize the entire dataset, which could easily lead to slowdowns of orders of magnitude. Note that the proposed subset readers still read and deserialize all blocks. 

However, you're right that there are scenarios where we currently perform unnecessary reads. To understand these, let me give you a little bit of background: (1) You don't have to worry about OOMs because these in-memory indexing instructions are only compiled if the input and output fits into memory of the given context (driver or remote executors). (2) If the data is partitionable (row/column/batch) and does not fit into memory, we apply automatic partitioning, which allows to read individual partitions (see readMatrixPartition in MatrixObject). (3) By default, these partitions are created (locally or by a Spark job), materialized on HDFS, and read by remote workers. However, there is also a fused datapartition-execute Spark job, that partitions your data on-the-fly and executes the parfor body on each partition - there are special constraints but it should apply here (and that's what we eventually want to achieve for this script). I think there are multiple underlying issues that prevent us from compiling this right now (e.g., unknown batch sizes, forced execution, etc)

Moving forward, I would recommend to address the following two major things, which would actually improve performance of a whole lot of scripts: 
1) Improve parfor block indexing analysis [1]: Once, the parfor optimizer can identify (and guarantee) a constant batch size for the batch indexing expressions, a lot of problems (including the need to force any specific execution strategies) will go away.  
2) Parfor broadcast exploitation [2]: Coming back to the actual read issue. Indeed right now each worker reads shared inputs redundantly. The right way to address this in the general case is to send these inputs via broadcast variables to the executors, which transfer the data just once for all tasks to each executor. In order to do that, we need to know all memory estimates (see 1) because we would need to ensure that we can pin this shared variable into memory and still execute all other operations (with its inputs and output). This would help all scenarios that are not amenable to partitioning (data is small enough to fit in each executor).
3) Data Locality: One issue with data partitioning is that the parfor tasks might not run on the same node where their input data resides. Our MapReduce backend addresses this via explicit data locality, where we created a custom wrapper file splits that mapped iterations to partitions and reported the nodes where these partitions reside. We could extend our Spark parfor backend with a similar technique. This would help all scenarios where we apply partioning but the fused dataparittion-execute does not apply.

We want to do these things for a while now, so its probably a good idea to get it done right now. If you'd like to take these over, I'm happy to assist; otherwise I will make some time for them in the next weeks. 

[1] https://issues.apache.org/jira/browse/SYSTEMML-1782
[2] https://issues.apache.org/jira/browse/SYSTEMML-1313

feihugis:
Thanks for your detailed explanation, @mboehm7! It helps me get the whole picture of SystemML. I'm glad to work on the two issues. Your assistance will be definitely helpful.

For the subset reading, I double check the source code of Hadoop Sequence File Reader. You are right. `org.apache.hadoop.io.SequenceFile.Reader#next(org.apache.hadoop.io.Writable)` still needs to read the whole block, even though it skips the value. 

I will start to work on SYSTEMML-1782.

PR #614, opened 2017-08-10
[MINOR] Do not print statistics if an error has occured while execution

nakul02:
Ping @mboehm7, @niketanpansare, @deroneriksson 
This is a tiny change, which helps with scripts. If the program crashed, then statistics are not printed out.

Is this ok to merge?
Was there any reason to print the stats even if the program crashed?


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1920/

dusenberrymw:
But... if it fails quickly, then the script will run really fast!  It's a feature, not a bug.  ;)

mboehm7:
Well, I actually like to get the statistics for failed runs. This gives you at least some hints if the script failed on the first iteration or later, which is useful for debugging OOMs, memory leaks, and leaks of temporary variables with files in the scratch_space.

nakul02:
@mboehm7 - i shall close this PR then.

mboehm7:
for the future: I just expressed my personal opinion here but it was still up for discussion - if there is a good reason, we can change anything.

PR #627, opened 2017-08-22
[SYSTEMML-445] Integrate GPU exectype selection into our existing infrastructure

niketanpansare:

### DIFF
--- a/src/main/java/org/apache/sysml/hops/Hop.java
+++ b/src/main/java/org/apache/sysml/hops/Hop.java
@@ -192,7 +192,9 @@ public abstract class Hop
 	
 	public void checkAndSetForcedPlatform()
 	{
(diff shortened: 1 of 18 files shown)


src/main/java/org/apache/sysml/hops/Hop.java
```diff
@@ -850,6 +848,8 @@ protected abstract ExecType optFindExecType()
 	
 	public abstract String getOpString();
 
+	public abstract boolean isGPUEnabled();
```
nakul02: Could you please add a little documentation for the `isGPUEnabled` method?

nakul02: Also, thank you for not providing a default `return false`. 👍

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1979/

PR #630, opened 2017-08-23
[MINOR] Function with no args gives ArrayIndexOutOfBoundsException

deroneriksson:
Handle built-in function calls with no args (such as 'x=min();'), since
this could give an ArrayIndexOutOfBoundsException in the toString() of
BuiltinFunctionExpression during live variable analysis.


deroneriksson:
The following DML:
```
a=min();
```

previously gave:
```
ArrayIndexOutOfBoundsException: 0
	at org.apache.sysml.parser.BuiltinFunctionExpression.toString(BuiltinFunctionExpression.java:1333)
```

now outputs:
```
LanguageException: ERROR: dml/a.dml -- line 2, column 0 -- Function MIN() has no arguments.
```

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1983/

deroneriksson:
Thanks for reviewing @gweidner

PR #631, opened 2017-08-23
[MINOR] Do not print statistics if an error has occured while execution

nakul02:


nakul02:
@mboehm7 - Could we please not print stats if an error has occured?
This would greatly benefit @krishnakalyan3 with his perf test scripts.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1987/

deroneriksson:
@nakul02 Perhaps if an error has occurred, the statistics could be sent to logging (maybe log.info() level)? That way, the stats info could be available to @mboehm7 if he needs it, but standard output does not have the stats, which would handle @krishnakalyan3's use case.

nakul02:
Thanks @deron. I would have liked to put it into `log.error()` or `log.warn()` because `log.info()` contains other useful information that I might need. Most people would probably not have either of `warn` or `error` turned. 
Here is what I've done instead - redirected it to stderr. I've updated the PR with this.

@mboehm7 - letting it stay the way it is and complicating the perf scripts is not a good option.

deroneriksson:
@nakul02 Redirecting the stats to standard error if an error occurs seems like a simple, elegant solution that meets all requirements and minimizes impact on the perf scripts. Great idea.

nakul02:
Thanks @deroneriksson

deroneriksson:
LGTM. Thanks @nakul02

mboehm7:
This seems just a bit odd to me. If you want to check for errors, please directly check for errors instead of relying on a brittle indirection over the stats output.

deroneriksson:
@mboehm7 Grepping for "our distinct error message" actually seems brittle to me. Why couple perf test code to strings generated in another part of the project?

mboehm7:
Just define proper error codes and constants for relevant messages. This would actually allow you to run the perf tests without stats, which will be necessary for algorithms like decision tree (with fine grained access), where time measurements for all operations dominate performance by more than 3x.

nakul02:
@mboehm7 - this solution solves 2 problems - making the perf script simpler and not hindering your workflow. 
I am struggling to understand your objection to this solution.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2008/

mboehm7:
Well, although I'd prefer to handle this via a proper error handling, I'll not stand in the way here. So, please go ahead and we'll clean this up once the modified perftest is in use.

nakul02:
I shall merge.
Thanks @mboehm7 & @deroneriksson

PR #633, opened 2017-08-23
[SYSTEMML-1859] Throw error for function calls with extra input args

deroneriksson:
If a function call contains more input arguments than defined by the function
signature, throw an error.

Example:
```
triple = function(double a) return (double z) { z = a*3; }
x = triple(10,11,12)
print(x)
```

Before this PR, outputs:
```
30.0
```

After this PR, outputs:
```
ERROR: dml/a.dml -- line 4, column 0 -- Wrong number of function input arguments: 3 found, but 1 expected.
```


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1990/

deroneriksson:
Thanks for reviewing @dusenberrymw

PR #634, opened 2017-08-23
[MINOR] Fix case of externalFunction classname message

deroneriksson:
Use constant for 'classname' in externalFunction error message, since
the previous 'className' in the error message does not work.


deroneriksson:
Can test with:
```
time = externalFunction() return (double B)
  implemented in (classname="org.apache.sysml.udf.lib.TimeWrapper");
t = time();
print("Time: " + t);
```
(Using `className` instead of `classname` does not work.)

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1991/

deroneriksson:
Thanks for reviewing this @gweidner

PR #636, opened 2017-08-24
[MINOR] Use default namespace constant rather than String literal

deroneriksson:
Make DMLProgram's DEFAULT_NAMESPACE and INTERNAL_NAMESPACE final.
Use DEFAULT_NAMESPACE rather than String literal in FunctionCallIdentifier.


deroneriksson:
Thanks for reviewing @gweidner!

PR #637, opened 2017-08-24
[SYSTEMML-1793] Support right indexing on GPU

niketanpansare:
- This PR supports right indexing without requiring sparse to dense conversion of inputs. Note: this PR only supports dense output.
- Also, added RightIndexingTests in gpu package.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/1999/

src/main/java/org/apache/sysml/runtime/matrix/data/LibMatrixCUDA.java
```diff
@@ -3086,14 +3086,67 @@ private static int toInt(long num) throws DMLRuntimeException {
 	//**************** Matrix Manipulation Functions *********************/
 	//********************************************************************/
```
nakul02: A little javadoc here would be nice.

src/main/cpp/kernels/SystemML.cu
```diff
@@ -28,6 +28,27 @@ nvcc -ptx -arch=sm_30 SystemML.cu
 
 
 /**
+ * Performs a slice operation where the input matrix is sparse and the output matrix is dense.
+ * This function avoids unnecessary sparse to dense conversion of the input matrix.
+ */
+extern "C"
```
nakul02: A little javadoc explaining the overall logic and the parameters would be nice here.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2004/

PR #644, opened 2017-08-29
[SYSTEMML-1873] Update instruction constructor scopes to reflect use

deroneriksson:
Instructions are usually created using static parseInstruction methods.
Therefore, change constructor scopes to either private or protected
where appropriate to reflect this actual usage.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2022/

deroneriksson:
Thank you for reviewing @gweidner!

PR #664, opened 2017-09-14
[SYSTEMML-445] Add optimization flag for lazy cudafree and also refactor locks

niketanpansare:
- Guard the cudaFree with an optimization flag to be consistent with other SystemML features.
- Refactored the locks to seperate out read and write lock. This gives clear description of the error for different cases. 
- Also, I have added mem info (a  developer debugging utility).

### DIFF
--- a/src/main/java/org/apache/sysml/hops/OptimizerUtils.java
+++ b/src/main/java/org/apache/sysml/hops/OptimizerUtils.java
@@ -204,6 +204,13 @@ public class OptimizerUtils
 	 * 
 	 */
 	public static final boolean ALLOW_COMBINE_FILE_INPUT_FORMAT = true;
+	
+	/**
+	 * Enables lazy freeing of rmvar-ed GPU pointers.
+	 * If the subsequent instructions potential have matrices of same sizes, this optimization
+	 * allows reuse of previously allocated GPU pointer. Thus, saving the cost of unnecessary CUDA alloc and free.   
+	 */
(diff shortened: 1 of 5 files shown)


nakul02:
@niketanpansare - the optimization flags look good.
The read/write locks may not be needed. 

Read/write locks are applied only to not evict a particular block because it will be used in the *current* operation. Whether it is a read or a write is irrelevant.
There should never be any sort of lock (read/write) after the operation is over. If there is, then we aren't updating our data structures.

If it is for adding correctness checks, you could keep them, though, what we have already is sort of simpler.

Also, I don't like the way eviction policy logic is implemented. In a uncalled for need for brevity, it has become complex.
If we plan on sticking with eviction policy set to LRU, we can just use [LRUCacheMap](https://github.com/apache/systemml/blob/master/src/main/java/org/apache/sysml/utils/LRUCacheMap.java) as the type for `allocatedObjects` and get rid of redundant stuff like timestamp.
If we do decide to go to a another scheme, we can use another data structure later.

niketanpansare:
@nakul02 I have removed ALLOW_GPU_LAZY_CUDA_FREE logic from OptimizerUtils. We can revisit the issue with lazy cuda free in a separate PR. I would recommend keeping read/write locks as it eliminates future bugs. I am in favor of simplifying the eviction policy, but we can do that in a separate PR as well.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2098/

PR #670, opened 2017-09-21
[SYSTEMML-1929] Update deploy-mode in sparkDML.sh and docs

gweidner:
Updated spark-submit description to use `--deploy-mode` instead of deprecated parameters and corrected batch mode copy/paste error.

### DIFF
--- a/bin/systemml-spark-submit.py
+++ b/bin/systemml-spark-submit.py
@@ -129,7 +129,8 @@ if __name__ == '__main__':
     cparser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter,
                                       description='System-ML Spark Submit Script')
     # SPARK-SUBMIT Options
-    cparser.add_argument('--master', default='local[*]', help='local, yarn-client, yarn-cluster', metavar='')
+    cparser.add_argument('--master', default='local[*]', help='local, yarn', metavar='')
+    cparser.add_argument('--deploy-mode', default='client', help='client, cluster', metavar='')
     cparser.add_argument('--driver-memory', default='8G', help='Memory for driver (e.g. 512M, 1G)', metavar='')
     cparser.add_argument('--num-executors', nargs=1, help='Number of executors to launch', metavar='')
     cparser.add_argument('--executor-memory', nargs=1, help='Memory per executor', metavar='')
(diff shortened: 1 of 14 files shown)


mboehm7:
Could we do the same for scripts/sparkDML.sh, incl the removal of `spark.akka.frameSize`?

gweidner:
Good catch Matthias!  I'll add commit for sparkDML.sh to this PR too.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2118/

gweidner:
I updated sparkDML.sh to use `--master yarn --deploy-mode client` as default and removed 
`--conf spark.akka.frameSize=128`.  Also removed spark.akka.frameSize reference from recommended Spark configuration settings in spark-batch-mode and spark-mlcontext-programming-guide docs.

gweidner:
Note although I removed `spark.akka.frameSize` from sparkDML.sh and recommended settings, I'm not sure if I should instead replace with `--conf spark.rpc.message.maxSize=128` since it also means:
"Maximum message size (in MB) to allow in "control plane" communication; generally only applies to map output size information sent between executors and the driver. Increase this if you are running jobs with many thousands of map and reduce tasks and see messages about the RPC message size."

gweidner:
Applied similar changes to python performance test scripts.

nakul02:
It turns out that we don't need to specify the `spark.akka.frameSize=128` or `spark.rpc.message.maxSize=128`. According to the documentation for [Spark 1.6 ](https://spark.apache.org/docs/1.6.1/configuration.html), that is already the default size. Also, in the current spark master, the [value defaults](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/RpcUtils.scala#L59) to 128.
I also discussed this in an offline conversation with @niketanpansare.

I think we should remove this parameters from all our scripts.

nakul02:
For the python scripts, we'll need to pass the newly introduced `--deploy-mode` parameter to other parts of the script. 
Ping @krishnakalyan3.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2158/

gweidner:
Thank you @nakul02 for reviewing and clarifying!  I've incorporated the comments where I similarly modified sparkDML.sh in src/main/resources/scripts and also updated all algorithms docs.  For the python performance scripts, I reverted the `--deploy-mode` parameter which can be addressed in a separate PR.

krishnakalyan3:
I will start working on `deploy-mode` for `perf test suit`.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2159/

gweidner:
I opened a separate JIRA (https://issues.apache.org/jira/browse/SYSTEMML-1945) which can be used for the performance test scripts.

PR #683, opened 2017-10-12
[MINOR] Bug in foreign parent check in RewriteElementwiseMultChainOptimization

dhutchis:
The check for foreign parents in the interior of an element-wise multiply chain is incorrect.
Interior element-wise multiply nodes that have a foreign parent which is not a BinaryOp (for example, a write DataOp) are missed.
This leads to incorrect rewrites in unlucky DAGs.


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2177/

mboehm7:
this is awesome - thanks for catching and the fix @dhutchis.

PR #691, opened 2017-10-27
[SYSTEMML-540] Avoid redundant computation of cudnnPoolingForward in max_pool_backward

niketanpansare:
If the `max_pool` is invoked in the forward pass, then its output can be reused by the `max_pool_backward` rather than calling `cudnnPoolingForward` again. For sentence CNN with 2 epochs, this reduces the time for `max_pool_backward` from  6.361 to  2.966 seconds. Here are the stats:

```
With this PR (temporarily addressed issue 1 described below by rewriting DML for getting the statistics):
 5  gpu_maxpooling_backward [100:2-101:70] 2.966    936 sync[2.906s,936], nnc[0.000s,936], ad[0.017s,936], nnmb[0.027s,936]

Current Master:
  5  gpu_maxpooling_backward [100:2-101:70] 6.361    936 nnc[0.001s,936], nni[0.006s,936], nnmf[0.015s,936], nnmb[0.015s,936], ad[0.018s,936], sync[6.287s,936]
```

However, this optimization is disabled anytime there is a persistent read/write. So there are two (potentially overlapping) options here:
1. Fix following two persistent read/write issues:
a. `if loop` in  `nn/layers/dropout.dml`. 
b. A separate `cnn-forward()` method in the script.

2. Add optional argument `max_pool_forward` that accepts output of `max_pool`. This will work even if there are any lingering persistent read/write issues. However, it requires extending the `nn/layers/max_pool2d_builtin.dml` to accept the output of the forward function.

@bertholdreinwald @mboehm7 @prithvirajsen  @nakul02  @dusenberrymw

### DIFF
--- a/src/main/java/org/apache/sysml/hops/ConvolutionOp.java
+++ b/src/main/java/org/apache/sysml/hops/ConvolutionOp.java
@@ -47,14 +47,23 @@ public class ConvolutionOp extends Hop  implements MultiThreadedHop
 	private static final boolean THROW_ERROR_IF_INFERRED_SHAPE_MISMATCH = true;
 	// -------------------------------------------------------------------------
 	
+	// Specifies the type of this hop
 	private Hop.ConvOp op;
-
 	private int _maxNumThreads = -1; //-1 for unlimited
 
 	private ConvolutionOp() {
 		//default constructor for clone
 	}
 
+	/**
+	 * Create a hop from the builtin expression
+	 * 
+	 * @param l name of the hop
+	 * @param dt datatype (only supports matrix datatype)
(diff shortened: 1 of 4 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2208/

mboehm7:
this is very nice - but please leave `nn/layers/dropout.dml` as it is - this can very easily be done automatically. After IPA, constant folding and branch removal this function already boils down to a single statement - the only reason why it makes a difference right is that inlining happens before IPA. I'll add another IPA pass that inlines such single statement functions.

PR #700, opened 2017-11-12
[SYSTEMML-1969] Perform float-to-double conversion during eviction on host

niketanpansare:
We invoke transfer matrix from device to host in two cases:
1. During eviction of unlocked matrices
2. During acquireHostRead

If the single-precision support is enabled, then float-to-double conversion is required as CP expects the data to be in double format. This conversion can be done on host or on device. We typically prefer to do this conversion on device due to GPU's high-memory bandwidth. However, the conversion requires an additional space to be allocated for the conversion, which can lead to infinite recursion during eviction: `evict -> devictToHost -> float2double -> allocate -> ensureFreeSpace -> evict`. To avoid this recursion, it is necessary to perform this conversion in host.


nakul02:
Seems like an important patch. LGTM 👍

niketanpansare:
Thanks @nakul02  ... I will merge :+1:

PR #701, opened 2017-11-12
[MINOR] Merge extra GPU and extra DNN flags into single "sysml.stats.finegrained" flag to simplify usage

niketanpansare:
@nakul02 @bertholdreinwald  can you please review this PR ?

### DIFF
--- a/conf/SystemML-config.xml.template
+++ b/conf/SystemML-config.xml.template
@@ -75,15 +75,9 @@
    <!-- enables native blas for matrix multiplication and convolution, experimental feature (options: auto, mkl, openblas, none) -->
    <sysml.native.blas>none</sysml.native.blas>
 
-   <!-- prints finegrained statistics information -->
+   <!-- prints finegrained statistics information (includes extra GPU information and extra statistics information for Deep Neural Networks done in CP mode) -->
    <sysml.stats.finegrained>false</sysml.stats.finegrained>
    
-   <!-- prints extra statistics information for GPU -->
-   <sysml.stats.extraGPU>false</sysml.stats.extraGPU>
-
-   <!-- prints extra statistics information for Deep Neural Networks done in CP mode -->
-   <sysml.stats.extraDNN>false</sysml.stats.extraDNN>
-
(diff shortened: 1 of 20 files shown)


nakul02:
While I don’t see any particular reason to merge the two (you are making control of the extra statistics less fine grained 😀), if you really wanted to, I can suggest that internally, you can let the GPUStatistics.DISPLAY_STATISTICS and LibMatrixDNN.DISPLAY_STATISTICS remain and set both of them true when your newly introduced parameter “FINEGRAINED_STATISTICS” is true. That way if one wanted to go back to more fine grained control of these fine grained statistics, it would be possible.

niketanpansare:
The reason is usability and forcing developers not to add additional flags.

The user has to know only one flag: fine grained stats. All others are dev flags n should not be exposed. This avoids permutations such as finegrained but not gpu stats n so on.

 The dev has to decide if the stats has to be exposed and if yes, she has only two options: stats or finegrained stats. 

For the above reasons, I would prefer to remove gpu stats n DNN stats completely to avoid technical debt. If it has to be enabled in future, the person HAS to do that extra work.

niketanpansare:
And I would prefer to add this before 1.0 as this is an API feature

nakul02:
There are two sets of people here - developers and users. 

If the task is to make it easier for users, what I suggested might be enough and you can have a single user facing system setting and the internals can expose two different sets of global variables. 


If the task is to make it so that developers need not add an extra global variable for another set of fine grained statistics, then unifying all internal fine grained statistics is fine by me. 

# begin of 2 cents. 
Please note that if we go this route, we will forgo the ability to control which part of the system collects fine grained stats and it will be an all or nothing. If for future experiments, we wish to turn off fine grained stats for a particular subsystem while focusing on another, it will not be possible with this approach. 
#end of 2 cents

niketanpansare:
I agree, not collecting statistics from custom components is a conscious decision. If it were to be revisited, then it has to be addressed more wholistically. According to me, GPU statistics is just fine grained stats when SystemML ends up using gpu. Similarly for DNN operations.

nakul02:
Alright, since this PR takes into account the discussion we had above, LGTM :+1:

PR #704, opened 2017-12-04
[Doc] svd local builtin function to the dml lang reference

j143:
Let's add this for now. Hopefully, we will implement gpu version & scaled version in Jan, 2018.


mboehm7:
LGTM - thanks @j143

PR #706, opened 2017-12-06
[MINOR][PERF] String new line error

krishnakalyan3:
While running performance tests on `AWS`, I noticed that a new line was being added to the end of the string. 
This caused weird file names (`m-svm.multinomial.dense.10k_100?.0.json`) and errors during the perf test execution.

By stripping the string this problem was resolved and could run my performance test workloads.


krishnakalyan3:
thanks for the review @gweidner

j143:
Hi @krishnakalyan3 , can the commit message be a little more explanatory, when we look at the git log. 
Thanks.

krishnakalyan3:
@j143 yes. Thanks for the feedback.

PR #724, opened 2018-02-03
Gaussian Process Classification Script.

j143-zz:
_what is here?_
Gaussian Process Classification script.

**result:**
1. calculation of gradients.
2. gives marginal likelihoods, & their gradients.


j143:
Jenkins, please do not test this PR.

PR #725, opened 2018-02-05
[minor] name change for sparse-sparse case

j143:
I am not sure, whether I am correct. Because, even with out the change, your changeset is correct. Thanks.

BTW, I am trying to fix the udf in expressions[1444] from last 24 hrs. But, still trying to figure a way out. Thanks.
cc @mboehm7


mboehm7:
LGTM - thanks for catching this.

j143:
Also, for my understanding why are we assuming the `b` have likely less zeros.

j143:
Can you give an example matrix, I will try to understand this operation. So, that I will try to use this as base for the coming operations.  cc @mboehm7 .

mboehm7:
well, all combinations of sparse and dense are relevant but its more common to have sparse (filters), at least in the applications I've seen so far.

mboehm7:
This operation is actually nothing else than `(X > 0) * dout` but Niketan added an operator to avoid the allocation and materialization of the intermediate. Once codegen is enabled by default (and supported by the GPU backend), we could actually remove this custom fused operators. If you like, you can start with SYSTEMML-2110 that aims to support relu_backward in codegen (in row and cell templates, simply compiled to > and * operations at CPlan level).

PR #726, opened 2018-02-07
[WIP][SYSTEMML-2110] Codegen support for relu operation

j143:
`(X > 0) * Y`,  integration support for codegeneration.

1. spoof primitives.
2. tests

**Question**
But, how `libmatrixDNN` class's operator can be invoked, is it a rewrite or is there any function name such as `relu_backward()` .

### DIFF
--- a/src/test/java/org/apache/sysml/test/integration/functions/codegen/CellwiseTmplTest.java
+++ b/src/test/java/org/apache/sysml/test/integration/functions/codegen/CellwiseTmplTest.java
@@ -56,6 +56,7 @@ public class CellwiseTmplTest extends AutomatedTestBase
 	private static final String TEST_NAME18 = TEST_NAME+18; //sum(ifelse(X,Y,Z))
 	private static final String TEST_NAME19 = TEST_NAME+19; //sum(ifelse(true,Y,Z))+sum(ifelse(false,Y,Z))
 	private static final String TEST_NAME20 = TEST_NAME+20; //bitwAnd() operation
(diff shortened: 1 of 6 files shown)


j143:
Sorry, I will try to map the expression to >, * operators instead of a primitive operator. Thanks.

mboehm7:
well right, there is no need for any of these extensions - just add the tests and a TODO comment to `BinaryOp.constructLopsBinaryDefault` (line 662) to remove RELU_BACKWARD once codegen is supported for GPUs as well. The reason is that relu_backward was not added as a hop rewrite but a hop-lop rewrite, which means its not even visible during codegen.

j143:
Hi @mboehm7 , the tests here seem to give wrong result to me, wrt R scripts.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2276/

mboehm7:
yes, the reason is that we use a row-wise default reshape while R uses a column-wise reshape. This was a conscious decision because we internally use row-major representations while R uses column-major formats. You can simply use fix your R script via 
`X = matrix(seq(1, 6000)/600, 300, 20, byrow=TRUE);`
instead of 
`X = matrix(seq(1, 6000)/600, 300, 20);`

While you are at it, could you also remove the other merged changes. Thanks.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2278/

mboehm7:
LGTM - thanks @j143

PR #728, opened 2018-02-09
[SYSTEMML-2068] codegen BitwAnd support

j143:
_what is here?_
1. change of opcode's naming to trigger builtin function `hop.OpOp2.BITWAND` to validate against.
2. `dense-dense` matrix primitive for `BitwAnd` operation.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/codegen/cplan/CNodeBinary.java
+++ b/src/main/java/org/apache/sysml/hops/codegen/cplan/CNodeBinary.java
@@ -239,7 +239,7 @@ public class CNodeBinary extends CNode
 				|| this == VECT_EQUAL || this == VECT_NOTEQUAL
 				|| this == VECT_LESS || this == VECT_LESSEQUAL
 				|| this == VECT_GREATER || this == VECT_GREATEREQUAL
-				|| this == VECT_XOR || this == VECT_BITWAND_SCALAR;
+				|| this == VECT_XOR || this == VECT_BITWAND;
 		}
 		public boolean isVectorMatrixPrimitive() {
 			return this == VECT_MATRIXMULT
--- a/src/main/java/org/apache/sysml/hops/codegen/cplan/CNodeBinary.java
+++ b/src/main/java/org/apache/sysml/hops/codegen/cplan/CNodeBinary.java
@@ -42,15 +42,16 @@ public class CNodeBinary extends CNode
 		VECT_EQUAL_SCALAR, VECT_NOTEQUAL_SCALAR, VECT_LESS_SCALAR, 
 		VECT_LESSEQUAL_SCALAR, VECT_GREATER_SCALAR, VECT_GREATEREQUAL_SCALAR,
 		VECT_CBIND,
-		VECT_XOR_SCALAR,
+		VECT_XOR_SCALAR, VECT_BITWAND_SCALAR,
 		//vector-vector operations
 		VECT_MULT, VECT_DIV, VECT_MINUS, VECT_PLUS, VECT_MIN, VECT_MAX, VECT_EQUAL, 
 		VECT_NOTEQUAL, VECT_LESS, VECT_LESSEQUAL, VECT_GREATER, VECT_GREATEREQUAL,
-		VECT_XOR,
+		VECT_XOR, VECT_BITWAND,
 		//scalar-scalar operations
 		MULT, DIV, PLUS, MINUS, MODULUS, INTDIV, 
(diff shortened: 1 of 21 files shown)


j143:
Hi @mboehm7 , can this preliminary one be merged. I am figuring out `sparse` cases,

But, as the bit wise operations happens at bitwise, how can we take advantage of this?.

j143:
Test this Jenkins.

src/main/java/org/apache/sysml/hops/codegen/cplan/CNodeBinary.java
```diff
@@ -234,7 +239,7 @@ public boolean isVectorVectorPrimitive() {
 				|| this == VECT_EQUAL || this == VECT_NOTEQUAL
 				|| this == VECT_LESS || this == VECT_LESSEQUAL
 				|| this == VECT_GREATER || this == VECT_GREATEREQUAL
-				|| this == VECT_XOR;
+				|| this == VECT_XOR || this == VECT_BITWAND_SCALAR;
```
mboehm7: Shouldn't this be VECT_BITWAND instead?

src/main/java/org/apache/sysml/runtime/codegen/LibSpoofPrimitives.java
```diff
@@ -1914,7 +1915,40 @@ public static void vectGreaterequalAdd(double bval, double[] a, double[] c, int[
...
+	 * BitwiseAnd Operator
+	 *
+	 * @return
+	 */
+
+	//1. dense vector vs. scalar
+	public static double[] vectBitwandWrite(double[] a, double bval, int ai, int len) {
+		double[] c = allocVector(len, false);
+		for( int j = 0; j < len; j++ )
+			c[j] = (double)( UtilFunctions.toInt(a[ai+j]) & UtilFunctions.toInt(bval) );
```
mboehm7: Please use a simple cast to int instead of this machine-precision-aware primitive for consistency with the builtin functions and scalar codegen template.

mboehm7: Also please prepare the int value of bval outside the loop.

src/main/java/org/apache/sysml/runtime/codegen/LibSpoofPrimitives.java
```diff
@@ -1914,7 +1915,40 @@ public static void vectGreaterequalAdd(double bval, double[] a, double[] c, int[
...
+		return c;
+	}
+
+	//2. scalar vs. dense vector
+	public static double[] vectBitwandWrite(double bval, double[] a, int ai, int len) {
+		double[] c = allocVector(len, false);
+		for( int j = 0; j < len; j++ )
+			c[j] = ( UtilFunctions.toInt(bval) & UtilFunctions.toInt(a[ai+j]) );
```
mboehm7: As the bitwise AND is commutative please simply call the vector-scalar primitive instead.

src/main/java/org/apache/sysml/runtime/codegen/LibSpoofPrimitives.java
```diff
@@ -1914,7 +1915,40 @@ public static void vectGreaterequalAdd(double bval, double[] a, double[] c, int[
...
+		return c;
+	}
+
+	//3. dense vector vs. dense vector
+	public static double[] vectBitwandWrite(double[] a, double[] b, int ai, int bi, int len) {
+		double[] c= allocVector(len, false);
+		for( int j = 0; j < len; j++ )
+			c[j] = (double)( UtilFunctions.toInt(a[ai+j]) & UtilFunctions.toInt(b[bi+j]));
```
mboehm7: Same casting as described above.

j143: Hi @mboehm7 ,
tried to cast for int as `(int)a[ai+j]`, without success. 
1. `Math.floor(a[ai+j])` is giving correct result.
2. It can be that, I am doing some mistake else where.

j143:
All tests passed locally.

j143:
Test this Jenkins

mboehm7:
LGTM - I removed the sparse block related changes and fixed the casting issues. You were right about the use of `UtilFunctions.toInt` because our builtin function object uses the same approach which was the reason for the discrepancies in the tests. I now use simply used the function object in all related codegen operations to ensure consistency.

PR #731, opened 2018-02-17
[SYSTEMML-1994] GP regression script, with predictive mean & variance

j143-zz:
_what is here?_
this is the gaussian process regression script, which 
1. gives **predictive mean** and **predictive variance** of a test point, later in the bayesian optimization 
2. To see whether the point is good to proceed with.


mboehm7:
There seem to be unsupported characters in this script - I would recommend to try running it over some example inputs.

akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2321/

PR #736, opened 2018-03-02
[SYSTEMML-2077] Add eval built-in function

EdgarLGB:
Hi, I have added an eval built-in function and a test for it. Thanks!


akchinSTC:

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -19,1749 +19,1718 @@
...
+            raiseValidateError("Function " + this + " has no arguments.", false);
+        }
+        Identifier id = this.getFirstExpr().getOutput();
+        output.setProperties(this.getFirstExpr().getOutput());
+        output.setNnz(-1); //conservatively, cannot use input nnz!
+        this.setOutput(output);
+
+        switch (this.getOpCode()) {
+            case EVAL:
+                checkNumParameters(Integer.MAX_VALUE);
```
mboehm7: This check has no affect if you look at the implementation of `checkNumParameters`. I would recommend to check that `_args.length >= 1 ` directly (and maybe that the first parameter is a string or function identifier);

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -19,1749 +19,1718 @@
...
+                output.setDataType(DataType.MATRIX);
+                output.setValueType(ValueType.DOUBLE);
+                output.setDimensions(1, 1);
+                output.setBlockDimensions(1000, 1000);
```
mboehm7: Please derive the blocksize from the input if there is at least one matrix input or use the default block size from ConfigurationManager.getBlocksize()

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -19,1749 +19,1718 @@
...
+        output.setProperties(this.getFirstExpr().getOutput());
+        output.setNnz(-1); //conservatively, cannot use input nnz!
+        this.setOutput(output);
+
+        switch (this.getOpCode()) {
+            case EVAL:
+                checkNumParameters(Integer.MAX_VALUE);
+                output.setDataType(DataType.MATRIX);
+                output.setValueType(ValueType.DOUBLE);
+                output.setDimensions(1, 1);
```
mboehm7: If you fix the output dimensions to 1x1 this could lead to incorrect results if subsequent rewrites exploit this "known" size information. Please, use -1 to indicate unknowns instead.

src/main/java/org/apache/sysml/parser/DMLTranslator.java
```diff
@@ -88,3129 +82,3038 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+
+public class DMLTranslator {
+    private static final Log LOG = LogFactory.getLog(DMLTranslator.class.getName());
+    private DMLProgram _dmlProg = null;
```
mboehm7: This file has a similar formatting issue (removed tabs) as described above.

src/main/java/org/apache/sysml/api/mlcontext/ScriptExecutor.java
```diff
@@ -253,7 +253,12 @@ protected void setGlobalFlags() {
 		
 		// set the global compiler configuration
 		try {
+
 			OptimizerUtils.resetStaticCompilerFlags();
+
+			//Workaround: for eval func to avoid cleaning the unused UDF
+			OptimizerUtils.ALLOW_INTER_PROCEDURAL_ANALYSIS = false;
```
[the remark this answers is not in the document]
mboehm7: Once your PR is merged in, I will then extend IPA accordingly, to conditionally disable this flag only if there is at least one eval function in the program.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Eval built-in function instruction
+ * Note: it supports only single matrix[double] output
+ */
+public class EvalBuiltinNaryCPInstruction extends BuiltinNaryCPInstruction {
+
+    public EvalBuiltinNaryCPInstruction(Operator op, String opcode, String istr, CPOperand output, CPOperand... inputs) {
+        super(op, opcode, istr, output, inputs);
+    }
```
mboehm7: Please fix the formatting (tabs over spaces) for the entire file.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void processInstruction(ExecutionContext ec) throws DMLRuntimeException {
+        //1. inside the new instruction, get the correspond function statement block
+        FunctionProgramBlock fpb = ec.getProgram().getFunctionProgramBlock(null, inputs[0].getName());
```
mboehm7: You might want to check if the given name contains a "::" to recognize given namespaces - in this case you would need to split it and pass namespace and function name separately to getFunctionProgramBlock. Also note if you where not given a string but a function identifier at parser level, you need to capture both namespace and function from the function identifier.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+            Data d;
+            if (di.getDataType().isScalar()) {
+                d = ec.getScalarInput(arg.getName(), arg.getValueType(), true);
+            } else if (di.getDataType().isMatrix()) {
+                d = ec.getMatrixObject(arg.getName());
+            } else if (di.getDataType().isFrame()) {
+                d = ec.getFrameObject(arg.getName());
+            } else {
+                throw new DMLRuntimeException(String.format("processInstruction():: unknown argument %s", arg.getName()));
+            }
```
mboehm7: You can simply get the variables via ec.getVariable which directly returns you any type of data object.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+        fpb.execute(ec);
+        //3. get the single func output and convert it to a matrix object
+        String funcOutputName = fpb.getOutputParams().get(0).getName();
+        Data outputData = ec.getVariable(funcOutputName);
+        MatrixObject mo;
+        if (outputData.getDataType().isScalar()) {
+            ScalarObject so = (ScalarObject) outputData;
+            //convert scalar to matrix
+            mo = MLContextConversionUtil.doubleMatrixToMatrixObject(output.getName(), new double[][]{{so.getDoubleValue()}});
```
mboehm7: Please don't use MLContextConversionUtil to prevent dependencies from the runtime to APIs. Specifically MLContext imports hadoop and spark classes, which might not be available when invoking SystemML through a different API such as JMLC for low-latency scoring.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+            if (di.getDataType().isScalar()) {
+                d = ec.getScalarInput(arg.getName(), arg.getValueType(), true);
+            } else if (di.getDataType().isMatrix()) {
+                d = ec.getMatrixObject(arg.getName());
+            } else if (di.getDataType().isFrame()) {
+                d = ec.getFrameObject(arg.getName());
+            } else {
+                throw new DMLRuntimeException(String.format("processInstruction():: unknown argument %s", arg.getName()));
+            }
+            lvm.put(di.getName(), d);
```
mboehm7: You must not modify the original symbol table (i.e., variable map) because this could overwrite other variables of the same name. Instead please create a new ExecutionContext and VariableMap and only copy the input references into this new VariableMap. You might want to have a look into `FunctionCallCPInstruction` - maybe we can refactor the code there to simply call a common code path from the eval function.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,63 @@
...
+        Data outputData = ec.getVariable(funcOutputName);
+        MatrixObject mo;
+        if (outputData.getDataType().isScalar()) {
+            ScalarObject so = (ScalarObject) outputData;
+            //convert scalar to matrix
+            mo = MLContextConversionUtil.doubleMatrixToMatrixObject(output.getName(), new double[][]{{so.getDoubleValue()}});
+        } else {
+            mo = (MatrixObject) outputData;
+        }
+        ec.getVariables().put(output.getName(), mo);
```
mboehm7: This could cause leaks of missing cleanups for potential intermediate files of existing variables. We would need to first use `old = ec.removeVariable` and `ec.cleanupCacheableData(old)` to properly clean up existing variables that are overwritten by the function output before putting the new variable onto the output name. Again, maybe the best way would be to simply refactor the existing function call logic and reuse it.

src/test/scripts/org/apache/sysml/api/mlcontext/eval-test.dml
```diff
@@ -0,0 +1,58 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+
+# product between matrix and scalar
+f1 = function (matrix[double] M, double factor) return (double res) {
+    res = prod(M) * factor
+}
```
mboehm7: For dml scripts, we use 2-space indentation.

src/test/scripts/org/apache/sysml/api/mlcontext/eval-test.dml
```diff
@@ -0,0 +1,58 @@
...
+
+# create a result matrix
+R = matrix(1, rows=2, cols=1)
+
+R[1, 1] = eval(f1, X, y)
+R[2, 1] = eval(f2, X, y)
+
+print(toString(R))
+
+M3 = eval(f3, X, X)
```
mboehm7: Maybe you could add tests where the functionname is given as string for both with and without namespace.

src/test/scripts/org/apache/sysml/api/mlcontext/eval-test.dml
```diff
@@ -0,0 +1,58 @@
...
+print(toString(M3))
+
+
+# a function frame
+#F = read("func.csv", data_type="frame", rows=2, cols=1, format="csv")
```
mboehm7: Please remove commented code to keep the tests and source code clean.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
EdgarLGB: But it seems that the parser doesn't accept the function identifier as the function argument.

j143:
Add to whitelist

j143:
Test this Jenkins

j143:
Hi @EdgarLGB , Please allow the SystemML default config settings for IDE. This will take care of whitespace issues.

EdgarLGB:
OK. I fixed the format problem and I'd like to create a new PR.

PR #737, opened 2018-03-03
[SYSTEMML-2167] Support for nary rbind operation

j143:
_what is here?_
- support for nary rbind operation
- work in progress

_what is my idea?_
- for every new row copy the whole row to the `c[]`

### DIFF
--- a/scripts/test.dml
+++ b/scripts/test.dml
@@ -1,0 +1,6 @@
+X = matrix(seq(1,6000)/6000, 20, 300);
+Y = matrix(2, 2, 300);
+Z = matrix(3, 3, 300);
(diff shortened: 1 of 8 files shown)


akchinSTC:
https://sparktc.ibmcloud.com/jenkins/job/SystemML-PullRequestBuilder/2300/

mboehm7:
RBind is actually not applicable to row-wise fusion because the rows from two or more inputs constitute different rows in the output. So unless we support a new column-wise template (for which I've not seen that many use cases) there is no need to add binary or n-ary rbind support.

j143:
closing this one. Thanks.

PR #740, opened 2018-03-05
[SYSTEMML-2077] New eval builtin function

EdgarLGB:
Hi,
I made an another PR for the new eval built-in function. I made a code refactoring and fixed the code format problem. And I put all the commits into the master branch. Thanks for review.

Regards,
Guobao


akchinSTC:

src/main/java/org/apache/sysml/hops/NaryOp.java
```diff
@@ -159,7 +159,8 @@ else if ( areDimsBelowThreshold() )
 		setRequiresRecompileIfNecessary();
 		
 		//ensure cp exec type for single-node operations
-		if ( _op == OpOpN.PRINTF )
+		//TODO eval function only supports CP execution type
```
mboehm7: please remove the TODO as eval will be be similar to all other function calls always be executed in CP.

src/main/java/org/apache/sysml/hops/ipa/InterProceduralAnalysis.java
```diff
@@ -89,7 +89,7 @@
 	protected static final boolean INTRA_PROCEDURAL_ANALYSIS      = true; //propagate statistics across statement blocks (main/functions)	
 	protected static final boolean PROPAGATE_KNOWN_UDF_STATISTICS = true; //propagate statistics for known external functions 
 	protected static final boolean ALLOW_MULTIPLE_FUNCTION_CALLS  = true; //propagate consistent statistics from multiple calls 
-	protected static final boolean REMOVE_UNUSED_FUNCTIONS        = true; //remove unused functions (inlined or never called)
+	protected static final boolean REMOVE_UNUSED_FUNCTIONS        = false; //remove unused functions (inlined or never called)
 	protected static final boolean FLAG_FUNCTION_RECOMPILE_ONCE   = true; //flag functions which require recompilation inside a loop for full function recompile
```
mboehm7: Once we merge this in, we would need to make this conditional on the presence of eval functions. I'm happy to take this over, but if you want to give it a shot, you can try extending the existing function call graph to capture this information and create the IPA rewrite set accordingly.

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -379,6 +380,14 @@ public void validateExpression(HashMap<String, DataIdentifier> ids, HashMap<Stri
 		this.setOutput(output);
 		
 		switch (this.getOpCode()) {
+		case EVAL:
+			if (_args.length == 0) {
+				raiseValidateError("Function eval should provide at least one arguments.", false);
```
mboehm7: Typo: "one arguments" -> "one argument, i.e., the function name".

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
```diff
@@ -1792,6 +1801,12 @@ else if ( functionName.equals("bitwShiftR") )
 			bifop = Expression.BuiltinFunctionOp.BITWSHIFTR;
 		else if ( functionName.equals("ifelse") )
 			bifop = Expression.BuiltinFunctionOp.IFELSE;
+		else if (functionName.equals("eval")) {
+			bifop = Expression.BuiltinFunctionOp.EVAL;
+			//Need to cast the first param "name of the referenced func" into a string identifier
+			DataIdentifier func = (DataIdentifier) paramExprsPassed.get(0).getExpr();
+			paramExprsPassed.get(0).setExpr(new StringIdentifier(func.toString(), func));
```
mboehm7: Please try to avoid this explicit modification. This should not be necessary and would probably crash if the argument is an expression such as "foo" + X[i,1]. Maybe you can add a test for such expressions to your unit test.

src/main/java/org/apache/sysml/parser/DMLTranslator.java
```diff
@@ -1272,6 +1272,10 @@ public void constructHops(StatementBlock sb)
 					if (!(target instanceof IndexedIdentifier)) {
 						//process right hand side and accumulation
 						Hop ae = processExpression(source, target, ids);
+						// if it is eval built-in func, add hop to output
+						if (ae instanceof NaryOp && ((NaryOp) ae).getOp().equals(OpOpN.EVAL)) {
+							output.add(ae);
+						}
```
mboehm7: Why was this necessary and not sufficiently handled by the default case for setting the output properties?

src/main/java/org/apache/sysml/parser/dml/DmlSyntacticValidator.java
```diff
@@ -519,7 +532,8 @@ public void exitBuiltinFunctionExpression(BuiltinFunctionExpressionContext ctx)
 		String functionName = names[1];
 
 		ArrayList<ParameterExpression> paramExpression = getParameterExpressionList(ctx.paramExprs);
-
+		//convert the namespace for the function identifier arguments
+		convertNamespace(paramExpression);
```
mboehm7: Any modification to the syntactic parser for DML would also need to go into the syntactic parser for PyDML. Hence, if possible, it would be good to leave the syntactic parsers untouched and move any preprocessing into the semantic parsing, i.e., parse tree validation.

src/main/java/org/apache/sysml/parser/dml/DmlSyntacticValidator.java
```diff
@@ -507,6 +507,19 @@ private void castAsScalarDeprecationCheck(String functionName, ParserRuleContext
 		}
 	}
 
+	private void convertNamespace(ArrayList<ParameterExpression> params) {
+		for (ParameterExpression p : params) {
```
mboehm7: Doing that for all parameters of all builtin functions seems dangerous as it has the potential to break user-provided string parameters as well. Please restrict these modifications to the first parameter of eval calls.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	@Override
+	public void processInstruction(ExecutionContext ec) throws DMLRuntimeException {
+		//1. get the namespace and func
+		CPOperand func = inputs[0];
+		String funcName = func.getName();
+		String namespace = null;
+		if (func.getName().contains("::")) {
+			String[] split = func.getName().split("::");
```
mboehm7: Please replace with `DMLProgram.splitFunctionKey` to have a single place where these symbols are defined.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,82 @@
...
+				Matcher matcher = pattern.matcher(func.getName());
+				if (matcher.matches()) {
+					funcName = matcher.group(0);
+				}
+			} else {
+				funcName = funcSignature;
+			}
+		}
+		// bound the inputs to avoiding being deleted after the function call
+		CPOperand[] boundInputs = Arrays.asList(inputs).subList(1, inputs.length).toArray(new CPOperand[]{});
```
mboehm7: Use `Arrays.copyOfRange` instead.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,82 @@
...
+		ArrayList<String> boundOutputNames = new ArrayList<>();
+		boundOutputNames.add(output.getName());
+		ArrayList<String> boundInputNames = new ArrayList<>();
+		for (CPOperand input : boundInputs) {
+			boundInputNames.add(input.getName());
+		}
+
+		//2. copy the created output matrix
+		MatrixObject outputMO = new MatrixObject(ec.getMatrixObject(output.getName()));
```
mboehm7: There is no need to create a copy or even obtain this output object.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,82 @@
...
+		//4. convert the result to matrix
+		Data newOutput = ec.getVariable(output);
+		if (newOutput instanceof MatrixObject) {
+			return;
+		}
+		MatrixBlock mb = null;
+		if (newOutput instanceof ScalarObject) {
+			//convert scalar to matrix
+			mb = new MatrixBlock(1, 1, false);
+			mb.setValue(0, 0, ((ScalarObject) newOutput).getDoubleValue());
```
mboehm7: use `new MatrixBlock(((ScalarObject) newOutput).getDoubleValue())` instead.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
```diff
@@ -0,0 +1,82 @@
...
+		} else if (newOutput instanceof FrameObject) {
+			//convert frame to matrix
+			mb = DataConverter.convertToMatrixBlock(((FrameObject) newOutput).acquireRead());
+			ec.cleanupCacheableData((FrameObject) newOutput);
+		}
+		outputMO.acquireModify(mb);
+		outputMO.release();
+		ec.setVariable(output.getName(), outputMO);
```
mboehm7: Use `ec.setMatrixOutput(output.getName(), mb);` instead

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
EdgarLGB: I have a question about the first argument of eval func. Should it be better to always have a string of func name as the first argument, i.e, eval("f1", X, y)? Because if it accepts any expression as the first argument, i.e, eval("f1" + X[i, 1], X, y), it will be complicated to deal with this expression of any type. And in fact, I think will can use it in an another way, i.e, eval("f1", X, y) + X[i, 1].

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
EdgarLGB: I think that the eval func always returns a matrix. Then I set the ouput type with Matrix. So I copy the created output matrix object to avoid overriding it when the func call returns a scalar object.

EdgarLGB:
Hi,
I have made the refactoring. And I'd like to consider the alias of the imported source as a macro. Hence, when meeting the namespace alias in symantic validation of eval func, I can replace the alias with the real complete path of the namespace.

src/main/java/org/apache/sysml/runtime/instructions/cp/EvalBuiltinNaryCPInstruction.java
mboehm7: Yes, you right, I forgot about the scalar case. Since the copy constructor creates a different ID maybe you alternatively, get the object and disable its cleanup with `mo.enableCleanup(false)`, which preserves the contents but might drop it from the symbol table.

src/main/java/org/apache/sysml/parser/BuiltinFunctionExpression.java
mboehm7: Let us separate two aspects: the type of the input, and the script level restrictions. The type should indeed always be a string which also guarantees that you will always get a string during runtime. 

However, we should not artificially restrict the possibility to have arbitrary expressions as input at script level because the common case will be either selections eval(X[i,]) or concatenations eval("foo"+i). These expressions will be handled automatically by the compiler, as done for any other builtin function - so you don't need to worry about it. 

Similarly, we will also allow eval in expression, but as mentioned before I will add a rewrite that splits the DAGs after eval in order to create a hook for dynamic recompilation because the sizes (dimensions and sparsity) of an unknown function will be unknown. Without dynamic recompilation this would lead to unnecessary distributed operations. In your tests you did not encounter this issue because toString is only supported as in-memory operation.

mboehm7:
ok thanks @EdgarLGB - once the validation of input parameters is generalized for expressions I can merge this in. Note that it might take a day or two before it'll be pushed to master because I need to add the other features first (conditionally disable function removal, rewrite for DAG split after eval).

EdgarLGB:
ok thanks @mboehm7

mboehm7:
LGTM - thanks @EdgarLGB I'll merge this in together with the additional features. 

Just to answer your question: In practice, a lot of scripts depend on dynamic recompilation to adapt runtime plans for initial unknowns or changing sizes (dimensions and sparsity). We compile at the granularity of entire HOP DAGs. Since the output of eval (which calls an unknown function) will in most cases be unknown during initial compilation, all operations with data dependencies on the eval output would fallback to conservative distributed operations. Hence, we want - similar as we do it for data-dependent operators like table or removeEmpty (see `RewriteSplitDagDataDependentOperators`) - cut the DAG after this operator to create a hook for dynamic recompilation. At the same time, we don't want to cut the DAG into too many pieces because this limits rewrite potential - accordingly, the above rewrite tries to find a minimal cut in case there are multiple relevant operators in the same DAG.

mboehm7:
Thanks again @EdgarLGB for your contribution - it worked like a charm. I further added the support for function name expressions, but removed the support for function calls outside the default namespace for two reasons. First, there were potential conflicts with constant variables, which could have caused incorrect results and runtime failures. Second, the call to namespace functions would still be limited to static names, which could simply be called directly. However, I took a note and will extend that once we cleaned up the entire meta data management of namespaces. Thanks.

EdgarLGB:
cool @mboehm7  And if we could support the global constant variables to resolve the conflicts?

mboehm7:
@EdgarLGB no global constants will not solve this issue. Doing namespace resolution by putting the mapping into the data structures used for constant propagation could overwrite constants with the same variable name as the namespace (not that variables can have the same name as namespaces or even builtin functions). 

Originally, we used the logical names as namespace, but it was later changed to the filenames to address issues with recursive namespace and function resolution. However, down the road (after our SystemML 1.1 release), we should simply maintain this additional meta data in the function directory. Once this is done, we could simply allow namespace functions in eval as well.

PR #750, opened 2018-03-28
[SYSTEMML-2078] Support for global variables

EdgarLGB:
I have added three builtin constant, i.e., PI, INF, NaN. And it supports also global variables.


akchinSTC:

PR #751, opened 2018-03-29
[SYSTEMML-2078] Support for global variables 

EdgarLGB:
Hi, I have cleaned up and closed the last pull request for this issue because I put the old commits in it. And I open this new PR for it.
In this PR, I add the support for global variables and some builtin global constants(PI, INF, NaN). 

Thanks for review

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DMLProgram.java
+++ b/src/main/java/org/apache/sysml/parser/DMLProgram.java
@@ -33,14 +33,44 @@ public class DMLProgram
 	private ArrayList<StatementBlock> _blocks;
 	private HashMap<String, FunctionStatementBlock> _functionBlocks;
 	private HashMap<String,DMLProgram> _namespaces;
+	private VariableSet _globals; // global variables
 	public static final String DEFAULT_NAMESPACE = ".defaultNS";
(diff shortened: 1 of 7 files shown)


akchinSTC:

mboehm7:
thanks for the cleanup @EdgarLGB - in general, this is a good start. However, I would recommend to split the handling of builtin constants and global variables into two PRs and start with the builtin constants.

The current approach prepends assignment statements such `NaN = 0.0/0.0` which does work but could lead to weird behavior because users are still able to assign NaN=7 afterwards - this will likely be very confusing. I would follow R's semantics here and disallow the assignment to NaN and similar constants altogether (R returns the error "Error in NaN = 7 : invalid (do_set) left-hand side to assignment"). You can realize this restriction either by extending the ID specification in our ANTLR grammar (e.g., dml.g4) or by extending the validation of assignment and multi-assignment statements. The latter is perhaps the simplest and thus best approach. Once this is done, you could easily replace NaN, PI, INF with const identifiers during syntactic parsing without the need for variable handling.

EdgarLGB:
Thanks for the review @mboehm7 
I would close this PR and create a new splitted PR for builtin constants.

PR #754, opened 2018-04-02
[SYSTEMML-2078] Add support for global variables

EdgarLGB:
Hi,
Here is a PR for adding support of global variables. We can define variables in global scope or function scope. For a function call, it can reference variables from its parent scope, i.e., global or its superior function scope.

Thanks for review

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DMLProgram.java
+++ b/src/main/java/org/apache/sysml/parser/DMLProgram.java
@@ -33,6 +33,7 @@ public class DMLProgram
 	private ArrayList<StatementBlock> _blocks;
 	private HashMap<String, FunctionStatementBlock> _functionBlocks;
 	private HashMap<String,DMLProgram> _namespaces;
+	private VariableSet _globals; // global variables
 	public static final String DEFAULT_NAMESPACE = ".defaultNS";
 	public static final String INTERNAL_NAMESPACE = "_internal"; // used for multi-return builtin functions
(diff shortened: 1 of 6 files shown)


akchinSTC:

mboehm7:
Sorry for the late response but I would like to play around with this a bit before doing the review. I'll come back to this in a few days.

krishnakalyan3:
@EdgarLGB things look good :). Tested `global-variables-test.dml`. 

I was wondering why you choose to use `+` operator instead of a function call.

krishnakalyan3:
@EdgarLGB thanks for the clarification.

j143:
Hi @Baunsgaard and @kev-inn , would you investigating this PR. 
Is this changeset in line with systemds direction.

cc @EdgarLGB

mboehm7:
Meanwhile, I think not supporting global variables might be better in the long run as it keeps functions (except for explicit reads/writes) side-effect free. Back when discussing this feature, it included two sub problems (1) builtin global constants like NaN, PI, INF, as well as (2) different function scoping rules (with global access). Since (1) has already been introduced (see https://github.com/apache/systemml/commit/4a822a22c88b579e935a1b09051c1fe236e18a4b) and meanwhile we extended the list functionality to pass parameters easier along function call graphs, there is no more a real need for globals. However, I would leave this PR open, in case we come back to it - then this would be an excellent starting point.

PR #757, opened 2018-04-08
[SYSTEMML-2197] Multi-threaded broadcast creation

EdgarLGB:
Hi,
Here is the PR for allowing to create the broadcast in multi-threaded way. After launching the FullDistributedMatrixMultiplicationTest, there are 4 impacted tests (testDenseSparseMapmmSpark, testSparseDenseMapmmSpark, testDenseDenseMapmmSpark, testSparseSparseMapmmSpark) slightly improved in terms of execution time.

Thanks for review


akchinSTC:

mboehm7:
LGTM - thanks @EdgarLGB this is awesome. I made a minor cleanup of replacing the `intstream.range.parallel.foreach` uses with `Arrays.parallelSetAll` (which is implemented exactly as you had it) to improve readability and factored out the common code to avoid redundancy.

PR #760, opened 2018-05-08
[MINOR] Fix native BLAS directory to lower case

EdgarLGB:
Hi,
I tried to set the custom blas directory and found it would be unnecessary to convert the path into lower case.

Thanks for review,
Guobao


akchinSTC:

PR #763, opened 2018-05-16
Replaced tabs with spaces for consistency

JeffMacaluso:
Replaced the tabs on line 949 with spaces for consistency. This prevents conflict with some IDEs causing errors when mixing tabs and spaces. For example:
systemmml example

### DIFF
--- a/src/main/python/systemml/mllearn/estimators.py
+++ b/src/main/python/systemml/mllearn/estimators.py
@@ -946,7 +946,7 @@ class Keras2DML(Caffe2DML):
         self.name = keras_model.name
         createJavaObject(sparkSession._sc, 'dummy')
         if not hasattr(keras_model, 'optimizer'):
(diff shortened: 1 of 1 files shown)


akchinSTC:

krishnakalyan3:
@JeffMacaluso thanks for catching this. It would be really nice if you could check for this kind of inconsitency in other python files.  You could use `pylint` to automatically check for this.

JeffMacaluso:
@krishnakalyan3 That's the only instance I found in my test on running `pylint` for all Python scripts in the repo.

@niketanpansare  You're welcome!

krishnakalyan3:
@niketanpansare yes. Thanks for working on this @JeffMacaluso. LGTM

j143:
@krishnakalyan3 add the author of this PR to whitelist, for jenkins to test this one.

consider copy pasting the following, as the comments.
Add to whitelist
Test this Jenkins

krishnakalyan3:
Test this Jenkins

PR #764, opened 2018-05-17
[Systemml 2084] Implementation of language and compiler extension for paramserv function

EdgarLGB:
Hi @mboehm7 ,

Here is the extension of language and compiler for the new paramserv builtin function. And I also added some basic tests and then will add another test that the paramserv function is included inside parfor after the support of list type for parfor.

Thanks for review,
Guobao


akchinSTC:

src/main/java/org/apache/sysml/hops/ipa/FunctionCallGraph.java
```diff
@@ -64,8 +68,9 @@
 	
 	//subset of direct or indirect recursive functions
 	private final HashSet<String> _fRecursive;
-	
-	private final boolean _containsEval;
+
+	// a boolean value to indicate if exists the second order function (e.g. eval, paramserv)
+	private final boolean _containsSecondOrder;
```
mboehm7: yes, that is a better name - could you please extend the comment to indicate that this also covers UDFs that are marked `secondorder="true"`.

src/main/java/org/apache/sysml/hops/ipa/FunctionCallGraph.java
```diff
@@ -364,9 +369,9 @@ else if (sb instanceof FunctionStatementBlock) {
 						&& ((ExternalFunctionStatement)fsb.getStatement(0)).isSecondOrder() ) {
 						ret = true;
 					}
-				}
-				else if( HopRewriteUtils.isData(h, DataOpTypes.TRANSIENTWRITE)
-					&& HopRewriteUtils.isNary(h.getInput().get(0), OpOpN.EVAL) ) {
+				} else if (HopRewriteUtils.isData(h, DataOpTypes.TRANSIENTWRITE)
+						&& (HopRewriteUtils.isNary(h.getInput().get(0), OpOpN.EVAL)
+						|| HopRewriteUtils.isParameterBuiltinOp(h.getInput().get(0), Hop.ParamBuiltinOp.PARAMSERV))) {
```
mboehm7: Since ParameterizedBuiltinOps can appear anyway in the DAG (not just under the root node) we need a to a DAG traversal to ensure that we catch it in the general case. Please move the probe outside the for loop though to ensure each DAG is traversed just once here.

src/main/java/org/apache/sysml/parser/ParameterizedBuiltinFunctionExpression.java
```diff
@@ -307,7 +314,65 @@ public void validateExpression(MultiAssignmentStatement stmt, HashMap<String, Da
...
 		}
 	}
-	
+
+
+	private void validateParamserv(DataIdentifier output, boolean conditional) {
+		String fname = getOpCode().name();
+		// validate the first five arguments
+		if (getVarParams().size() < 1) {
+			raiseValidateError("Should provide more arguments for function " + fname, conditional, LanguageErrorCodes.INVALID_PARAMETERS);
```
mboehm7: You can make this error unconditional (pass false instead of conditional) as otherwise it would turn into a warning if it is not guaranteed to be executed e.g., in an if branch.

src/main/java/org/apache/sysml/parser/ParameterizedBuiltinFunctionExpression.java
```diff
@@ -307,7 +314,65 @@ public void validateExpression(MultiAssignmentStatement stmt, HashMap<String, Da
...
+		checkDataValueType(true, fname, Statement.PS_BATCH_SIZE, DataType.SCALAR, ValueType.INT, conditional);
+		checkDataValueType(false, fname, Statement.PS_PARALLELISM, DataType.SCALAR, ValueType.INT, conditional);
+		checkStringParam(false, fname, Statement.PS_SCHEME, UtilFunctions.asSet(Statement.PS_SCHEME_DC, Statement.PS_SCHEME_DRR, Statement.PS_SCHEME_DR, Statement.PS_SCHEME_OR), conditional);
+		checkDataValueType(true, fname, Statement.PS_HYPER_PARAMS, DataType.LIST, ValueType.UNKNOWN, conditional);
+		checkStringParam(true, fname, Statement.PS_CHECKPOINTING, UtilFunctions.asSet(Statement.PS_CHECKPOINTING_NONE, Statement.PS_CHECKPOINTING_EPOCH, Statement.PS_CHECKPOINTING_EPOCH10), conditional);
+
+		// set output characteristics
+		output.setDataType(DataType.LIST);
+		output.setValueType(ValueType.UNKNOWN);
+		output.setDimensions(getVarParam(null).getOutput().getDim1(), 1);
```
mboehm7: Set the row dimension to the row dimension of the input parameter Statement.PS_MODEL if specified.

src/main/java/org/apache/sysml/parser/ParameterizedBuiltinFunctionExpression.java
```diff
@@ -663,7 +724,14 @@ private void checkOptionalBooleanParam(Expression param, String name, boolean co
 				+"["+param.getOutput().getValueType()+"].", conditional, LanguageErrorCodes.INVALID_PARAMETERS);
 		}
 	}
-	
+
+	private void checkInvalidParameters(ParameterizedBuiltinFunctionOp op, Set<String> params, Set<String> valid) {
+		Set<String> invalid = params.stream().filter(k -> !valid.contains(k)).collect(Collectors.toSet());
+		if (!invalid.isEmpty())
+			raiseValidateError("Invalid parameters for " + op.name() + ": " + Arrays.toString(invalid.toArray(new String[0])),
+					false);
```
mboehm7: Formatting - please break the line in the middle to make it more readable.

src/main/java/org/apache/sysml/parser/Statement.java
```diff
@@ -61,7 +61,39 @@
...
+	public static final String PS_MODEL = "model";
+	public static final String PS_FEATURES = "features";
+	public static final String PS_LABELS = "labels";
+	public static final String PS_VAL_FEATURES = "val_features";
+	public static final String PS_VAL_LABELS = "val_labels";
+	public static final String PS_UPDATE_FUN = "upd";
+	public static final String PS_AGGREGATION_FUN = "agg";
+	public static final String PS_MODE = "mode";
+	public static final String PS_MODE_LOCAL = "LOCAL";
+	public static final String PS_MODE_REMOTE_SPARK = "REMOTE_SPARK";
```
mboehm7: You might want to define an enum such as PSModeType because we will use that many times.

src/main/java/org/apache/sysml/parser/Statement.java
```diff
@@ -61,7 +61,39 @@
...
+	public static final String PS_UPDATE_TYPE = "utype";
+	public static final String PS_UPDATE_TYPE_BSP = "BSP";
+	public static final String PS_UPDATE_TYPE_ASP = "ASP";
+	public static final String PS_UPDATE_TYPE_SSP = "SSP";
```
mboehm7: Similarly, please convert all options for these parameters into separate enums.

src/main/java/org/apache/sysml/runtime/instructions/SPInstructionParser.java
```diff
@@ -63,6 +63,7 @@
 import org.apache.sysml.runtime.instructions.spark.MultiReturnParameterizedBuiltinSPInstruction;
 import org.apache.sysml.runtime.instructions.spark.PMapmmSPInstruction;
 import org.apache.sysml.runtime.instructions.spark.ParameterizedBuiltinSPInstruction;
+import org.apache.sysml.runtime.instructions.spark.ParamservBuiltinSPInstruction;
```
mboehm7: As discussed before, I would recommend to delete this unnecessary spark instruction. Looking forward, the central CP paramserv function will also allow us later to automatically determine the mode in an automatic manner on execute.

src/main/java/org/apache/sysml/runtime/instructions/spark/ParamservBuiltinSPInstruction.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.sysml.runtime.instructions.spark;
+
+import java.util.HashMap;
+
+import org.apache.sysml.runtime.controlprogram.context.ExecutionContext;
+import org.apache.sysml.runtime.instructions.cp.CPOperand;
+import org.apache.sysml.runtime.matrix.operators.Operator;
+
+public class ParamservBuiltinSPInstruction extends ParameterizedBuiltinSPInstruction {
```
mboehm7: Please remove.

src/test/java/org/apache/sysml/test/integration/functions/paramserv/ParamservFuncTest.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package org.apache.sysml.test.integration.functions.paramserv;
+
+import org.apache.sysml.api.DMLException;
+import org.apache.sysml.test.integration.AutomatedTestBase;
+import org.apache.sysml.test.integration.TestConfiguration;
+import org.junit.Test;
+
+public class ParamservFuncTest extends AutomatedTestBase {
```
mboehm7: Related to the IPA extension, it would be good to add a test where paramserv is not used immediately under a DAG root node. For example, you could so something like `print(length(paramserv()))`.

EdgarLGB:
Thanks @mboehm7 for the reviews. I will go over them.

src/main/java/org/apache/sysml/parser/ParameterizedBuiltinFunctionExpression.java
EdgarLGB: In fact, I want to make a change in the API. I would like to ask the user to always put the parameterized "model=modelList" instead of the non-parameterized "modelList" as the first argument. I think it would be more simple and all the arguments are in the same format.

src/test/java/org/apache/sysml/test/integration/functions/paramserv/ParamservFuncTest.java
EdgarLGB: In fact, I have tried ```print(length(paramserv()))````but it won't add the paramserv dag into the tree. Because I found that the builtin length function could directly get the length of the paramserv output without executing the paramserv function. Except that I try to print the element of the result list where it will add the paramserv dag.

mboehm7: good point - in these situations I usually add a if/else with some data-dependent conditions (e.g., sum(A)>7) and the assignment of lists with different lengths into if and else which causes that the length cannot be evaluated during runtime. You might also need to disable recompilation to actually execute the paramserv function.

EdgarLGB:
@mboehm7 I have pushed some modifications. Overall, I removed the ParamservSPInstruction, added a BFS traversal to probe the paramserv operator and appended a IPA test for the paramserv function.

mboehm7:
LGTM - thanks @EdgarLGB. As an FYI: during the merge I fixed two remaining minor issues (unnecessary import, and unnecessary guava dependency, which we removed in the past due to version conflicts) - something to keep an eye for in the future.

PR #766, opened 2018-05-23
[MINOR] Fix the list invalid lowest index 

EdgarLGB:
Hi,

Here is a patch for fixing the list problem which is found when trying to get the first parameterized argument of the list. And I also added two functions to get the parameterized names which is for paramserv function.

Thanks for review,
Guobao


akchinSTC:

EdgarLGB:
@mboehm7

mboehm7:
LGTM - thanks for catching this issue @EdgarLGB. One nitpick: for these kinds of fixes, let's refrain from reformatting in the future (this also reminded me that we should update our style guides).

PR #767, opened 2018-05-23
Added more test cases for ARIMA

TobiasSchmidtDE:
ARIMA implementation is now tested for all possible combinations of
arguments. However, each argument (e.g. p,d,q,...) has final amount of
possible values.

### DIFF
--- a/src/test/java/org/apache/sysml/test/integration/applications/ArimaTest.java
+++ b/src/test/java/org/apache/sysml/test/integration/applications/ArimaTest.java
@@ -23,14 +23,18 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
-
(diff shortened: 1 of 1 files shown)


akchinSTC:

mboehm7:
Welcome and thanks for catching these issues @TobiasSchmidtDE. Unfortunately, SYSTEMML-2340 did not resolve all issues yet. I'll work on a fix and merge this PR afterwards. In the meantime, it would be great if you could replace the introduced guava dependency with an alternative if possible. In the past, we faced version conflicts with Spark/Hadoop (pulling in different guava versions) and thus, decided to remove this unnecessary dependency.

mboehm7:
LGTM - thanks again @TobiasSchmidtDE - instead of creating all combinations (some of which are invalid), I included a subset of problematic cases. These cases are currently failing due to incorrect loop range handling (i.e., missing increment). Once the ARIMA scripts are fixed, we can enable these tests. Thanks.

PR #769, opened 2018-05-23
Bugfix: No differncing when d/D = 0

TobiasSchmidtDE:
The previous implementation of the differencing made the script crash for d or D equal to 0.
This occurred because ``` for (i in 1:0) {...}``` runs the code of the for loop instead of skipping it. However the code is only supposed to run ```if d > 0```.


akchinSTC:

TobiasSchmidtDE:
Also added implementation with seq for arguments q, p, Q and P.
As soon as for loops are fixed, the test cases introduced in #767 can be activated

mboehm7:
LGTM - thanks @TobiasSchmidtDE. I created SYSTEMML-2342 for everything related to the ARIMA script - feel free to create sub tasks as needed. Also during the merge I made a minor modification to make it work for the additional tests. However, I did not enable these tests yet because we would also need to modify the pydml and R scripts accordingly.

PR #770, opened 2018-05-25
[Systemml 2085] Initial version of local backend for paramserv builtin function

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for the initial version of local backend. To summarize, I implemented the "_BSP_" as well as "_BATCH_" update frequency which you can see in the new added test. In general, the idea is to launch a local ps which maintains a map. And the exchange of intermediates is done by looking up the corresponding key. On one hand, the local workers will work out and push the gradients to ps and then wait for a pending new global parameter pushed by the aggregation service. On the other hand, the aggregation service will wait for the pending gradients and then update and broadcast the updated model. "_Broadcast_" here means to deep copy the model and push them with each worker's "global parameter" key. So each worker will not impact the other's global parameter.

Thanks for the review,
Guobao


akchinSTC:

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalAggregationService.java
```diff
@@ -0,0 +1,151 @@
...
+					aggregate((ListObject) gradients);
+					finished++;
+					tables[i] = true;
+				}
+				try {
+					TimeUnit.SECONDS.sleep(POLL_FREQUENCY);
+				} catch (InterruptedException e) {
+					throw new DMLRuntimeException("Aggregation service: Failed to sleep when polling the gradients.",
+							e);
+				}
```
mboehm7: It should generally not be necessary to poll anything here. This might be an indicator of a suboptimal design - i.e., maybe by merging the aggregation service with the parameter server (or using it inside the parameter server), we get a leaner design where we could simply trigger aggregation whenever a worker pushed new gradients. Also, for multi-threaded environments this polling can often be replaced via wait/notify.

src/test/scripts/functions/paramserv/mnist_lenet_paramserv.dml
```diff
@@ -0,0 +1,349 @@
...
+                   int C, int Hin,
+                   int Win, int Hf,
+                   int Wf, int stride,
+                   int pad, double lambda,
+                   int F1, int F2, int N3,
+                   matrix[double] W1, matrix[double] b1,
+                   matrix[double] W2, matrix[double] b2,
+                   matrix[double] W3, matrix[double] b3,
+                   matrix[double] W4, matrix[double] b4)
+          return (matrix[double] dW1, matrix[double] db1,
```
mboehm7: Why not pass the model and hyper parameters as two lists as we discussed in the design?

src/test/scripts/functions/paramserv/mnist_lenet_paramserv.dml
```diff
@@ -0,0 +1,349 @@
...
+  dW3 = dW3 + dW3_reg
+  dW4 = dW4 + dW4_reg
+}
+
+aggregation = function(matrix[double] W1, matrix[double] b1,
+                       matrix[double] W2, matrix[double] b2,
+                       matrix[double] W3, matrix[double] b3,
+                       matrix[double] W4, matrix[double] b4,
+                       matrix[double] dW1, matrix[double] db1,
+                       matrix[double] dW2, matrix[double] db2,
```
mboehm7: Same as above.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalAggregationService.java
```diff
@@ -0,0 +1,151 @@
...
+		// Populate the variables table with the gradients
+		ParamservUtils.populate(_ec, gradients, _inputs);
+
+		// Invoke the aggregate function
+		FunctionCallCPInstruction inst = new FunctionCallCPInstruction(_funcNS, _funcName, _boundInputs, _inputNames,
+				_outputNames, "aggregate function");
+		inst.processInstruction(_ec);
+
+		// Get the output
+		List<Data> paramsList = _outputs.stream().map(id -> _ec.getVariable(id.getName())).collect(Collectors.toList());
```
mboehm7: Shouldn't the output be a single list object, the updated model?

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalAggregationService.java
```diff
@@ -0,0 +1,151 @@
...
+		ParamservUtils.cleanupListObject(_ec, gradients);
+	}
+
+	/**
+	 * Push the global parameters to ps
+	 */
+	private void pushGlobalParams() {
+		IntStream.range(0, _workerNum).forEach(i -> {
+			// Copy the model
+			ListObject copiedLO = ParamservUtils.copyList(_model);
```
mboehm7: No need to copy anything here due to copy-on-write semantics - i.e., whenever a worker tries to update a model, we create a copy of it first.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
```diff
@@ -0,0 +1,157 @@
...
+				long end = begin + _batchSize;
+				if (j == totalIter - 1) {
+					// If it is the last iteration
+					end = dataSize;
+				}
+
+				// Get batch features and labels
+				MatrixObject bFeatures = ParamservUtils.sliceMatrix(_features, begin, end);
+				MatrixObject bLabels = ParamservUtils.sliceMatrix(_labels, begin, end);
+				_ec.setVariable(Statement.PS_FEATURES, bFeatures);
```
mboehm7: Please do not compile custom DAGs per batch - either compile a parameterized instruction where you beed begin and end as scalar variables or use MatrixBlock.slice directly.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
```diff
@@ -0,0 +1,157 @@
...
+				ParamservUtils.cleanupListObject(_ec, globalParams);
+				ParamservUtils.cleanupListObject(_ec, gradients);
+				ParamservUtils.cleanupData(bFeatures);
+				ParamservUtils.cleanupData(bLabels);
+			}
+			LOG.info(String.format("Local worker_%d: Finished %d epoch.", _workerID, i + 1));
+		}
+
+		LOG.info(String.format("Local worker_%d: Job finished.", _workerID));
+		ParamservUtils.cleanupListObject(_ec, _hyperParams);
```
mboehm7: Again, don't clean up data objects that you did not create.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamServer.java
```diff
@@ -0,0 +1,45 @@
...
+
+	public static final String GLOBAL_PREFIX = "global_";
+	public static final String GRADIENTS_PREFIX = "gradients_";
+	public static final String RESULT_MODEL = "result_model";
+
+	private ConcurrentHashMap<String, Data> _params;
+
+	ParamServer() {
+		this._params = new ConcurrentHashMap<>();
+	}
```
mboehm7: Right now this parameter server is really just a hash map while the aggregation service handles the actual parameter server functionality in a very indirect way.

src/main/java/org/apache/sysml/runtime/instructions/cp/ListObject.java
```diff
@@ -44,30 +46,44 @@ public ListObject(List<Data> data, List<String> names) {
 		_data = data;
 		_names = names;
 	}
-	
+
+	public boolean isParameterizedList() {
```
mboehm7: Please rename to NamedList - a list object can be either an unnamed or a named list.

src/main/java/org/apache/sysml/runtime/instructions/cp/ListObject.java
```diff
@@ -44,30 +46,44 @@ public ListObject(List<Data> data, List<String> names) {
...
+
+	public List<Data> getData() {
+		return _data;
+	}
+
+	public long getDataSize() {
+		Optional<Long> result = _data.stream().filter(data -> data instanceof CacheableData)
+				.map(data -> ((CacheableData) data).getDataSize()).reduce((l1, l2) -> l1 + l2);
+		return result.get();
+	}
```
mboehm7: 1) Note that the getDataSize method only returns the in-memory data size if a matrix/frame is pinned into memory.

2) Simply the reduce.get to a plain sum

src/main/java/org/apache/sysml/runtime/instructions/cp/ListObject.java
```diff
@@ -81,7 +97,34 @@ public Data slice(String name) {
 		//return existing entry
 		return slice(pos);
 	}
-	
+
+	public Data slice(String name, ValueType vt, DataType dt) {
```
mboehm7: This method should not be necessary because the function class does any necessary value type casts automatically. So, let's please neither duplicate this logic nor create unnecessary copies here.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -19,23 +19,264 @@
...
+		List<MatrixObject> pvls = disjointContiguous(valLabels);
+		for (int i = 0; i < workers.size(); i++) {
+			workers.get(i).setValFeatures(pvfs.get(i));
+			workers.get(i).setValLabels(pvls.get(i));
+		}
+	}
+
+	private List<MatrixObject> disjointContiguous(MatrixObject mo) {
+		List<MatrixObject> list = new ArrayList<>();
+		long stepSize = mo.getNumRows() / _workerNum;
```
mboehm7: Please use a Math.ceil - otherwise you would get more partitions than workers or miss the last batch.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -19,23 +19,264 @@
...
+		long begin = 1;
+		while (begin < mo.getNumRows()) {
+			long end = begin + stepSize;
+			if (end > mo.getNumRows()) {
+				end = mo.getNumRows();
+			}
```
mboehm7: Simplify to min(begin+stepSize, mo.getNumRows)

src/test/java/org/apache/sysml/test/integration/functions/paramserv/ParamservFuncTest.java
```diff
@@ -32,7 +32,9 @@
 	private static final String TEST_NAME4 = "paramserv-wrong-type-args";
 	private static final String TEST_NAME5 = "paramserv-wrong-args";
 	private static final String TEST_NAME6 = "paramserv-wrong-args2";
-	private static final String TEST_NAME7 = "paramserv-ipa-test";
+	private static final String TEST_NAME7 = "paramserv-wrong-list-test";
+	private static final String TEST_NAME8 = "paramserv-nn-test";
+	private static final String TEST_NAME9 = "paramserv-minimum-version";
```
mboehm7: When I executed these tests, test 8 and 9 were still failing - could you please have a look?

src/main/java/org/apache/sysml/runtime/instructions/cp/ListObject.java
```diff
@@ -21,18 +21,20 @@
...
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.sysml.parser.Expression.DataType;
 import org.apache.sysml.parser.Expression.ValueType;
 import org.apache.sysml.runtime.DMLRuntimeException;
+import org.apache.sysml.runtime.controlprogram.caching.CacheableData;
 
 public class ListObject extends Data {
```
mboehm7: Please remove all introduced warnings related to raw types and missing switch cases.

EdgarLGB:
Thanks @mboehm7 for the review. I will go over your points and make another PR for the aggregation service and parameter server.

PR #772, opened 2018-05-30
[MINOR] include applications integration test suite

bertholdreinwald:
[MINOR] including applications test suite


akchinSTC:

akchinSTC:

PR #774, opened 2018-05-30
[SYSTEMML-445] Major refactoring of the GPU Memory Manager

niketanpansare:
- Several bugfixes found during recent experiments with ResNet200.
- Added align_memory eviction policy
- Added GPU usage documentation
- Refactored the GPU Memory Manager into distinct components

@nakul02 Can you please review this PR? This PR only deals with refactoring during the recent ResNet200 experiments and doesnot contain LSTM or Batch Normalization kernel. Once this PR is in, it will simplify the LSTM PR.


akchinSTC:

src/main/cpp/kernels/SystemML.cu
```diff
@@ -1961,3 +1961,100 @@ extern "C" __global__ void matrix_sigmoid_f(float *A, float *C,
                                          unsigned int size) {
   matrix_sigmoid(A, C, size);
 }
+
+// We can later fold it in our reduce method
+template <typename T>
+__device__ void compute_nnz(
```
nakul02: This should reuse our templatized reduction method, as you mentioned in the comment above.

niketanpansare:
Thanks @nakul02 ... I will merge. We can reuse our templatized reduction in a differen PR.

PR #777, opened 2018-06-02
[SYSTEMML-2344, 2348, 2349, 2352] Improvement for the initial local paramserv builtin

EdgarLGB:
Hi @mboehm7 ,

Here is some improvement for the initial local backend for the paramserv function. In general, I rewrite the ps with blocking queue instead of the classical lock. And also the error handler is added to each thread. At last, I added the preparation of execution context for workers and the agg service.

Thanks for the review,
Guobao


src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalParamServer.java
```diff
@@ -24,36 +24,37 @@
...
 	}
 
 	@Override
 	public void push(long workerID, ListObject gradients) {
-		synchronized (_lock) {
-			_queue.add(new Gradient(workerID, gradients));
-			_lock.notifyAll();
+		try {
+			_gradientsQueue.put(new Gradient(workerID, gradients));
+			launchService();
```
mboehm7: It seems that there is actually no need to run the aggregation service as a separate thread. The single-threaded thread pool provides you serialization of incoming gradients. You might be able to simple call the code of the run method as a synchronized method (serialized on the parameter server instance).

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
```diff
@@ -50,66 +48,68 @@
 	private MatrixObject _valFeatures;
 	private MatrixObject _valLabels;
 
-	ArrayList<DataIdentifier> _outputs;
+	DataIdentifier _output;
 	FunctionCallCPInstruction _inst;
```
mboehm7: Please use private or protected and final where appropriate.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
```diff
@@ -50,66 +48,68 @@
...
 		this._updFunc = updFunc;
 		this._freq = freq;
 		this._epochs = epochs;
 		this._batchSize = batchSize;
-		this._ec = ExecutionContextFactory.createContext(ec.getProgram());
-		if (hyperParams != null) {
-			this._ec.setVariable(Statement.PS_HYPER_PARAMS, hyperParams);
-		}
+		this._ec = ec;
 		this._ps = ps;
```
mboehm7: Avoid unnecessary `this.` - the naming schemes of a "_"-prefix for instance members makes this redundant.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamServer.java
```diff
@@ -180,52 +197,65 @@ private void setFinishedState(int workerID) {
...
+							String.format("Param server: failed to broadcast the model for worker_%d", id), e);
+				}
+				return q;
+			}));
+		}
+
 		@Override
 		public void run() {
-			synchronized (_lock) {
-				while (isAlive()) {
```
mboehm7: Please remove unused methods such as `isAlive()`; in case you need it later, we would need to make the alive flag volatile to ensure that updates are seen from all threads (unless used in a synchronized which implicitly creates a memory barrier as well).

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalParamServer.java
```diff
@@ -24,36 +24,37 @@
...
+		} catch (InterruptedException e) {
+			throw new DMLRuntimeException(
+					String.format("Local param server: failed to push the gradients for worker_%d.", workerID), e);
 		}
```
mboehm7: I leave it up to you but there is no need for explicit exception handling for each queue interaction. You have a surrounding exception handling and the issue would simply show up in the stack trace.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamServer.java
```diff
@@ -180,52 +197,65 @@ private void setFinishedState(int workerID) {
 			_finishedStates[workerID] = true;
 		}
 
+		private void broadcastModel() {
+			IntStream.range(0, _finishedStates.length).forEach(i -> _modelMap.compute(i, (id, q) -> {
+				if (q == null) {
+					q = new ArrayBlockingQueue<>(1);
+				}
```
mboehm7: No need to create a new queue here - please simply put the new model into the queue.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamServer.java
```diff
@@ -180,52 +197,65 @@ private void setFinishedState(int workerID) {
...
+				try {
+					q.put(_model);
+				} catch (InterruptedException e) {
+					throw new DMLRuntimeException(
```
mboehm7: I usually use streams only when the called methods throw runtime exceptions - for other exceptions a for loop without explicit exception handling is probably simpler.

src/test/scripts/functions/paramserv/mnist_lenet_paramserv_agg_service_failed.dml
```diff
@@ -0,0 +1,379 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+/*
+ * MNIST LeNet Example
+ */
```
mboehm7: For these error tests, it would be sufficient to test on toy examples instead of full application level scripts. You can leave them as they are if you like, just something to keep in mind for the future.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -95,44 +150,92 @@ public void processInstruction(ExecutionContext ec) {
...
 
 		// Do data partition
 		doDataPartition(ec, workers);
 
 		// Create the worker threads
-		List<Thread> threads = workers.stream().map(Thread::new).collect(Collectors.toList());
+		workers.forEach(w -> CompletableFuture.runAsync(w, _es).exceptionally(handler));
+
+		// Wait for the worker finishing
+		_es.shutdown();
```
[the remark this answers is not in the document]
EdgarLGB: Ok, it will be more simple to handle the error in try/catch. And in fact, I'd like to leverage the feature of CompletableFuture which can combine all the futures instead of joining the thread one by one so that we could avoid to have some blocking case, i.e., if the second worker failed, maybe the main thread will always be blocked in joining the first worker but the first worker as well as the agg service are also blocked due to some reason. Thus, it will fall into a case never stopped.
Alternatively, I'd like to use:
```
		try {
			CompletableFuture[] futures = es.invokeAll(workers).stream().map(ParamservUtils::makeCompletableFuture)
					.toArray(CompletableFuture[]::new);
			CompletableFuture<Void> combinedFuture = CompletableFuture.allOf(futures);
			combinedFuture.get();
		} catch (InterruptedException | ExecutionException e) {
			throw new DMLRuntimeException("ParamservBuiltinCPInstruction: some error occurred: ", e);
		} finally {
			// shutdown the workers thread pool
			es.shutdownNow();
		}

```

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
```diff
@@ -31,67 +32,70 @@
...
+		try {
+			long dataSize = _features.getNumRows();
 
-		long dataSize = _features.getNumRows();
+			for (int i = 0; i < _epochs; i++) {
+				int totalIter = (int) Math.ceil(dataSize / _batchSize);
+				for (int j = 0; j < totalIter; j++) {
+					// Pull the global parameters from ps
+					// Need to copy the global parameter
+					ListObject globalParams = ParamservUtils.copyList((ListObject) _ps.pull(_workerID));
```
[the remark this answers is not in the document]
EdgarLGB: Well, I think the problem is that agg service will broadcast the references of the model to all the workers. And then when the worker finish its iteration, it will destruct this pulled model which is also referenced by other workers. Hence, to avoid impacting the global model in other workers, I'd like to copy the global model in each worker so that it can have its own local model.

EdgarLGB:
Thanks @mboehm7 for the code review. I'm going through the feedback and will make another PR consisting of the implementation of ASP and EPOCH.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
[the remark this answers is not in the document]
mboehm7: Thanks for the clarification - the function call and list indexing should already take care of disabling the cleanup. If this is not working then it simply needs fixing. I'll have a look during the merge.

mboehm7:
Thanks for incorporating the comments - I'll make another pass later today. Also, just for the future, you don't need to close and reopen your PRs - you could simply push additional commits onto this branch.

mboehm7:
LGTM - thanks @EdgarLGB this is awesome. There are a couple minor cleanups as well as a fix for race conditions on model update remaining but I'll do these during the merge.

PR #779, opened 2018-06-04
Arima cleanup

TobiasSchmidtDE:
The complete refactoring of this script is not done yet. However this should be a stable version, that can already be merged back into the master


mboehm7:
Apparently you used two different accounts here - under which account do you want his PR to be merged?

TobiasSchmidtDE:
Oh sorry, I didn't notice. Please merge under the TobiasSchmidtDE account, that is the one associated to my IBM Mail

mboehm7:
LGTM - thanks @TobiasSchmidtDE. For now, I will also keep the old dml script to ensure we're not breaking the tests. I already added (but commented) the modified named parameter handling in `ArimaTest`. Apart from that, three issues showed up - I fixed the major one in SYSTEMML-2361 and addressed the other two via minor workarounds at script level. Once SYSTEMML-2362 and SYSTEMML-2363 are fixed, we can remove the workarounds again.

PR #780, opened 2018-06-04
[SYSTEMML-2359] Extend the epoch update frequency

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for extension of epoch update frequency. At the same time, I cleaned up some dml script so that we could play with the combination of update strategy and frequency.

Thanks for review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB. Overall, this is a very elegant and useful extension. As the number (and diversity) of tests increases, you might want to split the tests into separate classes in the future (e.g., failures, frequency, etc). Furthermore, I had to change the degree of parallelism of `paramserv-worker-failed` and `paramserv-agg-service-failed` again to 1 because otherwise the returned error message it not as expected - this indicates some side effect between workers, which we will have to address separately.

PR #781, opened 2018-06-08
[SYSTEMML-2364, 66, 67, 68] Extend the data partition schemes

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for extending the data partition schemes (Disjoint_Round_Robin, Disjoint_Random, Overlap_Reshuffle).

Thanks for review,
Guobao


src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerDR.java
```diff
@@ -0,0 +1,56 @@
...
+		// Generate the permutation
+		MatrixObject permutation = ParamservUtils.generatePermutation(mo, ec);
+
+		// Slice the original matrix and make data partition by permutation multiply
+		AggregateOperator agg = new AggregateOperator(0, Plus.getPlusFnObject());
+		AggregateBinaryOperator aggbin = new AggregateBinaryOperator(Multiply.getMultiplyFnObject(), agg, 1);
+		AggregateBinaryCPInstruction multiInst = new AggregateBinaryCPInstruction(aggbin,
+				new CPOperand("permutation", Expression.ValueType.DOUBLE, Expression.DataType.MATRIX),
+				new CPOperand("data"), new CPOperand("result", Expression.ValueType.DOUBLE, Expression.DataType.MATRIX),
+				"ba+*", "permutation multiply");
```
mboehm7: Since you're anyway calling the CP, i.e., in-memory instructions here, you could directly call matrix block operations instead to simplify this code. For example, you can use the following for indexing and matrix multiply:

```
AggregateBinaryOperator op = InstructionUtils.getMatMultOperator(k);
MatrixBlock mPermk = mPerm.slice(begin-1, end-1, 0, mPerm.getNumColumns()-1, new MatrixBlock());
MatrixBlock out = mPermk.aggregateBinaryOperations(mPermk, data, new MatrixBlock(), op);

```

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerDRR.java
```diff
@@ -0,0 +1,43 @@
+package org.apache.sysml.runtime.controlprogram.paramserv;
```
mboehm7: Missing apache license header.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerDRR.java
```diff
@@ -0,0 +1,43 @@
...
+public class DataPartitionerDRR extends DataPartitioner {
+	@Override
+	public List<MatrixObject> doPartition(int k, MatrixObject mo) {
+		return IntStream.range(0, k).mapToObj(i -> removeEmpty(mo, k, i)).collect(Collectors.toList());
+	}
+
+	private MatrixObject removeEmpty(MatrixObject mo, int k, int workerId) {
+		MatrixObject result = new MatrixObject(Expression.ValueType.DOUBLE, null,
+				new MetaDataFormat(new MatrixCharacteristics(-1, -1, -1, -1), OutputInfo.BinaryBlockOutputInfo,
+						InputInfo.BinaryBlockInputInfo));
```
mboehm7: Use you `ParamservUtils.newMatrixObject()` to avoid code duplication.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerOR.java
```diff
@@ -0,0 +1,45 @@
+package org.apache.sysml.runtime.controlprogram.paramserv;
```
mboehm7: Missing apache license header.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerOR.java
```diff
@@ -0,0 +1,45 @@
...
+	public List<MatrixObject> doPartition(int k, MatrixObject mo) {
+		ExecutionContext ec = ExecutionContextFactory.createContext();
+		ec.setVariable("data", mo);
+
+		AggregateOperator agg = new AggregateOperator(0, Plus.getPlusFnObject());
+		AggregateBinaryOperator aggbin = new AggregateBinaryOperator(Multiply.getMultiplyFnObject(), agg, 1);
+		AggregateBinaryCPInstruction multiInst = new AggregateBinaryCPInstruction(aggbin,
+				new CPOperand("permutation", Expression.ValueType.DOUBLE, Expression.DataType.MATRIX),
+				new CPOperand("data"), new CPOperand("result", Expression.ValueType.DOUBLE, Expression.DataType.MATRIX),
+				"ba+*", "permutation multiply");
```
mboehm7: As mentioned in a comment above, you can directly call the block operations here to avoid this unnecessarily convoluted setup.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamservUtils.java
```diff
@@ -77,6 +83,12 @@ public static void cleanupData(Data data) {
 		cd.clearData();
 	}
 
+	public static MatrixObject newMatrixObject() {
+		return new MatrixObject(Expression.ValueType.DOUBLE, null,
```
mboehm7: Please use `OptimizerUtils.getUniqueTempFileName()` instead of null for the filename.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/DataPartitionerOR.java
```diff
@@ -0,0 +1,45 @@
...
+
+		return IntStream.range(0, k).mapToObj(i -> {
+			// Generate the permutation
+			MatrixObject permutation = ParamservUtils.generatePermutation(mo, ec);
```
mboehm7: The permutation matrix is not used at all here. Please add a junit tests that calls all data partitioners and checks for correct outputs.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamservUtils.java
```diff
@@ -86,14 +98,44 @@ public static void cleanupData(Data data) {
...
+		MatrixObject seq = ParamservUtils.newMatrixObject();
+		seq.acquireModify(seqMB);
+		seq.release();
+		ec.setVariable("seq", seq);
+
+		// Generate a sample
+		DataGenCPInstruction sampleInst = new DataGenCPInstruction(null, Hop.DataGenMethod.SAMPLE, null,
+				new CPOperand("sample", Expression.ValueType.DOUBLE, Expression.DataType.MATRIX),
+				new CPOperand(String.valueOf(mo.getNumRows()), Expression.ValueType.INT, Expression.DataType.SCALAR,
+						true), new CPOperand("1", Expression.ValueType.INT, Expression.DataType.SCALAR, true),
```
mboehm7: You may directly call `MatrixBlock.sampleOperations` and `ternaryOperations` for sample and table.

src/main/java/org/apache/sysml/runtime/instructions/cp/AggregateBinaryCPInstruction.java
```diff
@@ -34,7 +34,8 @@
 
 public class AggregateBinaryCPInstruction extends BinaryCPInstruction {
 
-	private AggregateBinaryCPInstruction(Operator op, CPOperand in1, CPOperand in2, CPOperand out, String opcode, String istr) {
+	public AggregateBinaryCPInstruction(Operator op, CPOperand in1, CPOperand in2, CPOperand out, String opcode,
+			String istr) {
```
mboehm7: Please don't introduce such a corrupted formatting (only break the line if there are sufficiently many inputs and try to break lines evenly).

src/main/java/org/apache/sysml/runtime/instructions/cp/CtableCPInstruction.java
```diff
@@ -40,7 +40,7 @@
 	private final boolean _isExpand;
 	private final boolean _ignoreZeros;
 
-	private CtableCPInstruction(CPOperand in1, CPOperand in2, CPOperand in3, CPOperand out,
```
mboehm7: Revert these changes to the visibility if you call the block operations directly.

src/main/java/org/apache/sysml/runtime/instructions/cp/CtableCPInstruction.java
```diff
@@ -40,7 +40,7 @@
...
+	public CtableCPInstruction(CPOperand in1, CPOperand in2, CPOperand in3, CPOperand out,
```
mboehm7: Same as above.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -381,26 +405,17 @@ private void doDataPartition(ExecutionContext ec, List<LocalPSWorker> workers) {
...
 	}
 
-	private void disjointContiguous(MatrixObject features, MatrixObject labels, MatrixObject valFeatures,
+	private void doDataPartition(DataPartitioner dp, MatrixObject features, MatrixObject labels, MatrixObject valFeatures,
 			MatrixObject valLabels, List<LocalPSWorker> workers) {
-		// training data
-		List<MatrixObject> pfs = disjointContiguous(workers.size(), features);
-		List<MatrixObject> pls = disjointContiguous(workers.size(), labels);
+		List<MatrixObject> pfs = dp.doPartition(workers.size(), features);
+		List<MatrixObject> pls = dp.doPartition(workers.size(), labels);
```
mboehm7: This separate partitioning worked well for disjoint contiguous because it is a deterministic partitioning scheme. However, for random reshuffling (disjoint random and overlap reshuffle) this creates two different permutation matrices and thus mixes observations and labels arbitrarily. In general, we would need to reuse the same permutation for features and labels. Hence, I would recommend to extend the abstraction of `DataPartitioner` accordingly.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -381,26 +405,17 @@ private void doDataPartition(ExecutionContext ec, List<LocalPSWorker> workers) {
...
 		if (pfs.size() < workers.size()) {
 			if (LOG.isWarnEnabled()) {
 				LOG.warn(String.format("There is only %d batches of data but has %d workers. "
-					+ "Hence, reset the number of workers with %d.", pfs.size(), workers.size(), pfs.size()));
+						+ "Hence, reset the number of workers with %d.", pfs.size(), workers.size(), pfs.size()));
```
mboehm7: Minor nitpick (because I've seen this in previous PRs): the indentation for single line breaks is one not two tabs.

src/main/java/org/apache/sysml/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -410,24 +425,11 @@ private void disjointContiguous(MatrixObject features, MatrixObject labels, Matr
 		}
 
 		// validation data
-		List<MatrixObject> pvfs = disjointContiguous(workers.size(), valFeatures);
-		List<MatrixObject> pvls = disjointContiguous(workers.size(), valLabels);
+		List<MatrixObject> pvfs = dp.doPartition(workers.size(), valFeatures);
+		List<MatrixObject> pvls = dp.doPartition(workers.size(), valLabels);
```
mboehm7: Sorry for not catching this before - we don't need to partition the validation data at all. All workers should get the same read-only validation data.

EdgarLGB:
Thanks @mboehm7 for the code review. I've added some modification and hoping to have some feedback. Thanks!

mboehm7:
LGTM - thanks @EdgarLGB this is a very nice extensions. Just for the record here are the changes I made during the merge
* Fixed DataPartitionerDRR to produce partitions of size 1/k * rows. The issues was `l % k == workerId ? 0 : 1` instead of `workerId ? 1 : 0`, which created partitions of size (k-1)/k. All tests used two workers where this does not show up because it both result in the same sizes.
* Fixed DataPartitionerOR to use different permutation matrices for each worker. Before every worker received a copy of exactly the same reshuffled data. I refactored the hierarchy of DataPartitionerDR and DataPartitionerDR accordingly.
* Fixed tests by moving the unit tests into the paramserv package, included it into the test suite, and fixed the DRR tests. However, I left additional TODOs which we should address in a subsequent OR.
* Minor cleanups (e.g., simplified slice, sequence operations instead of streams for performance). Also sorry for my wrong comment regarding ternary vs ctable operations - we just recently renamed ternary to ctable to make room for general-purpose ternary operations.

PR #782, opened 2018-06-11
[SYSTEMML-2380] Should shutdown the thread pool of agg service

EdgarLGB:
Hi @mboehm7 ,

Here is a patch for fixing the problem of hanging at the end of the paramserv dml script execution. In general, the reason is that the thread pool of agg service is not shutdowned at the end of execution.

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB for catching this issue.

PR #783, opened 2018-06-12
[SYSTEMML-2381] Reorganise the API of data partitioner

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for reorganising the API of data partitioner by removing the workers as arguments and also fixing the unit test.

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB for the cleanup and fixing the tests.

PR #784, opened 2018-06-13
fixed nullpointer exception when sparseBlock of matrix is null

johannaSommer:


src/main/java/org/apache/sysml/hops/estim/EstimatorBitsetMM.java
```diff
@@ -29,118 +29,116 @@
...
-		
+
 		public long getNonZeros() {
 			return _nonZeros;
 		}
-		
+
 		private void init(MatrixBlock in) {
-			if( in.isInSparseFormat() ) {
+			if (in.isInSparseFormat() && in.isEmptyBlock(false)) {
```
mboehm7: This would lead to all empty (i.e., not allocated blocks) blocks enter this branch and subsequently fail on accessing the sparse block. Similarly non-empty sparse blocks would enter the dense case and also fail. Please move the check for empty blocks up and directly return in this case.

mboehm7:
LGTM - thanks again for catching and fixing this @johannaSommer.

PR #785, opened 2018-06-13
[SYSTEMML-2389] Fix the calculation of total iteration number for workers

EdgarLGB:
Hi @mboehm7 ,

Here is a PR for fixing the bug of calculating the total iteration number.

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks for catching this (which applies to all datasets unaligned with the batchsize).

PR #787, opened 2018-06-17
Rework the recompilation and fix the epoch strategy

EdgarLGB:
Hi @mboehm7 ,

I'd like to create a new PR due to some new different commits. In general, I reused some logic of parfor parallelism recompilation and fix an important problem wrt epoch updating strategy.  The idea of this fix is to accumulate the intermediate gradients generated in each worker in one epoch and push the gradients of this accumulated one.

Thanks for the review,
Guobao

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
@@ -23,11 +23,14 @@ import java.util.concurrent.Callable;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.parser.Statement;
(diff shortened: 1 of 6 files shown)


mboehm7:
LGTM - thanks @EdgarLGB for catching and fixing this issue. The rework for update per epoch is absolutely right - sorry for not catching this during the previous PRs. During the merge I made some additional improvements, which mostly address performance issues:
* SYSTEMML-2398: The reason why we've seen such a large aggregation overhead w/ 80 workers, was because the `updateModel` method was synchronized on the single `AggregationService` instance, which led to a serialization of all local and global model updates and thus heavy thread contention. We now use separate execution context and unsynchronized model updates which improved the overhead for ASP EPOCH 2 epochs 80 workers from `3179.774/63.368 secs` (local/global) to `123.369/9.508 secs`. 
* Furthermore, we now use inplace binary operations for accruing the local gradients to avoid unnecessary reallocations. 
* I also made all time measurements conditional on enabled statistics because for very small operations the time measurement itself can become a bottleneck.

PR #789, opened 2018-06-20
Improve the function copy

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for improving the update and aggregation function copy. In general, I create a new program and put the needed function respectively for workers and agg service. And also, I think that the dags will be always deep copied when they need to be recompiled during runtime. And then I fixed the batchsize of validation function.

Thanks for the review,
Guobao

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
@@ -19,7 +19,7 @@
 
 package org.apache.sysml.runtime.controlprogram.paramserv;
 
-import static org.apache.sysml.runtime.controlprogram.paramserv.ParamservUtils.UPDATE_FUNC_PREFIX;
+import static org.apache.sysml.runtime.controlprogram.paramserv.ParamservUtils.PS_FUNC_PREFIX;
 
 import java.util.ArrayList;
 import java.util.stream.Collectors;
@@ -65,7 +65,7 @@ public abstract class PSWorker {
 		_ps = ps;
 
 		// Get the update function
-		String[] cfn = ParamservUtils.getCompleteFuncName(updFunc, UPDATE_FUNC_PREFIX + workerID + "_");
+		String[] cfn = ParamservUtils.getCompleteFuncName(updFunc, PS_FUNC_PREFIX);
 		String ns = cfn[0];
 		String fname = cfn[1];
 		FunctionProgramBlock func = ec.getProgram().getFunctionProgramBlock(ns, fname);
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
@@ -24,7 +24,6 @@ import static org.apache.sysml.runtime.controlprogram.paramserv.ParamservUtils.U
 import java.util.ArrayList;
 import java.util.stream.Collectors;
 
-import org.apache.sysml.parser.DMLProgram;
 import org.apache.sysml.parser.DataIdentifier;
 import org.apache.sysml.parser.Expression;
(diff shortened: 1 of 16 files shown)


src/test/scripts/functions/paramserv/mnist_lenet_paramserv_minimum_version.dml
```diff
@@ -295,9 +295,8 @@ predict = function(matrix[double] X, int C, int Hin, int Win,
 
   # Compute predictions over mini-batches
   probs = matrix(0, rows=N, cols=K)
-  batch_size = 64
   iters = ceil(N / batch_size)
-  for(i in 1:iters) {
+  parfor(i in 1:iters, check=0) {
```
mboehm7: That is great - on a scenario of mnist60K, 2 epochs per worker, ASP, EPOCH, batchsize=64, 80 workers, this change alone change the end-to-end runtime from 61s to 36s.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/PSWorker.java
```diff
@@ -66,14 +65,10 @@ protected PSWorker(int workerID, String updFunc, Statement.PSFrequency freq, int
...
 		// Get the update function
-		String[] keys = DMLProgram.splitFunctionKey(updFunc);
-		String funcName = keys[0];
-		String funcNS = null;
-		if (keys.length == 2) {
-			funcNS = keys[0];
-			funcName = keys[1];
-		}
-		FunctionProgramBlock func = ec.getProgram().getFunctionProgramBlock(funcNS, UPDATE_FUNC_PREFIX + _workerID + "_" + funcName);
+		String[] cfn = ParamservUtils.getCompleteFuncName(updFunc, UPDATE_FUNC_PREFIX + workerID + "_");
```
mboehm7: I think this naming is the source of the misunderstanding. When we talked about the "update" function I was not referring to the "gradient" function per worker (currently called update function), but the "aggregate" function of the parameter server (which we call "model update time" in the statistics output). For the update strategy per EPOCH, all workers share the same "aggregate" function. The recompiler synchronizes any function or DAG recompilation on the program blocks and DAGs respectively in order to ensure isolation for all scenarios. I would recommend to give each worker (at least for updates per EPOCH) a separate update (i.e., aggregate) function.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/ParamservUtils.java
```diff
@@ -131,7 +133,22 @@ public static MatrixBlock generatePermutation(int numEntries) {
...
 
-	public static ExecutionContext createExecutionContext(ExecutionContext ec, String updFunc, String aggFunc, int workerNum, int k) {
+	public static String[] getCompleteFuncName(String funcName, String prefix) {
+		String[] keys = DMLProgram.splitFunctionKey(funcName);
+		String namespace = null;
+		String func = keys[0];
+		if (keys.length == 2) {
+			namespace = keys[0];
+			func = keys[1];
+		}
```
mboehm7: You can simplify that because we're only dealing with out own function key - so there should be no need to handle empty namespaces.
```
return StringUtils.isEmpty(prefix) ? keys :
	new String[] { keys[0], prefix + keys[1] };
```

EdgarLGB: Well, in my negative test, I used some dummy update function which is in the empty namespace. Hence, I'd like to keep it in order to handle this case.

EdgarLGB:
Thanks @mboehm7 for the code review. I have pushed some modification related to the feedback.

mboehm7:
just as an FYI - I also fixed the {{ParamservRuntimeNegativeTest}} as I changed the list indexing error messages, when introducing right indexing on lists yesterday.

PR #790, opened 2018-06-23
[SYSTEMML-2416] Use synchronized method instead of single thread pool

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for leveraging the synchronized update method instead of the callable method. I think that we do not need the single thread pool because the agg service is always invoked in a synchronized manner.

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB. This is great. There wasn't a significant change in performance (i.e., fluctuations larger than difference) - in these situations, we should always prefer the simpler approach. During the merge, I additionally flattened the aggregation service into the parameter server to reduce unnecessary indirections.

PR #791, opened 2018-06-23
[SYSTEMML-2403] Fix the occurred low accuracy 

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for fixing the low accuracy occurred in the BATCH strategy. In general, the idea is to accrue the gradients in the ps side for the BATCH strategy just as what has been done for the local worker's accrued gradients. In this way, there is less computation to do in ps side and it improved slicely the performance. And then there is no more low accuracy occurred in my env. So I feel that something went wrong when intensely calling the agg function. But I don't know why.

Thanks for the review,
Guobao

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
+++ b/src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/LocalPSWorker.java
@@ -20,7 +20,6 @@
 package org.apache.sysml.runtime.controlprogram.paramserv;
 
 import java.util.concurrent.Callable;
-import java.util.stream.IntStream;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


mboehm7:
Thanks @EdgarLGB - in general, this LGTM. However, before merging it in, we need to understand why this was happening. I could imagine it's related to the used optimizer (SGD nesterov here) but I need to run a couple of experiments first. 

In the meantime, I would recommend to focus on the design of the distributed spark backend and it's break down into individual tasks. You should not be blocked by this issue here, right?

EdgarLGB:
@mboehm7 sure, I'm working on it.

mboehm7:
LGTM - thanks @EdgarLGB. Although I was not able to reproduce the accuracy issue (probably fixed by a previous patch that removed side effects), it's safe to use this patch as it is also beneficial for performance. During the merge, I added an internal flag such that we can play around with this configuration if needed, and parallelized the accrueGadients method to reduce thread contention in BSP.

PR #793, opened 2018-06-29
[SYSTEMML-2418] Add spark distributed data partitioner

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for adding the spark distributed data partitioner as well as the unit test for it. And I also made some refactoring on the local data partitioner in order to be reused in spark distributed context.

Thanks for the review,
Guobao


mboehm7:
thanks @EdgarLGB - I'm traveling right now but will provide feedback later next week.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/spark/DataPartitionerSparkMapper.java
```diff
@@ -0,0 +1,82 @@
...
+	 * @return WorkerID -> partitioned "features, labels"
+	 * @throws Exception Some exception
+	 */
+	@Override
+	public Iterator<Tuple2<Integer, Tuple2<MatrixBlock, MatrixBlock>>> call(Tuple2<Long,Tuple2<MatrixBlock,MatrixBlock>> input)
+			throws Exception {
+		List<Tuple2<Integer, Tuple2<MatrixBlock, MatrixBlock>>> partitions = new LinkedList<>();
+		MatrixBlock features = input._2._1;
+		MatrixBlock labels = input._2._2;
+		DataPartitionSparkScheme.Result result = _scheme.doPartitioning(_workersNum, features, labels);
```
mboehm7: Following the above comment, we would need to extend all the spark partitioning schemes to properly return matrix indexes and matrix blocks, potentially shifted to the target location.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/spark/DataPartitionerSparkReducer.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class DataPartitionerSparkReducer implements Function2<Tuple2<MatrixBlock, MatrixBlock>, Tuple2<MatrixBlock, MatrixBlock>, Tuple2<MatrixBlock, MatrixBlock>> {
+
+	private static final long serialVersionUID = -6581711018365107364L;
+
+	@Override
+	public Tuple2<MatrixBlock, MatrixBlock> call(Tuple2<MatrixBlock, MatrixBlock> input1, Tuple2<MatrixBlock, MatrixBlock> input2) throws Exception {
+		MatrixBlock features = ParamservUtils.rbindMatrix(input1._1, input2._1);
+		MatrixBlock labels = ParamservUtils.rbindMatrix(input1._2, input2._2);
+		return new Tuple2<>(features, labels);
```
mboehm7: Since reduceByKey assumes associative and commutative functions, these blocks are combined in arbitrary order depending on their initial location (e.g., blocks on a given node are combined first before they are shuffled and combined with blocks from other nodes).

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/spark/ORSparkScheme.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	@Override
+	public Result doPartitioning(int workersNum, MatrixBlock features, MatrixBlock labels) {
+		// Generate a different permutation matrix for each worker
+		List<MatrixBlock> permutations = IntStream.range(0, workersNum).mapToObj(
+				i -> ParamservUtils.generatePermutation((int) features.getNumRows())).collect(Collectors.toList());
+		List<MatrixBlock> pfs = ORLocalScheme.partition(workersNum, features, permutations);
+		List<MatrixBlock> pls = ORLocalScheme.partition(workersNum, labels, permutations);
+		return new Result(pfs, pls);
```
mboehm7: This implementation only shuffles rows per block. Instead we need global randomization (i.e., across blocks). Maybe you could create a global permutation vector (source-target row ids), broadcast this likely very small vector and shuffle the blocks accordingly. This requires that the result will hold potentially many MatrixIndexes-MatrixBlock pairs per input block.

src/main/java/org/apache/sysml/runtime/controlprogram/paramserv/spark/DRSparkScheme.java
```diff
@@ -0,0 +1,50 @@
...
+		// No-args constructor used for deserialization
+	}
+
+	@Override
+	public Result doPartitioning(int workersNum, MatrixBlock features, MatrixBlock labels) {
+		// Generate a single permutation matrix (workers use slices)
+		MatrixBlock permutation = ParamservUtils.generatePermutation((int) features.getNumRows());
+		List<MatrixBlock> pfs = DRLocalScheme.partition(workersNum, features, permutation);
+		List<MatrixBlock> pls = DRLocalScheme.partition(workersNum, labels, permutation);
+		return new Result(pfs, pls);
```
mboehm7: Similar to ORSparkScheme (see comment below), we would need a global permutation.

src/test/java/org/apache/sysml/test/integration/functions/paramserv/SparkDataPartitionerTest.java
```diff
@@ -0,0 +1,186 @@
...
+			for (int i = 0; i < v._1.getNumRows(); i++) {
+				for (int j = 0; j < v._1.getNumColumns(); j++) {
+					Assert.assertEquals((int)labels[i], (int)features[i * v._1.getNumColumns() + j] / v._1.getNumColumns());
+				}
+			}
+		});
+	}
+
+	@Test
+	public void testSparkDataPartitionerDR() {
```
mboehm7: Please compare with local DR partitioner to check for consistent global ordering.

src/test/java/org/apache/sysml/test/integration/functions/paramserv/SparkDataPartitionerTest.java
```diff
@@ -0,0 +1,186 @@
...
+					Assert.assertEquals((int) ld[i],
+							(int) fd[i * features.getNumColumns() + j] / features.getNumColumns());
+				}
+			}
+		});
+	}
+
+
+	@Test
+	public void testSparkDataPartitionerOR() {
```
mboehm7: Please compare with local OR partitioner to check for consistent global ordering.

EdgarLGB:
Thanks @mboehm7 for the review. I have pushed the modifications.

EdgarLGB:
I added the rework on the spark data partitioner and both the local and spark unit test. I'm looking forward to your review @mboehm7 . Thanks!

mboehm7:
Thanks @EdgarLGB - in general this looks fine. In the next couple of days, I'll make an additional detailed pass, run some cluster experiments, and merge it in afterwards.

In the meantime, could you please (1) resolve the conflicts of this PR, (2) simplify the creation of global permutation vectors (instead of using sample-ctable-extract_vector you could simple create the vector via sample), and (3) change the spark row aggregation from rbind to in-place left indexing (performing rbind for n rows copies O(n^2) rows while in-place indexing only copies O(n) rows). Thanks.

EdgarLGB:
Thanks @mboehm7 - I've pushed the modifications. And I will continue on the remote worker setup and cleanup.

mboehm7:
LGTM - thanks @EdgarLGB. I only made a few minor modifications, most notably the handling of seeds in `OVERLAP_RESHUFFLE` to ensure all workers get different permutations (with the same seed, they would all use the same reshuffled but copied dataset).

PR #794, opened 2018-07-03
Profile memory use in JMLC execution

thomas9t:
This PR adds utilities to profile memory use during execution in JMLC. Specifically, the following changes were made:

1. Added options `setStatistics()` and `gatherMemStats()` to `api.jmlc.Connection` which control whether or not statistics should be gathered, and if so, whether memory use should be profiled. Also added an appropriate method to `api.jmlc.PreparedScript` to display the resulting statistics. The following points are only applicable when running in JMLC mode, and memory statistics have been enabled. Both these options are `false` by default.
2. Modified `utils.Statistics` to track the memory used by **distinct** `CacheBlock` objects. At the conclusion of the script, the maximum memory use is reported. Memory use is computed by calling the object's `getInMemorySize()` method. This will generally be a slight over-estimate of the actual memory used by the object. 
3. If `FINEGRAINED_STATISTICS` are enabled, `Statistics` will also track the memory use by each named variable in a DML script and report this in a table as in `heavy hitter instructions`. The goal of this is to detect unexpected large intermediate matrices (e.g. resulting from an outer product `X %*% t(X)`).
4. If `FINEGRAINED_STATISTICS` are enabled, `Statistics` will attempt to measure more accurate memory use by checking to see if an object has been garbage collected. This is done by maintaining a soft reference to the object and periodically checking to see if it has become null. This is enabled only when using fine-grained statistics since it introduces potentially non-trivial overheads by scanning a list of live objects. Note that simply using `rmvar` to remove a live variable results in a substantial underestimate of memory used by the program and so this method is not used. When finegrained statistics are not enabled, the resulting statistics will be an overestimate.

Potential impacts to performance: when finegrained statistics are enabled there will be some performance degradation from maintaining the set of live variables.

Potential Improvements: Related to the above, it would be nice to find a way of accurately tracking when an object is actually released without resorting to checking whether a soft reference has become null. It might also be nice to include a line number indicating where a "heavy hitting object" was created to make debugging easier.

@niketanpansare can you please take a look at these changes when you have time and let me know what you think? Below is an example of what statistics output looks like from running a DML script under JMLC:

```
SystemML Statistics:
Total elapsed time:		0.000 sec.
Total compilation time:		0.000 sec.
Total execution time:		0.000 sec.
Number of compiled MR Jobs:	0.
Number of executed MR Jobs:	0.
Cache hits (Mem, WB, FS, HDFS):	227/0/0/0.
Cache writes (WB, FS, HDFS):	0/0/0.
Cache times (ACQr/m, RLS, EXP):	0.001/0.007/0.001/0.000 sec.
Max size of objects in CP memory:	3.794 GB (73 total objects)
Total JIT compile time:		4.119 sec.
Total JVM GC count:		14.
Total JVM GC time:		1.18 sec.
LibMatrixDNN dense count (conv/bwdF/bwdD/im2col/maxBwd):	0/0/0/0/0.
LibMatrixDNN sparse count (conv/bwdF/bwdD/im2col/maxBwd):	0/0/0/0/0.
LibMatrixDNN conv(im2col/matmult), bwdF (im2col/matmult), bwdD (col2im/matmult) time:	0.000/0.000/0.000/0.000/0.000/0.000 sec.
Heavy hitter instructions:
  #  Instruction            Time(s)  Count  Misc Timers
  1  lmSolve [96:0-96:16]    32.115      1                               
  2  tsmm [41:22-41:25]      26.177      8 rlsi[0.000s,8], aqrd[0.000s,8]
  3  rand [36:12-36:42]       5.030      8                               
  4  jacobian [25:8-25:24]    0.473      8                               
  5  * [79:10-79:10]          0.423      8 aqrd[0.000s,8], rlsi[0.000s,8]
  6  tsmm [31:27-31:27]       0.253      8 rlsi[0.000s,8], aqrd[0.000s,8]
  7  uak+ [80:8-80:13]        0.048      8                               
  8  uak+ [37:16-37:21]       0.040      8                               
  9  uak+ [42:16-42:21]       0.039      8                               
 10  solve [33:12-33:34]      0.036      8                               
Heavy hitter objects:
  #   Object   Memory
  1   _mVar41   762.939 MB
  2   P         762.939 MB
  3   _mVar35   190.735 MB
  4   J         7.629 MB
  5   X         7.629 MB
  6   _mVar32   7.629 MB
  7   _mVar0    7.629 MB
  8   _mVar66   78.168 KB
  9   _mVar68   78.168 KB
  10  _mVar48   78.168 KB
```

### DIFF
--- a/src/main/java/org/apache/sysml/api/jmlc/Connection.java
+++ b/src/main/java/org/apache/sysml/api/jmlc/Connection.java
@@ -178,6 +178,24 @@ public class Connection implements Closeable
 		
 		setLocalConfigs();
 	}
+
+	/**
+	 * Sets a boolean flag indicating if runtime statistics should be gathered
+	 * Same behavior as in "MLContext.setStatistics()"
+	 *
+	 * @param stats boolean value with true indicating statistics should be gathered
+	 */
+	public void setStatistics(boolean stats) { DMLScript.STATISTICS = stats; }
+
+	/**
+	 * Sets a boolean flag indicating if memory profiling statistics should be
+	 * gathered.
+	 * @param stats
+	 */
+	public void gatherMemStats(boolean stats) {
+		DMLScript.STATISTICS = stats || DMLScript.STATISTICS;
+		DMLScript.JMLC_MEMORY_STATISTICS = stats;
+	}
 	
 	/**
 	 * Prepares (precompiles) a script and registers input and output variables.
--- a/src/main/java/org/apache/sysml/utils/Statistics.java
+++ b/src/main/java/org/apache/sysml/utils/Statistics.java
@@ -22,6 +22,7 @@ package org.apache.sysml.utils;
 import java.lang.management.CompilationMXBean;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
+import java.lang.ref.SoftReference;
 import java.text.DecimalFormat;
 import java.util.Arrays;
 import java.util.Comparator;
@@ -29,11 +30,13 @@ import java.util.List;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.DoubleAdder;
 import java.util.concurrent.atomic.LongAdder;
 
 import org.apache.sysml.api.DMLScript;
 import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.hops.OptimizerUtils;
+import org.apache.sysml.runtime.controlprogram.caching.CacheBlock;
 import org.apache.sysml.runtime.controlprogram.caching.CacheStatistics;
 import org.apache.sysml.runtime.controlprogram.context.SparkExecutionContext;
 import org.apache.sysml.runtime.instructions.Instruction;
@@ -68,6 +71,16 @@ public class Statistics
 	// number of compiled/executed SP instructions
 	private static final LongAdder numExecutedSPInst = new LongAdder();
 	private static final LongAdder numCompiledSPInst = new LongAdder();
+
+	// number and size of pinned objects in scope
+	private static final DoubleAdder sizeofPinnedObjects = new DoubleAdder();
+	private static long maxNumPinnedObjects = 0;
+	private static double maxSizeofPinnedObjects = 0;
+
+	// Maps to keep track of CP memory objects for JMLC (e.g. in memory matrices and frames)
+	private static final ConcurrentHashMap<String,Double> _cpMemObjs = new ConcurrentHashMap<>();
+	private static final ConcurrentHashMap<Integer,Double> _currCPMemObjs = new ConcurrentHashMap<>();
+	private static final ConcurrentHashMap<Integer,SoftReference<CacheBlock>> _liveObjects = new ConcurrentHashMap<>();
 
 	//JVM stats (low frequency updates)
 	private static long jitCompileTime = 0; //in milli sec
@@ -553,7 +566,7 @@ public class Statistics
 	public static void accPSBatchIndexingTime(long t) {
 		psBatchIndexTime.add(t);
 	}
-	
+
 	public static String getCPHeavyHitterCode( Instruction inst )
 	{
 		String opcode = null;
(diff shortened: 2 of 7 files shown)


src/main/java/org/apache/sysml/api/jmlc/Connection.java
```diff
@@ -178,6 +178,24 @@ public Connection(DMLConfig dmlconfig) {
...
+	 * Same behavior as in "MLContext.setStatistics()"
+	 *
+	 * @param stats boolean value with true indicating statistics should be gathered
+	 */
+	public void setStatistics(boolean stats) { DMLScript.STATISTICS = stats; }
+
+	/**
+	 * Sets a boolean flag indicating if memory profiling statistics should be
+	 * gathered.
+	 * @param stats
```
niketanpansare: missing comment

src/main/java/org/apache/sysml/utils/Statistics.java
```diff
@@ -69,6 +72,16 @@
...
 
+	// number and size of pinned objects in scope
+	private static final DoubleAdder sizeofPinnedObjects = new DoubleAdder();
+	private static long maxNumPinnedObjects = 0;
+	private static double maxSizeofPinnedObjects = 0;
+
+	// Maps to keep track of CP memory objects for JMLC (e.g. in memory matrices and frames)
+	private static final ConcurrentHashMap<String,Double> _cpMemObjs = new ConcurrentHashMap<>();
+	private static final ConcurrentHashMap<Integer,Double> _currCPMemObjs = new ConcurrentHashMap<>();
+	private static final ConcurrentHashMap<Integer,SoftReference<CacheBlock>> _liveObjects = new ConcurrentHashMap<>();
```
niketanpansare: Can you add comment here about underutlization if this was handled via rmvar mechanism ?

niketanpansare:
Also, can you please update the [JMLC documentation](https://github.com/apache/systemml/blob/master/docs/jmlc.md)? ... Other than that, LGTM :+1:

src/main/java/org/apache/sysml/api/jmlc/Connection.java
thomas9t: Done!

src/main/java/org/apache/sysml/utils/Statistics.java
thomas9t: Done!

thomas9t:
Thanks @niketanpansare - I've addressed your comments above. Please let me know if you'd like to see any other changes.

thomas9t:
Thanks for the suggestions @mboehm7! I had considered placing the logic for removal of pinned objects in `cleanUpCacheableData` - however it seemed in practice that this substantially underestimated the actual memory used by the program. For example, consider the following (admittedly trivial) DML script:

```
b = rand(rows=100, cols=1)
for (ii in 1:30)
{
    X = rand(rows=1000000, cols=100)
    y = X %*% b
    print(as.scalar(y[1,1])) 
}
```

If I place the logic to remove pinned objects in `cleanUpCacheableData` then the reported memory use is `763MB`. However, according to VisualVM, the actual peak heap use of the program is closer to `4.8GB`. Running the program with `-Xmx1g` results in an `OutOfHeapSpace` for me. By contrast, using `WeakReference` (see comment below regarding `SoftReference`) reports a peak memory use of `3.75GB`. While the above example is contrived, a similar situation could occur in an iterative estimation algorithm. However, given that JMLC is primarily used for prediction rather than estimation this may not be an issue.  Let me know what you think - it's entirely possible I'm not understanding how JVM memory management works correctly.

I see your point about `SoftReferences` - they indeed will discourage garbage collection unless there is memory pressure. If we do want to stick with this approach than I think a `WeakReference` would be better as it (at least to my understanding) does not change garbage collection.

mboehm7:
@thomas9t thanks for the clarification and yes, weak references would solve one of the issues. Btw, instead of scanning for cleared objects, you would simply implement this via reference queues.

However, from my perspective ~800MB is exactly what we should report here. The JVM in server mode will likely consume all available memory before GC is triggered. Capturing actual memory consumption (including live garbage) would mean that the statistics would report different max memory consumption under different heap sizes - this would be confusing and defeat the purpose of this tool (automatically determine necessary max heap sizes). By specifying `-Xmx1g -Xmn100m` your scenario will work just fine. I think the openjdk/hotspot is too aggressive sizing the young generation (which also affects the memory available to the application). In hybrid mode, we guarantee not to run out of memory whenever the young generation is less or equal 10% of the max heap size.

thomas9t:
@mboehm7 - understood. I see your point about the confusion stemming from gathering statistics under different max heap sizes. Per this discussion and the comments above, I have done the following:

1. Moved the logic in `CacheableData.acquireModifyIntern` to the public wrapper. Note as well that since `CacheableData.isCachingActive()` will always return `false` in JMLC, I think we don't need to do anything in `CacheableData.release` - please correct me if I'm wrong.
2. Removed the use of `SoftReference` to track when objects have been removed and instead remove an object when `ExecutionContext.cleanUpCacheableData` is called on the object.
3. Updated documentation accordingly.

Should I submit this as another PR?

mboehm7:
thanks @thomas9t - this sounds good to me. Yes, I'd appreciate if you could submit another PR. 

Wrt to release, we still would want to use this tool through all APIs. However, always reporting the number of live objects (independent if pinned or not) is generally useful - so, yes let's go with your proposal.

PR #795, opened 2018-07-09
Updates to JMLC Memory Profiling

thomas9t:
Following the conversation in https://github.com/apache/systemml/pull/794 with @mboehm7 I have a made a couple improvements to the method used to profile memory use in JMLC:

1. Removed the use of `SoftReference` to track when an object has been garbage collected. Objects are now removed (i.e. the memory used by them is subtracted from the total) when `ExectutionContext.cleanUpCacheableData` is called. This ensures that we always get a consistent maximum memory use which does not change depending on the max heap size used to initialize the JVM.
2. Removed some unneeded code which won't be called during JMLC execution.

Documentation was also updated accordingly.


src/main/java/org/apache/sysml/runtime/controlprogram/context/ExecutionContext.java
```diff
@@ -600,6 +601,9 @@ public void unpinVariables(List<String> varList, boolean[] varsState) {
 	}
 	
 	public void cleanupCacheableData(CacheableData<?> mo) {
+		if (DMLScript.JMLC_MEMORY_STATISTICS)
+			Statistics.removeCPMemObject(System.identityHashCode(mo));
```
mboehm7: Please, add the same in the overloaded method in `SparkExecutionContext` such that it also works through the other APIs.

thomas9t:
Closing this PR and creating a new one to deal with merge conflict issues.

PR #796, opened 2018-07-11
wip: Cohen Estimator

johannaSommer:


mboehm7:
SYSTEMML-2291

mboehm7:
LGTM - thanks @johannaSommer. As it turned out the accuracy issues was due to a syntactic issue on final aggregation (`result=+LGraph.nodes.get(h).getValue()` instead of `result += ...`). Furthermore, I made  a couple of minor simplifications and cleanups. For example, we now create just a single distribution function (whose seed is initialized based on current time in milliseconds) in order to ensure that we generate pseudo-random numbers.

PR #797, opened 2018-07-13
Updates to JMLC Memory Profiling

thomas9t:
Following the conversation in https://github.com/apache/systemml/pull/794 with @mboehm7 I have a made a couple improvements to the method used to profile memory use in JMLC:

1. Removed the use of `SoftReference` to track when an object has been garbage collected. Objects are now removed (i.e. the memory used by them is subtracted from the total) when `ExectutionContext.cleanUpCacheableData` is called. This ensures that we always get a consistent maximum memory use which does not change depending on the max heap size used to initialize the JVM.
2. Removed some unneeded code which won't be called during JMLC execution.

Documentation was also updated accordingly.


mboehm7:
LGTM - thanks @thomas9t, this works very well. During the merge, I also added a command line argument `-mem` and made some minor cleanups.

PR #802, opened 2018-07-18
[SYSTEMML-2446] Fix the cleanup of list object

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for fixing the cleanup of list object. The idea is to not remove the matrix block which is referenced in the previous matrix object. Because when doing the partial model update, there will be some parameters unchanged (i.e., the matrix block remains the same).

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB. After running a number of experiments with different configurations, I can confirm that we don't seem to miss any cleanups. I tracked the number of managed objects in the bufferpool, and this number stabilizes even for 10 epochs at ~450 matrices when running ASP with BATCH and EPOCH and the different optimizers. So this looks good to me.

PR #803, opened 2018-07-18
[SYSTEMML-2339] Docker support

j143-zz:
Hi, this adds base image of Ubuntu16.04 with CUDA 8.0, CuDNN 5.1.10 for now. Please, provide your feedback on dependencies.

- [x] Base image
- [x] Dependencies (spark)
- [x] notebook
- [ ] Directing to browser
- [ ] installing systemml from target files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+#1. Base image, contains OS with CUDA 8.0, cuDNN 5.1.10 prebuilt
+FROM nvidia/cuda:8.0-cudnn5-runtime-ubuntu16.04
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    python3-setuptools
+
+# 2. Install spark backend
+RUN pip3 install pyspark
+
+# 3. Install systemml
+RUN pip3 install systemml
+
+# 4. Install utilities
+RUN pip3 install jupyter matplotlib numpy
```
bdwyer2: You could combine all the `pip3 install` commands to reduce the size of the image.

PR #807, opened 2018-07-25
changed rounding on propagation to be probabilistic

johannaSommer:


mboehm7:
LGTM - thanks @johannaSommer. The failed tests were just due to a flipped comparison.

PR #808, opened 2018-07-25
[SYSTEMML-2420, 2457] Rework the rpc communication and add the statistics

EdgarLGB:
Hi @mboehm7 ,

Here is the PR which consists of the rework for rpc data exchange by doing the deep serialization and the modification of statistics for remote workers via accumulators. Also the spark ps test has been fixed.

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB. Overall, this fixes all the issues. I just made some additional cleanups (serialization/deserialization via data streams for UTF encoding and to avoid copies) and statistics management (given your architecture, we don't need to handle local mode differently, additive accumulators for number of batches/epochs).

PR #811, opened 2018-08-01
Single operations pt. 1

johannaSommer:
Added estimation for operations: elementw. mult, elementw. plus, cbind & rbind. 
You can now call the estimator with an input string to determine the operation.

### DIFF
--- a/src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
+++ b/src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
@@ -84,20 +84,26 @@ public class EstimatorMatrixHistogram extends SparsityEstimator
 		MatrixHistogram h1 = new MatrixHistogram(m1, _useExcepts);
 		MatrixHistogram h2 = (m1 == m2) ? //self product
 			h1 : new MatrixHistogram(m2, _useExcepts);
+		double r_sparsity = 0;
+		double c_sparsity = 0;
+		int msize = h1.getCols()*h1.getRows();
 		switch (op) {
 		case "mult":
-			double r_sparsity = 0;
-			double c_sparsity = 0;
-			int msize = h1.getCols()*h1.getRows();
 			for(int k=0; k<h1.getCols();k++) {
-				c_sparsity += h1.cNnz[k]/msize + h1.cNnz[k]/msize - (h1.cNnz[k]/msize)*(h1.cNnz[k]/msize);
+				c_sparsity += (h1.cNnz[k]/msize)*(h2.cNnz[k]/msize);
 			}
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
```diff
@@ -71,15 +71,66 @@ public double estim(MMNode root) {
...
+	
+	@Override 
 	public double estim(MatrixBlock m1, MatrixBlock m2) {
 		MatrixHistogram h1 = new MatrixHistogram(m1, _useExcepts);
 		MatrixHistogram h2 = (m1 == m2) ? //self product
 			h1 : new MatrixHistogram(m2, _useExcepts);
 		return estimIntern(h1, h2);
 	}
 
+	public double estim(MatrixBlock m1, MatrixBlock m2, String op) {
```
mboehm7: 1) Please add an enum for the operation type to avoid this string handling.
2) Extend the estimator class to have overloaded estim method with and without op code
3) The default estim without opcode should simply call the other estim method with default opcode MM (matrix mult)

src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
```diff
@@ -71,15 +71,66 @@ public double estim(MMNode root) {
...
+				c_sparsity = c_sparsity + (double)h1.cNnz[k]/msize + (double)h2.cNnz[k]/msize - ((double)h1.cNnz[k]/msize)*((double)h2.cNnz[k]/msize);
+			}
+			for(int j=0; j<h1.getRows();j++) {
+				r_sparsity = r_sparsity + (double)h1.rNnz[j]/msize + (double)h2.rNnz[j]/msize - ((double)h1.rNnz[j]/msize)*((double)h2.rNnz[j]/msize);
+			}
+			return Math.min(c_sparsity, r_sparsity);
+		case "cbind":
+			return (double) m1.getNonZeros()+m2.getNonZeros()/(m1.getNumRows()*(m1.getNumColumns()+1));
+		case "rbind":
+			return (double) m1.getNonZeros()+m2.getNonZeros()/((m1.getNumRows()+1)*m1.getNumColumns());
```
mboehm7: The operations cbind and rbind may concatenate arbitrary matrices - hence, we need to take the exact number of columns and rows form m2 into account to compute the sparsity.

mboehm7: Also try to use OptimizerUtils.getSparsity() to prevent int overflows when doing rows*cols.

mboehm7:
LGTM - thanks @johannaSommer. During the merge I only some minor cleanups and extensions.

PR #813, opened 2018-08-03
single operations for Average Case estimator

johannaSommer:
finished the Histogram operations, Avg. Case operations are now implemented


mboehm7:
LGTM - thanks @johannaSommer.

PR #815, opened 2018-08-03
Improvement of elem-wise multiplication for Histogram

johannaSommer:


mboehm7:
LGTM - thanks @johannaSommer. Just for the future: please refrain from reformatting existing code, especially via auto formatting because (1) it causes unnecessary merge conflicts, and (2) some of the reformatted code really does not look good.

src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
```diff
@@ -31,254 +31,256 @@
...
-		private final int rMaxNnz, cMaxNnz;     //max nnz per row/row
-		private final int rN1, cN1;             //number of rows/cols with nnz=1
-		private final int rNonEmpty, cNonEmpty; //number of non-empty rows/cols (w/ empty is nnz=0)
-		private final int rNdiv2, cNdiv2;       //number of rows/cols with nnz > #cols/2 and #rows/2
-		private boolean fullDiag;               //true if there exists a full diagonal of nonzeros
-		
+		private final int rMaxNnz, cMaxNnz; // max nnz per row/row
+		private final int rN1, cN1; // number of rows/cols with nnz=1
+		private final int rNonEmpty, cNonEmpty; // number of non-empty rows/cols
+												// (w/ empty is nnz=0)
```
mboehm7: Please refrain from formatting existing code, especially via auto formatting as it (1) creates unnecessary merge conflicts, and (2) really does not look good.

src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
```diff
@@ -31,254 +31,256 @@
...
-			case NEQZERO:
-			case TRANS:
-			case RESHAPE:
-				return OptimizerUtils.getSparsity(h1.getRows(), h1.getCols(), h1.getNonZeros());
-			default:
-				throw new NotImplementedException();
+		case MM:
+			return estimInternMM(h1, h2);
+		case MULT:
+			double[] out_r = null;
```
mboehm7: This missing allocation would lead to null pointer exceptions on the first access. Also, there is no need to materialize this array as we can directly compute the aggregate.

src/main/java/org/apache/sysml/hops/estim/EstimatorMatrixHistogram.java
```diff
@@ -31,254 +31,256 @@
...
+			for (int i = 0; i < h1.getRows(); i++) {
+				out_r[i] = (h1.rNnz[i] * h2.rNnz[i]
+						* IntStream.range(0, h1.getCols()).mapToDouble(k -> h1.cNnz[k] * h2.cNnz[k]).sum())
+						/ (Arrays.stream(h1.rNnz).sum() * Arrays.stream(h1.rNnz).sum());
```
mboehm7: These repeated scans of the count vectors for each row in the lhs, would cause a time complexity in the number of cells of the output. Luckily, these are constants with regard to different rows and hence can be computed once.

PR #816, opened 2018-08-04
[DOC] Add the language doc for paramserv func

EdgarLGB:
Hi @mboehm7 ,

Here is the PR for adding the language doc for paramserv function. And I also put it on my personal github page to review it visually. Here is the [link](https://edgarlgb.github.io/systemml/dml-language-reference.html#parameter-server-built-in-function).

Thanks for the review,
Guobao

### DIFF
--- a/docs/dml-language-reference.md
+++ b/docs/dml-language-reference.md
@@ -53,6 +53,7 @@ limitations under the License.
     * [Read/Write Built-In Functions](dml-language-reference.html#readwrite-built-in-functions)
     * [Data Pre-Processing Built-In Functions](dml-language-reference.html#data-pre-processing-built-in-functions)
     * [Deep Learning Built-In Functions](dml-language-reference.html#deep-learning-built-in-functions)
+    * [Parameter Server Built-In Function](dml-language-reference.html#parameter-server-built-in-function)
     * [Other Built-In Functions](dml-language-reference.html#other-built-in-functions)
   * [Frames](dml-language-reference.html#frames)
     * [Creating Frames](dml-language-reference.html#creating-frames)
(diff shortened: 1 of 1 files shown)


docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
 | bias_add             |                             | `ones = matrix(1, rows=1, cols=height*width); output = input + matrix(bias %*% ones, rows=1, cols=numChannels*height*width)`                                |
 | bias_multiply        |                             | `ones = matrix(1, rows=1, cols=height*width); output = input * matrix(bias %*% ones, rows=1, cols=numChannels*height*width)`                                |
 
+### Parameter Server Built-in Function
+SystemML supports large-scale machine learning architecture **data-parallel Parameter Server**, introduced by a built-in function **paramserv**. Currently both local multi-threaded and spark distributed backend are supported to execute the **paramserv** function. For example, in order to train a model in local backend with update strategy BSP, 10 epochs, 64 batchsize, 10 workers, **paramserv** function should look like this:
```
mboehm7: I would recommend to start like that "Apart from data-parallel operations and task-parallel parfor loops, SystemML also supports a **data-parallel Parameter Server** via a built-in function **paramserv**".

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+
+
+    model'=paramserv(model=params, features=X, labels=Y, val_features=X_val, val_labels=Y_val, upd="fun1", agg="fun2", mode="LOCAL", utype="BSP", epochs=10, batchsize=64, k=10, hyperparams=hParams)
```
mboehm7: Please use valid script syntax; i.e., use model2 instead of model', and maybe call params better model1 to make clear it's the initial model.

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+
+**Table**: Inputs of paramserv function
+
+Parameters | Description | Type | Mandatory | Options
+-------- | ----------- | ---------- | ---------- | -------
+model | All the parameters (e.g., the weight and bias matrices) | list | yes | 
+features | Training features | matrix | yes 
+labels | Training labels | matrix | yes
+val_features | Validation features | matrix | yes
+val_labels | Validation labels | matrix | yes
```
mboehm7: We could make the validation data (features and labels) optional as we're not using them internally, right?

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+upd | Physical name of gradient calculation function. The format should be "related path:func name". For example, "./mnist_lenet_paramserv_sgd.dml::gradients". | string | yes
+agg | Physical name of gradient aggregation function. The format should be "related path:func name". For example, "./mnist_lenet_paramserv_sgd.dml::aggregation". | string | yes
+mode | Execution backend where the parameter function is executed | string | yes | "LOCAL", "REMOTE_SPARK"
```
mboehm7: It would be good to assign a default, e.g., LOCAL

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+utype | Update strategy | string | yes | "BSP", "ASP"
```
mboehm7: It would be good to assign a default, e.g., ASP

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+freq | Frequency of parameters updating. When setting to "BATCH", the gradients will be pushed to server for updating the parameters after each mini-batch. When setting to "EPOCH", the gradients will be pushed to server after each epoch. | string | no | "BATCH"(default), "EPOCH"
```
mboehm7: Could we please make EPOCH the default, as it's probably the more common usage.

docs/dml-language-reference.md
mboehm7: Also please provide a description (maybe below the table) for BSP and ASP.

docs/dml-language-reference.md
mboehm7: Here the same, please provide description for BATCH and EPOCH

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+epochs | Number of epoch | integer | yes |
+batchsize | Size of mini-batch, if the update frequence is setted to "EPOCH", this parameter will be ignored. | integer | no | 64(default)
```
mboehm7: Please clarify that we use batch of rows

mboehm7: Also it's "set" not "setted" and the parameter has still influence under EPOCH update frequency.

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+k | Degree of parallelism | integer | no | Number of vcores, otherwise vcores / 2 if using openblas(default)
```
mboehm7: Could you please verify that this workaround is still needed for openblas? I think after we fixed the degree of parallelism during recompilation this should be fine now.

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+scheme | Scheme of data partition, i.e., how the data is distributed across workers | string | no | "DISJOINT_CONTIGUOUS", "DISJOINT_ROUND_ROBIN", "DISJOINT_RANDOM", "OVERLAP_RESHUFFLE"
```
mboehm7: Please describe these schemes below the table. Perhaps you could define them in terms of linear algebra operations as we discussed them before.

mboehm7: How about we use DISJOINT_CONTIGUOUS as default?

docs/dml-language-reference.md
```diff
@@ -1536,6 +1537,62 @@ Examples:
...
+          # the output name can be something else than "gradients" but should always return a list
+```
+
+**Aggregate function:**
+
+The aggregate function aims to leverage some optimizers to update the model with the given gradients. The implementation of this function should be based on a function signature like this: (i.e., **the input parameter including both type and name should be exactly the same as the below, except that the output name could be different**)
+
+```sh
+aggregation = function(list[unknown] model, list[unknown] gradients,
+                       list[unknown] hyperparams)
```
mboehm7: Could we please cleanup this API and use the same order for model and hyper parameters in both functions?

mboehm7:
LGTM - thanks @EdgarLGB. I'll also push these changes to our website.

PR #817, opened 2018-08-05
[SYSTEMML-2299] Update the API of paramserv built-in func

EdgarLGB:
Hi @mboehm7 ,

According to the latest PR, here is the PR for updating the API of paramserv func which involves the arguments "val_features", "val_labels", "mode", "utype", "freq", "k".

Thanks for the review,
Guobao


mboehm7:
LGTM - thanks @EdgarLGB this is awesome.

PR #818, opened 2018-08-06
Avg. & Worst Case extension

johannaSommer:
Extended Average and Worst Case with single other operations and also chains of other operations.
Also changed structure of MMNode, so that all non-leaf nodes now have an OpCode handed in during construction. That is for special cases of multiplication, where there are not only multiplication operations in the chain.


mboehm7:
LGTM - thanks @johannaSommer. I made some additional API cleanups to avoid unnecessary code duplication. Furthermore, this also includes a HOTFIX for an issue in the bitset estimator I introduced earlier today.

PR #820, opened 2018-08-07
Histogram chain operations

johannaSommer:
Transferred the switch case for chains of operations into the deriveHistogram() function, as the estimation of the nnzs stays the same and only the propagation of the vectors stays the same.


mboehm7:
LGTM - thanks @johannaSommer. I fixed the allocated dimensions, unnecessary overwrites of allocated vectors, and rbind/cbind which can be composed of basic addition instead of probabilistic propagation. Furthermore, I added a TODO for the remaining unary operations. To simplify development and ensure we don't introduce bugs, it would be good to complement these additional operations with appropriate junit tests.

PR #823, opened 2018-08-09
JUnit tests and bug fixes for estimators

johannaSommer:
added junit tests for single matrix operations. Minor bug fixes on wrong function call in the density map and wrong loop boundaries in the MatrixHistogram. 
Also changed the implementation of MatrixHistogram elementwise plus, it is now more accurate.


mboehm7:
LGTM - thanks @johannaSommer. These fixes and new tests are great. Just for the future please try to avoid unnecessary warnings. 

Furthermore, when playing around with different dimension sizes for the MNC plus and mult estimation, I noticed an accuracy issue whenever the number of rows larger than the number of columns. Maybe it's worth experimenting with a symmetric version over columns instead of rows whenever this is the case.

PR #824, opened 2018-08-10
Bitset single operations

johannaSommer:
First extension to Bitset estimator, added single operations and structure for MMNode, as well as a first suggestion for rbind.


mboehm7:
LGTM - thanks @johannaSommer. In order to avoid code duplication, we should always use a single estimInternal method together with a probe for exact meta data operations (incl diag). I reworked all estimators accordingly. Also regarding the rbind, the idea was good but we have to construct the actual BitsetMatrix instance and need special handling for both implementations. I modified the rbind operations accordingly - please follow this scheme for all other operations along with dedicated tests.

PR #825, opened 2018-08-10
Finished single operations for Sample estimator

johannaSommer:
WIP: almost finished sample single operation estimation, reshape and diag still missing.


johannaSommer:
he ComputeRowNz function has bugs. Please let me fix them before merging this PR. As you can see in the updated test the general approach works, as well as the ComputeColumnNz function. Either one is called depending on which dimension is the larger one (accuracy reasons).

johannaSommer:
This PR now includes the single op extension for the sample estimator, as well as a bug fix for the right hand side call of MMNodes and a change to the MNC elementwise plus and minus (casting to double instead of long)

PR #826, opened 2018-08-10
Histogram finalize

johannaSommer:
Finished Histogram chain operations, excluding reshape and diag sketch propagation. 
Still to be implemented: tests for all estimators for chains of operations.


johannaSommer:
The ComputeRowNz function has bugs. Please let me fix them before merging this PR. As you can see in the updated test the general approach works, as well as the ComputeColumnNz function. Either one is called depending on which dimension is the larger one (accuracy reasons).

johannaSommer:
will create a new pr as soon as histogram is completely done.

PR #829, opened 2018-08-17
Layered Graph extension for chains

johannaSommer:
Added the estim method for a MMNode input, so that the graph is built for every element in an array list. Also added tests for Layered Graph.


mboehm7:
LGTM - thanks @johannaSommer. During the merge I just made some minor cleanups and a fix for the correct output dimensions.

PR #831, opened 2018-08-23
Adds Commons-Lang3 to Standalone JAR Dependencies

thomas9t:
JMLC has a dependency on `org/apache/commons/lang3/EnumUtils` which causes a `NoClassDefFoundError` when running with the standalone jar. This PR adds this dependency. @bertholdreinwald @niketanpansare can you please take a look and let me know if you approve.


niketanpansare:
LGTM. Thanks @thomas9t for catching this 👍

PR #834, opened 2018-09-17
Fixes a minor bug in LRU cache

thomas9t:
@niketanpansare There is a minor bug in `PersistentLRUCache` which causes an exception to be improperly thrown when caching `MatrixBlock`. This PR should fix the issue.


niketanpansare:
LGTM 👍

PR #835, opened 2018-09-18
[MINOR] Adds method to unpin data in JMLC

thomas9t:
There are instances in which it's useful to be able to "unpin" variables in JMLC. For example, when executing several different JMLC scripts in the same JVM, I manage model weights externally using a cache of `SoftReferences` to the underlying matrix blocks. Each script gets a slice of time on a CPU during which I want the weights pinned. However, once I am done with a script I would like the weights to be free for garbage collection under memory pressure which is not possible if the script holds a reference to the weight as a reused variable. For the sake of consistency, I have also switched from storing references to pinned data in a `HashMap<String,Data>` to a `LocalVariableMap`.

### DIFF
--- a/src/main/java/org/apache/sysml/api/jmlc/PreparedScript.java
+++ b/src/main/java/org/apache/sysml/api/jmlc/PreparedScript.java
@@ -73,7 +73,7 @@ public class PreparedScript implements ConfigurableAPI
 	//input/output specification
 	private final HashSet<String> _inVarnames;
 	private final HashSet<String> _outVarnames;
-	private final HashMap<String,Data> _inVarReuse;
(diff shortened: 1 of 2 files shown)


thomas9t:
@niketanpansare can you take a quick look when you have a chance?

thomas9t:
Also just FYI - the CI build failed due to `The job exceeded the maximum log length, and has been terminated` - I ran `mvn clean verify` on my machine and didn't encounter any errors.

niketanpansare:
@thomas9t  Overall, this looks good to me. If you still want to merge it, can you please resolve the conflicts or create a new PR?

mboehm7:
LGTM - thanks @thomas9t and sorry for the delay.

PR #832, opened 2018-08-24
[SYSTEMML-1325] Cleanup static variables in DMLScript

niketanpansare:
We use ThreadLocal DMLOptions and DMLConfig instead of static variables in DMLScript class. It allows different JMLC instances (or MLCContext instances) to run with different options (such as with GPU, with CPU, etc).

@nakul02 @thomas9t @mboehm7 Please let me know if you have suggestions or concerns regarding this PR?

PR #836, opened 2018-10-11
[SYSTEMML-1325] Harmonize Compilation Execution Pipelines and Add GPU Support to JMLC

thomas9t:
This PR adds support for compilation and execution of GPU enabled scripts in JMLC and harmonizes the pipeline used to compile and execute DML programs across the JMLC, MLContext and DMLScript. Specifically, the following changes were made:

1. All three APIs now call ScriptExecutorUtils.compileRuntimeProgram to compile DML scripts. The original logic in MLContext and JMLC for pinning inputs and persisting outputs has been preserved.
1. All three APIs now use ScriptExecutorUtils.executeRuntimeProgram to execute the compiled program. Previously, JMLC called the Script.execute method directly.
1.  jmlc.Connection.prepareScript now supports compiling a script to use GPU. Note that following #832 the issue noted in #830 has been resolved. 
1. A PreparedScript is now statically assigned a GPU context when it is compiled and instatiated. This has potential performance implications because it means that a PreparedScript must be executed on a specific GPU. However, it reduces overhead from creating a GPU context each time a script is executed and unsures that a user cannot compile a script to use GPU and then forget to assign a GPU context when the script is run.
1. Per (3) I have added a unit test which compiles and executes a GPU enabled script in JMLC both with and without pinned data and just asserts that no errors occur. 

I ran `mvn clean verify` on a GPU enabled machine and encountered no exceptions.

@niketanpansare can you please take a look and let me know what you think? @mboehm7 you may also want to look at this.


niketanpansare:
Thanks @thomas9t ... the PR looks good to me 👍 

Let's wait until tomorrow before merging in case @mboehm7 has additional comments.

PR #837, opened 2018-10-15
[SYSTEMML-1325] Bugfix when cleaning up temporary GPU memory via JMLC

niketanpansare:
- When cleaning up the temporary GPU memory, we should skip unpinned
matrices.

@thomas9t can you please review this PR ?

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
@@ -94,10 +94,12 @@ public class GPUMatrixMemoryManager {
 	 * Get pointers from the first memory sections "Matrix Memory"
 	 * @param locked return locked pointers if true
 	 * @param dirty return dirty pointers if true
+	 * @param isCleanupEnabled return pointers marked for cleanup if true
 	 * @return set of pointers
 	 */
-	Set<Pointer> getPointers(boolean locked, boolean dirty) {
-		return gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty).flatMap(gObj -> getPointers(gObj).stream()).collect(Collectors.toSet());
+	Set<Pointer> getPointers(boolean locked, boolean dirty, boolean isCleanupEnabled) {
+		return gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty
+				&& gObj.mat.isCleanupEnabled() == isCleanupEnabled).flatMap(gObj -> getPointers(gObj).stream()).collect(Collectors.toSet());
 	}
 	
 	/**
--- a/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMemoryManager.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMemoryManager.java
@@ -516,8 +516,10 @@ public class GPUMemoryManager {
 	 * Clears up the memory used by non-dirty pointers.
 	 */
 	public void clearTemporaryMemory() {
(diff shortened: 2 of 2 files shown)


thomas9t:
@niketanpansare - sorry for the slow movement on this. I'm waiting until I have access to a GPU again to test.

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
```diff
@@ -94,10 +94,12 @@ else if(sparsePtr.val != null)
...
 	 * @param locked return locked pointers if true
 	 * @param dirty return dirty pointers if true
+	 * @param isCleanupEnabled return pointers marked for cleanup if true
 	 * @return set of pointers
 	 */
-	Set<Pointer> getPointers(boolean locked, boolean dirty) {
-		return gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty).flatMap(gObj -> getPointers(gObj).stream()).collect(Collectors.toSet());
+	Set<Pointer> getPointers(boolean locked, boolean dirty, boolean isCleanupEnabled) {
+		return gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty
+				&& gObj.mat.isCleanupEnabled() == isCleanupEnabled).flatMap(gObj -> getPointers(gObj).stream()).collect(Collectors.toSet());
```
thomas9t: This will end up clearing pointers for pinned weight matrices as they are unlocked and not-dirty (since they're read only) at the end of a script invocation. This ends up causing a `CUDNN_STATUS_MAPPING_ERROR`. I think we should just never clean up pointers for pinned data (i.e. when cleanup is not enabled). We can fix this by either changing the above to:

```
gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty || gObj.mat.isCleanupEnabled() == isCleanupEnabled)
```

Or by adding a separate check to see if a pointer being cleared is associated with a GPU object which is marked as pinned. My preference is for the latter as it seems easier to understand.

thomas9t:
@niketanpansare - there's one additional problem here. In JMLC we can have inputs which are not cleaned up by `rmvar` but for which `isCleanupEnabled = true`. So these pointers end up being freed by `guardedCudaFree` in `clearTemporaryMemory`, but the associated GPUObject remains in `GPUMatrixMemoryManager.gpuObjects`. An example would be the feature vector for prediction. These change each time a prepared script is invoked (so `isCleanupEnabled = true`), but are not cleaned up by an `rmvar` instruction. 

This causes a single `Pointer` to be associated with several different `GPUObjects` which can cause us to try and free a pointer twice leading to an exception. We could fix this issue by adding a method which removes non-dirty, unlocked, GPUObjects for which cleanup is enabled from the memory manager as part of the logic for `GPUContext.cleanupTemporaryMemory()`. However, this will add a performance cost as we need to do an extra scan of the list of GPUObjects. Note that it's not clear to me why these variables are not cleaned up by `rmvar` in JMLC.

mboehm7:
well, although I don't know the details here, if the prediction vectors are passed from outside into JMLC then we can't compile a rmvar because a user might still want to use the variable afterwards (or even pass it into another prepared script invocation). 

If the GPU cleanup is an issue, you might want to think about something like Spark's ContextCleaner, which simply registers a weak reference on the object and listens on its queue - if the weak reference is cleared, you know the application program can't reach the object anymore and hence it can be lazily cleaned up from the GPU memory manager. Spark uses the same for RDDs, broadcasts, accumulators, etc.

thomas9t:
Thanks for the clarification @mboehm7

niketanpansare:
@mboehm7 The issue was with the GPU memory manager and @thomas9t fixed it in the PR https://github.com/apache/systemml/pull/839 👍

PR #838, opened 2018-10-19
[SYSTEMML-1325] Fixes formatting issues and warnings.

thomas9t:
Fixes some minor formatting issues (spaces -> tabs) and warnings from #836 Removes no longer needed code for compilation and execution from `org.apache.sysml.api.mlcontext.ScriptExecutor` Fixes a bug from #836 that was causing explain plans to not be printed in ML context. @niketanpansare @mboehm7 can you please take a look when you have a chance and let me know if you have any comments?


niketanpansare:
Overall, this looks good to me. Let's wait for few days in case @mboehm7 has additional comments.

thomas9t:
@niketanpansare anything else we need to do here?

PR #839, opened 2018-10-22
[SYSTEMML-1325] Bugfix for GPU memory manager clear temporary memory

thomas9t:
In `GPUMemoryManager.clearTemporaryMemory()` we deallocate pointers but do not set the corresponding `Pointer` slots to `null` in the associated `GPUObject` instances. This can lead to attempted double-freeing of Pointers which results in an exception. This PR fixes this issue by creating a list of GPU objects associated with Pointers that have been freed as part of `clearTemporaryMemory()` and setting the corresponding pointer slots to `null`. This PR also addresses a minor issue with cleanup in JMLC which was causing Pointers for pinned data to be improperly cleared. Note this PR will reduce performance of `GPUMemoryManager.clearTemporaryMemory()` because it is now necessary to search through the list of managed `GPUObjects` to find the ones corresponding to pointers being freed. However, this method is only called once at the end of script invocation and so the performance cost will be small.

@niketanpansare can you please take a look and let me know what you think?

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
@@ -18,7 +18,9 @@
  */
 package org.apache.sysml.runtime.instructions.gpu.context;
 
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -43,8 +45,7 @@ public class GPUMatrixMemoryManager {
 	void addGPUObject(GPUObject gpuObj) {
 		gpuObjects.add(gpuObj);
 	}
-	
-	
+
 	/**
 	 * Get list of all Pointers in a GPUObject 
 	 * @param gObj gpu object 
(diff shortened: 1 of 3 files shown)


niketanpansare:
Looks good to me. Thanks @thomas9t 👍

src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMatrixMemoryManager.java
```diff
@@ -94,10 +109,14 @@ else if(sparsePtr.val != null)
...
 	 * @param dirty return dirty pointers if true
+	 * @param isCleanupEnabled return pointers marked for cleanup if true
 	 * @return set of pointers
 	 */
-	Set<Pointer> getPointers(boolean locked, boolean dirty) {
-		return gpuObjects.stream().filter(gObj -> gObj.isLocked() == locked && gObj.isDirty() == dirty).flatMap(gObj -> getPointers(gObj).stream()).collect(Collectors.toSet());
+	Set<Pointer> getPointers(boolean locked, boolean dirty, boolean isCleanupEnabled) {
+		return gpuObjects.stream().filter(
+				gObj -> (gObj.isLocked() == locked && gObj.isDirty() == dirty) ||
+						(gObj.mat.isCleanupEnabled() == isCleanupEnabled)).flatMap(
```
niketanpansare: @thomas9t  can you please double-check this logic with other APIs?

thomas9t: @niketanpansare sorry for the slow movement on this. Cleanup is disabled on inputs to MLContext scripts, so we would also not be cleaning up GPU pointers associated with these objects. However, I think this is appropriate as it's possible to invoke an ML context script multiple times as well (as in JMLC).

thomas9t:
@niketanpansare - anything else we need to do here?

PR #842, opened 2018-11-06
 [SYSTEMML-1325] Fix a performance bug when using GPU backend with JMLC

niketanpansare:
- This PR simplifies the clearTemporaryMemory logic in GPU memory manager and performs aggressive cleanup to reduce the memory pressure.

@thomas9t can you please review this PR?

@bertholdreinwald

### DIFF
--- a/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMemoryManager.java
+++ b/src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMemoryManager.java
@@ -39,6 +39,7 @@ import org.apache.sysml.conf.ConfigurationManager;
 import org.apache.sysml.conf.DMLConfig;
 import org.apache.sysml.hops.OptimizerUtils;
 import org.apache.sysml.runtime.DMLRuntimeException;
+import org.apache.sysml.runtime.controlprogram.caching.MatrixObject;
 import org.apache.sysml.runtime.instructions.gpu.GPUInstruction;
 import org.apache.sysml.utils.GPUStatistics;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/sysml/runtime/instructions/gpu/context/GPUMemoryManager.java
```diff
@@ -518,21 +519,28 @@ public void clearMemory() {
...
-	public void clearTemporaryMemory() {
-		// To record the cuda block sizes needed by allocatedGPUObjects, others are cleared up.
-		Set<Pointer> unlockedDirtyPointers = matrixMemoryManager.getPointers(false, true, false);
-		Set<Pointer> temporaryPointers = nonIn(allPointers.keySet(), unlockedDirtyPointers);
+	public void clearTemporaryMemory(HashSet<MatrixObject> outputMatrixObjects) {
+		Set<Pointer> donotClearPointers =  new HashSet<>();
+		// First clean up all GPU objects except:
+		// 1. Output matrix objects
+		// 2. GPU objects that are currently being used (i.e. locked)
+		// 3. Matrix object are
```
thomas9t: Just for the sake of clarity might want to comment that we also do not clean up pinned inputs

thomas9t:
Thanks @niketanpansare looks good to me - :+1: A quick comparison indicates that JMLC is now in the ballpark of TF/Keras at small batch sizes. However, at larger batch sizes we're still a bit slower - I have not yet had a chance to dig into what exactly is going on in detail. However, `nvprof` indicates that the most costly operations are convolution followed by device to host transfer, so there may still be some unnecessary copying going on. We can address this in follow up though. See below for numbers (all computation in single precision):

| Batch Size | SystemML (Sec.) | TensorFlow (Sec.) |
|------------|-----------------|-------------------|
| 1          | 0.031           | 0.024             |
| 2          | 0.043           | 0.021             |
| 4          | 0.076           | 0.037             |
| 8          | 0.137           | 0.076             |
| 16         | 0.257           | 0.146             |
| 32         | 0.468           | 0.273             |
| 64         | 0.818           | 0.426             |
| 128        | 2.97            | 0.849             |

PR #844, opened 2018-11-08
[MINOR] Added an external UDF to split string

niketanpansare:
- Also, updated ListObject to specify the valuetype of the list. This takes care of the "wrong value type warning".

@bertholdreinwald can you please review this PR?


bertholdreinwald:
thanks for fixing the valuetype. Now that returning a list of strings works, please add a limits parameter to fully exploit the Java function.

split = externalFunction(string s, string regex, integer limits) return (list[string] pl) implemented in (classname="org.apache.sysml.udf.lib.SplitWrapper", exectype="mem");

parts = split ("foo_boo_goo", "_", 5);

for (i in 1:length(parts))
{
    print ("p" + i + ": \"" + as.scalar(parts[i]) + "\"");
}

niketanpansare:
Done.

PR #854, opened 2019-03-04
[SYSTEMML-1325] Fix GPU cleanup bug when invoked with JMLC.

niketanpansare:
@thomas9t @bertholdreinwald @mboehm7 Can either of you please review this PR?


thomas9t:
@niketanpansare - this looks good to me and I confirm resolves the performance issues we noted earlier. I had originally thought we should extend this to objects allocated on the CPU as well. However, `rmvar` already removes references to these objects from the symbol table, so I think they can be garbage collected by the JVM under memory pressure. Do you agree?

niketanpansare:
Thanks @thomas9t. I will merge 👍 

I agree. We don't need to extend this logic to CPU objects.

PR #860, opened 2019-03-28
[SYSTEMML-2525] Initial implementation of RESTful model serving system

niketanpansare:
- Creating this PR on behalf of @thomas9t and hence will be merged with @thomas9t as the author.
- The current implementation extends JMLC's `readMatrix` and `GPUContext` API.
- The serving system is implemented in Scala using Akka and is available in the `org.apache.sysml.api.ml.serving`.
- Minor cleanup and refactoring required before it's ready to be used by the general public will be done in subsequent commits.
- It still remains unclear whether CUDA and Serving code should be included in future standalone releases. If yes, it will greatly simplify the deployment, else the user will have to build standalone jar before deployment.
- The serving system can be started by:
```
mvn -Djcuda.scope=compile -Dserving.scope=compile package -P standalone-jar
java -jar systemml-*-standalone.jar org.apache.sysml.api.ml.serving.PredictionService -port 8099 -scheduler scheduler -admin_password admin
```
- The model can be then registered using `http://localhost:8099/register-model` and user can invoke prediction using `http://localhost:8099/predict` service.

@thomas9t @bertholdreinwald Please review this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ before_script:
 
 script:
 #  - mvn clean verify jacoco:report coveralls:report
-  - mvn clean verify
+# The -q parameter tells mvn not to display anything other than ERROR level log messages. This is required because travis kills the job after the log length exceeds its maximum log length (usually 4 MB).
+  - mvn -q clean verify
 
 after_success:
 #  -  mvn test jacoco:report coveralls:report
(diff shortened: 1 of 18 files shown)


bertholdreinwald:
LGTM

thomas9t:
👍 LGTM. Thanks for creating the PR. I will follow up with a list of remaining TODOs for cleanup/refactoring.

niketanpansare:
Thanks @thomas9t ... I will merge as soon as the tests are completed 👍

PR #862, opened 2019-04-12
Systemml 2526 Date.getTime() can be changed to System.currentTimeMillis()

bd2019us:
new Date() is just a thin wrapper around System.currentTimeMillis(). Using System.currentTimeMillis() can help speed up the system

### DIFF
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
@@ -334,7 +334,7 @@ public class MLContext implements ConfigurableAPI
 		try {
 			executionScript = script;
 
-			Long time = System.currentTimeMillis();
+			Long time = new Long(System.currentTimeMillis());
 			if ((script.getName() == null) || (script.getName().equals(""))) {
 				script.setName(time.toString());
 			}
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
@@ -334,7 +334,7 @@ public class MLContext implements ConfigurableAPI
 		try {
 			executionScript = script;
 
-			Long time = new Long((new Date()).getTime());
+			Long time = System.currentTimeMillis();
 			if ((script.getName() == null) || (script.getName().equals(""))) {
 				script.setName(time.toString());
 			}
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -1255,7 +1255,7 @@ public final class MLContextUtil {
 			}
 			scriptExecutor.setMaintainSymbolTable(mlCtx.isMaintainSymbolTable());
 
-			Long time = new Long((new Date()).getTime());
+			Long time = System.currentTimeMillis();
 			if ((script.getName() == null) || (script.getName().equals(""))) {
 				script.setName(time.toString());
 			}
--- a/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
+++ b/src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
@@ -1255,7 +1255,7 @@ public final class MLContextUtil {
 			}
 			scriptExecutor.setMaintainSymbolTable(mlCtx.isMaintainSymbolTable());
 
-			Long time = System.currentTimeMillis();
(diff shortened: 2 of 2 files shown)


bd2019us:
Example
Hello, please see the example program.

xtern:
> Hello, please see the example program.

Hello! I re-checked these results with JMH on Ubuntu (jdk8) and found that the difference is not that significant: currentTimeMillis faster for 0.02 nanosecond per operation (less than the measurement error).
```
Benchmark             Mode  Cnt   Score   Error  Units
TimeTest.checkDate    avgt   10  22.733 ± 0.106  ns/op
TimeTest.checkMillis  avgt   10  22.753 ± 0.106  ns/op
```
code
```java
import java.util.Date;
import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Threads;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

@BenchmarkMode(Mode.AverageTime)
@Fork(value = 1, jvmArgsAppend = {"-Xms3g", "-Xmx3g", "-server", "-XX:+AggressiveOpts"})
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
@Threads(1)
@Warmup(iterations = 10)
@Measurement(iterations = 20)
public class TimeTest {
    @Benchmark
    public long checkDate() {
        return new Date().getTime();
    }

    @Benchmark
    public long checkMillis() {
        return System.currentTimeMillis();
    }

    public static void main(String[] args) throws RunnerException {
        final Options options = new OptionsBuilder()
            .include(TimeTest.class.getSimpleName())
            .build();

        new Runner(options).run();
    }
}
```

ex00:
Hi,
I assume that need to pay attention to converting time to string object too. 
My small test: 
```java
@Benchmark
    public String longObjectToString() {
        Long time = new Long(System.currentTimeMillis());
        return time.toString();
    }

    @Benchmark
    public String longToString() {
        return Long.toString(System.currentTimeMillis());
    }

    @Benchmark
    public String stringValueOf() {
        return String.valueOf(System.currentTimeMillis());
    }
```
results:
```
Benchmark                                                    Mode       Cnt        Score      Error   Units
MyBenchmark.longObjectToString                              thrpt        30        0.020 ±    0.001  ops/ns
MyBenchmark.longToString                                    thrpt        30        0.021 ±    0.001  ops/ns
MyBenchmark.stringValueOf                                   thrpt        30        0.021 ±    0.001  ops/ns
MyBenchmark.longObjectToString                               avgt        30       47.306 ±    0.048   ns/op
MyBenchmark.longToString                                     avgt        30       47.263 ±    0.226   ns/op
MyBenchmark.stringValueOf                                    avgt        30       47.002 ±    0.075   ns/op
```

src/main/java/org/apache/sysml/api/mlcontext/MLContext.java
```diff
@@ -334,7 +334,7 @@ public MLResults execute(Script script, ScriptExecutor scriptExecutor) {
 		try {
 			executionScript = script;
 
-			Long time = new Long((new Date()).getTime());
+			Long time = new Long(System.currentTimeMillis());
```
ex00: I suggest, better to move `time` variable to `if` block, because it uses only for setting script name

src/main/java/org/apache/sysml/api/mlcontext/MLContextUtil.java
```diff
@@ -1255,7 +1255,7 @@ public static String getHopDAG(MLContext mlCtx, Script script, ArrayList<Integer
 			}
 			scriptExecutor.setMaintainSymbolTable(mlCtx.isMaintainSymbolTable());
 
-			Long time = new Long((new Date()).getTime());
+			Long time = new Long(System.currentTimeMillis());
```
ex00: same

mboehm7:
LGTM - thanks @bd2019us, @ex00. Even though that's not in a performance-relevant path, it doesn't hurt to make this change.

PR #863, opened 2019-04-13
[SYSTEMML-2528] Change String.equals to Objects.equals to avoid NullPointerException

bd2019us:
Hello,
The String "execType" and "opcode" in the two files may have potential risk of 
NullPointerException since they are immediately used after initialization. One recommended API is Objects.equals(String,String) which can avoid this exception.


j143:
Closing this due to inactivity. Please reopen if needed. Thanks for contributing.

If you would like to contribute to the project 
1. Browse https://issues.apache.org/jira/projects/SYSTEMML/issues
2. or start a discussion on mailing list at dev@systemds.apache.org

PR #866, opened 2020-03-01
[SYSTEMML-2533] Error in running MNIST LeNet algorithm

HannanKan:
Fix bug in issue [SYSTEMML-2533]  https://issues.apache.org/jira/projects/SYSTEMML/issues/SYSTEMML-2533?filter=allopenissues.  This bug is caused by wrong DML script in `systemml-master/scripts/nn/examples/mnist_lenet.dml`.  I explain it in JIRA "https://issues.apache.org/jira/projects/SYSTEMML/issues/SYSTEMML-2533?filter=allopenissues"  comment in detail.

### DIFF
--- a/scripts/nn/examples/mnist_lenet.dml
+++ b/scripts/nn/examples/mnist_lenet.dml
@@ -118,13 +118,13 @@ train = function(matrix[double] X, matrix[double] Y,
                                                 stride, stride, pad, pad)
       outr1 = relu::forward(outc1)
       [outp1, Houtp1, Woutp1] = max_pool2d::forward(outr1, F1, Houtc1, Woutc1, Hf=2, Wf=2,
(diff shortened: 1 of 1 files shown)


mboehm7:
LGTM. Thanks for catching this issue @HannanKan. You're right and the reason for the conflict is that the related script max_pool2d_builtin::forward defines padh and padw but no pad. So, it might better to simply change the argument names at the calling sites.

mboehm7:
Also could you please associate your github account with the email of your commit or change your commit so you get properly linked when merging this in. Thanks.

HannanKan:
> LGTM. Thanks for catching this issue @HannanKan. You're right and the reason for the conflict is that the related script max_pool2d_builtin::forward defines padh and padw but no pad. So, it might better to simply change the argument names at the calling sites.

You are right. I have change the parameter name in the new pull request .

PR #868, opened 2020-03-25
[SYSTEMML-2537] Improve the consistency of predicates in "if-then-else" expressions

raymondtay:
The original issue i noticed was when the conditional predicate in the if-expression didn't seem what i was used to as in the C-programming language (where negative values were interpreted as _false_ while the converse values were interpreted as _true_) and that caused some discomfort.

Below is an example of the issue
```
// This is a session i ran from running SystemML
scala> import org.apache.sysml.api.mlcontext.ScriptFactory._
scala> val ml = new MLContext(spark)
scala> val s = dml(""" if(-42) print("Yes!") else print("No!") """)
scala> ml.execute(s)
Yes!
```
In the Scala programming language, this would be a type-error
```
scala> if (-31) println("yes") else println("no")
           ^
       error: type mismatch;
        found   : Int(-31)
        required: Boolean
```

So i started mucking around _Antlr4_ and start poking around and discovered a potential fix for it. Running this fix after a build on my local machine, here's the desired result:
```
Using Scala version 2.11.12 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_202)

scala> import org.apache.sysml.api.mlcontext.ScriptFactory._

scala> val ml = new MLContext(spark)

Welcome to Apache SystemML!
Version 1.3.0-SNAPSHOT
ml: org.apache.sysml.api.mlcontext.MLContext = org.apache.sysml.api.mlcontext.MLContext@380cd266

scala> val s = dml(""" if(-42) print("Yes!") else print("No!") """)
s: org.apache.sysml.api.mlcontext.Script =
Inputs:
None

Outputs:
None

scala> ml.execute(s)
ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.7ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.7ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.7ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.720/03/25 09:10:54 WARN DMLTranslator: WARNING: [line 1:4] -> -42 -- Numerical value '-42' (!= 0/1) is converted to boolean FALSE by DML
No!
SystemML Statistics:
Total execution time:		0.010 sec.

res0: org.apache.sysml.api.mlcontext.MLResults =
None

```

### DIFF
--- a/src/main/java/org/apache/sysml/parser/DMLTranslator.java
+++ b/src/main/java/org/apache/sysml/parser/DMLTranslator.java
@@ -1546,15 +1546,28 @@ public class DMLTranslator
 			} else if ((predicate instanceof IntIdentifier && ((IntIdentifier) predicate).getValue() == 1)
 					|| (predicate instanceof DoubleIdentifier && ((DoubleIdentifier) predicate).getValue() == 1.0)) {
 				cp.setPredicate(new BooleanIdentifier(true, predicate));
-			} else if (predicate instanceof IntIdentifier || predicate instanceof DoubleIdentifier) {
-				cp.setPredicate(new BooleanIdentifier(true, predicate));
-				LOG.warn(predicate.printWarningLocation() + "Numerical value '" + predicate.toString()
-						+ "' (!= 0/1) is converted to boolean TRUE by DML");
-			} else if (predicate instanceof StringIdentifier) {
-				throw new ParseException(predicate.printErrorLocation() + "String value '" + predicate.toString()
-						+ "' is not allowed for iterable predicate");
-			}
-			predicateHops = processExpression(cp.getPredicate(), null, _ids);
+                        } else if (predicate instanceof IntIdentifier) {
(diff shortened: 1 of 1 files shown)


mboehm7:
thanks for creating the issue, but I'm afraid that DML follows R semantics here where both if conditions and ifelse functions interpret negative values as FALSE. Do you think adding this to the docu would help.

Btw, in the next days we'll merge SystemDS back into Apache SystemML with many new features and a packed roadmap.

raymondtay:
> thanks for creating the issue, but I'm afraid that DML follows R semantics here where both if conditions and ifelse functions interpret negative values as TRUE. Do you think adding this to the docu would help.
> 
> Btw, in the next days we'll merge SystemDS back into Apache SystemML with many new features and a packed roadmap.

I think adding the documentation to SystemML would definitely clarify the issue. Thank you for enlightening me on this fact w.r.t **R programing language**.

PR #869, opened 2020-03-28
[MINOR] GitHub Workflows Cache Dependencies

Baunsgaard:
Workflows cache dependencies.
In small test cases this reduce the time to execute tests to half.
Especially true in the python tests cases.

Note:
- Cache size is at max 5 GB for a repository, and it will slowly build up.
- Cache change if changes happen in
  - pom.xml for maven dependencies.
  - src/main/python/setup.py or src/main/python/docs/requires-docs.txt for python.

First Pull request Apache ! :+1: :smiley: :racehorse:


mboehm7:
LGTM - thanks @Baunsgaard.

PR #871, opened 2020-03-29
[SYSTEMDS-315] Python Federated Matrices

Baunsgaard:
This commit contains:

- Easy start of federated worker in /bin/systemds.sh
- Setup of tests for the Python language bindings federated matrices
- Initial tutorial on Python federated matrices
- Minor :bug: fix in federated matrix, not allowing multiple sources
- Github workflow action for automated federated tests


Baunsgaard:
It would be good to have a minor release with at least the federated worker startup part of this commit.

Baunsgaard:
Also containing documentation update such that it is using relative paths to documentation.
This makes the references work in different forks, pointing to the forks docs.

Baunsgaard:
Note, some tests are failing in python federated.
This is intentionally left in, so that we at some point can go back and fix the underlying functionality.

1. bug is if you create a federated matrix that is not 100% filled in by federated workers.
2. bug is if you create a federated matrix that is larger than the input size, it crashes, while expected behavior is to pad the matrix with 0's for consistent behavior.
3. feature/bug, currently if a single federated matrix is having overlapping sub matrices, see tests ["federated basic test"](/src/main/python/tests/federated/test_federated_basic) for examples, the system does not complain, and in some cases even works by overwriting the matrix values of the first selected federated matrix with the later matrices.

docker/pythonsysds.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+#
+#-------------------------------------------------------------
+
+FROM openjdk:8-alpine
+
+RUN apk add --no-cache --virtual .build-deps g++ python3-dev libffi-dev openssl-dev && \
+    apk add --no-cache python3 && \
+    pip3 install --upgrade pip setuptools
+
+RUN pip3 install systemds
```
corepointer: The github code browser widget suggests a newline at eof

src/main/python/tests/federated/runFedTest.sh
```diff
@@ -0,0 +1,65 @@
...
+#FIELDS
+workerdir="tests/federated/worker/"
+outputdir="tests/federated/output/"
+mkdir $workerdir
+mkdir $outputdir
+w1_Output="$workerdir/w1"
+w2_Output="$workerdir/w2"
+log="$outputdir/out.log"
+
+# make the workers start quiertly
```
corepointer: typo: quietly

docker/pythonsysds.Dockerfile
Baunsgaard: Sure Thanks

src/main/python/tests/federated/runFedTest.sh
Baunsgaard: Done!

Baunsgaard:
> LGTM except for the failing tests. In your comment, you mention that the failing tests are left in as a reminder to fix some bugs. Why not create a separate PR marked WIP or an issue in the tracker for that?

it makes sense to have tests fail showing the bugs, especially because it is functionality that we 'claim' to support in master.

i also think that @kev-inn would be suited for the fixing the bug. Therefore there i see two ways of going forward, either wait for that fix while being a pull request, or allow the pull request in that shows a bug in the system now. (i vote for the later (not that i have a vote) :smile: ).

corepointer:
> 
> it makes sense to have tests fail showing the bugs, especially because it is functionality that we 'claim' to support in master.
> 

Yes, failing tests should show bugs (in the code, not in the tests ;-) ) Since the bugs are already in master, we can merge the tests to point that out. But we should make an exception in situations like this one and not have new code merged with failing tests.

Baunsgaard:
@mboehm7 
I have commented out the "advanced" federated matrix tests.
This means it now verifies that the python federated basics work, and we can extend from that.

Baunsgaard:
Fixed pull request to use the brand new SystemDSContext !! Thanks @kev-inn !

kev-inn:
Ok, so currently we have the bugs:
1. ~~bug is if you create a federated matrix that is not 100% filled in by federated workers, then operations does not take into account the remaining cells.~~ This should be fixed, I committed the fix to your sb_fed_python branch, seems like the change didn't make it in to this PR. Should we do a separate one?
2. bug is if you create a federated matrix that is larger than the input size, it crashes, while expected behavior is to pad the matrix with 0's for consistent behavior, with other functionality.
3. ~~feature~~/bug, currently if a single federated matrix is having overlapping sub matrices, see tests "federated basic test" for examples, the system does not complain, and in some cases even works by overwriting the matrix values of the first selected federated matrix with the later matrices.

I am not sure if the second bug is actually a bug or a problem of our function definition. Our definition (the way I implemented it) expects the ranges to match the actual sizes of the matrices on the workers. And the actual size of the federated matrix goes from (0,0) to (max row, max col) given in the ranges, making it therefore not possible to get 0 rows and 0 cols after our ranges are finished. It is also bad that I forgot some very important checks to actually throw errors if the function definition is used wrong (3. could be regarded as fixed if we just check for overlaps).

Baunsgaard:
>     1. ~bug is if you create a federated matrix that is not 100% filled in by federated workers, then operations does not take into account the remaining cells.~ This should be fixed, I committed the fix to your sb_fed_python branch, seems like the change didn't make it in to this PR. Should we do a separate one?

I don't see the change, did you commit to an old repo?

kev-inn:
> > ```
> > 1. ~bug is if you create a federated matrix that is not 100% filled in by federated workers, then operations does not take into account the remaining cells.~ This should be fixed, I committed the fix to your sb_fed_python branch, seems like the change didn't make it in to this PR. Should we do a separate one?
> > ```
> 
> I don't see the change, did you commit to an old repo?

I committed to git@github.com:Baunsgaard/systemml.git on the branch sb_fed_python.

Baunsgaard:
> > > ```
> > > 1. ~bug is if you create a federated matrix that is not 100% filled in by federated workers, then operations does not take into account the remaining cells.~ This should be fixed, I committed the fix to your sb_fed_python branch, seems like the change didn't make it in to this PR. Should we do a separate one?
> > > ```
> > 
> > 
> > I don't see the change, did you commit to an old repo?
> 
> I committed to [git@github.com](mailto:git@github.com):Baunsgaard/systemml.git on the branch sb_fed_python.

This pull request is linked to  fedPython branch , I made the other branch because i wanted a branch containing all the tests, (including the ones that shows the bugs), while this one in this pull request only have the minimum requirements parsing.

mboehm7:
LGTM - that's awesome. Thanks for the thoughtful addition of tests, docs, and scripts. 

Minor things for future modifications: First, maybe it make sense to add a badge for the federated python tests to our readme as well. Second, the docker examples currently refer to 'sebaba' - maybe we can make this more generic.

PR #872, opened 2020-03-29
[SYSTEMDS-273] Refactor Compressed Package

Baunsgaard:
Hi, I think this is ready to merge :grinning:, @mboehm7 , and I will continue work in other branches :factory_worker:.

Note that tests of:

Sample based estimation tests are disabled, but can be enabled inside:
`src/test/java/org/apache/sysds/test/component/compress/colgroup/JolEstimateTest.java`

DDC share dictionary is disabled, but can be enabled inside:
`src/test/java/org/apache/sysds/test/component/compress/CompressedTestBase.java`

both are disabled because of :bug:.

Further work is added as tasks inside `/docs/task`.

SYSTEMDS-270 Compressed Matrix Blocks
 * 273a Redesign allocation of ColGroups in ColGroupFactory
 * 274 Make the DDC Compression dictionary share correctly
 * 275 Include compressionSettings in DMLConfiguration
 * 276 Allow Uncompressed Columns to be in sparse formats
 * 277 Sampling based estimators fix
 * 278 Compression-CoCode algorithm optimization
 * 278a Return ColGroups estimated compression ratio to Factory
 * 279 Add missing standard lossless compression techniques
 * 279a ColGroup FOR (Frame of reference) encoding
 * 279b ColGroup DEL (Delta) encoding
 * MINOR Reduce memory usage for compression statistics.
 * MINOR Make ContainsAZero() method in UncompressedBitMap

SYSTEMDS-360 Lossy Compression Blocks
 * 361 ColGroup Quantization
 * 362 ColGroup Base Data change (from Double to ??)

COMMIT MESSAGE:

Major refactor of compressed Matrix Block to "simplify" responsibilities.
The main goal of the refactoring was to reduce the amount of bindings
to CompressedMatrixBlock from the different types of ColGroups, to enable
easy addition of new types of groups. (without reducing performance!)

A list of new groups and other tasks associated with this is added to
/docs/tasks.txt .

Furthermore, the tests added for the compression work based on an
estimation tolerance, allowing the estimations inside the system to be
off by a set factor. In the tests, this has to be expanded upon to
enable better precision in how the estimations are calculated.

Bellow is an extract of the different changes:

- Separated sub-parts of compression into different packages.
- Array memory footprint worst case calculations.
- Moved Compressed Size Estimation Calculation to specific ColGroups

- Extensive testing of size Estimation of ColGroups and compression
  - Jol Memory Estimate tests for compression blocks
  - Using worst case Jol Estimate JVM using uncompressed 64-bit JVM
  - Ideal input generator for testing Col groups compression.

- Factory pattern added for selected constructors
  - ColGroups
    - NameChange from ColGroupCompressor to ColGroupFactory
  - CompressedMatrixBlock

- Enable the parallel execution of the ColGrouping (previously disabled)

- Settings File added for Compression to enable selection of specific
  compression types. This can be extended with settings loaded from
  DMLConfig
  - This is constructed using a builder pattern such that invocation is
  private.

- Added abstract compressed block for overwriting default MatrixBlock

- Add Test Libs to pom.xml
  - Memory estimator framework JOL from OpenJDK to measure object sizes


Baunsgaard:
Major refactor of compressed Matrix Block to "simplify" responsibilities.
Many changes in compression planning especially in memory estimation.

Feedback appreciated :+1:

The remaining errors, are mainly in the sparse estimation of compression sizes. But include

- Sparse estimation of Number of Distinct values is off when the input is sparse the wrong sample based estimators are used.
- The bitmaps encoded for extracting column facts does not contain information of if there is a 0 present in the column.
- A bug (intend to fix today) in unary operators when compressing with a specific compression scheme.

Hopefully if we merge the bugs mentioned above can be fixed within reasonable time.

Bellow is an extract of the different changes:

- Separated sub-parts of compression into different packages.
- Array memory footprint worst case calculations.
- Moved Compressed Size Estimation Calculation to specific ColGroups

- Extensive testing of size Estimation of ColGroups and compression
  - Jol Memory Estimate tests for compression blocks
  - Using worst case Jol Estimate JVM using uncompressed 64-bit JVM
  - Ideal input generator for testing Col groups compression.

- Factory pattern added for selected constructors
  - ColGroups
    - NameChange from ColGroupCompressor to ColGroupFactory
  - CompressedMatrixBlock

- Enable the parallel execution of the ColGrouping

- Settings File added for Compression to enable selection of specific
  compression types.

- Added abstract compressed block for overwriting default MatrixBlock

- Add Test Libs to pom.xml
  - Memory estimator framework JOL from OpenJDK to measure object sizes

Baunsgaard:
Furthermore there are some remaining issues in the estimation of joined colGroups, containing multiple columns that estimate to large sizes.

mboehm7:
LGTM - overall this is great @Baunsgaard. During the merge, I just fixed a few typos, and minor warnings (generics, static methods).

Additionally, you might want to consider moving the `ColGroupSizes` into the class hierarchy of column groups to avoid string-based dispatching (including abstract classes that could never be instantiated) and bring this closer to their implementation. Also, please reconsider the unnecessary indirection of `AbstractCompressedMatrixBlock` and `CompressedMatrixBlockFactory`. Is there any intention behind them other than splitting `CompressedMatrixBlock` into smaller files?

Baunsgaard:
> LGTM - overall this is great @Baunsgaard. During the merge, I just fixed a few typos, and minor warnings (generics, static methods).

Thanks @mboehm7  for the fixes :+1:, our editors detect different things, so I think it is a fine trade-off unless it required significant changes :smile:.

> Additionally, you might want to consider moving the `ColGroupSizes` into the class hierarchy of column groups to avoid string-based dispatching (including abstract classes that could never be instantiated) and bring this closer to their implementation. 

Yes I could move it back into the specific ColGroup classes, but the important thing is that they are static methods, not requiring you to construct an object of the ColGroup to calculate the sizes. Therefore I had some difficulties because static methods if can't be overwritten in java. Therefore I decided to have the logic separated into another class, instead of having "Semi Random" names for getting size of objects inside the individual classes. I hope you have a better suggestion, because I two do not like the ColGroupSizes class as it is now.

>Also, please reconsider the unnecessary indirection of `AbstractCompressedMatrixBlock` and `CompressedMatrixBlockFactory`. Is there any intention behind them other than splitting `CompressedMatrixBlock` into smaller files?

The reason was that I wanted to make an LossyMatrixBlock object, that also pointed to the abstract CompressedMatrixBlock, such that we had the option to select compression also based on Object type. I did this before our discussion on keeping everything colocated inside CompressedMatrixBlock,  but while adding the lossy compression logic I will reconsider what could be smartest. That said, i 100% agree that the separation right now is unnecessary.

mboehm7:
ok that makes a lot of sense - I though of lossy compression techniques as ColGroups but you're right it might be better to indicate that at global level (lossy if at least one lossy ColGroup).

PR #875, opened 2020-03-31
[WIP] Kmv k minimum values (Destinct value estimation algoritm)

Baunsgaard:
Kevin S. Beyer, Peter J. Haas, Berthold Reinwald, Yannis Sismanis, Rainer Gemulla: On synopses for distinct‐value estimation under multiset operations. SIGMOD 2007

I have tried the k minimum values algorithm in a basic setup. 

Feedback appreciated!

### DIFF
--- a/src/main/java/org/apache/sysds/common/Builtins.java
+++ b/src/main/java/org/apache/sysds/common/Builtins.java
@@ -170,6 +170,7 @@ public enum Builtins {
 	TANH("tanh", false),
 	TRACE("trace", false),
 	TYPEOF("typeOf", false),
+	UNIQUE_LENGTH("unique_length",false),
 	VAR("var", false),
 	XOR("xor", false),
 	WINSORIZE("winsorize", true, false), //TODO parameterize w/ prob, min/max val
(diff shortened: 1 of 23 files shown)


Baunsgaard:
Also including bit array data type utilizing long to contain the bits.
Initially i thought i needed it, but apparently the KVM does not use it.

j143:
Hi @Baunsgaard , let us move this forward. With each improvement, if we track the test results in this PR itself in the phase by phase manner (Might take a few weeks :), like with any good PR.). 

So, for this basic setup _shall we track_:
1. computational cost (how much lower?)
2. The accuracy of DV estimate!

Once we are satisfied with the results, we could request perhaps _Berthold Reinwald_ for a more involved review. Thanks.

Baunsgaard:
> Hi @Baunsgaard , let us move this forward. With each improvement, if we track the test results in this PR itself in the phase by phase manner (Might take a few weeks :), like with any good PR.).
> 
> So, for this basic setup _shall we track_:
> 
>     1. computational cost (how much lower?)
> 
>     2. The accuracy of DV estimate!
> 
> 
> Once we are satisfied with the results, we could request perhaps _Berthold Reinwald_ for a more involved review. Thanks.

currently this pr is on a hold, because of multiple reasons:

1. The intention was to implement a function and use it in the compression planning as well, so that we have more granularity in our estimations, and not relying only on the current compression sample estimators. Therefore i want to make the compression planning accurate first without estimations before coming back to these estimation algorithms.
2. The accuracy achieved with the current implementation of KMV so far achieves worse estimates than the ShlosserJackKnifeestimator inside the compression.
3. I'm working on HyperLogLog ( and variations) on the side to compare with this as well.

I see no reason to contact anyone, and thinks it will be counter productive in this specific case, since the target for these implementations at the moment is exploratory in nature.

Baunsgaard:
Closing Pull request, to be opened in other, (was easiest, to fix merge errors this way)

PR #880, opened 2020-04-08
[MINOR] [DOC] skip IntelliJ, and mvn build config files

j143:
This PR is for discussion and work related to few more add the to be ignored files.
Configuration files related to 
1. maven build
  - [x] dependency-reduced-pom.xml
  - [x] temp/
2. intellij
  - [x] .idea/

Feel free to add more extensions. thank you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+## Jekyll build files
+
 .classpath
 .jekyll-cache/
 .project
@@ -9,7 +11,16 @@ Gemfile.lock
 .settings/
 .vscode/
 
-## Inproper copy when switching branch
+## IntelliJ
+
+.idea/
+
+## Improper copy when switching branch
+
+src/
+
+## Maven build files
 
 target/
-src/ 
+temp/
+dependency-reduced-pom.xml


.gitignore
```diff
@@ -9,7 +11,16 @@ Gemfile.lock
 .settings/
 .vscode/
 
-## Inproper copy when switching branch
+## IntelliJ
+
+.idea/
+
+## Improper copy when switching branch
```
Baunsgaard, line 18: thanks for the correction :+1:

.gitignore
```diff
@@ -9,7 +11,16 @@ Gemfile.lock
...
+
+src/
+
+## Maven build files
 
 target/
-src/ 
\ No newline at end of file
+temp/
+dependency-reduced-pom.xml
```
Baunsgaard, line 26: dependency-reduced-pom is gitignored on main branch yes, but i don't think it is removed completely. I would suggest you remove it from your fork, unless it is needed for something.

.gitignore
```diff
@@ -1,3 +1,5 @@
+## Jekyll build files
```
Baunsgaard, line 1: It is not just Jekyll, it is also artifacts from mvn.

maybe split it up, if you want to mark it as such.

PR #883, opened 2020-04-09
Builtin "is_correct" for validation of data against a given schema

Shafaq-Siddiqi:
The built-in will validate the data in frame against a given schema. All non-conforming values will be converted into missing values (null in this case).
usage: is_correct(Frame[String] data_frame, Frame[String] schema); schema is a row vector 
ncol(data_frame) = ncol(schema).


src/main/java/org/apache/sysds/runtime/instructions/SPInstructionParser.java
```diff
@@ -39,55 +39,10 @@
...
-import org.apache.sysds.runtime.instructions.spark.MatrixReshapeSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.MultiReturnParameterizedBuiltinSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.PMapmmSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.ParameterizedBuiltinSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.PmmSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.QuantilePickSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.QuantileSortSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.QuaternarySPInstruction;
-import org.apache.sysds.runtime.instructions.spark.RandSPInstruction;
-import org.apache.sysds.runtime.instructions.spark.ReblockSPInstruction;
```
j143, line 74: A change for imports can be done in a separate PR with `[MINOR]` tag. Otherwise this would impede the PR progress.

So, for this PR only these _deletions_ be reverted.

mboehm7:
LGTM - thanks @Shafaq-Siddiqi, that's very useful for extending our data cleaning primitives. During the merge, I just changed the name to `dropInvalid`, fixed some formatting issues (e.g., place of java annotations) and created consistency between CP and Spark instructions.

PR #885, opened 2020-04-11
Initial design suggestion for an ONNX importer

lukas-jkl:
@mboehm7  PR for initial design thoughts on an ONNX importer. 
Since ONNX does support conditional operators (loop, if) I've tailored the design towards a command-line tool that generates a script as discussed.


mboehm7:
LGTM - I made some minor wording improvements, and added the export functionality (not necessarily part of your project). While building the Python command line tool, please already plan for exposing it in our Python API as well.

PR #887, opened 2020-04-14
Built-in for missing value imputation using Functional Dependency.

Shafaq-Siddiqi:
For two functionally dependent attributes S -> T the imputeFD will impute the missing values in T based on the information from S given a threshold value t.

### DIFF
--- a/scripts/builtin/imputeFD.dml
+++ b/scripts/builtin/imputeFD.dml
@@ -1,0 +1,105 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


scripts/builtin/imputeFD.dml
```diff
@@ -0,0 +1,105 @@
...
+  # Get the most frequent mapped value of Y 
+  ans = (ctab == rowMaxs(ctab)) * t(seq(1, ncol(ctab)))
+  tabMax = rowSums(ans) != (ncol(ans) * ((ncol(ans))+1)/2) # vector for controlling max(0)
+  filled = rowMaxs(ans) * tabMax 
+  for(i in 1:nrow(X))
+  {
+    index = as.scalar(XY[i,1])
+    Y[i,1] = as.scalar(filled[index,1])
+  }
+  imputed_Y = Y
```
mboehm7, line 104: This loop can be replaced via a simple selection matrix multiply:
`imputed_Y = table(seq(1,nrow(X)), XY[,1]) %*% filled`

scripts/builtin/imputeFD.dml
```diff
@@ -0,0 +1,105 @@
...
+ 
+  # remove the table column representing missing values
+  if(sum(missing_mask[,2]) > 0)
+    ctab = ctab[,1:ncol(ctab)-1]
+
+  ctab = ctab/(rowSums(ctab)) > threshold 
+  
+  # Get the most frequent mapped value of Y 
+  ans = (ctab == rowMaxs(ctab)) * t(seq(1, ncol(ctab)))
+  tabMax = rowSums(ans) != (ncol(ans) * ((ncol(ans))+1)/2) # vector for controlling max(0)
```
mboehm7, line 97: I don't understand this comparison with n*(n+1)/2, which looks like a condition for all pairs but why is this needed? Multiple values with the same max? If we only want to get the single max the previous line (`(ctab == rowMaxs(ctab)) * t(seq(1, ncol(ctab)))`) may get simplified to `rowIndexMax(ctab)`

mboehm7:
When I fixed the entire frame handling for built-in functions, it would be good to provide both functions for matrices and frames, where in case of matrices we can avoid the transformencode/decode.

PR #891, opened 2020-04-17
[SYSTEMDS-333, 337]Improved lineage cache eviction

phaniarnab:
This patch improves lineage cache eviction by taking into account the actual
execution time of instructions/functions. The disk I/Os start with default speeds and then eventually converge to the actual speed of the running system. The ordering policy is still
LRU.


phaniarnab:
This PR is ready to be merged in.

mboehm7:
LGTM - thanks for the fine tuning. 

Just for awareness: I would recommend to avoid statistics maintenance for simple cost formulas as the time measurement via `System.nanotime()` creates system calls and with that is much slower than these few floating point operations.

mboehm7:
During the merge, I fixed few warnings.

PR #894, opened 2020-04-21
Privacy metadata

sebwrede:
This pull request adds privacy constraints as metadata. The privacy constraints are propagated from input to result of matrix multiplications. The privacy constraints currently only contain a simple boolean, which represents whether the data may be shared or not, but more complex constraints will later be added to the PrivacyConstraint class.


j143:
Thanks for updating the license files. :)

A minor name change to `org.apache.sysds` might be required.

sebwrede:
I have replaced this pull request [with a new one](https://github.com/apache/systemml/pull/895).

PR #896, opened 2020-04-23
[MINOR] Python Lineage Trace Test Change

Baunsgaard:
[MINOR] Python Lineage Trace Test Change
    
Change to the way lineage trace tests are executed, such that instead of
having files to compare to, the python trace is compared to a trace made
from systemds directly.
    
Motivated by the fact that the previous tests were failing, because of
inconsistencies between new traces and old.


Baunsgaard:
Currently failing because python interface call operations with 1 thread only, while sysds calls with max number of threads.

Baunsgaard:
@phaniarnab comments, please?

src/main/python/tests/lineage/test_lineagetrace.py
```diff
@@ -0,0 +1,98 @@
...
+    def tearDown(self):
+        warnings.filterwarnings(
+            action="ignore", message="unclosed", category=ResourceWarning
+        )
+
+    def test_compare_trace1(self):  # test getLineageTrace() on an intermediate
+        m = sds.full((10, 10), 1)
+        m_res = m + m
+
+        python_trace = [x.strip() for x in m_res.get_lineage_trace().split("\n")]
```
Baunsgaard, line 51: Do we really want the trace to be returned as one large string as default, or something like a list?

phaniarnab: I guess that depends on what one wants to do with the trace. String is ubiquitous, woks in dml and python. Using other data structures might need different handling. But can be done if that helps the querying interface on top the lineage storage.

phaniarnab:
This looks very good. Thanks for making the test better. As discussed offline, comparing only the internal nodes might solve the problem of metadata mismatch.

Baunsgaard:
LGTM :grinning: 
@mboehm7 or @j143 , could you merge ? (It would fix the few issues we have on master now.)

mboehm7:
LGTM - thanks @Baunsgaard. In general this is fine, but in the future, it would be good to preserve the history for renamed/moved files (`src/main/python/tests/test_lineagetrace.py` to
`src/main/python/tests/lineage/test_lineagetrace.py`). What usually works is to first rename it, commit, and make the necessary changes in a subsequent commit.

j143:
Normally, the `refactor` option in the IDEs should preserve git history.

mboehm7:
Yes, git is able to detect plain refactorings - the problem usually arises, if a commit does refactoring AND makes major code changes.

Baunsgaard:
I will note it down, and try to make it so next time :)

PR #898, opened 2020-04-25
[MINOR][DOC] Append 'Apache' to 'SystemDS'

j143:
This is as per the discussion at [dev@systemml](https://www.mail-archive.com/dev@systemml.apache.org/msg01051.html)

[Live Preview](https://github.com/apache/systemml/blob/ecaba47314e73447963c69bd640b615f355582eb/README.md)


j143:
@Baunsgaard - any comments about badges?

j143:
Thank you. Yes, things like this are too difficult agree upon by everyone 😸 .

PR #902, opened 2020-04-27
Modify Python API Testcases (rand, lm)

juliale-15:
Just a few changes to the testcase (increasing tolerance) so that the tc doesn't fail randomly.
Remove mutiple definitions of run in python workflow file.


.github/workflows/python.yml
```diff
@@ -74,9 +74,10 @@ jobs:
           ${{ runner.os }}-pip-${{ matrix.python-version }}-
   
     - name: Install pip Dependencies
-      run: pip install numpy py4j wheel
-      run: pip install scipy
-      run: pip install sklearn
+      run: |
+        pip install numpy py4j wheel
+        pip install scipy
+        pip install sklearn
```
Baunsgaard, line 80: while merging change this to 

`run: pip install numpy py4j wheel scipy sklean`

mboehm7:
Thanks for patch @juliale-15 and for the review @Baunsgaard

PR #905, opened 2020-04-30
[SYSTEMDS-335] Weighted eviction policy

phaniarnab:
This patch contains a new eviction policy for lineage cache. A min-heap
based priority queue over a function of computation time and
size is maintained to define the order of evictions.The idea is to evict
large matrices, which take little time to recompute. This weighted
scheme significantly reduces the number of evictions (including disk spilling).
This patch also refactors the LineageCache class to hide the eviction
policy related maintenance.


mboehm7:
LGTM. Thanks for the patch @phaniarnab - I think this is a great direction. During the merge, I fixed the conflicts and simplified `LineageCacheEviction.isSparse` to get the sparsity flag from the actual object not an estimate over the objects metadata. Down the road, it would be good to further generalize the different eviction policies to better separate the different code paths (otherwise, it would get complex and convoluted very quickly) - I left a TODO in `makeSpace`.

PR #906, opened 2020-05-01
[MINOR] Refactor component tests (move layout tests into componentx)

mboehm7:
Just for testing if moving all layout tests into an excluded `componentx` would fix the github workflows


mboehm7:
ok, the following reconfiguration seems to have fixed it - next should be individual ablation studies.

```
<reportFormat>brief</reportFormat>
<shutdown>kill</shutdown>
```

PR #908, opened 2020-05-10
Slicing review fixes

gilgenbergg:
Fixes according to last @mboehm7 review:
- Broadcast handling with reading from local variables by driver
- Removal of inner loops in filtering by mask and checking validity of node (union approach)
- Removal of exceptions handling in method that updates bounds


mboehm7:
LGTM - thanks @gilgenbergg for addressing the comments.

PR #911, opened 2020-05-13
[SYSTEMDS-344]Unmark functions/SBs containing non-determinism

phaniarnab:
This patch moves the fragile and less efficient non-determinism check in runtime to compile time. This adds a new IPA rewrite to unmark the functions and StatementBlocks containing direct or transitive nondeterministic calls (e.g. rand with UNSPECIFIED_SEED) for lineage caching.


mboehm7:
LGTM - thanks @phaniarnab. The only missing piece was to copy this information during parfor worker creation (where we deep copy functions and statement blocks to allow to independent, and thus uncontended recompilation) - dropping it would have potentially created incorrect results. Other than that and a few unnecessary imports this is great!

PR #915, opened 2020-05-18
[SYSTEMDS-335]Updated weighted scheme for lineage cache eviction.

phaniarnab:
This patch updates the weighted scheme by adding an elaborated scoring function. The function has two components - a ratio of compute time, in-memory size, and a last-used timestamp. The components are associated with weights, which can tune the eviction policies (e.g. weights 0 and 1 for time/size and timestamp respectively translate to LRU scheme). This patch also replaces the earlier PriorityQueye by a TreeSet.

### DIFF
--- a/docs/Tasks.txt
+++ b/docs/Tasks.txt
@@ -268,7 +268,7 @@ SYSTEMDS-330 Lineage Tracing, Reuse and Integration
  * 332 Parfor integration with multi-level reuse                      OK
  * 333 Improve cache eviction with actual compute time                OK
  * 334 Cache scalars only with atleast one matrix inputs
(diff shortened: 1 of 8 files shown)


mboehm7:
Great - thanks @phaniarnab, I like the generalization, but could you please also add a test that systematically uses all available cache eviction policies (via some use case that quickly fills the available lineage cache)?

phaniarnab:
This patch is ready to be merged in.

mboehm7:
LGTM - thanks for adding the test. During the merge I only fixed some minor formatting issues (tabs between code and comments, ordering of members, enums, static initializers, and methods) as well as a slight modification of DNN tests to make them less prone to surefire failures.

PR #916, opened 2020-05-19
Add to_one_hot builtin function

deutschmn:
Adds a builtin function `to_one_hot` which transforms a vector containing integers into a one-hot-encoded matrix


scripts/builtin/to_one_hot.dml
```diff
@@ -0,0 +1,42 @@
...
+# Output: 
+# --------------------------------------------------------------------------------------------
+# NAME          TYPE     MEANING
+# -------------------------------------------------------------------------------------------
+# Y             matrix   one-hot-encoded matrix with shape (N, num_classes)
+# -------------------------------------------------------------------------------------------
+
+m_to_one_hot = function(matrix[double] X, integer num_classes)
+        return (matrix[double] Y) {
+    assert(num_classes >= max(X));
```
Baunsgaard, line 40: this would be nice if changedto:
```
if(condition)
  stop("Good Error Message")
```

scripts/builtin/to_one_hot.dml
```diff
@@ -0,0 +1,42 @@
...
+    Y = table(seq(1, nrow(X)), X, nrow(X), num_classes);
+}
```
Baunsgaard, line 42: new line in end of file would be nice.

src/main/java/org/apache/sysds/common/Builtins.java
```diff
@@ -173,6 +173,7 @@
 	TAN("tan", false),
 	TANH("tanh", false),
 	TRACE("trace", false),
+	TO_ONE_HOT("to_one_hot", true),
```
Baunsgaard, line 176: This is a very Pythonic way of naming the function, I Just looked at the other function definitions and it is not very consistent in the naming. But the few fundamental functions taken from R is named using camelCase. I would suggest using that.
Any opinions @Shafaq-Siddiqi , is there an official decision?

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
```diff
@@ -0,0 +1,77 @@
+package org.apache.sysds.test.functions.builtin;
```
Baunsgaard, line 1: Missing License

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final String TEST_CLASS_DIR = TEST_DIR + BuiltinToOneHotTest.class.getSimpleName() + "/";
+
+    private final static double eps = 0;
+    private final static int rows = 10;
+    private final static int cols = 1;
+    private final static int numClasses = 10;
+
+    @Override
+    public void setUp() {
+        addTestConfiguration(TEST_NAME,new TestConfiguration(TEST_CLASS_DIR, TEST_NAME, new String[]{"B"}));
```
Baunsgaard, line 26: I would say that one test case of a "nice" input is insufficient. If you want to see behavior, since i just added a Buildin function myself i suggest something in the direction of (Up for debate):
https://github.com/apache/systemml/blob/e80145b344068d68e53094e437524108d5f0f00a/src/test/java/org/apache/sysds/test/functions/builtin/BuiltinConfusionMatrixTest.java

Baunsgaard:
Also it seems like the commit is not claimed by you. This means that you wont accumulate credit associated with your account. So you might want to fix that.

src/main/java/org/apache/sysds/common/Builtins.java
deutschmn: I have got no particular opinion on this, but I found a function called `one_hot` in an R package: https://www.rdocumentation.org/packages/mltools/versions/0.3.5/topics/one_hot 

Just let me know your final decision and I will change it accordingly.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
deutschmn: I added another test case that checks the stop condition if an incorrect number of classes is passed. The simple test case already runs on random input and verifies what the function does using a Java implementation in the test case. However, feel free to add more, if you deem it necessary.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
```diff
@@ -0,0 +1,77 @@
...
+            //generate actual dataset
+            double[][] doubles = getRandomMatrix(rows, cols, 1, numClasses, 1, 7);
+
+            // round them
+            double[][] A = new double[rows][cols];
+            for(int i = 0; i < rows; i++) {
+                for(int j = 0; j < cols; j++) {
+                    A[i][j] = Math.round(doubles[i][j]);
+                }
+            }
```
Baunsgaard, line 78: We have a util function to round all values. `TestUtils.round();` then you don't need this double for loop.

`TestUtils.round(getRandomMatrix(rows, cols, 1, numClasses, 1, 7));`

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+            String HOME = SCRIPT_DIR + TEST_DIR;
+            fullDMLScriptName = HOME + TEST_NAME + ".dml";
+            programArgs = new String[]{"-explain", "-args", input("A"), String.format("%d", numClassesPassed),
+                    output("B") };
+
+
+            writeInputMatrixWithMTD("A", A, false);
+
+            runTest(true, false, null, -1);
```
Baunsgaard, line 99: our run Test has an argument to detect if the execution should fail.

`runTest(true, false, DMLScriptException.class, -1);`

This will eliminate your if else bellow.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+            if(!shouldFail) {
+                HashMap<MatrixValue.CellIndex, Double> expected = new HashMap<MatrixValue.CellIndex, Double>();
+                for(int i = 0; i < A.length; i++) {
+                    for(int j = 0; j < A[i].length; j++) {
+                        // indices start with 1 here
+                        expected.put(new MatrixValue.CellIndex(i + 1, (int) A[i][j]), 1.0);
+                    }
+                }
```
Baunsgaard, line 108: maybe make this into an helper function to construct the result?
(the double for loop +  expected construction)

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
deutschmn: I did actually try that, but `stop` doesn't seem to throw an exception. Therefore I detect it below by checking if the out-file hasn't been written. Any suggestions how I else could do it?

Baunsgaard: oh, so I did not notice that thanks for pointing it out. :+1: I have to fix this in my own tests as well.

The second argument have to be true if an exception is expected, (so weird really have to be modified). But there is no exception but then again as you say there is no exception in a stop condition.

So conclusion is that there is no great way of detecting if a stop is called in SystemDS in testing other than looking at if the script produced the output file? @mboehm7  ?

src/main/java/org/apache/sysds/common/Builtins.java
Baunsgaard: After discussing this with Matthias, we think that we should try to stick to camelCase. The same goes for the parameters. for the internal enum you can keep it as TO_ONE_HOT.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinToOneHotTest.java
Baunsgaard: Update, I have looked into options so that if we call the system it does not require us to either check stdOut, or look if a file has been created. Unfortunately neither is available at the moment, so we will have to make some.

A current idea is that the exception thrown from a stop should not be caught and just print, but be thrown all the way out to a user, but with the twist that it should still provide the same print as before if executed normally not a stack trace.

Since it is another feature, I suggest not to include it in this PR.

src/main/java/org/apache/sysds/common/Builtins.java
deutschmn: Alright 👍 I just updated the naming in the most recent commit.

Baunsgaard:
LGTM

mboehm7:
LGTM - thanks @deutschmn for the patch and @Baunsgaard for the review. During the merge I just fixed minor issues in the test (tabs over spaces in Java code, static `computeExpectedResult`, and less verbose data generation).

PR #917, opened 2020-05-20
[MINOR][WIP] R-lib dependency use for workflows

Baunsgaard:


Baunsgaard:
Closing, because the time it takes for startup is fundamentally worse than the docker image ATM.

PR #918, opened 2020-05-22
[MINOR] docker update

Baunsgaard:
Minor docker update, to use new R dependencies for testing. 
Should fix integration tests of: https://github.com/apache/systemml/pull/914


mboehm7:
LGTM - thanks for the patch @Baunsgaard

PR #927, opened 2020-05-30
[SYSTEMDS-274] Fix compressed colMins/colMaxs w/ shared dictionary

mboehm7:
This patch fixes remaining issues of incorrect results for colMins and
colMaxs over compressed matrix blocks with shared DDC1 dictionaries.
Specifically, if the individual column groups have only partial overlap,
the shared dictionary contains a superset of column group distinct
values. Since aggregation functions like min and max are executed only
over the dictionary (without touching the compressed data), it led to
incorrect results as we find extreme values that do not actually exist
in the column group.

Three alternatives approaches could solve this: (1) drop shared
dictionaries, (2) execute colMins and colMaxs over the compressed data,
or (3) refactor the double array dictionary into a proper class
hierarchy and maintain additional meta data for shared dictionaries. We
decided for (3) in order to keep predictable performance, irrespective
of shared dictionaries and because this class hierarchy allows for
further improvements of shared dictionaries between any subsets of
column groups.

Additionally, this fix also cleanups incorrect estimates of the
individual column groups (because getValueSize was used in the estimates
as a number of values, although it gave the size in bytes) as well as
some of the Class-layout size estimation tests.

### DIFF
--- a/dev/docs/Tasks.txt
+++ b/dev/docs/Tasks.txt
@@ -223,7 +223,7 @@ SYSTEMDS-270 Compressed Matrix Blocks
  * 272 Simplify and speedup compression tests                         OK
  * 273 Refactor compressed Matrix Block to simplify responsibilities  OK
  * 273a Redesign allocation of ColGroups in ColGroupFactory
- * 274 Make the DDC Compression dictionary share correctly
+ * 274 Make the DDC Compression dictionary share correctly            OK
  * 275 Include compressionSettings in DMLConfiguration
  * 276 Allow Uncompressed Columns to be in sparse formats
  * 277 Sampling based estimators fix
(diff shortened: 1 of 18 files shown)


mboehm7:
cc @Baunsgaard - this fixes the issue we discussed, but we should discuss which of the JolSize estimation tests we should drop or update. So far I fixed everything except for the `JolEstimateTest.instanceSize()` (marked as `@Ignore`), which expected estimates I don't fully understand yet.

Baunsgaard:
Thank you so much for looking into it! 

Yes the test you ignored, is annoying to setup.
I will look if i can think of good alternatives other than completely removing it.

Also thanks for fixing some of my formatting issues, when we merged the entire rework it was a little much to look through :smirk:. I will try to look through in depth tomorrow, and finish up the ColGroupQuan i'm making.

mboehm7:
sounds good and thanks for catching this issue - it has shown that these brute force component tests find things that previously slipped through.

PR #932, opened 2020-05-31
[SYSTEMDS-???]Efficient cache management for multilevel caching.

phaniarnab:
This patch contains handling of multiple cache entries pointing to the
same data (due to multilevel caching).
 1) All the entries with the same values are connected with a linkedlist. Even though
 they output same data, they have different computation time.
 2) Eviction logic marks an entry for deferred spilling/removal if other
 entries are linked to that. If all the entries in a list are marked for
 spilling or removal, only then we evict the item.
 3) Disk write and read happen only once for all the items connected to a
 single matrix. This way single read and write restores multiple entries
 to cache and clears more space respectively.
 4) Initial experiments show huge improvements in cache management. Now
 the cache can store many more entries (this patch fixes duplicate size
 calculations), need reduced number of disk I/O. These changes overall improve
 cache hit count.


mboehm7:
LGTM - thanks @phaniarnab. I like the idea of linking them together and only spilling if no more non-spilled references exist. Just some minor nitpicks: please also use a free line between the git message title and description - otherwise interactive rebase gets messy. Also try to keep the public/protected interfaces as narrow as possible - we need to keep the caching logic as simple as possible and having narrow entry points makes it much easier to reason about synchronization, etc.

PR #941, opened 2020-06-03
[DOC] Documentation for builtin gnmf function

escher-m:
I have tried adding documentation for the gnmf function. Please review changes.

### DIFF
--- a/dev/docs/builtins-reference.md
+++ b/dev/docs/builtins-reference.md
@@ -417,8 +417,9 @@ y = normalize(X = X)
 ```
 ## `gnmf`-Function
 
-The `gnmf`-function does Generalized Non-Negative Matrix Factorization.
-It performs generalized non-negative matrix factorization based on Renyi Divergence
+The `gnmf`-function does Gaussian Non-Negative Matrix Factorization.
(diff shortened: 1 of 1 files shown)


dev/docs/builtins-reference.md
```diff
@@ -411,7 +412,38 @@ normalize(X);
...
 ```r
 X = rand(rows = 50, cols = 10)
-y = X %*% rand(rows=ncol(X), cols=1)
+y = X %*% rand(rows = ncol(X), cols = 1)
 y = normalize(X = X)
+```
+## `gnmf`-Function
+
+The `gnmf`-function does Generalized Non-Negative Matrix Factorization.
+It performs generalized non-negative matrix factorization based on Renyi Divergence
```
j143, line 421: I am not sure whether we are using `Renyi Divergence` or not.

Perhaps the information provided at this link could give some more info. https://github.com/apache/systemml/blob/faec15810e18fc497e950b830811a919f5aa5c5f/scripts/builtin/gnmf.dml#L24

escher-m: Made changes

PR #936, opened 2020-06-02
[DOC] Documentation for builtin msvm function

h0901:
Edit (by @j143):
This PR adds documentation for multiclass svm (`msvm`) builtin function.

PR #942, opened 2020-06-03
builtins-reference.md

h0901:


j143:
Already resolved by #936

PR #949, opened 2020-06-04
Propose: Bring docs to master branch. 

j143:
Previously, the docs used to be in the `1.2.0` release and was removed by https://github.com/apache/systemml/commit/6775e5f576f30339f2baed4e54fc0d6f25511f4b

Should we bring them back?

**Votes:**
@Baunsgaard: +1
@j143: +1


Baunsgaard:
This is one strategy to just bring all.
but it is also an opportunity to make something new, and not be forced to continue using all the old that might not work completely, and have to be debugged.
We should seriously consider only to take select elements from the previous.

j143:
Yes, Sebastian. We can a start fresh. There many broken pieces. :smiling_face_with_three_hearts:

mboehm7:
I strong second @Baunsgaard opinion. During the fork we removed a lot of old baggage and it would be counter-productive to bring this back. Also, bringing old (out-dated) documentation back means people get a false sense of "we have some documentation available".

mboehm7:
of course, one we settle on the target doc scheme, we can selectively bring e.g., parts of the algorithm documentation back (in a way that ports it to the new builtin functions)

Baunsgaard:
@j143, How about we then close this PR? 
and make an concrete design in a another PR?

PR #954, opened 2020-06-06
[DOC] Documentation for builtin intersect function

t201:
This pull request adds documentation for builtin intersect function

## `intersect`-Function

This is a generic function, extended from the intersect function in the base package. The elements of intersect(x,y) are those elements in x and in y. 
The original definition is preserved in the case that x and y are vectors of the same mode.
 
 ### Usage
 ```r
intersect(x, y)

 ```
 
 ### Arguments
 | Name    | Type           | Default  | Description |
 | :------ | :------------- | -------- | :---------- |
 | X       | Matrix[Double] | --       | a set of matrix A |
 | Y       | Matrix[Double] | --       | a set of matrix B | 
 
### Returns
| Type           | Description |
| :------------- | :---------- |
| Matrix[Double] | intersection matrix, set of intersecting items|

### Example
```
intersect = function(Matrix[Double] X, Matrix[Double] Y)
  return(Matrix[Double] A)
{
X = rand (rows = 50, cols = 10)
Y = rand (rows = 50, cols = 20)
y = X %*% rand(rows=ncol(X), 1)
y = Y
intersect(X,Y)
```

### DIFF
--- a/dev/docs/builtins-reference.md
+++ b/dev/docs/builtins-reference.md
@@ -23,7 +23,8 @@ limitations under the License.
   * [Built-In Construction Functions](#built-in-construction-functions)
     * [`tensor`-Function](#tensor-function)
   * [DML-Bodied Built-In functions](#dml-bodied-built-in-functions)
-    * [`KMeans`-Function](#KMeans-function)
+    * [`intersect`-Function](#intersect-function)
+	* [`KMeans`-Function](#KMeans-function)
     * [`lm`-Function](#lm-function)
     * [`lmDS`-Function](#lmds-function)
     * [`lmCG`-Function](#lmcg-function)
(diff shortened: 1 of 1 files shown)


t201:
Could you tell if this logic is correct for intersect function? :simple_smile:

t201:
Could you look if this logic is correct for intersect func. ?

j143:
Well done @t201 , I slightly modified the snippet. 

```r
X = rand (rows = 50, cols = 10)
y = X %*% rand(rows=ncol(X), 1)
Result = intersect(X,y)

print(to string(Result))
```

t201:
Screenshot (150)

Collab test run
Is it good to go ? I can implement more changes if suggested

j143:
The example is not valid and hence not included.

j143:
Thank you, @t201 - LGTM. 👍 
🎉 🚀

PR #963, opened 2020-06-08
[MINOR] Extend reusable instructions and bug fixes

phaniarnab:
This patch
 - extends reusable opcodes, which primarily improves multilevel (statementblock) cache hits,
 - removes most of the System.nanoTime calls from cache logic,
 - replaces operand names with placeholders in datagen lineage items,
   (Note: this fix is temporarily commented due to a bug in parfor-lineage)
 - fixes a bug in lineage item creation for multilevel caching,
 - update grid search lineage test with a loss function.


phaniarnab:
This is a temporary PR -- opened to verify if tests are passing.

PR #946, opened 2020-06-03
[SYSTEMDS-365] Privacy Runtime Extended

sebwrede:
- Privacy Constraint support for GLM
- Privacy tests for GLM
- Improved exception handling of federated responses
- Log of checked privacy constraints

PR #966, opened 2020-06-08
[SYSTEMDS-2558][SYSTEMDS-2554][SYSTEMDS-2561] Fed frame recode transform (encode) support

kev-inn:
This adds functionality for a federated recode transform. The resulting matrix will also be federated. 
Currently none of the other transform modes are supported and will result in failure (note: WIP). The decode works, but it is not a federated operation, instead the federated matrices are just sent back to the master and a local decode is executed.
This is again is quite a large PR so an early look (@Baunsgaard ?) would be appreciated while I continue to adapt things to support other encodings.
Federated decode will be done in a separate PR.

I already added the initial privacy features, but @sebwrede you might want to take a closer look since the current recode implementation leaks all the distinct values of columns to the master which could be reason for another level of privacy.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
@@ -39,17 +39,21 @@ import org.apache.sysds.runtime.controlprogram.caching.FrameObject;
 import org.apache.sysds.runtime.controlprogram.caching.MatrixObject;
 import org.apache.sysds.runtime.controlprogram.context.ExecutionContext;
 import org.apache.sysds.runtime.controlprogram.federated.FederatedRequest.RequestType;
-import org.apache.sysds.runtime.controlprogram.federated.FederatedResponse.ResponseType;
 import org.apache.sysds.runtime.instructions.InstructionParser;
 import org.apache.sysds.runtime.instructions.cp.Data;
 import org.apache.sysds.runtime.instructions.cp.ListObject;
 import org.apache.sysds.runtime.instructions.cp.ScalarObject;
+import org.apache.sysds.runtime.io.FileFormatPropertiesCSV;
 import org.apache.sysds.runtime.io.IOUtilFunctions;
+import org.apache.sysds.runtime.matrix.data.FrameBlock;
+import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 import org.apache.sysds.runtime.meta.MatrixCharacteristics;
 import org.apache.sysds.runtime.meta.MetaDataFormat;
 import org.apache.sysds.runtime.privacy.DMLPrivacyException;
 import org.apache.sysds.runtime.privacy.PrivacyMonitor;
 import org.apache.sysds.runtime.privacy.PrivacyPropagator;
+import org.apache.sysds.runtime.transform.encode.Encoder;
+import org.apache.sysds.runtime.transform.encode.EncoderFactory;
 import org.apache.sysds.utils.JSONHelper;
 import org.apache.wink.json4j.JSONObject;
 
@@ -108,6 +112,10 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 					return getVariable(request);
 				case EXEC_INST:
 					return execInstruction(request);
+				case ENCODE_META:
+					return createFrameEncodeMeta(request);
+				case FRAME_ENCODE:
+					return executeFrameEncode(request);
 				default:
 					String message = String.format("Method %s is not supported.", method);
 					return new FederatedResponse(FederatedResponse.ResponseType.ERROR, new FederatedWorkerHandlerException(message));
@@ -117,12 +125,83 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 			return new FederatedResponse(FederatedResponse.ResponseType.ERROR, ex);
 		}
 		catch (Exception ex) {
-			return new FederatedResponse(FederatedResponse.ResponseType.ERROR, 
-				new FederatedWorkerHandlerException("Exception of type " 
+			return new FederatedResponse(FederatedResponse.ResponseType.ERROR,
+				new FederatedWorkerHandlerException("Exception of type "
 				+ ex.getClass() + " thrown when processing request", ex));
 		}
 	}
-
+	
+	private FederatedResponse createFrameEncodeMeta(FederatedRequest request) {
+		// param parsing
+		checkNumParams(request.getNumParams(), 3);
+		String spec = (String) request.getParam(0);
+		int globalOffset = (int) request.getParam(1);
+		long varID = (long) request.getParam(2);
+		
+		FrameObject fo = (FrameObject) PrivacyMonitor.handlePrivacy(_vars.get(varID));
+		FrameBlock data = fo.acquireRead();
+		String[] colNames = data.getColumnNames();
+		
+		// create the encoder
+		Encoder encoder = EncoderFactory.createEncoder(spec,
+				colNames,
+				data.getNumColumns(),
+				null,
+				globalOffset,
+				globalOffset + data.getNumColumns());
+		// build necessary structures for encoding
+		encoder.build(data);
+		// just get the meta frame block
+		FrameBlock meta = encoder.getMetaData(new FrameBlock(data.getNumColumns(), Types.ValueType.STRING));
+		meta.setColumnNames(colNames);
+		// otherwise data of FrameBlock would be null, therefore it would fail
+		// hack because serialization of FrameBlock does not function if Arrays are not allocated
+		meta.ensureAllocatedColumns(meta.getNumRows());
+		fo.release();
+		
+		return new FederatedResponse(FederatedResponse.Type.SUCCESS, meta);
+	}
+	
+	private FederatedResponse executeFrameEncode(FederatedRequest request) {
+		checkNumParams(request.getNumParams(), 4);
+		FrameBlock meta = (FrameBlock) request.getParam(0);
+		String spec = (String) request.getParam(1);
+		// offset of local column indexes compared to global (due to federation)
+		int globalOffset = (int) request.getParam(2);
+		long varID = (long) request.getParam(3);
+		
+		FrameObject fo = (FrameObject) PrivacyMonitor.handlePrivacy(_vars.get(varID));
+		FrameBlock data = fo.acquireRead();
+		String[] colNames = data.getColumnNames();
+		
+		// compute encoding with given meta data
+		Encoder encoder = EncoderFactory.createEncoder(spec,
+				colNames,
+				data.getNumColumns(),
+				meta,
+				globalOffset,
+				globalOffset + data.getNumColumns());
+		
+		// apply transformation
+		MatrixBlock mbout = encoder.apply(data, new MatrixBlock(data.getNumRows(), data.getNumColumns(), false));
+		
+		// copy characteristics
+		MatrixCharacteristics mc = new MatrixCharacteristics(fo.getDataCharacteristics());
+		MatrixObject mo = new MatrixObject(Types.ValueType.FP64, OptimizerUtils.getUniqueTempFileName(),
+				new MetaDataFormat(mc, FileFormat.BINARY));
+		// set the encoded data
+		mo.acquireModify(mbout);
+		mo.release();
+		fo.release();
+		
+		// create a new id handle
+		long id = _seq.getNextID();
+		// add it to the list of variables
+		_vars.put(id, mo);
+		// return id handle
+		return new FederatedResponse(FederatedResponse.Type.SUCCESS, id);
+	}
+	
 	private FederatedResponse readData(FederatedRequest request) {
 		checkNumParams(request.getNumParams(), 2);
 		String filename = (String) request.getParam(0);
@@ -143,7 +222,7 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 				break;
 			default:
 				// should NEVER happen (if we keep request codes in sync with actual behaviour)
-				return new FederatedResponse(FederatedResponse.ResponseType.ERROR, 
+				return new FederatedResponse(FederatedResponse.ResponseType.ERROR,
 					new FederatedWorkerHandlerException("Could not recognize datatype"));
 		}
 		
@@ -168,6 +247,8 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 			throw new DMLRuntimeException(ex);
 		}
 		cd.setMetaData(new MetaDataFormat(mc, fmt));
+		// TODO send FileFormatProperties with request and use them for CSV
+		cd.setFileFormatProperties(new FileFormatPropertiesCSV());
 		cd.acquireRead();
 		cd.refreshMetaData(); //in pinned state
 		cd.release();
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
@@ -113,7 +113,7 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 				case EXEC_INST:
 					return execInstruction(request);
 				case ENCODE_META:
-					return createFrameEncodeMeta(request);
+					return createFrameEncoder(request);
 				case FRAME_ENCODE:
 					return executeFrameEncode(request);
 				default:
@@ -130,8 +130,8 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 				+ ex.getClass() + " thrown when processing request", ex));
 		}
 	}
-	
-	private FederatedResponse createFrameEncodeMeta(FederatedRequest request) {
+
+	private FederatedResponse createFrameEncoder(FederatedRequest request) {
 		// param parsing
 		checkNumParams(request.getNumParams(), 3);
 		String spec = (String) request.getParam(0);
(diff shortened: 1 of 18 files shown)


sebwrede:
It seems like it would throw an exception whenever privacy level Private or PrivateAggregation is set in the current implementation. This should be fine for now. Did you test it with privacy constraints on the input? I tested the other federated methods in FederatedWorkerHandlerTest.java.

I am also making changes to the FederatedWorkerHandler in another PR: https://github.com/apache/systemml/pull/946.

kev-inn:
> It seems like it would throw an exception whenever privacy level Private or PrivateAggregation is set in the current implementation. This should be fine for now. Did you test it with privacy constraints on the input? I tested the other federated methods in FederatedWorkerHandlerTest.java.
> 
> I am also making changes to the FederatedWorkerHandler in another PR: #946.

No I actually didn't really test it with privacy settings, but I guess if I set it to PrivateAggregation that would be the correct level right, because taking the unique elements would be an aggregation?

sebwrede:
I still have some work to do on actually determining whether something is aggregated enough to be called an aggregation. If you want to make it pass with PrivateAggregation, then you need to call the other "handlePrivacy" method, but right now I set it up to only take MatrixObject as parameter, which may not be what is needed in this case. I think it is acceptable to just throw the exception at this point and then I can improve the PrivacyMonitor later so that it can handle the new cases better.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+		// first we use the spec to construct a meta frame which will provide us with info about the encodings
+		List<Pair<FederatedRange, Future<FederatedResponse>>> metaFutures = new ArrayList<>();
+		for(Map.Entry<FederatedRange, FederatedData> entry : fedMapping.entrySet()) {
+			Future<FederatedResponse> response = entry.getValue().executeFederatedOperation(new FederatedRequest(
+				FederatedRequest.FedMethod.ENCODE_META, spec, entry.getKey().getBeginDimsInt()[1] + 1), true);
+			metaFutures.add(new ImmutablePair<>(entry.getKey(), response));
+		}
+
+		// TODO support encodings other than recode
+		// the combined mappings for the frame columns (because we only support recode)
```
Baunsgaard, line 97: I think it is important that we already think about the extensions. 
Since you implemented it for recode, how would you extend for others?
If the answer is complicated we should consider changing the code to make it less so.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+			return new MultiReturnParameterizedBuiltinFEDInstruction(null, in1, in2, outputs, opcode, str);
+		}
+		else {
+			throw new DMLRuntimeException("Invalid opcode in MultiReturnBuiltin instruction: " + opcode);
+		}
+
+	}
+
+	@Override
+	public void processInstruction(ExecutionContext ec) {
```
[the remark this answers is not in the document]
Baunsgaard: Looking at the reference `MultiReturnParameterizedBuiltinCPInstruction`

It looks like the design there moves the logic entirely to the Encoder. Is there a reason not to do such a thing in this case?

```java

	@Override 
	public void processInstruction(ExecutionContext ec) {
		//obtain and pin input frame
		FrameBlock fin = ec.getFrameInput(input1.getName());
		String spec = ec.getScalarInput(input2).getStringValue();
		String[] colnames = fin.getColumnNames(); 
		
		//execute block transform encode
		Encoder encoder = EncoderFactory.createEncoder(spec, colnames, fin.getNumColumns(), null);
		MatrixBlock data = encoder.encode(fin, new MatrixBlock(fin.getNumRows(), fin.getNumColumns(), false)); //build and apply
		FrameBlock meta = encoder.getMetaData(new FrameBlock(fin.getNumColumns(), ValueType.STRING));
		meta.setColumnNames(colnames);
		
		//release input and outputs
		ec.releaseFrameInput(input1.getName());
		ec.setMatrixOutput(getOutput(0).getName(), data);
		ec.setFrameOutput(getOutput(1).getName(), meta);
	}
```

src/main/java/org/apache/sysds/runtime/matrix/data/FrameBlock.java
```diff
@@ -642,6 +642,9 @@ public void setColumn(int c, Array column) {
 
 	///////
 	// serialization / deserialization (implementation of writable and externalizable)
+	// FIXME for FrameBlock fix write and readFields, it does not work if the Arrays are not yet
+	// allocated (after fixing remove hack in FederatedWorkerHandler.createFrameEncodeMeta(FederatedRequest) call to
+	// FrameBlock.ensureAllocatedColumns())
```
Baunsgaard, line 647: Maybe make a Jira report?

src/test/java/org/apache/sysds/test/functions/federated/FederatedConstructionTest.java
```diff
@@ -142,7 +142,7 @@ public void federatedConstruction(Types.ExecMode execMode, String testFile, Stri
 			DMLScript.USE_LOCAL_SPARK_CONFIG = true;
 		}
 		fullDMLScriptName = HOME + testFile + ".dml";
-		programArgs = new String[] {"-args", "\"localhost:" + port + "/" + input(inputIdentifier) + "\"",
+		programArgs = new String[] {"-args", TestUtils.federatedAddress("localhost", port, input(inputIdentifier)),
```
Baunsgaard, line 145: I Like this TestUtils for the federated Address. Could you change this for more of the federated tests?

src/test/java/org/apache/sysds/test/functions/transform/TransformFederatedEncodeDecodeTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+			// B will be recoded and will be the column that will be grouped by
+			Arrays.fill(schema, Types.ValueType.STRING);
+			// we set sparsity to 1.0 to ensure all the string labels exist
+			double[][] B = TestUtils.round(getRandomMatrix(rows, cols / 2, 1, 15, 1.0, 8));
+			writeInputFrameWithMTD("B", B, false, schema, format);
+
+			fullDMLScriptName = SCRIPT_DIR + TEST_DIR + TEST_NAME1 + ".dml";
+
+			programArgs = new String[] {"-explain", "-args", TestUtils.federatedAddress("localhost", port1, input("A")),
```
Baunsgaard, line 115: If you can live without calling explain in tests it would be great. since it has a tendency to write alittle to much to the terminal, that in turn crash our test suite on workflows.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -115,6 +131,77 @@ private FederatedResponse constructResponse(FederatedRequest request) {
 		}
 	}
 
+	private FederatedResponse createFrameEncodeMeta(FederatedRequest request) {
```
Baunsgaard, line 134: Im not really a fan of this method located inside the workerHandler.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -115,6 +131,77 @@ private FederatedResponse constructResponse(FederatedRequest request) {
...
+		meta.setColumnNames(colNames);
+		// otherwise data of FrameBlock would be null, therefore it would fail
+		// hack because serialization of FrameBlock does not function if Arrays are not allocated
+		meta.ensureAllocatedColumns(meta.getNumRows());
+		fo.release();
+
+		return new FederatedResponse(FederatedResponse.Type.SUCCESS, meta);
+	}
+
+	private FederatedResponse executeFrameEncode(FederatedRequest request) {
```
Baunsgaard, line 165: Im not really a fan of this method located inside the workerHandler.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+		ArrayList<CPOperand> outputs = new ArrayList<>();
+		String opcode = parts[0];
+
+		if(opcode.equalsIgnoreCase("transformencode")) {
+			// one input and two outputs
+			CPOperand in1 = new CPOperand(parts[1]);
+			CPOperand in2 = new CPOperand(parts[2]);
+			outputs.add(new CPOperand(parts[3], Types.ValueType.FP64, Types.DataType.MATRIX));
+			outputs.add(new CPOperand(parts[4], Types.ValueType.STRING, Types.DataType.FRAME));
+			return new MultiReturnParameterizedBuiltinFEDInstruction(null, in1, in2, outputs, opcode, str);
```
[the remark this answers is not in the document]
kev-inn: Oh you mean create `Operator`s for Transform? We could definitely do that, if we categorize Transform's as an `Operator`, which I believe makes sense. I like the idea, but I think it would be smart to do this in a seperate PR right?

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+		String[] parts = InstructionUtils.getInstructionPartsWithValueType(str);
+		ArrayList<CPOperand> outputs = new ArrayList<>();
+		String opcode = parts[0];
+
+		if(opcode.equalsIgnoreCase("transformencode")) {
+			// one input and two outputs
+			CPOperand in1 = new CPOperand(parts[1]);
+			CPOperand in2 = new CPOperand(parts[2]);
+			outputs.add(new CPOperand(parts[3], Types.ValueType.FP64, Types.DataType.MATRIX));
+			outputs.add(new CPOperand(parts[4], Types.ValueType.STRING, Types.DataType.FRAME));
```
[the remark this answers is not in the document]
kev-inn: I don't quite get what you mean by "make it federated", could you explain?

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.sysds.runtime.matrix.data.FrameBlock;
+import org.apache.sysds.runtime.matrix.operators.Operator;
+import org.apache.sysds.runtime.transform.encode.EncoderRecode;
+
+public class MultiReturnParameterizedBuiltinFEDInstruction extends ComputationFEDInstruction {
+	protected final ArrayList<CPOperand> _outputs;
+
+	private MultiReturnParameterizedBuiltinFEDInstruction(Operator op, CPOperand input1, CPOperand input2,
+		ArrayList<CPOperand> outputs, String opcode, String istr) {
+		super(FEDType.MultiReturnParameterizedBuiltin, op, input1, input2, outputs.get(0), opcode, istr);
```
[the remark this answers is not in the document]
kev-inn: I don't quite get how a null would help, but I agree that this is not quite good at the moment. We probably should add options for multi return functions?

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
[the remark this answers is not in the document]
kev-inn: Yes, I agree (I should really adapt the single responsibility principle better). On a more general note, I believe I might refactor it to actually move the functionality to Encoders (with new abstract methods to combine encoders created on the federated workers and split them to distribute the global encoder). I was scared about this at first, because it would be a more intrusive approach, but I now believe that is the way to go. Thoughts?

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
```diff
@@ -0,0 +1,215 @@
...
+
+	private FrameBlock frameBlockFromRecodeMaps(Map<String, Long>[] combinedRecodeMaps) {
+		int rows = 0;
+		for(Map<String, Long> map : combinedRecodeMaps) {
+			if(map != null) {
+				rows = Integer.max(rows, map.size());
+			}
+		}
+		FrameBlock fb = new FrameBlock(combinedRecodeMaps.length, Types.ValueType.STRING);
+		fb.ensureAllocatedColumns(rows);
```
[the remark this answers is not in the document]
kev-inn: There is no constructor for rows and columns, actually this loop anyway was a bit hard to develop, maybe because I wasn't as familiar with `Frame`s, or because the interface has a few pitfalls that I had problems with.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
kev-inn: Yes, I refer to my comment above.

src/test/java/org/apache/sysds/test/functions/federated/FederatedConstructionTest.java
kev-inn: sure, I guess I should note it down in Jira and do a seperate PR?

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
kev-inn: Do you mean you don't like what the function does, or you would prefer if we but it in a different class?
If the first is the problem, maybe some details (or I refer to the comment further above), otherwise I believe we should anyway start thinking about splitting `FederatedWorkerHandler` up into some/one class that handles some types of tasks. Thoughts?

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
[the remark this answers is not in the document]
Baunsgaard: well It is the super class that should needs a new the declaration using lists or collections of input and outputs. such that potential nulls in output.get(1), 2... n, doesn't fail.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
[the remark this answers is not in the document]
Baunsgaard: 1. thing (log time ago i wrote this comment)
2. It does not make sense while i'm looking at it again. since our `common/Types.java` doesn't contain such a DataType of FederatedMatrix nor FederatedFrame. (which i thought they did when i wrote the comment)
3. I think it might make sense to have such abstractions making it easier in general to code federated operations, oppinions?

If it is implemented like currently this I'm concerned about how many places it has to interact.
Currently it requires us to put information about if the matrix or frame is federated into those objects, rather than leveraging that the object itself by type defines it.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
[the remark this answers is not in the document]
Baunsgaard: Separation sounds good :)

src/test/java/org/apache/sysds/test/functions/federated/FederatedConstructionTest.java
Baunsgaard: You don't have to do the change if you just make a Jira task and mark it as an warmup.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
Baunsgaard: It is purely a comment on location.
I think adding a FedEncoder file inside `sysds/runtim/transform/encode/fedencode.java`, could be a good because it keeps the functionality collected at the same place the other transform operations are located.

`anyway start thinking about splitting FederatedWorkerHandler` 
- I'm always a fan of splitting things apart :smile: (maybe to much)

src/test/java/org/apache/sysds/test/functions/federated/FederatedConstructionTest.java
kev-inn: How does one mark something as warmup? Do you mean with a label, because I would need to create a new one I guess? I added it as an improvement with low priority.

src/main/java/org/apache/sysds/runtime/instructions/fed/MultiReturnParameterizedBuiltinFEDInstruction.java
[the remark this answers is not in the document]
kev-inn: I guess the reason we currently don't have it, is due to us using the enum to figure out which cast we are going to do, which is independent of it's federation status. If we plan on creating new classes for `FederatedMatrice`s, then I do believe a new type makes perfect sense, but even if we don't, and instead keep it in `MatrixObject`, I believe it still would be beneficial to have actual types for them. As you said, the code would get simpler (checks are easier) and we can use it to decide on operations (the current replacement should anyway be removed rather sooner than later).
So I agree with your opinion on the matter.

kev-inn:
I decided to not yet split up the methods in `FederatedWorkerHandler` since they got a lot shorter and I think the modularization of our of the `FederatedWorkerHandler` is more work, maybe the way the handler is "abused" for all the work is also a problem, since I don't know what exactly the intended way to compute large workloads with netty communication is.

mboehm7:
@kev-inn is this PR on purpose in WIP status - if not, I could have a look and merge it in. The rework of `FederatedWorkerHandler`, I would then do tomorrow.

kev-inn:
I will adapt this implementation to the reworked `FederatedWorkerHandler` and then it should be ready for review.

mboehm7:
LGTM - thanks for the patch @kev-inn. Tomorrow, I'll try to implement a first version of the UDF framework and will use this transform encode as a test case. Thanks.

PR #970, opened 2020-06-09
[MINOR] Add federated Prefix for statistics 

Baunsgaard:


sebwrede:
LGTM.
Why is  `-Dlog4j.configuration=file:` removed?

Baunsgaard:
Because previously it required the whole line for the logging file. but now it has been changed to construct the `-Dlog4j.configuration=file:` part inside a variable. If you compare to line 303, you can see the distributed does it the same way.

Baunsgaard:
@sebwrede, Thanks i forgot about this PR, I will merge it soon.

PR #971, opened 2020-06-09
[WIP] - Add protobuf support to write and read FrameBlocks to HDFS

muehlburger:
This pull requests adds a first protobuf support to read and write FrameBlocks to HDFS.

Appreciate any feedback.


src/main/java/org/apache/sysds/parser/dml/DmlBaseListener.java
```diff
@@ -1,4 +1,4 @@
-// Generated from org\apache\sysds\parser\dml\Dml.g4 by ANTLR 4.5.3
+// Generated from org/apache/sysds/parser/dml/Dml.g4 by ANTLR 4.5.3
```
Baunsgaard, line 1: These files change because they are generated on your system not windows.
If you revert the change on you branch in a new commit, and thereafter look inside our git ignore.
then it will show how to ignore these files for future commits :).

Same comment for all `src/main/java/org/apache/sysds/parser/dml/XXX.java` files.

pom.xml
```diff
@@ -1021,6 +1021,25 @@
 			<artifactId>maven-gpg-plugin</artifactId>
 			<version>1.6</version>
 		</dependency>
+
+		<!-- https://github.com/protocolbuffers/protobuf/tree/master/java -->
+		<dependency>
+			<groupId>com.google.protobuf</groupId>
+			<artifactId>protobuf-java</artifactId>
+			<version>3.11.0</version>
```
Baunsgaard, line 1029: It is not 100% updated on their GitHub.
cold you try the 3.12.X out?
<https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java>

src/main/java/org/apache/sysds/protobuf/SysdsProtos.java
```diff
@@ -0,0 +1,2985 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: src/main/resources/protobuf/Frame.proto
+
+package org.apache.sysds.protobuf;
+
+public final class SysdsProtos {
+  private SysdsProtos() {}
```
Baunsgaard, line 25: - If you change indentation to Tabs that would be great I don't know if it is possible while it is auto generated.. 
- If you write the Generated in the first line above license then we have the same format as the DML files auto-generated
- Also we have a code-style file in `/dev/docs/CodeStyle_eclipse.xml`, that could be applied again if possible

src/main/java/org/apache/sysds/runtime/io/FrameReaderFactory.java
```diff
@@ -58,6 +58,9 @@ public static FrameReader createFrameReader( FileFormat fmt, FileFormatPropertie
 				else
 					reader = new FrameReaderBinaryBlock();
 				break;
+			case PROTO:
+				reader = new FrameReaderProto();
+				break;
```
Baunsgaard, line 63: I know it is out of scope, but could you check out if a parallel readers is possible, if so you don't have to add it but just write a TODO specifying it is, or if not write not possible and reason?

src/main/resources/protobuf/Frame.proto
```diff
@@ -0,0 +1,46 @@
...
+
+message Schema {
+    enum ValueType {
+        FP32 = 0;
+        FP64 = 1;
+        INT32 = 2;
+        INT64 = 3;
+        BOOLEAN = 4;
+        STRING = 5;
+        UNKNOWN = 6;
```
Baunsgaard, line 43: Is this list somewhat the same as in `src/main/python/systemds/onnx_systemds/onnx_helper.py`? 
or rather is there some predetermined specific schema needed, and why is this specific one used here?

src/test/java/org/apache/sysds/test/functions/io/proto/FrameReaderWriterProtoTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.Test;
+
+public class FrameReaderWriterProtoTest
+{
+	private static final String FILENAME_SINGLE = "target/testTemp/functions/data/FrameReaderWriterProtoTest/testFrameBlock.proto";
+
+	@Test
+	public void testWriteReadFrameBlockSingleSingleFromHDFS() throws IOException, JSONException {
+		testWriteReadFrameBlockComplete(1,1, 4669201);
+	}
```
Baunsgaard, line 45: Since these tests are the same as in `i0/json/freameReaderWriterJSONLTest`.
You could make it into a class that extends and overwrites the json FrameReaderWriter, or another super class for both of them.

src/main/resources/protobuf/Frame.proto
muehlburger: I need this enum specified in this .proto File. This is used to generate the language specific (in our case Java) glue code.

This list is taken from [ValueType](https://github.com/apache/systemml/blob/master/src/main/java/org/apache/sysds/common/Types.java#L72)

src/test/java/org/apache/sysds/test/functions/io/proto/FrameReaderWriterProtoTest.java
muehlburger: Good point, will refactor that.

muehlburger: I refactored and simplified the tests:
- removed unnecessary JSONException which is never thrown in this code (so it does not make sense to use inheritance any more).
- simplified method calls

With the current implementation it should be clear what happens in every test.

muehlburger:
@Baunsgaard could you have another look at the changes? Would be happy to get your feedback! :-)

src/main/java/org/apache/sysds/runtime/io/FrameReaderProto.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.sysds.conf.ConfigurationManager;
+import org.apache.sysds.protobuf.SysdsProtos;
+import org.apache.sysds.runtime.DMLRuntimeException;
+import org.apache.sysds.runtime.matrix.data.FrameBlock;
+import org.apache.sysds.runtime.util.UtilFunctions;
+
+public class FrameReaderProto extends FrameReader {
+	@Override
+	public FrameBlock readFrameFromHDFS(String fname, Types.ValueType[] schema, String[] names, long rlen, long clen)
+		throws IOException, DMLRuntimeException {
```
Baunsgaard, line 39: Runtime Exceptions does not need to be declared as thrown.

src/main/java/org/apache/sysds/runtime/io/FrameWriterFactory.java
```diff
@@ -55,6 +55,10 @@ public static FrameWriter createFrameWriter( FileFormat fmt, FileFormatPropertie
 				else
 					writer = new FrameWriterBinaryBlock();
 				break;
+
+			case PROTO:
+				writer = new FrameWriterProto();
```
Baunsgaard, line 60: maybe add a parallel todo again.

src/main/java/org/apache/sysds/runtime/io/FrameWriterProto.java
```diff
@@ -0,0 +1,80 @@
...
+	protected void writeProtoFrameToHDFS(Path path, JobConf jobConf, FrameBlock src, long rlen, long clen)
+		throws IOException {
+		FileSystem fileSystem = IOUtilFunctions.getFileSystem(path, jobConf);
+		writeProtoFrameToFile(path, fileSystem, src, 0, (int) rlen);
+		IOUtilFunctions.deleteCrcFilesFromLocalFileSystem(fileSystem, path);
+	}
+
+	protected void writeProtoFrameToFile(Path path, FileSystem fileSystem, FrameBlock src, int lowerRowBound,
+		int upperRowBound) throws IOException {
+		// what about > 2GB Protobuf Messages?
```
Baunsgaard, line 65: does this mean that >2GB is impossible?

src/main/resources/protobuf/Frame.proto
```diff
@@ -0,0 +1,46 @@
...
+    }
+    repeated ValueType valueType = 1;
+}
```
Baunsgaard, line 46: Just to make Git happy, add a single newline

Baunsgaard:
Also A nice SystemDS name of the PR and add it to a task inside `/docs/dev/task.txt`

Baunsgaard:
Hi @muehlburger !
Thanks for your contribution! :superhero: :star: 
I have squashed all your commits into one, and intend to push this to master after the testsuite is done on my fork.
The only change i made (hopefully) is to add to the comment of the 2GB limit. We will have to see if that is going to be a problem in java as well.

I hope you enjoyed committing to opensource and we hope to see you contributing again another time.

muehlburger:
Thank you! It was a pleasure!

PR #972, opened 2020-06-09
[MINOR] MICE now accepts a recoded matrix as input

Shafaq-Siddiqi:
MICE now accept a matrix as input instead of frame. The categorical features could be recoded first before calling MICE.

### DIFF
--- a/scripts/builtin/mice.dml
+++ b/scripts/builtin/mice.dml
@@ -19,16 +19,15 @@
 #
 #-------------------------------------------------------------
 
-# Builtin function Implements Multiple Imputation using Chained Equations (MICE) for nominal data
+# Built-in function Implements Multiple Imputation using Chained Equations (MICE) 
 #
 # INPUT PARAMETERS:
 # ---------------------------------------------------------------------------------------------
 # NAME            TYPE    DEFAULT     MEANING
 # ---------------------------------------------------------------------------------------------
-# F               String    ---        Data Frame
-# cMask           Double    ---        A 0/1 row vector for identifying numeric (0) adn categorical features (1)
+# X               String    ---        Data Matrix (Recoded Matrix for categorical features)
+# cMask           Double    ---        A 0/1 row vector for identifying numeric (0) and categorical features (1)
 # iter            Integer    3         Number of iteration for multiple imputations 
-# complete        Integer    3         A complete dataset generated though a specific iteration
 # ---------------------------------------------------------------------------------------------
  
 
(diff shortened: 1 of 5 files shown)


src/test/java/org/apache/sysds/test/functions/builtin/BuiltinMiceTest.java
```diff
@@ -50,33 +50,47 @@ public void testMiceMixCP() {
 		runMiceNominalTest(mask, 1, LopProperties.ExecType.CP);
 	}
 
+//	@Test
```
Shafaq-Siddiqi, line 53: @mboehm7  as discussed, the Spark tests are commented out due to "NullPointerException" in one-hot-encoding. The error could be reproduced by running the spark tests.

j143:
Can this file https://github.com/apache/systemml/blob/master/dev/docs/builtins-reference.md#mice-function also be updated? 

Thank you.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinMiceTest.java
```diff
@@ -36,7 +36,7 @@
 	private static final String TEST_CLASS_DIR = TEST_DIR + BuiltinMiceTest.class.getSimpleName() + "/";
 
 	private final static String DATASET = SCRIPT_DIR +"functions/transform/input/ChickWeight.csv";
-	private final static double eps = 0.2;
+	private final static double eps = 0.16;
 	private final static int iter = 3;
 	private final static int com = 2;
```
phaniarnab, line 41: Please remove the unused variable com.

scripts/builtin/mice.dml
```diff
@@ -19,266 +19,176 @@
...
+  for(k in 1:iter) # start iterative imputation
   {
-    Mask_Filled_n = Mask_n;
-    Mask_Filled_c = Mask_c
-    in_n = 1; in_c = 1; i=1; j=1; # variables for index selection
-    while(i <= ncol(dX))
+    Mask_Filled = Mask1
+    inverseMask = Mask1 == 0
+    # OHE of categorical features
+    [dX, dM] = transformencode(target=as.frame(X1), spec=jspecDC);
```
phaniarnab, line 106: Sorry for being naive, can we not move this transformencode outside, to have no frame operations?

j143:
Shall we move this forward? 
cc @Shafaq-Siddiqi @phaniarnab

Shafaq-Siddiqi:
> Shall we move this forward?
> cc @Shafaq-Siddiqi @phaniarnab

yes from my side.

src/test/scripts/functions/caching/BufferpoolLeak.dml
```diff
@@ -22,7 +22,7 @@
 X = rand(rows=$1, cols=$2, min=1, max=10);
 for(i in 1:500) {
   #print("executed iteration "+i)
-  [m1,m2] = mice(as.frame(X), matrix(0,1,ncol(X)),3,3, FALSE)
+  [m1,m2] = mice(X), matrix(0,1,ncol(X)), 3, FALSE)
```
Baunsgaard, line 25: It seems like the parenthesis is incorrect here.
Does this not crash the system?

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinMiceTest.java
```diff
@@ -50,33 +50,47 @@ public void testMiceMixCP() {
...
-	//	@Test
-	//	public void testMiceSpark() {
-	//		runMiceNominalTest( LopProperties.ExecType.SPARK);
-	//	}
+
+//	@Test
+//	public void testMiceCategoricalSpark() {
+//		double[][] mask = {{ 1.0, 1.0, 1.0, 1.0, 1.0}};
+//		runMiceNominalTest(mask, 3, LopProperties.ExecType.SPARK);
+//	}
```
Baunsgaard, line 81: Could you mark these tests with @Ignore so that our test framework skips them, but reports that they are skipped. This will notify us that there is a problem we need to fix.

src/test/scripts/functions/builtin/mice.dml
```diff
@@ -19,26 +19,44 @@
 #
 #-------------------------------------------------------------
 
-X = read($X, data_type="frame", format="csv");
```
Baunsgaard, line 22: Looking at this test, it seems very complicated to use MICE.
does it make sense to split this up into multiple smaller tests?
(does not have to be in this PR)

scripts/builtin/mice.dml
Baunsgaard: Im not sure if i get the location either, Is it located here because it needs to be applied multiple times because X1 is cleaned?

src/test/scripts/functions/builtin/mice.dml
Shafaq-Siddiqi: I think some comments will help to understand.

src/test/scripts/functions/caching/BufferpoolLeak.dml
Shafaq-Siddiqi: > It seems like the parenthesis is incorrect here.
> Does this not crash the system?

I fixed the call, surprisingly it was not crashing the test on git. After fixing the call now the cache is leaning and BufferpoolleakTest is failing.

j143:
bb197a6c9556cb082d580282361319918ac04c7b
bb197a6c9556cb082d580282361319918ac04c7b 
fixes the bufferpool leaks in mvvar instructions. This make this PR ready for merge!

j143:
Hi @Shafaq-Siddiqi ,

The caching test is still not passing for me with the latest changeset!. Any updates?

mboehm7:
LGTM - thanks @Shafaq-Siddiqi. I made some smaller cleanups to avoid unnecessary branches but left the spark issues for now. Once the spark transform is fixed, we should also fix the script-level string concatenation ([1,2,]) and investigate if we still need the transformencode inside the mice implementation.

PR #973, opened 2020-06-10
[SYSTEMML-2536][WIP] Found CVEs in dependencies

Baunsgaard:
Major update of Hadoop version to 3.2.1 from 2.6.0.
Will have to update servers installation to fit with this PR.

https://issues.apache.org/jira/browse/SYSTEMML-2536

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,9 +43,9 @@
 	<properties>
 		<hadoop.version>3.2.1</hadoop.version>
 		<antlr.version>4.5.3</antlr.version>
-		<spark.version>2.1.0</spark.version>
+		<spark.version>3.0.0-preview2</spark.version>
 		<scala.version>2.11.8</scala.version>
-		<scala.binary.version>2.11</scala.binary.version>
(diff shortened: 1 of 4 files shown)


j143:
- [ ] For `Spark 2.*`, only `Hadoop 2.*` seems to compatible as seen at https://spark.apache.org/downloads.html?
- [ ] For `Spark 3.*` there seems to be no problem though!

Baunsgaard:
well there are multiple issues with that. Switching to Hadoop 3.0 requires yes spark 3.0 but that is currently not in a stable release. (since December 2019).
Also exploring that currently, we have some uses of the spark in paramserver rpc that does not work with spark 3.0 because:

- SparkTransportConf & SparkPSProxy has changed.

and Hadoop because they removed the function copyMerge.
I have "taken" the original souce for the copyMerge method and included it in the merge, and made the rpc parts of Parameter server throw an NotImplementedException.

Baunsgaard:
Priority ATM is not to switch to Spark and hadoop 3.0 Therefore closing this PR.

PR #975, opened 2020-06-15
Extend the Python API

juliale-15:
Extend the Python API by 
- Trigonometric Built-In Functions such as sin, cos, tan etc
- Other Build-In Functions: reverse, transpose, order
- Linear Algebra Built-In Functions: Cholesky


src/main/python/systemds/matrix/data_gen.py
```diff
@@ -19,7 +19,7 @@
 #
 # -------------------------------------------------------------
 
-__all__ = ['full', 'seq', 'rand']
+__all__ = ['full', 'seq', 'rand', 'rev', 'order', 't', 'cholesky']
```
Baunsgaard, line 22: This file contains methods for data generation, not for operations.
I think it would be nice if you could move the `rev` `order` `t` and `cholesky` to the matrix.py object. thereby removing the need and potential error associated with calling these functions with an SystemDSContext.

src/main/python/systemds/matrix/data_gen.py
```diff
@@ -29,6 +29,8 @@
 
 from systemds.operator import OperationNode
 from systemds.context import SystemDSContext
+from systemds.matrix import Matrix
+import numpy as np
 
 def full(sds_context: SystemDSContext, shape: Tuple[int, int], value: Union[float, int]) -> OperationNode:
```
Baunsgaard, line 35: Also if you want to, then maybe change these data_gen functions to return matrix Objects, not Operation nods.

src/main/python/systemds/matrix/data_gen.py
```diff
@@ -102,3 +105,70 @@ def rand(sds_context: SystemDSContext, rows: int, cols: int,
         named_input_nodes['seed'] = seed
 
     return OperationNode(sds_context, 'rand', [], named_input_nodes=named_input_nodes)
+
+
+def rev(sds_context: 'SystemDSContext', X: Matrix) -> 'OperationNode':
```
Baunsgaard, line 110: In the Methods, i would appreciate if the syntax is:

` def rev(sds_context: SystemDSContext, X: Matrix) -> OperationNode:`

Since it allows the system to complain earlier about missing imports and such not resulting in a potential runtime error.

src/main/python/systemds/matrix/data_gen.py
```diff
@@ -102,3 +105,70 @@ def rand(sds_context: SystemDSContext, rows: int, cols: int,
...
+    if X._np_array.shape[0] != X._np_array.shape[1]:
+        raise ValueError("Last 2 dimensions of the array must be square")
+
+    # check if mat is positive definite
+    if not np.all(np.linalg.eigvals(X._np_array)>0):
+        raise ValueError("Matrix is not positive definite")
+
+    # check if mat is symmetric
+    if not np.allclose(X._np_array, X._np_array.transpose()):
+        raise ValueError("Matrix is not symmetric")
```
Baunsgaard, line 172: I like the checks, but the complexity of these operations grow in size of X.  (except for the first check).
This makes the overall performance of the operation slow, maybe we add an optional flag saying unsafe, that skip these checks?

src/main/python/tests/test_matrix_cholesky.py
```diff
@@ -0,0 +1,75 @@
...
+        with self.assertRaises(ValueError) as context:
+            cholesky(self.sds, Matrix(self.sds, m2)).compute()
+
+    def test_asymetric_dim(self):
+        with self.assertRaises(ValueError) as context:
+            cholesky(self.sds, Matrix(self.sds, m3)).compute()
+
+
+if __name__ == "__main__":
+    unittest.main(exit=False)
```
Baunsgaard, line 75: If you want to make git happy, then add a single newline in the end of the files.

src/main/python/tests/test_matrix_reverse.py
```diff
@@ -0,0 +1,68 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.sds = SystemDSContext()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.sds.close()
+
+    def test_basic(self):
+        r = rev(self.sds, Matrix(self.sds, m)).compute()
```
Baunsgaard, line 52: Just a single example of the syntax i think would be great:

`r = Matrix(self.sds, m).rev().compute()`

src/main/python/systemds/matrix/data_gen.py
juliale-15: Ah ok I get your point, but I'm not sure how to catch these two scenarios (positive definite, symmetric) without these check since SystemDS will throw an exception which is not easily readable after calling `compute()`.

Baunsgaard: The system have changed alittle, so that it should be possible to catch these exceptions.
once `compute()` is called. But i don't expect to be able to use it here in this operations content. Therefore, i suggest to still just have a flag, that specifies unsafe, that simply skips these checks.

juliale-15:
> Hi Juliale!
> 
> Great job, ⭐ i like the addition, and nice to see that you also consider the invalid input cases!
> I have a few comments to be addressed but nothing major.
> 
> Question. How did you like the new test setup?

Great! I like it it is more readable now :)

src/main/python/systemds/matrix/data_gen.py
juliale-15: Can you specify what you meant by using a flag unsafe? I could just put a warning into this function saying it is unsafe to use such a matrix and just let the program continue even if the user violates the conditions. Did you mean something like this?

Baunsgaard: Something along the lines of:
```python
def cholesky(sds_context: SystemDSContext, X: Matrix, safe: boolean = false) -> OperationNode:
```

But then again + old comments it becomes:
```python
def cholesky(safe: boolean = false) -> OperationNode:
```
if you move the operations into the matrix object

Baunsgaard:
Hi @juliale-15,
Thank you for your PR :+1: , while merging I only updated the documentation.
I hope you feel motivated to commit more to Open Source and SystemDS in the future!

PR #922, opened 2020-05-27
[WIP][SYSTEMML-2524] Dedicated documentation for source installation on windows

j143:
This doc is dedicated to complete windows dev, end-to-end. This is a preliminary one, though.

cc @Baunsgaard

PR #977, opened 2020-06-18
[SYSTEMDS-501] Documentation webpage framework

Baunsgaard:
The Basic documentation framework from gh-pages back on master branch
removing all previous except basic design.

If possible i would like that the API docs are generated and not pushed like this. but if we can't find a better solution we will keep this one.

This is as disscussed in PRs #949  #922 

and dev mailing list.
http://mail-archives.apache.org/mod_mbox/systemml-dev/202006.mbox/%3Cf787ae9d276247e4a95f7271cc53364d%40tugraz.at%3E


Baunsgaard:
If you want to see how it looks go to: https://baunsgaard.github.io/systemds/

PR #978, opened 2020-06-21
print max memory, spilled items.

phaniarnab:
This PR is to investigate the test failure.


phaniarnab:
Didn't reproduce. Trying again.

phaniarnab:
Didn't reproduce. 
From the stats output, it seemed that the assertion initially failed due to disk spilling - which is weird because those small instructions shouldn't take more than 100ms to execute. I blame it on the cloud system and close this PR for now.

PR #980, opened 2020-06-25
[SYSTEMDS-135] log4j.properties simplified

Baunsgaard:
@corepointer I think you made some changes to the execution?

https://issues.apache.org/jira/browse/SYSTEMML-135

I would like to go further and simply remove the template file `log4j.properties.template`, but i think that is up for discussion.


phaniarnab:
Any reason to use SYSTEMML tag?

Baunsgaard:
it was because it linked to JIRA, task 135, but it has changed now to become SYSTEMDS-135

https://issues.apache.org/jira/browse/SYSTEMDS-135

kev-inn:
What exactly was changed by this? I get that the file is now much smaller, but I don't get why we had all those settings. Are they safe to delete?
I am also in favor of deleting of the template files.

Baunsgaard:
> What exactly was changed by this? I get that the file is now much smaller, but I don't get why we had all those settings. Are they safe to delete?

They should all be safe to delete, all of them are customization of where to log information to. I much prefer the simplicity and control of this simplification. The downside (in my opinion not so much) it is not leveraging some of the "fancy features of log4j", for instance you can split up different logging to different outputs such as saying Hadoop should log to `x.log` while spark to `y.log`.

The one difference that could be debated is the security logging (from hadoop only nothing with out system) that is customized in the old settings, to log to the Hadoop security log.

> I am also in favor of deleting of the template files.

Historic reason for the template files seems to be that a user downloading the system would copy these into real files, that they could then customize for their own settings while being git ignored. Also the release scripts seems to refer to these template files, to copy them into a release as the official settings. 

I think for simplicity in reproducing settings for each user all should share the same settings file, and we therefore remove the template files.

kev-inn:
Thanks for the explanation, sounds good.
Leaving configurations up to the user is better in my opinion and I also agree with your sentiment on the template files.
LGTM :+1:

mboehm7:
LGTM - thanks @Baunsgaard for the cleanup. Just for the historical background: we actually created this log4j file as a copy from the hadoop log4j file (that's were the detailed hadoop configurations came from) and we kept INFO for SystemML by default, because all our messages went through logging (instead of stdout) in order to avoid filling up the stdout queue if nobody is reading from it. Meanwhile, most of our explain (including runtime explain), stats, etc go to stdout so this change is fine. However, I just double checked and a number of warnings/errors (e.g., of the parser - index identifier, parfor stats, codegen explani) go to LOG.info and thus will be silent by default. Let's clean these places up to put all errors/warnings into error level and explain/stats consistently to stdout.

PR #982, opened 2020-07-01
Fine-Grained Privacy Constraints

sebwrede:


mboehm7:
LGTM - thanks for the patch @sebwrede. Sorry for the delay, I think it was about time to merge this in as the PR already started diverging from main. I resolved the conflicts, fixed some minor unnecessary imports, fixed the already changed federated L2SVM cmd parameters, and removed unnecessary uses of parallel streams (we try to avoid that for tiny collections because the performance impact is negative, and there is a danger of deadlocks if the surrounding code already uses the common fork-join pool). Furthermore, I reduce the datasizes in the tests such that the test time for the privacy package came down from 2500s to 170s on my laptop. 

However, I also had to disable the L2SVM and individual operation tests as they were failing with privacy exceptions. Once you're back, let's look over them and see what changed are necessary wrt the rework of the federated backend.

mboehm7:
Also, I created the JIRA epic SYSTEMDS-2604 to keep track of all changes related to privacy constraints.

PR #983, opened 2020-07-02
[MINOR] Fix Tasks

sebwrede:
Changed numbers of tasks to fit the general structure and marked two completed tasks as completed.


sebwrede:
@Baunsgaard, I think this involves your tasks.

Baunsgaard:
yup I think it is just because we are more people, and therefore we overlap these IDs sometime. (a good reason to use JIRA), Will merge it later.

PR #984, opened 2020-07-03
[SYSTEMDS-2540] Stop Exception and Test improvements

Baunsgaard:
This PR resolves an Issue I've had for some time in the testing. It comes with a some changes in the system behavior, and improves the testing experience. 
This PR, has dependencies on #980 (Merge that first!)

Here is a list:

- When calling `runTest` in `AutomatedTestBase` the Std out is caught from when the system is initialized through `DMLScript.main`, This results in a `ByteArrayOutputStream` that is returned, by overwriting the java `System.out`. After the test is executed the standard `System.out` is reinstated. This change improves the ability to check for certain print statements executed in the DML script, while also keeping the Test from printing to console, and in turn crashing our regression tests on GitHub(because of to much printing).
- If a test fails, the standard output is then directed to the error message returned from the systems output from the failing test reducing the amount of text to sift through to find the failing tests output.
- Another major change is that the system now when calling `DMLScript.main` returns exceptions instead of writing them to `System.err`. This should have no impact on user experience because they still get the entire stack trace as before, but it makes it easier for debugging, and testing for specific Exception types that are not handled inside the system.

- A minor change that was in the initial direction this Branch went is NAN handling. This has resulting in this PR also containing a change in the mtd file flag for NAN values to change from 'na.string' to 'naString'. To enable an new argument `naString` when reading csv files in subsequent PRs since `.` is not allowed in variable names or argument names.

**ISSUE** https://issues.apache.org/jira/browse/SYSTEMDS-2540


src/test/java/org/apache/sysds/test/functions/misc/DataTypeCastingTest.java
```diff
@@ -85,10 +85,10 @@ public void testMatrixToMatrix()
 	 * @param cfc
 	 * @param vt
 	 */
-	private void runTest( String testName, boolean matrixInput, boolean exceptionExpected ) 
+	private void runTest( String testName, boolean matrixInput, Class<?> exceptionExpected )
```
kev-inn, line 88: A variable name change would be good, maybe just `exception`?

src/test/java/org/apache/sysds/test/functions/misc/ConditionalValidateTest.java
```diff
@@ -55,52 +55,52 @@ public void setUp() {
...
 	
 	@Test
 	public void testWhileConditionalReadNoErrorNotExists() 
 	{ 
-		runTest( TEST_NAME4, false, false ); 
+		runTest( TEST_NAME4, null, false ); 
 	}
 	
-	private void runTest( String testName, boolean exceptionExpected, boolean fileExists )
+	private void runTest( String testName, Class<?> exceptionExpected, boolean fileExists )
```
kev-inn, line 103: Variable name

src/test/java/org/apache/sysds/test/functions/io/csv/ReadCSVTest1.java
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	protected String getTestClassDir() {
+		return TEST_CLASS_DIR;
+	}
+
+	protected int getId() {
+		return 1;
+	}
+}
```
kev-inn, line 38: I preferred the old version, although it could have used arrays for the IDs/test_numbers and parameterized tests. It comes down to personal preference tough. Same with the other tests.

src/test/java/org/apache/sysds/test/functions/codegenalg/partone/AlgorithmKMeans.java
```diff
@@ -39,14 +39,14 @@
...
-	private final static int cols = 972;
+	private final static int rows = 1241;
+	private final static int cols = 83;
 		
 	private final static double sparsity1 = 0.7; //dense
 	private final static double sparsity2 = 0.1; //sparse
 	
 	private final static double epsilon = 0.000000001;
-	private final static double maxiter = 10;
+	private final static double maxiter = 50;
```
kev-inn, line 49: What is the reason for changed params? Speed?

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinKmeansTest.java
```diff
@@ -28,17 +28,21 @@
...
 	private static final String TEST_CLASS_DIR = TEST_DIR + BuiltinKmeansTest.class.getSimpleName() + "/";
 	private final static double eps = 1e-10;
-	private final static int rows = 3972;
-	private final static int cols = 972;
+	private final static int rows = 1320;
+	private final static int cols = 32;
 	private final static double spSparse = 0.3;
 	private final static double spDense = 0.7;
-	private final static double max_iter = 10;
+	private final static double max_iter = 50;
```
kev-inn, line 45: Again, what is the reason for the param change? Just out of curiosity.

src/test/java/org/apache/sysds/test/applications/GLMTest.java
```diff
@@ -140,7 +140,7 @@ public GLMTest (int numRecords_, int numFeatures_, int distFamilyType_, double d
 				
 
 		// THIS IS TO TEST "INTERCEPT AND SHIFT/SCALE" OPTION ("icpt=2"):
-			{ 200000,   50,  1,  0.0,  1,  0.0,  0.01, 3.0,  10.0,  2.0,  2.5 },   // Gaussian.log	 // CHECK DEVIANCE !!!
+			// { 200000,   50,  1,  0.0,  1,  0.0,  0.01, 3.0,  10.0,  2.0,  2.5 },   // Gaussian.log	 // CHECK DEVIANCE !!!
```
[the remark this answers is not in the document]
kev-inn: Since a fix is probably of interest, maybe a TODO would be appropriate?

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
```diff
@@ -978,19 +989,26 @@ protected void runRScript(boolean newWay) {
...
+			}else {
+				e.printStackTrace();
+				StringBuilder errorMessage = new StringBuilder();
+				errorMessage.append("failed to run script " + executionFile);
+				errorMessage.append("\nexception: " + e.toString());
+				errorMessage.append("\nmessage: " + e.getMessage());
+				errorMessage.append("\nstack trace:");
+				for(StackTraceElement ste : e.getStackTrace()) {
+					errorMessage.append("\n>" + ste);
+				}
```
kev-inn, line 1009: Why do we use `getStackTraceString` at line 1194, but here we don't?

src/main/java/org/apache/sysds/hops/rewrite/RewriteConstantFolding.java
```diff
@@ -98,13 +98,7 @@ private Hop rConstantFoldingExpression( Hop root ) {
...
 			&& ( isApplicableBinaryOp(root) || isApplicableUnaryOp(root) ) )
 		{ 
-			//core constant folding via runtime instructions
-			try {
-				literal = evalScalarOperation(root); 
-			}
-			catch(Exception ex) {
-				LOG.error("Failed to execute constant folding instructions. No abort.", ex);
-			}
+			literal = evalScalarOperation(root);
```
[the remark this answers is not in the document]
kev-inn: It would anyway fail at some point not only if it tries to rewrite, right? Did you check if the error message still is informative enough if it fails here?

scripts/algorithms/Kmeans.dml
```diff
@@ -188,9 +187,9 @@ print ("Number of successful runs = " + as.integer (as.scalar (termination_stats
 print ("Number of incomplete runs = " + as.integer (as.scalar (termination_stats [1, 2])));
 print ("Number of failed runs (with lost centroids) = " + as.integer (as.scalar (termination_stats [1, 3])));
 
-num_successful_runs = as.scalar (termination_stats [1, 1]);
+num_successful_runs = as.integer(as.scalar (termination_stats [1, 1]));
 if (num_successful_runs > 0) {
-    final_wcss_successful = final_wcss * termination_bitmap [, 1];
+    final_wcss_successful = replace(target = final_wcss, pattern = Inf, replacement = 0)* termination_bitmap [, 1];
```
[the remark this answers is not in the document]
kev-inn: LGTM

scripts/algorithms/StepGLM.dml
```diff
@@ -173,15 +173,15 @@ if (dir == "forward") {
 		# Subsequent passes over the features
 		parfor (i in 1:num_features) { 
 			if (as.scalar(columns_fixed[1,i]) == 0) {	
-        
+		
 				# Construct the feature matrix
 				X = cbind (X_global, X_orig[,i]);
-
```
kev-inn, line 179: Are tabs our standard for sure? I see a lot of DML files with spaces.

src/main/java/org/apache/sysds/hops/rewrite/RewriteConstantFolding.java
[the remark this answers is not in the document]
Baunsgaard: The main issue is that previously it did not fail. because the error was caught and not handled, while the program would crash at some arbitrary place afterwards (or not at all). When the system crashed the exception had no reference to this place unless your logging settings were set to ERROR level and in that case you had to still sift through the logging messages to find the exception. that was the actual root course. (to much work for a single exception)

src/test/java/org/apache/sysds/test/applications/GLMTest.java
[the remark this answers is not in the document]
Baunsgaard: Up for debate, since the tests previously would just parse if the execution stopped or stop was called. So technically i changed the foundation of the test making this one fail while it maybe should not. Since we have so many GLM tests i would just leave it out.

src/test/java/org/apache/sysds/test/functions/io/csv/ReadCSVTest1.java
Baunsgaard: I preferred the old as well for syntax reasons, but not for performance reasons (not really a good argument i guess). When i extend like this the test can run in parallel, before they did not, and since i intended at the time to extend the CSV Testing in the next commits i made the changes here.

Then again this is a roundabout way of fixing it in general, because it is still using the awkward AutomatedTestBase and setup, that to make this really cool needs to be rethought.

for reference when executing:
`mvn surefire:test -DskipTests=false -Dtest=org.apache.sysds.test.functions.io.csv.**`
it went from total time of `22.836 s` to `16.548 s`,

src/test/java/org/apache/sysds/test/functions/codegenalg/partone/AlgorithmKMeans.java
Baunsgaard: Kmeans arguments had to be changed to make it possible/more likely for kmeans to actually converge to find some clusters. The previous settings made this highly unlikely therefore the tests ended calling stop resulting in failing tests. This did not crash previously because calling stop would not make the tests fail.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinKmeansTest.java
Baunsgaard: same argument as above.

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
Baunsgaard: Im not sure if i get this comment, at line 1194 for me is: 
```
errorMessage.append("\nfailed to run script: " + executionFile);
```

If i understand correctly, then you are getting confused about the method name at line 1198

```
errorMessage.append(getStackTraceString(e, 0));
```

This is not javas standard getStackTrace. but a method i added to return a custom string of the stack trace. 
Previously the tests would only print one stack trace layer, resulting in a trace that most of the time was useless, and furthermore it would not give out the actual exception method of the lowest level stack. 
This means in practice that the error message most of the time was at script block level saying error in lines between 31 - 33 instead of index out of bounds at exactly line x in file y.

scripts/algorithms/StepGLM.dml
Baunsgaard: The reason why i changed this in this file was because it was inconsistent, and i was debugging why it was crashing.
I should probably revert this, because now that i look at it, it is not done correctly either by me.

src/main/java/org/apache/sysds/hops/rewrite/RewriteConstantFolding.java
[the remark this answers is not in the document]
kev-inn: If it did not fail was the result as expected? Anyway if the error is more obvious this way then this seems better, although it is strange that the rewrite fails even though the code runs.

src/test/java/org/apache/sysds/test/applications/GLMTest.java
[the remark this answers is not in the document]
kev-inn: I also checked some other tests and it is not a rare sight that cases are just commented out. :+1:

src/test/java/org/apache/sysds/test/functions/io/csv/ReadCSVTest1.java
kev-inn: Performance actually seems like a good argument, since we have sooo many tests.

Okay, so they will get another rework? It is not really bad, although changes now affect quite a few classes, but tests anyway are not overworked as often so it should be fine, right?

Yes, the `AutomatedTestBase` really suffers from legacy code, maybe we could select some standard and mark some methods deprecated.

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
kev-inn: Oh, I might have confused the lines. I understood that this is a method you made yourself, but shouldn't you be able to use the same function, because the segment you replaced below is equal (if I am not mistaken) to the one here.

src/test/java/org/apache/sysds/test/functions/io/csv/ReadCSVTest1.java
Baunsgaard: yes deprecation markings could be a way forward.

src/test/java/org/apache/sysds/test/applications/GLMTest.java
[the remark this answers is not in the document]
Baunsgaard: I have added a task on Jira, for the future.
https://issues.apache.org/jira/browse/SYSTEMDS-2580

mboehm7:
Thanks for the initiative @Baunsgaard. Some of these changes are very good, on others I'm kind of split. I would recommend we merge it in (with the changes I made) and see how it works out - if it doesn't we change the defaults accordingly.

* Logging: I changed some of the logging, especially in `VariableCPInstruction` (which makes a substantial portion of all execution instructions) to guard any logging of concatenated messages behind `if( LOG.isDebugEnabled() )`. Yes, the `LOG.debug` would discard the message, but cannot (without inlining) avoid the string concatenation. Would scalar and small-matrix operations this logging would add significant overhead.

* Test Output Buffering: The new output buffering is useful for some scripts that want to check the output. However, for scripts with large outputs this poses a scalability bottleneck (collect all output instead of streaming print & discard), developers have to wait for the output until a test finishes, and we might not see certain problems during development (e.g., development debug output). I now added a method `setOutputBuffering(false)` to disable this output buffering. We go with your new default and let people disable it during development. If this does not work for the team, we simply change the default.

* Rewrite Errors: The constant folding rewrite (applied on compilation and dynamic recompilation during runtime) has been change from reporting errors and continuing execution to a hard crash. Meanwhile this is fine as these rewrites are stable enough and our exceptions have been changed to subclassing `RuntimeException`, so we can leave it like that. 

* Error Handling: Stop and parser issues were on purpose not reported as exceptions but reported in a more user-friendly manner. Now, that we propagate everything as exceptions, a run from command line would throw potentially very deep stacktraces for these. I'm not sure this is a good idea as it simplifies development and testing at the expense of our users. Again, let's give it a try - if it turns out to be unacceptable, we change it to two slightly different entry points for tests and command line invocations. 

* Minor Issues: Furthermore, I removed some unnecessary imports, changed instance methods to static methods where possible, and replaced some loops with more succinct lambda expressions. Regarding formatting, please avoid `} else`.

PR #987, opened 2020-07-10
Check for extra values in test matrix comparison.

skogler:
In the case of expected matrices with zero values, TestUtils was not
correctly checking whether the actual test output matrix does have
values in these places.

### DIFF
--- a/src/test/java/org/apache/sysds/test/TestUtils.java
+++ b/src/test/java/org/apache/sysds/test/TestUtils.java
@@ -217,10 +217,13 @@ public class TestUtils
 
 				readValuesFromFileStreamAndPut(outIn, actualValues);
 			}
-			
+
+			Set<CellIndex> allKeys = new HashSet<>();
+			allKeys.addAll(expectedValues.keySet());
+			allKeys.addAll(actualValues.keySet());
 
 			int countErrors = 0;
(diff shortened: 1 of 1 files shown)


Baunsgaard:
Hi @skogler,

I see that this is a valid point and something we need ! 
Could it maybe make sense to compare the lengths of the HashMaps for better performance?
Since if the number differs then the output is anyway incorrect.

Best regards
Sebastian

skogler:
Hi, yeah sure if its needed I can add it. But then you would not get the
output showing which elements are incorrect. For unit tests I would prefer
getting it, but I don't know all the test cases, can't say if it would be a
problem or not.

Baunsgaard:
Hi @skogler,

Thanks for the PR! :+1: .

I ended up changing it slightly to not unnecessarily increase testing time. the most common case is that the output is correct and the sizes are the same, therefore the edge case where the number of values are different the test fails, while also reporting the incorrect indexes.

The only case that does not report all the incorrect indexes is if the actual and expected contains the same number of elements, but then the test would fail now, since the indexes would have to be different, or the values contained different.

```java
allKeys.addAll(expectedValues.keySet());
if(expectedValues.size() != actualValues.size())
	allKeys.addAll(actualValues.keySet());
```

PR #988, opened 2020-07-13
[SYSTEMDS-2658] Synthetic Minority Over-sampling Technique (SMOTE)

Shafaq-Siddiqi:
Function for oversampling the minority class to deal with class imbalance problem.


phaniarnab:
I know this is in progress, but still commenting: is it possible to replace the `for` blocks with `parfor`?

Shafaq-Siddiqi:
> I know this is in progress, but still commenting: is it possible to replace the `for` blocks with `parfor`?


I tried but there are some matrix dependencies carried out that do not allow parallel run. I will try to refactor the code.

PR #857, opened 2019-03-19
[SYSTEMML-2523] Update SystemML to Support Spark 2.3.0

niketanpansare:
Spark 2.3 (released on February 28, 2018) has updated the Antlr version from 4.3 to 4.7, which throws a warning every time we invoke SystemML.

@bertholdreinwald @romeokienzler @mboehm7 @prithvirajsen @nakul02 @j143  Let's use this PR for discussion and raising potential concerns.

PR #992, opened 2020-07-17
[SYSTEMDS-2523] Update to Spark 2.4.6

Baunsgaard:
- Also bumps Hadoop version to 2.10
- Update to Index docs to reflect the change
- Update Netty to 4.1.47.Final because spark use it.

See: 
#857 

https://github.com/apache/systemds/pull/857#issuecomment-658371761

### DIFF
--- a/.github/workflows/componentTests.yml
+++ b/.github/workflows/componentTests.yml
@@ -52,19 +52,5 @@ jobs:
       run: mvn clean compile test-compile
 
     - name: Component Tests
-      run: |
-        log="/tmp/sysdstest.log"
-        echo "Starting Tests"
-        mvn surefire:test -DskipTests=false -Dtest=org.apache.sysds.test.component.*.** 2>&1 > $log
-        grep_args="SUCCESS"
-        grepvals="$( tail -n 100 $log | grep $grep_args)"
-        if [[ $grepvals == *"SUCCESS"* ]]; then
-        	echo "--------------------- last 100 lines from test ------------------------"
-        	tail -n 100 $log
-        	echo "------------------ last 100 lines from test end -----------------------"
-        	exit 0
-        else
-        	echo "\n $(cat $log)"
-        	exit 1
-        fi
-
+      run: mvn surefire:test -DskipTests=false -Dtest=org.apache.sysds.test.component.** 
+        
--- a/docs/index.md
+++ b/docs/index.md
@@ -28,9 +28,9 @@ SystemDS's distinguishing characteristics are:
   2. **Multiple execution modes**, including Spark MLContext, Spark Batch, Standalone, and JMLC.
   3. **Automatic optimization** based on data and cluster characteristics to ensure both efficiency and scalability.
 
-This version of SystemDS supports: Java 8+,  Python 3.5+, Hadoop 2.6+ (Not 3.X), and Spark 2.1+ (Not 3.X).
+This version of SystemDS supports: Java 8+,  Python 3.5+, Hadoop 2.10+ (Not 3.X), and Spark 2.4.6+ (Not 3.X).
 
-# Links
+## Links
 
 Various forms of documentation for SystemDS are available.
 
@@ -41,4 +41,3 @@ Various forms of documentation for SystemDS are available.
 - The [javadoc API](./api/java/index) contains internal documentation of the system source code.
 - [Install from Source](./site/install) guides through setup from git download to running system.
 - If you want to contribute take a look at [Contributing](https://github.com/apache/systemds/blob/master/CONTRIBUTING.md)
-
--- a/pom.xml
+++ b/pom.xml
@@ -41,10 +41,10 @@
 	</licenses>
 
 	<properties>
-		<hadoop.version>2.6.0</hadoop.version>
+		<hadoop.version>2.10.0</hadoop.version>
 		<antlr.version>4.5.3</antlr.version>
-		<spark.version>2.1.0</spark.version>
-		<scala.version>2.11.8</scala.version>
+		<spark.version>2.4.6</spark.version>
+		<!-- <scala.version>2.11.8</scala.version> -->
 		<scala.binary.version>2.11</scala.binary.version>
 		<maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss z</maven.build.timestamp.format>
 		<enableGPU>false</enableGPU>
@@ -989,7 +989,7 @@
 		<dependency>
 			<groupId>org.codehaus.janino</groupId>
 			<artifactId>janino</artifactId>
-			<version>3.0.8</version>
+			<version>3.0.16</version>
 			<scope>provided</scope>
 		</dependency>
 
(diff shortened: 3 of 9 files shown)


Baunsgaard:
Current error in tests is only in python api that is missing some Java files in its distribution.


``` code
ERROR: setUpClass (test_matrix_rand.TestRand)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/baunsgaard/github/systemds/src/main/python/tests/test_matrix_rand.py", line 50, in setUpClass
    cls.sds = SystemDSContext()
  File "/home/baunsgaard/github/systemds/src/main/python/systemds/context/systemds_context.py", line 97, in __init__
    raise Exception("Exception in startup of GatewayServer: " + stderr)
Exception: Exception in startup of GatewayServer: Exception in thread "main" java.lang.NoClassDefFoundError: com/ctc/wstx/io/InputBootstrapper
```

Baunsgaard:
The reason why it did not work was because the packaged jar file did not contain all the packages required now (due to me updating some dependencies). Therefore i decided to change the python to not decompress the distribution zip but use the environment variable SYSTEMDS_ROOT to locate the target folder and in turn all the jar files. This means we don't have to include java files in our python distribution, and the python execution comes closer to our `bin/systemds` distribution script.

The down side is that we have to install SystemDS or compile from scratch to make the Python package work, but i think it is fair to do that.

kev-inn:
Does this mean the python package (pip install) no longer contains the jar files? If you are using the target (as I already did at some point before) and use all those jars for the python package it becomes huge (>100MB), which I think is something we would like to not do and keep it relatively small.

Baunsgaard:
> Does this mean the python package (pip install) no longer contains the jar files? If you are using the target (as I already did at some point before) and use all those jars for the python package it becomes huge (>100MB), which I think is something we would like to not do and keep it relatively small.

Yes it will no longer contain any jar files, making it much smaller.

kev-inn:
Shouldn't pip packages not be dependant on other ressources? That would at least make it much simpler to install and I would be much in favor of a single `pip install` command.

Baunsgaard:
> Shouldn't pip packages not be dependant on other ressources? That would at least make it much simpler to install and I would be much in favor of a single `pip install` command.

I agree, but we have to have the user setup java anyway and in that context i think installing and setting up systemds is fine.
One solution is that on first execution download the dependencies if SYSTEMDS_ROOT not set? I could live with that.

kev-inn:
That also sounds like a better option (in my opinion), but then we would have to define a default path because setting an environment variable for the system will be hard. Also we would not like to download all the jars, which would save us no download work, but instead anyway set the jars to download, so why not do it when building the package and distribute it combined? That would also lead to all the data be in a single directory where pip packages are stored. Also setting up java is a single command (we should actually force java 8 execution, which we don't do atm jus `java ...` right?).

Baunsgaard:
> That also sounds like a better option (in my opinion), but then we would have to define a default path because setting an environment variable for the system will be hard.

I think this is just a decision we have to make at some point. there are many good options for install location `/etc/systemds/...` or `/lib/systemds` ? If we do make such a decision we could provide a install for Ubuntu, `sudo apt install systemds`!

> Also we would not like to download all the jars, which would save us no download work, but instead anyway set the jars to download, so why not do it when building the package and distribute it combined? That would also lead to all the data be in a single directory where pip packages are stored.

Well, currently i either had to figure out which packages specifically the Python lib require to execute, or use the packages from systemds that was downloaded anyway. To keep it simple in development i chose the later option. I agree that one install command through pip would be better, but i also would like that we don't install, copy or download the jars twice.

> Also setting up java is a single command (we should actually force java 8 execution, which we don't do atm jus `java ...` right?).

We need to move forward towards java 11. but not for this release: Hadoop 3.x and Spark 3.x use/support Java 11.
https://hadoop.apache.org/docs/current/
https://spark.apache.org/docs/3.0.0/

phaniarnab:
I didn't go through all the comments in the cited PRs, but I'm curious to see how this Spark/Hadoop upgrade impacts the performance. It might not improve anything, but also should not deteriorate. I am not sure if we have a performance suit. Usually, it is better to fire a few performance tests to be very sure for these kinds of upgrades. 
Please ignore if that is already done.

Baunsgaard:
> I didn't go through all the comments in the cited PRs, but I'm curious to see how this Spark/Hadoop upgrade impacts the performance. It might not improve anything, but also should not deteriorate. I am not sure if we have a performance suit. Usually, it is better to fire a few performance tests to be very sure for these kinds of upgrades.
> Please ignore if that is already done.

If you have some scripts you could test it, unfortunately i don't have such a test suite yet, and currently i don't have access to a cluster with these versions install, once we get this we should as you say test it. I would not expect huge performance increases, but the main point of upgrading is not performance but to address the open CVEs and use the versions other ppl are using.

phaniarnab:
> > I didn't go through all the comments in the cited PRs, but I'm curious to see how this Spark/Hadoop upgrade impacts the performance. It might not improve anything, but also should not deteriorate. I am not sure if we have a performance suit. Usually, it is better to fire a few performance tests to be very sure for these kinds of upgrades.
> > Please ignore if that is already done.
> 
> If you have some scripts you could test it, unfortunately i don't have such a test suite yet, and currently i don't have access to a cluster with these versions install, once we get this we should as you say test it. I would not expect huge performance increases, but the main point of upgrading is not performance but to address the open CVEs and use the versions other ppl are using.

True, and as you rightly said, we should move forward towards java 11 and Spark 3.x. Maybe not much needed for this PR, but extensive performance tests will be of the highest priority for java version upgrades. I have scripts, but those will not be sufficiently helpful for this, and also the cluster is a problem. I will work on this.
LGTM for this PR.

kev-inn:
> > That also sounds like a better option (in my opinion), but then we would have to define a default path because setting an environment variable for the system will be hard.
> 
> I think this is just a decision we have to make at some point. there are many good options for install location `/etc/systemds/...` or `/lib/systemds` ? If we do make such a decision we could provide a install for Ubuntu, `sudo apt install systemds`!
> 

Let's not forget about windows, but I agree that good options exist for all distros.
> > Also we would not like to download all the jars, which would save us no download work, but instead anyway set the jars to download, so why not do it when building the package and distribute it combined? That would also lead to all the data be in a single directory where pip packages are stored.
> 
> Well, currently i either had to figure out which packages specifically the Python lib require to execute, or use the packages from systemds that was downloaded anyway. To keep it simple in development i chose the later option. I agree that one install command through pip would be better, but i also would like that we don't install, copy or download the jars twice.
> 

That was also the option I choose at first, but it seems a bit extreme that even if I choose python for development, as I expect many of the end users will do, I have to download way more jars which I will never make use of. Downloading twice is also a problem, since if I choose to write code in dml and python (how big do we expect the end user overlap to be?) that would be unnecessary. I think we should either make a blacklist of packages to not download/copy or let mvn create use a collection of exactly those packages that we need, in either case it is necessary to figure out which package are needed.
> > Also setting up java is a single command (we should actually force java 8 execution, which we don't do atm jus `java ...` right?).
> 
> We need to move forward towards java 11. but not for this release: Hadoop 3.x and Spark 3.x use/support Java 11.
> https://hadoop.apache.org/docs/current/
> https://spark.apache.org/docs/3.0.0/

Then let's keep it as it is for now

Baunsgaard:
Closing, because of plans to update Spark version after 2.0 release of systemds.

PR #996, opened 2020-07-19
Implementation of Hyperband

OsChri:
-including test and documentation
-hardcoded for lmCG
-only works with continuous hyper parameters

2 Problems encountered with eval():
- eval("lmCG", ...) seems to call "lm" internally, but only when used inside a built-in function. when used in a user-defined dml-script, behaviour is as expected. calling lmCG() inside hyperband.dml directly solved the issue
- when calling ...= eval(...,list(W=w,...)) to validate a model, w can't be used for other purposes anymore, e.g. printing it or storing it in a matrix. calling the prediction function (lmpredict) directly solved the issue

### DIFF
--- a/docs/site/builtins-reference.md
+++ b/docs/site/builtins-reference.md
@@ -32,6 +32,7 @@ limitations under the License.
     * [`discoverFD`-Function](#discoverFD-function)
     * [`glm`-Function](#glm-function)
     * [`gridSearch`-Function](#gridSearch-function)
+    * [`hyperband`-Function](#hyperband-function)
     * [`img_brightness`-Function](#img_brightness-function)
     * [`img_crop`-Function](#img_crop-function)
     * [`img_mirror`-Function](#img_mirror-function)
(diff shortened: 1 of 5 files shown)


mboehm7:
LGTM - thanks for this great new builtin @OsChri. I just slightly changed the test to use fixed seeds and replaced the for loops for joint sorting of two matrices with vectorized permutation matrix multiplications:
```
reorder = order(target=scoreboard, index.return=TRUE);
P = table(seq(1,n_i), reorder); # permutation matrix
scoreboard = P %*% scoreboard;
candidateWeights = P %*% candidateWeights;
```

mboehm7:
Also thanks for catching the `eval` issues - I'll fix them in subsequent commits.

mboehm7:
Thanks again for catching theses issues @OsChri - the second one is fixed in master, while the first one originates from side effects of IPA as described here: https://issues.apache.org/jira/browse/SYSTEMDS-2576

mboehm7:
Just for closure - now all raised bugs have been fixed in master. @OsChri this was a great catch.

PR #999, opened 2020-07-22
Notebook for SystemDS on colab for developers

j143:
**What this PR do?**
* Creates a workspace with all the dependencies for project build.
* Helps prototype the DML code in browser.

Preview at [google colab ↗](https://colab.research.google.com/github/j143/systemds/blob/colab-notebook-initial/notebooks/systemds_dev.ipynb)

**Discussions:**
https://www.mail-archive.com/dev@systemml.apache.org/msg01763.html
https://www.mail-archive.com/dev@systemds.apache.org/msg00130.html

---
**TL;DR:** (Do not forget to say hi to kittens! 😸 )

image

### DIFF
--- a/notebooks/systemds_dev.ipynb
+++ b/notebooks/systemds_dev.ipynb
@@ -1,0 +1,882 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "SystemDS on Colaboratory.ipynb",
+      "provenance": [],
+      "collapsed_sections": [],
+      "include_colab_link": true
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github.com/apache/systemds/blob/master/notebooks/colab/systemds_dev_standalone.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "_BbCdLjRoy2A",
+        "colab_type": "text"
+      },
+      "source": [
+        "## Developer notebook for Apache SystemDS"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "I76oWp7foiyF",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        ""
+      ],
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "kvD4HBMi0ohY",
+        "colab_type": "text"
+      },
+      "source": [
+        "#### Install Java\n",
+        "This installs Java 8 of Open JDK"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "fUhBhrGmyAvs",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        "!apt-get install openjdk-8-jdk-headless -qq > /dev/null"
+      ],
+      "execution_count": 1,
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "b4Kjvk_h1AHl",
+        "colab_type": "text"
+      },
+      "source": [
+        "#### Set Environment Variables\n",
+        "Set the locations where Spark and Java are installed."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "8Xnb_ePUyQIL",
+        "colab_type": "code",
+        "colab": {
+          "base_uri": "https://localhost:8080/",
+          "height": 85
+        },
+        "outputId": "d5546e63-97df-42a9-ce53-f9f361eaff30"
+      },
+      "source": [
+        "import os\n",
+        "os.environ[\"JAVA_HOME\"] = \"/usr/lib/jvm/java-8-openjdk-amd64\"\n",
+        "!update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java\n",
+        "!java -version\n",
+        "# os.environ[\"SPARK_HOME\"] = \"/content/spark-2.4.5-bin-hadoop2.7\""
+      ],
+      "execution_count": 2,
+      "outputs": [
+        {
+          "output_type": "stream",
+          "text": [
+            "update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java to provide /usr/bin/java (java) in manual mode\n",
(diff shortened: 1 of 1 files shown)


notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "b4Kjvk_h1AHl",
+        "colab_type": "text"
+      },
+      "source": [
+        "#### Set Environment Variables\n",
+        "Set the locations where Spark and Java are installed."
```
Baunsgaard, line 82: At install the java should be on path, so that you don't need to set that environment variable.

notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+        "id": "I76oWp7foiyF",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        ""
+      ],
+      "execution_count": null,
+      "outputs": []
+    },
```
Baunsgaard, line 49: This above code cell contains nothing, Maybe make some introduction text saying ...

```code
In this notebook we show ... x ... becauase ... y etc. 
it is separated into 3 parts...
1. install [link]
2. code [link]
3. results [link]
```

notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "fUhBhrGmyAvs",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        "!apt-get install openjdk-8-jdk-headless -qq > /dev/null"
```
Baunsgaard, line 69: Could we not move most of the install command into this block. like:

```
sudo apt install openjdk-8-jdk-headless
sudo apt install maven
```

notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+        "spark_version = 'spark-2.4.6'\n",
+        "hadoop_version = 'hadoop2.7'\n",
+        "spark_path = f\"/opt/{spark_version}-bin-{hadoop_version}\"\n",
+        "if not os.path.exists(spark_path):\n",
+        "  run(f\"wget -q -nc -O apache-spark.tgz https://downloads.apache.org/spark/{spark_version}/{spark_version}-bin-{hadoop_version}.tgz\")\n",
+        "  run('tar zxf apache-spark.tgz -C /opt')\n",
+        "  run('rm -f apache-spark.tgz')\n",
+        "\n",
+        "os.environ[\"SPARK_HOME\"] = spark_path\n",
+        "os.environ[\"PATH\"] += \":$SPARK_HOME/bin\"\n"
```
Baunsgaard, line 271: I think a version using spark, and one without would be better. Since adding the spark install, unnecessarily complicates this notebook. Also using spark inside this notebook should lead to no improved performance only overhead, since it is executed in a single node.

notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+        "id": "SaPIprmg3lKE",
+        "colab_type": "code",
+        "colab": {
+          "base_uri": "https://localhost:8080/",
+          "height": 153
+        },
+        "outputId": "1eee45c1-cd92-480e-d27c-a2229086b5a0"
+      },
+      "source": [
+        "!git clone https://github.com/apache/systemds systemds\n",
```
Baunsgaard, line 311: Could we use the trick of only downloading the last commit `--depth=1`

notebooks/systemds_dev.ipynb
```diff
@@ -0,0 +1,882 @@
...
+      "metadata": {
+        "id": "Tcxkh8cdUy1V",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        "!echo 'export SYSTEMDS_ROOT='$(pwd) >> ~/.bashrc\n",
+        "!echo 'export PATH=$SYSTEMDS_ROOT/bin:$PATH' >> ~/.bashrc"
+      ],
+      "execution_count": 16,
```
Baunsgaard, line 520: It seems like after some debugging from my side that all the exports doesn't work inside this notebook. This also includes the echo to ~/.bashrc.
also fixing block `16` adding a `!source ~/.bashrc` does not work.

j143:
> as a final repeated note Install is way to large again, i think i have said this many times and In this case again, it does not make much sense to install spark for this execution.

1. Hey @Baunsgaard - Sorry that I am not complying there. This notebook is meant for developers (Engine) mainly.
   This acts as a companion to our documentation for example `systemds-distributed-guide`, `systemds-standalone`

2. We will have different notebook something like `systemds-user.ipynb`. In which all the setup will be done in two lines)
  a. `pip install systemds`
  b. `pip install spark` that's it.

(1st comment)

notebooks/systemds_dev.ipynb
j143: Yes, will do.

j143:
Let us confine this notebook to engine developers only.

About 
> what does the results mean? can we use them for anything?

Let us take this in a visualization `dml` developer focused notebook.

Is this okay for you?

notebooks/systemds_dev.ipynb
j143: I did `!cat ~/.bashrc`, could see `SYSTEMDS_ROOT`.

notebooks/systemds_dev.ipynb
j143: Yes, for now I am adding this as an optional.

notebooks/systemds_dev.ipynb
j143: done

notebooks/systemds_dev.ipynb
j143: I have reorganized the sections with more description!

notebooks/systemds_dev.ipynb
j143: about `JAVA_HOME`, `mvn` seems to use this variable.

notebooks/systemds_dev.ipynb
Baunsgaard: so the point here is not that it does not put the lines into .bashrc, the point is that it does not use them for the following commands, if they did you would be able to change the `./bin/systemds` to `systemds`.
i did not have success with making this work with either of the two methods.

Baunsgaard:
>     1. Let us confine this notebook to engine developers only.

I think that the engine developers should not be the ones such a notebook is directed at, but again it might just be my opinion.
Again a matter of priority, if you think that this notebook is needed for engine developers then fine but in general the 'beginner' user should be taken more into account, therefore addressing such a guide for them would be great. 

>     2. About
> > what does the results mean? can we use them for anything?
>
> Let us take this in a visualization, `dml` developer focused notebook.

I think that would be unnecessarily confusing and duplication of work elsewhere. The discussing results should be a necessary extension of executing anything. That said i might be the only one with that opinion, therefore if you think that discussion of results is not needed then fine.

> Is this okay for you?
> 
>     1. If you still want this notebook to be removed of spark, happy to do that.

Yes i think spark is an unnecessary addition in such a guide.
Our system is able to leverage spark and make operations faster, but it is not limited to only run in cases where spark is installed. In this specific case where you want to guide people into using a system it should be easy to do and jump into the interesting parts as early as possible, not fight with setup, which in many cases deter users.

j143:
Thank you for the review. :tada:

PR #1007, opened 2020-08-07
[WIP][DOC] DML to R translation guide

j143:
**What this PR do?**

1. some common differences between R and DML
2. The examples should be runnable to make the
  guide more interesting.

**TL;DR:**

Keep the differences b/w `dml` and `R` in a well-lit place

image


j143:
cc @mboehm7 @Shafaq-Siddiqi 

May be you would like to
1. add some more examples.
2. Any comments over this for arranging the doc

Thank you.

PR #1009, opened 2020-08-07
[WIP][SYSTEMDS-2601] Relational Operators Support in Frames

Shafaq-Siddiqi:
Support for frame comparisons (==, !=, >, < , >= ,<=)

### DIFF
--- a/src/main/java/org/apache/sysds/parser/DMLTranslator.java
+++ b/src/main/java/org/apache/sysds/parser/DMLTranslator.java
@@ -1812,6 +1812,10 @@ public class DMLTranslator
 				target.setDataType(DataType.MATRIX);
 				target.setValueType(ValueType.FP64);
 			}
+			else if(left.getDataType() == DataType.FRAME || right.getDataType() == DataType.FRAME) {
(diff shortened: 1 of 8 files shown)


mboehm7:
LGTM - thanks for the patch @Shafaq-Siddiqi, this is a great addition. I only made some minor modifications related to parser error handling (only frame-frame supported), the integration as `binaryOperations` (to make it consistent with matrix blocks), modified CP/Spark instructions to fix systematic typos ("comparision") and avoid unnecessary code redundancy, fixed the handling of RDD/broadcast dependencies (to avoid too eager cleanup), and fixed the formatting of tests (position of java annotations).

In a subsequent PR, please first run an experiment to measure runtime performance. Then, reimplement the frame block comparison operation to directly allocate and update `BooleanArray`s instead over going through the indirection of string arrays. This has the advantage of avoiding allocation, avoiding unnecessary string serialization and parsing, and avoiding cache-unfriendly operations (row vs column layout). Finally, measure the resulting performance improvement (e.g., for 100 binary frame comparisons at script level).

PR #1020, opened 2020-08-13
[SYSTEMDS-2617] Add columnNames builtin to get Frame col names

kev-inn:
This PR adds a new function `columnNames()` which takes a `Frame` and returns a `Frame` containing the names of the columns.
I did not yet add any documentation since I am not aware of the current process. Should documentation be added in the initial PR? Should a new PR be made just for documentation of this function? Is documentation added in bulk?
I also detected that the functions `typeOf` and `detectSchema`, which belong to the same "category" (instruction type) as this function are defined in `src/main/java/org/apache/sysds/runtime/functionobjects/Builtin.java`, but I believe those definitions are not used. Am I overlooking something and this function should be added there?

### DIFF
--- a/src/main/java/org/apache/sysds/common/Builtins.java
+++ b/src/main/java/org/apache/sysds/common/Builtins.java
@@ -84,6 +84,7 @@ public enum Builtins {
 	CONFUSIONMATRIX("confusionMatrix", true),
 	COR("cor", true),
 	DETECTSCHEMA("detectSchema", false),
+	COLUMN_NAMES("columnNames", false),
 	DIAG("diag", false),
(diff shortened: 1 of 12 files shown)


mboehm7:
LGTM - thanks for the addition @kev-inn. I only changed the name to `colnames(X)` for consistency with the existing R function. Similarly, I change `typeOf` to `typeof`, and removed some unnecessary imports. Down the road we might actually move all these frame meta data operations to unary aggregates as they fit the schema of column aggregates.

Regarding documentation, I personally would prefer documentation in a batches of builtin functions (one PR per batch), but there are no general guidelines yet (we want to modify the contributing guidelines with @j143  anyway).

PR #1023, opened 2020-08-17
[SYSTEMDS-2625]: Cleanup GMM [SYSTEMDS-200] with builtin for LogSumExp

Shafaq-Siddiqi:
This cleanup includes the functions reordering, branching cleanups and elimination of lists, now matrices are used for storing covariance. The private function logSumExp is replaced with a new builtin logSumExp().

### DIFF
--- a/scripts/builtin/gmm.dml
+++ b/scripts/builtin/gmm.dml
@@ -60,259 +60,192 @@ return (Matrix[Double] weights, Matrix[Double] labels, Integer df, Double bic)
   if(model != "VVV" & model != "EEE" & model != "VVI" & model != "VII")
     stop("model not supported, should be in VVV, EEE, VVI, VII");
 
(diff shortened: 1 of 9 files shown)


mboehm7:
Thanks for the cleanup @Shafaq-Siddiqi - when finalizing the rework, could you please also have a look into (1) making the default number of components consistent with the documentation (1 vs 3), and (2) add tests with default configurations but more than one component (right now it fails for me with dimension mismatches with n=10 but runs fine with n=1). Once this PR is out of WIP status, I'll have a look and merge it in.

mboehm7:
LGTM - thanks for the cleanup @Shafaq-Siddiqi. The scenario with 10 components was still failing, but after some debugging it turned out this was due to Kmeans not converging. During the merge I fixed the hard-coded maximum iterations for Kmeans, some formatting issues, and vectorized part of the cholesky computation. With those changes it ran fine.

PR #1025, opened 2020-08-17
[DISCUSSION][WIP] Algorithm documentation

j143:
**What does this PR do?**
1. Carefully read each algorithm and add to the codebase.
2. A discussion would be nice over how and where we would want to keep our 
algorithm doc.

**TL;DR:**
Top level algorithm documentation is not available yet.


j143:
cc @Baunsgaard

PR #1027, opened 2020-08-18
[SYSTEMDS-2558,2554,2561] Fed frame recode transform (decode) support

kev-inn:
Adds decode support for recode, pass-through and composite (containing only those two).

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
@@ -267,7 +267,8 @@ public class FederatedWorkerHandler extends ChannelInboundHandlerAdapter {
 			return udf.execute(ec, inputs);
 		}
 		catch(Exception ex) {
-			return new FederatedResponse(ResponseType.ERROR, ex.getMessage());
+			return new FederatedResponse(ResponseType.ERROR, new FederatedWorkerHandlerException(
+				"Exception of type " + ex.getClass() + " thrown when processing request", ex));
 		}
 	}
 
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
@@ -97,7 +97,7 @@ public class ParameterizedBuiltinFEDInstruction extends ComputationFEDInstructio
 			return new ParameterizedBuiltinFEDInstruction(null, paramsMap, out, opcode, str);
 		}
 		else {
-			throw new DMLRuntimeException("Unknown opcode (" + opcode + ") for ParameterizedBuiltin Instruction.");
+			throw new DMLRuntimeException("Unsupported opcode (" + opcode + ") for ParameterizedBuiltinFEDInstruction.");
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -268,7 +268,8 @@ private FederatedResponse execUDF(FederatedRequest request) {
 			return udf.execute(ec, inputs);
 		}
 		catch(Exception ex) {
-			return new FederatedResponse(ResponseType.ERROR, ex.getMessage());
+			return new FederatedResponse(ResponseType.ERROR, new FederatedWorkerHandlerException(
+				"Exception of type " + ex.getClass() + " thrown when processing request", ex));
```
kev-inn, line 272: `ex.getMessage()` would sometimes not suffice and instead be confusing. A `ArrayIndexOutOfBoundsException` would have for example just have the index as a message, therefore our response would just contain an arbitrary number and the problem would not be explained

src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
```diff
@@ -19,103 +19,217 @@
...
 		// process remaining parts and build a hash map
-		LinkedHashMap<String,String> paramsMap = constructParameterMap(parts);
-	
+		LinkedHashMap<String, String> paramsMap = constructParameterMap(parts);
+
 		// determine the appropriate value function
 		ValueFunction func = null;
-		if( opcode.equalsIgnoreCase("replace") ) {
-			func = ParameterizedBuiltin.getParameterizedBuiltinFnObject(opcode);
-			return new ParameterizedBuiltinFEDInstruction(new SimpleOperator(func), paramsMap, out, opcode, str);
```
kev-inn, line 86: I formatted this file (sorry reviewer), but I think since those files are quite new we should start by abiding to our code style.

src/main/java/org/apache/sysds/runtime/transform/encode/Encoder.java
```diff
@@ -145,7 +145,7 @@ public Encoder subRangeEncoder(int colStart, int colEnd) {
 	 */
 	protected void mergeColumnInfo(Encoder other, int col) {
 		// update number of columns
-		_clen = Math.max(_colList.length, col - 1 + other.getNumCols());
+		_clen = Math.max(_clen, col - 1 + other._clen);
```
kev-inn, line 148: This was my mistake in the last PR `_colList.length` is something different to `_clen` and `.getNumCols()` will be wrong once we add dummycoding (it gives us the cols after encoding).

src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
mboehm7: don't worry about the formatting - more severe was the above deletion of the recently added replace. ;-)

mboehm7:
LGTM - thanks for the patch @kev-inn. Overall this is nice, I just made some minor modifications: (1) fixed the export in `CacheableData` which did not handle federated objects (it might have worked due to previous transfer to the coordinator by chance), (2) fixed the removed replace handling in parsing of parameterized builtin instructions, (3) generalized the improved exception handling from UDFs to instruction execution and clear commands, and (4) fixed a javadoc issue I introduced in a previous commit.

PR #1031, opened 2020-08-19
[SYSTEMDS-2555] Federated transform dummycoding encoding decoding

kev-inn:
Adds dummycoding support for federated `transformencode` and `transformdecode`.
Merge PR #1027 first.


mboehm7:
LGTM - thanks @kev-inn for this solid patch. I only did some minor cleanups (arrays.fill/streams where applicable, removed csv property handling on read) and left a performance TODO. For dummycode decode, it would be important (once we come to performance improvements) to work with sparse matrices directly to avoid unnecessarily bad asymptotic behavior (O(m n log n) vs O(nnz))

PR #1036, opened 2020-08-23
[SYSTEMDS-2635] Builtin for missing value imputation using forward an…

Shafaq-Siddiqi:
…d backward fill


Shafaq-Siddiqi:
The test fails due to missing R dependency.

mboehm7:
@Shafaq-Siddiqi has the R dependency been resolved by adding it to the installed R packages, or do we still need to update the docker image for the github actions?

mboehm7:
LGTM - thanks for the additional builtin function @Shafaq-Siddiqi. I only slightly modified the formatting and changed the for loops over columns to parfor loops.

PR #1037, opened 2020-08-24
Federated worker stats

Baunsgaard:
This PR contains Federated Worker stats such that if the worker is started with the -stats argument it will print execution stats each time the CLEAR command is sent.

Also in this PR the test, now isolate the workers even more inside their own JVMs to no corrupt the stats from the main JVM.


Baunsgaard:
merged

PR #1040, opened 2020-08-26
Built-in function for bivariate statistics

OlgaOvcharenko:
Added bivar builtin


Baunsgaard (commented):
Hi @OlgaOvcharenko,

Thanks for the PR, :racehorse: the tests are all working the tests on git crashed due to the startup timer being to small.
Also the code it looks fine. But there is some changes that would make life easier down the road. Therefore i suggest you take a look at them.

The PR technically contains two different things, first the bivariate statistics, secondly min-max col and row support. It would be nice if that was two separated PRs, but not a requirement.

For the testing, i suggest taking a look at:

`src/test/java/org/apache/sysds/test/functions/aggregate/FullAggregateTest.java`
`src/test/java/org/apache/sysds/test/functions/aggregate/FullColAggregateTest.java`

OlgaOvcharenko:
@Baunsgaard I also pushed modified FederationUtils (with fed min, max, sum, mean), because it seems like privacy tests are failing without this changes

Baunsgaard:
@OlgaOvcharenko Thanks for the PR. :1st_place_medal: 

While merging i fixed the indentation in the test files, and reduced the number of tests run on RowCol aggregate since each test require startup of each worker of 1 second it really grows fast. Thanks again.

PR #1041, opened 2020-08-27
[WIP][top-k slicing] current update for review

gilgenbergg:
- Additional real data tests
- Unit test on synthetic data
- Bugfixes in pruning output
- Fixes on enumeration of nodes in both join and union cases.


mboehm7:
LGTM - thanks @gilgenbergg for the update and tests, as well as incorporating the licenses.

PR #1046, opened 2020-08-31
[SYSTEMDS-2556,2560] Add federated Encoder impute support and improve Omit

kev-inn:
Adds support for federated execution for the final encoder `EncoderMVImpute`. This should finish support for federated transform operations (perf and improvements being TODO).

## `EncoderMVImput`

Note that I removed quite a bit of code from `EncoderMVImpute` as it seems to not be in use at all, please confirm if this is fine.

## `EncoderOmit`

I added the perf improvement of the TODO, since omit anyway had some problems which needed a fix.


mboehm7:
LGTM - thanks @kev-inn  for the added impute, and general clean. I modified the Encoder interface though to keep it independent of data structures from the federated backend. Besides formatting, I also left another TODO in Omit (but low priority).

PR #1047, opened 2020-09-02
[SYSTEMDS-2651] Add faster multiple federated workers startup

kev-inn:
Adds a new function to start multiple federated workers. The function first starts multiple processes and then waits for all of them to be ready for a connection by pinging (by sending a `CLEAR`) them.

This method cannot be used for threads (running federated workers), because they share static variables, therefore their startup interferes with each other. Switching purely to processes should definitely be considered.

### DIFF
--- a/src/test/java/org/apache/sysds/test/AutomatedTestBase.java
+++ b/src/test/java/org/apache/sysds/test/AutomatedTestBase.java
@@ -111,7 +111,7 @@ public abstract class AutomatedTestBase {
 
 	public static final int FED_WORKER_WAIT = 1000; // in ms
 	// max time to wait when checking if worker accepts connections
-	public static final int FED_WORKER_CONNECTION_WAIT = 5000;
+	public static final int FED_WORKER_CONNECTION_WAIT = 10000;
 
 	// With OpenJDK 8u242 on Windows, the new changes in JDK are not allowing
 	// to set the native library paths internally thus breaking the code.
@@ -1391,12 +1391,14 @@ public abstract class AutomatedTestBase {
 		}
 	}
 
-	private static void waitForFederatedWorker(int port) throws TimeoutException {
+	protected static void waitForFederatedWorker(int port) throws TimeoutException {
 		long start = System.currentTimeMillis();
 		boolean failed = true;
 		while(System.currentTimeMillis() - start <= FED_WORKER_CONNECTION_WAIT) {
 			FederatedResponse response = null;
 			try {
+				// don't request too frequently
+				Thread.sleep(FED_WORKER_CONNECTION_WAIT / 1000);
 				response = executeFederatedOperation(new InetSocketAddress("localhost", port),
 					new FederatedRequest(FederatedRequest.RequestType.CLEAR)).get();
 			}
@@ -1422,10 +1424,12 @@ public abstract class AutomatedTestBase {
 	 * @param ports the ports to use
 	 * @return the processes associated with the workers
 	 */
-	protected Process[] startLocalFedWorkers(int... ports) throws TimeoutException {
-		Process[] processes = Arrays.stream(ports).mapToObj(port -> startLocalFedWorker(port, false))
-			.toArray(Process[]::new);
-		for (int port : ports) {
+	protected Process[] startLocalFedWorkers(int... ports) throws TimeoutException, IOException {
+		Process[] processes = new Process[ports.length];
+		for(int i = 0; i < ports.length; i++) {
+			processes[i] = startLocalFedWorker(ports[i], false);
+		}
+		for(int port : ports) {
 			waitForFederatedWorker(port);
 		}
 		return processes;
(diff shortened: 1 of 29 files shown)


Baunsgaard:
agree! also really like the PR!

Baunsgaard:
But the issue is that it is hard / impossible to debug the processes while it is possible when inside the same JVM using threads, and in that context if we really want to make it smart then we would need to change the system such that it does not produce these static variables and objects that potentially also lead to more bugs in other parts of the program.

kev-inn:
> But the issue is that it is hard / impossible to debug the processes while it is possible when inside the same JVM using threads, and in that context if we really want to make it smart then we would need to change the system such that it does not produce these static variables and objects that potentially also lead to more bugs in other parts of the program.

Yes, that is the one major downside. Intellij supports attaching to the debugger to process, but I am not sure if that process can be automated. [`ThreadLocal`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) would be another option for using threads, but I believe that would lead to performance loss, even though we only get debug improvements.

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
```diff
@@ -1283,44 +1288,89 @@ public static int getRandomAvailablePort() {
...
+				// failed, try again if fed worker wait has not passed yet
+			}
+			if(response != null) {
+				if(!(response.isSuccessful()))
+					LOG.error("Ping failed: " + response.getErrorMessage());
+				else
+					break;
+			}
+		}
+	}
```
sebwrede, line 1314: If the worker does not respond and the time limit has been reached then the execution would just continue as usual from here. This means that it will fail later when it tries to send a request to the worker. This will make it harder to debug compared to handling it when trying to create the worker.

src/test/java/org/apache/sysds/test/functions/privacy/FederatedL2SVMTest.java
```diff
@@ -314,7 +314,7 @@ private void federatedL2SVM(Types.ExecMode execMode, Map<String, PrivacyConstrai
 		if(rtplatform == Types.ExecMode.SPARK) {
 			DMLScript.USE_LOCAL_SPARK_CONFIG = true;
 		}
-		Process t1 = null, t2 = null;
+		Process[] ts = new Process[0];
```
sebwrede, line 317: This has been changed to use threads since the use of processes failed in GitHub Actions.

src/test/java/org/apache/sysds/test/functions/privacy/FederatedWorkerHandlerTest.java
```diff
@@ -199,7 +199,7 @@ public void federatedSum(Types.ExecMode execMode, PrivacyLevel privacyLevel, Cla
 
 		assert(checkedPrivacyConstraintsContains(privacyLevel));
 
-		TestUtils.shutdownThread(t);
+		TestUtils.shutdownProcess(t);
```
sebwrede, line 202: This has also been changed to threads.

sebwrede (commented):
I think it looks good. The privacy tests are conflicting with what we have on master at the moment because the tests have been changed to use threads, which is required for them to pass in GitHub Actions. When merging this, I think we should keep the thread-based implementation which is currently on master.

Baunsgaard:
PR, on hold since we haven't found the perfect setup, and therefore the optimization is invalid, since it disables debugging.

kev-inn:
I fixed the problem with debugging, please read the last commit message to get a view of the current state.
I am not sure if the enhancement (keeping workers alive) should be put into this PR or a new one, since I can imagine that there will be a few trade-offs which require discussion.

corepointer (commented):
A few questions to this PR: 
* Multiple processes could be debugged by a remote connection. But I like your solution because a remote debugger would not be very practical with many workers.
* Have you measured the before and after startup times? Is it really such a big issue?
* How about a test case that spawns *a lot* of workers?
* With your solution you try to circumvent a bad design (the static issue). But that discussion is out of scope here :see_no_evil: 
* Why declare so many exception throwing test cases instead of handling them (what about one more catch(Exception ignored){} ?  ;-)  )

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
```diff
@@ -1380,80 +1386,116 @@ public static int getRandomAvailablePort() {
 			return availableSocket.getLocalPort();
 		}
 		catch(IOException e) {
-			// If no port was found just use 9999
+			// If no port was found just use 9990
 			return 9990;
 		}
```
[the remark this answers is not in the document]
Baunsgaard: i think i did this, because i experienced weird behavior in cases where i searched for a random available port recursively, and i did not want a potential infinite loop in our tests startup.
But i do agree that searching for a valid port would be better, than just returning this default.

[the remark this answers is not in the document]
corepointer: If you get an IOException upon creating a socket, there are probably more severe issues pending than finding a nice port number :fearful:

Baunsgaard:
Im closing this PR, because i have just added a faster startup to the federated tests, by adding optional startup time to the threads spawned in a federated tests. This renders this PR obsolite.

If this conclusion is wrong feel free to reopen it @kev-inn .

PR #1049, opened 2020-09-06
[MINOR] Python API hardening, and stability

Baunsgaard:
Minor changes in Python API startup for ease of startup if systemds is installed somewhere else it will use that systemds.
This practically means that if you have systemds home set, it will allow the python to use that systemds, while if it is not set, it will default back to the installed jar files from the PIP install.

This is a debated topic in  #992, where it is argued that it would make it harder for a user if the PIP does not contain the jar files.

but it is now here in a new PR that can do both.


Baunsgaard:
Closing PR to reduce number of tests run while debugging.

PR #985, opened 2020-07-09
Fine-Grained Privacy Constraints 2

sebwrede:
Adds fine-grained privacy to the privacy constraints. This PR builds on [Fine-Grained Privacy Constraints 1](https://github.com/apache/systemds/pull/982).

PR #1051, opened 2020-09-09
[SYSTEMDS-2605] Fine-Grained Privacy Constraints 2 Rebased

sebwrede:
This PR replaces the [Fine-Grained Privacy Constraints 2 PR](https://github.com/apache/systemds/pull/985). 
This PR adds fine-grained privacy constraints and adapts the privacy constraint handling to the new federated architecture.

### DIFF
--- a/src/main/java/org/apache/sysds/parser/DataExpression.java
+++ b/src/main/java/org/apache/sysds/parser/DataExpression.java
@@ -2254,42 +2254,59 @@ public class DataExpression extends DataIdentifier
 		Expression eprivacy = getVarParam(PRIVACY);
 		Expression eFineGrainedPrivacy = getVarParam(FINE_GRAINED_PRIVACY);
 		if ( eprivacy != null || eFineGrainedPrivacy != null ){
-
 			PrivacyConstraint privacyConstraint = new PrivacyConstraint();
 			if ( eprivacy != null ){
 				privacyConstraint.setPrivacyLevel(PrivacyLevel.valueOf(eprivacy.toString()));
 			}
 			if ( eFineGrainedPrivacy != null ){
-				FineGrainedPrivacy fineGrainedPrivacy = privacyConstraint.getFineGrainedPrivacy();
-				StringIdentifier fgPrivacyIdentifier = (StringIdentifier) eFineGrainedPrivacy;
-				String fgPrivacyValue = fgPrivacyIdentifier.getValue();
-				try {
-					JSONArtifact fgPrivacyJson = JSON.parse(fgPrivacyValue);
-					JSONObject fgPrivacyObject = (JSONObject)fgPrivacyJson;
-					JSONArray keys = fgPrivacyObject.names();
-					for ( int i = 0; i < keys.length(); i++ ){
-						String key = keys.getString(i);
-						JSONArray privateArray = fgPrivacyObject.getJSONArray(key);
-						for (Object range : privateArray.toArray()){
-							JSONArray beginDims = ((JSONArray)range).getJSONArray(0);
-							JSONArray endDims = ((JSONArray)range).getJSONArray(1);
-							long[] beginDimsLong = new long[beginDims.length()];
-							long[] endDimsLong = new long[endDims.length()];
-							for ( int dimIndex = 0; dimIndex < beginDims.length(); dimIndex++ ){
-								beginDimsLong[dimIndex] = beginDims.getLong(dimIndex);
-								endDimsLong[dimIndex] = endDims.getLong(dimIndex);
-							}
-							DataRange dataRange = new DataRange(beginDimsLong, endDimsLong);
-							fineGrainedPrivacy.put(dataRange, PrivacyLevel.valueOf(key));
-						}
-					}
-				} catch (JSONException exception){
-					throw new DMLException("JSONException: " + exception);
-				}
-				privacyConstraint.setFineGrainedPrivacyConstraints(fineGrainedPrivacy);
+				setFineGrainedPrivacy(privacyConstraint, eFineGrainedPrivacy);
 			}
 			getOutput().setPrivacy(privacyConstraint);
 		}
+	}
+
+	private void setFineGrainedPrivacy(PrivacyConstraint privacyConstraint, Expression eFineGrainedPrivacy){
+		FineGrainedPrivacy fineGrainedPrivacy = privacyConstraint.getFineGrainedPrivacy();
+		StringIdentifier fgPrivacyIdentifier = (StringIdentifier) eFineGrainedPrivacy;
+		String fgPrivacyValue = fgPrivacyIdentifier.getValue();
+		try {
+			putFineGrainedConstraintsFromString(fineGrainedPrivacy, fgPrivacyValue);
+		} catch (JSONException exception){
+			throw new DMLException("JSONException: " + exception);
+		}
+		privacyConstraint.setFineGrainedPrivacyConstraints(fineGrainedPrivacy);
+	}
+
+	private void putFineGrainedConstraintsFromString(FineGrainedPrivacy fineGrainedPrivacy, String fgPrivacyValue)
+		throws JSONException {
+		JSONArtifact fgPrivacyJson = JSON.parse(fgPrivacyValue);
+		JSONObject fgPrivacyObject = (JSONObject)fgPrivacyJson;
+		JSONArray keys = fgPrivacyObject.names();
+		for ( int i = 0; i < keys.length(); i++ ){
+			String key = keys.getString(i);
+			putFineGrainedConstraint(fgPrivacyObject, fineGrainedPrivacy, key);
+		}
+	}
+
+	private void putFineGrainedConstraint(JSONObject fgPrivacyObject, FineGrainedPrivacy fineGrainedPrivacy, String key)
+		throws JSONException {
+		JSONArray privateArray = fgPrivacyObject.getJSONArray(key);
+		for (Object range : privateArray.toArray()){
+			DataRange dataRange = getDataRangeFromObject(range);
+			fineGrainedPrivacy.put(dataRange, PrivacyLevel.valueOf(key));
+		}
+	}
+
+	private DataRange getDataRangeFromObject(Object range) throws JSONException {
+		JSONArray beginDims = ((JSONArray)range).getJSONArray(0);
+		JSONArray endDims = ((JSONArray)range).getJSONArray(1);
+		long[] beginDimsLong = new long[beginDims.length()];
+		long[] endDimsLong = new long[endDims.length()];
+		for ( int dimIndex = 0; dimIndex < beginDims.length(); dimIndex++ ){
+			beginDimsLong[dimIndex] = beginDims.getLong(dimIndex);
+			endDimsLong[dimIndex] = endDims.getLong(dimIndex);
+		}
+		return new DataRange(beginDimsLong, endDimsLong);
 	}
 	
 } // end class
--- a/src/main/java/org/apache/sysds/parser/DataExpression.java
+++ b/src/main/java/org/apache/sysds/parser/DataExpression.java
@@ -2273,8 +2273,12 @@ public class DataExpression extends DataIdentifier
 						for (Object range : privateArray.toArray()){
 							JSONArray beginDims = ((JSONArray)range).getJSONArray(0);
 							JSONArray endDims = ((JSONArray)range).getJSONArray(1);
-							long[] beginDimsLong = Arrays.stream(beginDims.toArray()).mapToLong((x)->Long.valueOf(((Integer)x).longValue())).toArray();
-							long[] endDimsLong = Arrays.stream(endDims.toArray()).mapToLong((x)->Long.valueOf(((Integer)x).longValue())).toArray();
+							long[] beginDimsLong = new long[beginDims.length()];
+							long[] endDimsLong = new long[endDims.length()];
+							for ( int dimIndex = 0; dimIndex < beginDims.length(); dimIndex++ ){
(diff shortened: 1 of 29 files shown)


src/main/java/org/apache/sysds/runtime/instructions/Instruction.java
```diff
@@ -38,7 +41,21 @@
 		FEDERATED
 	}
 	
-	private static final Log LOG = LogFactory.getLog(Instruction.class.getName());
+	protected static final Log LOG = LogFactory.getLog(Instruction.class.getName());
```
Baunsgaard, line 44: I made this private to ensure that the LOG is located in the same file it is called from.
If you want to make use of it somewhere else make a new Log object in that file.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -280,8 +281,9 @@ private FederatedResponse execInstruction(FederatedRequest request) {
 		ExecutionContext ec = _ecm.get(request.getTID());
 		BasicProgramBlock pb = new BasicProgramBlock(null);
 		pb.getInstructions().clear();
-		pb.getInstructions().add(InstructionParser
-			.parseSingleInstruction((String)request.getParam(0)));
+		Instruction receivedInstruction = InstructionParser
+				.parseSingleInstruction((String)request.getParam(0));
+		pb.getInstructions().add(receivedInstruction);
```
Baunsgaard, line 286: I don't see what this change does.

src/main/java/org/apache/sysds/runtime/instructions/cp/MultiReturnBuiltinCPInstruction.java
```diff
@@ -47,6 +47,10 @@ public CPOperand getOutput(int i) {
 		return _outputs.get(i);
 	}
 
+	public ArrayList<CPOperand> getOutputs(){
```
Baunsgaard, line 50: Maybe change these to List type return?

src/main/java/org/apache/sysds/runtime/instructions/spark/SPInstruction.java
```diff
@@ -41,16 +41,15 @@
...
 	protected final SPType _sptype;
-	protected final Operator _optr;
 	protected final boolean _requiresLabelUpdate;
 
 	protected SPInstruction(SPType type, String opcode, String istr) {
 		this(type, null, opcode, istr);
 	}
 
 	protected SPInstruction(SPType type, Operator op, String opcode, String istr) {
+		super(op);
```
Baunsgaard, line 51: can you explain to me this super class call?

src/main/java/org/apache/sysds/runtime/privacy/FineGrained/FineGrainedPrivacyList.java
```diff
@@ -0,0 +1,122 @@
...
+	private ArrayList<Map.Entry<DataRange, PrivacyLevel>> constraintCollection = new ArrayList<>();
+
+	@Override
+	public void put(DataRange dataRange, PrivacyLevel privacyLevel) {
+		constraintCollection.add(new AbstractMap.SimpleEntry<DataRange, PrivacyLevel>(dataRange, privacyLevel));
+	}
+
+	@Override
+	public Map<DataRange,PrivacyLevel> getPrivacyLevel(DataRange searchRange) {
+		Map<DataRange, PrivacyLevel> matches = new LinkedHashMap<>();
```
Baunsgaard, line 46: if possible can it not be a linkedHashMap, since the performance of that is ... slow, in many cases.

src/main/java/org/apache/sysds/runtime/privacy/FineGrained/FineGrainedPrivacyList.java
```diff
@@ -0,0 +1,122 @@
...
+		constraintCollection.clear();
+	}
+
+	@Override
+	public boolean hasConstraints() {
+		return !constraintCollection.isEmpty();
+	}
+
+	@Override
+	public Map<String, long[][][]> getAllConstraints() {
```
Baunsgaard, line 82: also this array of array of array, seems extravagant. Preferably it would be nice in some other format.

src/main/java/org/apache/sysds/runtime/privacy/PrivacyPropagator.java
```diff
@@ -19,15 +19,19 @@
 
 package org.apache.sysds.runtime.privacy;
 
+import java.util.*;
```
Baunsgaard, line 22: no wildcard imports.

src/main/java/org/apache/sysds/runtime/privacy/PrivacyPropagator.java
```diff
@@ -414,4 +591,25 @@ else if ( inst instanceof SqlCPInstruction )
...
+			return ((MultiReturnParameterizedBuiltinCPInstruction) inst).getOutputs();
+		else if ( inst instanceof MultiReturnBuiltinCPInstruction )
+			return ((MultiReturnBuiltinCPInstruction) inst).getOutputs();
+		else if ( inst instanceof ComputationCPInstruction )
+			return getSingletonList(((ComputationCPInstruction) inst).getOutput());
+		else if ( inst instanceof VariableCPInstruction )
+			return getSingletonList(((VariableCPInstruction) inst).getOutput());
+		else if ( inst instanceof SqlCPInstruction )
+			return getSingletonList(((SqlCPInstruction) inst).getOutput());
+		return new ArrayList<CPOperand>();
```
Baunsgaard, line 607: not 100% sure, but the if else is instance of is probably slow, and there should be some alternative.

src/test/java/org/apache/sysds/test/functions/privacy/BuiltinGLMTest.java
```diff
@@ -228,6 +229,9 @@ public void runtestGLM(PrivacyConstraint privacyConstraint, Class<?> expectedExc
 				{  100,   10,  2,  1.0,  2,  0.0,  3.0,   0.0,  2.0,  2.5 },   // Binomial two-column.logit
 				{  200,   10,  2,  1.0,  3,  0.0,  3.0,   0.0,  2.0,  2.5 },   // Binomial two-column.probit
 		};
-		return Arrays.asList(data);
+		if ( runAll )
+			return Arrays.asList(data);
+		else
+			return Arrays.asList(new Object[][]{data[0]});
```
Baunsgaard, line 235: is it to slow to run all or?

src/test/java/org/apache/sysds/test/functions/privacy/GLMTest.java
```diff
@@ -154,7 +155,10 @@ public GLMTest (int numRecords_, int numFeatures_, int distFamilyType_, double d
 			{  100,  10,  2, -1.0,  4,  0.0,  0.01, 3.0,  -2.0,  1.0,  1.0, GLMType.Bernoullicloglog1 }, // Bernoulli {-1, 1}.cloglog
 			{  200,  10,  2, -1.0,  5,  0.0,  0.01, 3.0,   0.0,  2.0,  1.0, GLMType.Bernoullicauchit },  // Bernoulli {-1, 1}.cauchit
 		};
-		return Arrays.asList(data);
+		if ( runAll )
+			return Arrays.asList(data);
+		else 
+			return Arrays.asList( new Object[][]{data[0]} );
```
Baunsgaard, line 161: again the filter?

Baunsgaard (commented):
Looks fine to me, but i only did a rough pass.
Some of the data types chose, seems excessive such as long[][][], but if you really need it it should be fine.

We will as discussed wait with merging it until after 2.0 release?

src/main/java/org/apache/sysds/runtime/instructions/Instruction.java
sebwrede: Private is also good. I do not need the "protected" to be there.

src/main/java/org/apache/sysds/parser/DataExpression.java
```diff
@@ -2239,9 +2251,44 @@ public boolean isRead()
 	 * Sets privacy of identifier if privacy variable parameter is set.  
 	 */
 	private void setPrivacy(){
-		Expression eprivacy = getVarParam("privacy");
-		if ( eprivacy != null ){
-			getOutput().setPrivacy(PrivacyLevel.valueOf(eprivacy.toString()));
+		Expression eprivacy = getVarParam(PRIVACY);
```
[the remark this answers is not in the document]
sebwrede: Yes, I also considered splitting it up, but the current implementation actually fits the general style of this class :stuck_out_tongue: .
I can change it now, if it is too confusing like this.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
sebwrede: It does not do anything differently. I thought at some point that I would need to handle the receivedInstruction, but then I decided against it. I could change it back like it was before and the code would work the same way, so right now it is just a question of readability: is the new version easier to read for you, @Baunsgaard ?

src/main/java/org/apache/sysds/runtime/instructions/cp/MultiReturnBuiltinCPInstruction.java
sebwrede: I also thought about doing that. It will always be an ArrayList, so in the end it would not make much of a difference. It is more an API-design thing. Do you prefer the List type?

src/main/java/org/apache/sysds/runtime/instructions/spark/SPInstruction.java
sebwrede: It calls the super class constructor, which sets the Operator field (which is in the super class).

src/main/java/org/apache/sysds/runtime/instructions/cp/MultiReturnBuiltinCPInstruction.java
Baunsgaard: well, its an opinion thing, i like to return the highest abstraction, i think we had a course this was mentioned as a point.
Software Engineering... :+1:

src/main/java/org/apache/sysds/runtime/privacy/FineGrained/FineGrainedPrivacyList.java
sebwrede: This depends on how it is used. In this case, the "containsValue" is the most used call on the returned map.

src/main/java/org/apache/sysds/parser/DataExpression.java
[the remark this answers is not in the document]
Baunsgaard: i only ask because when i was going through all the code, this stood out, and was ...  to complicated to review.

src/main/java/org/apache/sysds/runtime/privacy/PrivacyPropagator.java
sebwrede: OK

src/main/java/org/apache/sysds/runtime/privacy/PrivacyPropagator.java
sebwrede: The alternative is to have a method in Instruction called "getOutputs" and then implement it in the subclasses. This is however not the general design we use at the moment, but it would make much of the code cleaner to look at. For instance, most of the switch statements in this class could also be avoided if we changed the design and used Java inheritance instead.

src/test/java/org/apache/sysds/test/functions/privacy/GLMTest.java
sebwrede: Indeed :smiley:

src/main/java/org/apache/sysds/runtime/instructions/cp/MultiReturnBuiltinCPInstruction.java
sebwrede: Yes, I usually also go for that option.

src/main/java/org/apache/sysds/runtime/instructions/spark/SPInstruction.java
sebwrede: Because they all have the same field, but it was written in each of the classes. I placed it into their super class (Instruction) and then decided that calling the super constructor is a good idea, since we would then be able to change the way the operator is set for all subclasses by changing the Instruction constructor.

src/main/java/org/apache/sysds/runtime/instructions/cp/MultiReturnBuiltinCPInstruction.java
sebwrede: I have changed it now.

src/main/java/org/apache/sysds/parser/DataExpression.java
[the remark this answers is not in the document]
sebwrede: Is it easier to read in the current version, @Baunsgaard ?

mboehm7:
LGTM. Thanks for the patch @sebwrede and reconciling the tests with the new federated backend. I only made some minor changes: removed warnings, fixed few formatting issues, renamed new package to `finegrained`, and moved static methods for constraint parsing from `DataExpression` to `PrivacyUtils`.

mboehm7:
Also thanks to @Baunsgaard for the earlier review.

PR #1053, opened 2020-09-11
[MINOR] Move Privacy Handling of UDFs

sebwrede:
The privacy handling of UDFs is moved from the individual instruction classes to the FederatedWorkerHandler.
This will handle the privacy constraints of the input before any execution of the UDF and it will be easier to maintain since the call to the PrivacyMonitor does not have to be written in every "execute" implementation of FederatedUDF.


mboehm7:
LGTM - thanks @sebwrede

PR #1056, opened 2020-09-17
[MINOR] Python API hardening, and stability

Baunsgaard:
Minor changes in Python API startup for ease of startup if SystemDS is
installed somewhere else it will use that SystemDS.
This practically means that if you have SystemDS home set, it will allow
the python to use that SystemDS, while if it is not set, it will default
back to the installed jar files from the PIP install.

This is a debated topic in #992, where it is argued that it would make
it harder for a user if the PIP does not contain the jar files.

- Fix dual setup of systemds_context
- Added to usertest


Baunsgaard:
@mboehm7  what we talked about

PR #1057, opened 2020-09-18
Fix FederatedWorkerHandlerTest

sebwrede:


sebwrede:
The failing tests could be caused by the shared target space used by both federated and privacy tests. The tests never fail when I run it locally and they only fail sometimes when running it through GitHub Actions, so even if the privacy tests pass now, we cannot know for sure if we only run it once. I fixed FederatedWorkerHandlerTest to use its own target space and I believe this has fixed the problem.

Baunsgaard:
> The failing tests could be caused by the shared target space used by both federated and privacy tests. The tests never fail when I run it locally and they only fail sometimes when running it through GitHub Actions, so even if the privacy tests pass now, we cannot know for sure if we only run it once. I fixed FederatedWorkerHandlerTest to use its own target space and I believe this has fixed the problem.

sounds reasonable, Lets try it out, we will see if it becomes stable by following commits!

also the pr LGTM, will merge shortly

PR #1060, opened 2020-09-21
[SYSTEMDS-2668] Fine-Grained Privacy Constraint Propagation for Matrix Multiplications

sebwrede:
This PR improves propagation of fine-grained privacy constraints for matrix multiplications. The PR also provides a new structure for fine-grained privacy propagations by introducing a "Propagator" interface which are implemented by different propagator classes. This interface will be used in the following implementations of privacy propagation for other operators. 

The new matrix multiplication propagation is more efficient than the previous implementation since it makes an array with the summarized privacy level of the rows of the first matrix and the columns of the second matrix. 
Furthermore, it takes the operator type into account. This means that if a row or column contains only a single non-zero value, it cannot be considered an aggregation, hence the output in case of the PrivateAggregation privacy level in the input should still be PrivateAggregation. 
The rules of propagation is implemented in the method "PrivacyPropagator.corePropagation", where the comment also details the privacy "truth table".


src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
```diff
@@ -35,11 +35,61 @@
 
 	private ArrayList<Map.Entry<DataRange, PrivacyLevel>> constraintCollection = new ArrayList<>();
```
Baunsgaard, line 36: why is it an Array List of Map Entries. 

You can use Immutable Pairs, if you want to stick to Lists.
or maybe use a Hash Map instead of array list?

Have you measured how it performs, since i assume that what you want it to search through ranges and get the privacy associated with that specific data?

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
```diff
@@ -35,11 +35,61 @@
...
+			long[] beginRange = new long[] {0, i};
+			long[] endRange = new long[] {numRows, i};
+			privacyLevels[i] = getStrictestPrivacyLevel(new DataRange(beginRange, endRange));
+		}
+		return privacyLevels;
+	}
+
+	private PrivacyLevel getStrictestPrivacyLevel(DataRange searchRange){
+		PrivacyLevel strictestLevel = PrivacyLevel.None;
+		for ( Map.Entry<DataRange, PrivacyLevel> constraint : constraintCollection ) {
```
Baunsgaard, line 62: because you have a for loop in the previous and loop through each entry here. it is going to be slow.
I think a future project would be an index structure for the privacy. maybe a tree structure would be good?
one for cols and one for rows?

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyMap.java
```diff
@@ -40,6 +40,18 @@ public void put(DataRange dataRange, PrivacyLevel privacyLevel) {
 		constraintCollection.put(dataRange, privacyLevel);
 	}
 
+	@Override public void putRow(int rowIndex, int rowLength, PrivacyLevel privacyLevel) {
```
Baunsgaard, line 43: misformatted, and does nothing inside the method.

src/main/java/org/apache/sysds/runtime/privacy/propagation/MatrixMultiplicationPropagatorNaive.java
```diff
@@ -0,0 +1,60 @@
...
+	public MatrixMultiplicationPropagatorNaive(){
+		super();
+	}
+
+	public MatrixMultiplicationPropagatorNaive(MatrixBlock input1, PrivacyConstraint privacyConstraint1,
+		MatrixBlock input2, PrivacyConstraint privacyConstraint2) {
+		super(input1, privacyConstraint1, input2, privacyConstraint2);
+	}
+
+	@Override
```
Baunsgaard, line 43: put annotations after jdoc

src/main/java/org/apache/sysds/runtime/privacy/propagation/OperatorType.java
```diff
@@ -0,0 +1,25 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sysds.runtime.privacy.propagation;
+
+public enum OperatorType {
```
Baunsgaard, line 22: this could be considered to be moved into common/Types.java
or leverage already existing Operator Types?

src/main/java/org/apache/sysds/runtime/privacy/propagation/PrivacyPropagator.java
```diff
@@ -64,7 +64,61 @@ public static Data parseAndSetPrivacyConstraint(Data cd, JSONObject mtd)
...
+	 * privAgg-priv 	| priv 			| priv
+	 * none-priv 		| priv 			| priv
+	 * privAgg-privAgg 	| privAgg 		| none
+	 * none-none 		| none 			| none
+	 * privAgg-none 	| privAgg 		| none
+	 * none-privAgg 	| privAgg 		| none
+	 *
+	 * @param inputLevels privacy levels of the input
+	 * @param operatorType type of the operator which is either an aggregation (Aggregate) or not an aggregation (NonAggregate)
+	 * @return output privacy level
```
Baunsgaard, line 98: :+1:  nice jdoc

src/main/java/org/apache/sysds/runtime/privacy/propagation/PrivacyPropagator.java
```diff
@@ -64,7 +64,61 @@ public static Data parseAndSetPrivacyConstraint(Data cd, JSONObject mtd)
...
+	 * Unary input:
+	 * Input   | NonAggregate | Aggregate
+	 * -----------------------------------
+	 * priv    | priv         | priv
+	 * privAgg | privAgg      | none
+	 * none    | none         | none
+	 *
+	 * Binary input:
+	 * Input   			| NonAggregate 	| Aggregate
+	 * -----------------------------------
```
Baunsgaard, line 85: minor alignment

Baunsgaard (commented):
LGTM, again i only looked through without full understanding.

I have some potential performance concerns, but i hope and assume that the lists are so small that they are not noticeable. But if the constraints become complicated better data structures are needed.

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
sebwrede: This is not really a part of the PR. I will look at other ways to implement the FineGrainedPrivacy interface when it becomes relevant. 
There are several classes that represent pairs that could be relevant to use. I picked Map.Entry since it was a part of java.util and supported the operations I needed.

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
sebwrede: I actually started implementing a tree structure when I added fine-grained privacy, but we decided that it was not relevant for now and that we could do it later when we have the propagation of fine-grained constraints ready.

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyMap.java
sebwrede: I know. I am not using that class and will perhaps delete it, but I wanted to get the propagation done and then see how that performs with the other implementation. Then I can try different implementations of FineGrainedPrivacy.

src/main/java/org/apache/sysds/runtime/privacy/propagation/MatrixMultiplicationPropagatorNaive.java
sebwrede: OK

src/main/java/org/apache/sysds/runtime/privacy/propagation/OperatorType.java
sebwrede: Right now, it is only used for privacy propagation, hence it is not a common type.

sebwrede: The existing operator types are not relevant, since I only deal with two types: aggregate and non-aggregate.

src/main/java/org/apache/sysds/runtime/privacy/propagation/PrivacyPropagator.java
sebwrede: I am not sure, what you believe should be aligned, but now I added additional "-------".

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
Baunsgaard: sounds reasonable, but maybe add a JIRA task to remember it later

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyList.java
Baunsgaard: same jira task as mentioned above

src/main/java/org/apache/sysds/runtime/privacy/finegrained/FineGrainedPrivacyMap.java
Baunsgaard: if it is not used, delete it :+1:

src/main/java/org/apache/sysds/runtime/privacy/propagation/OperatorType.java
Baunsgaard: okay, i do the same in compression. It was just if they matched up completely, then they could be reused, and furthermore used as a TODO list, for what is not supported.

PR #1062, opened 2020-09-22
[SYSTEMDS-2546,2547] Fix federated rbind/cbind with support for local data

kev-inn:
Adds `FederatedLocalData` so that we can use local data without the necessity to send it to a worker. This allows reusing a lot of code, but might lead to overhead. Other options to handle this scenario exist, please let me know what you think about this specific implementation.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
@@ -145,11 +145,17 @@ public class InstructionUtils
 		int numParts = parts.length;
 		int numFields = numParts - 1; //account for opcode
 		
-		if (Arrays.stream(expected).noneMatch((i) -> numFields == i))
-			throw new DMLRuntimeException(
-				"checkNumFields() -- expected number ("
-					+ Arrays.stream(expected).mapToObj(Integer::toString).reduce((a, b) -> a + ", " + b).orElse("")
-					+ ") != is not equal to actual number (" + numFields + ").");
+		if (Arrays.stream(expected).noneMatch((i) -> numFields == i)) {
+			StringBuilder sb = new StringBuilder();
+			sb.append("checkNumFields() -- expected number (");
+			for (int i = 0; i < expected.length; i++) {
+				sb.append(expected[i]);
+				if (i != expected.length - 1)
+					sb.append(", ");
+			}
+			sb.append(") != is not equal to actual number (").append(numFields).append(").");
+			throw new DMLRuntimeException(sb.toString());
+		}
 		
 		return numFields;
 	}
--- a/src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
@@ -19,6 +19,7 @@
 
 package org.apache.sysds.runtime.instructions;
 
+import java.util.Arrays;
 import java.util.StringTokenizer;
 
 import org.apache.sysds.common.Types;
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
```diff
@@ -140,6 +141,19 @@ public static int checkNumFields( String[] parts, int expected1, int expected2 )
...
 
+	public static int checkNumFields( String[] parts, int... expected ) {
+		int numParts = parts.length;
+		int numFields = numParts - 1; //account for opcode
+		
+		if (Arrays.stream(expected).noneMatch((i) -> numFields == i))
+			throw new DMLRuntimeException(
+				"checkNumFields() -- expected number ("
+					+ Arrays.stream(expected).mapToObj(Integer::toString).reduce((a, b) -> a + ", " + b).orElse("")
+					+ ") != is not equal to actual number (" + numFields + ").");
```
Baunsgaard, line 152: this looks overkill, maybe use a string builder

src/main/java/org/apache/sysds/runtime/instructions/fed/AppendFEDInstruction.java
```diff
@@ -33,71 +32,74 @@
...
-			throw new DMLRuntimeException(
-				"Append-rbind is not possible for federated input matrices " + input1.getName() + " and "
-				+ input2.getName() + " with different number of columns: " + mo1.getNumColumns()
-				+ " vs " + mo2.getNumColumns());
+
+		// check input dimensions
+		if(_cbind && mo1.getNumRows() != mo2.getNumRows()) {
+			throw new DMLRuntimeException("Append-cbind is not possible for federated input matrices "
+				+ input1.getName() + " and " + input2.getName() + " with different number of rows: " + mo1.getNumRows()
+				+ " vs " + mo2.getNumRows());
```
Baunsgaard, line 69: again, maybe string builder.

Baunsgaard (commented):
Hi @kev-inn,

The code looks fine :+1: , the only suggestion there is that you construct some large strings sometimes for the exceptions. This might be cleaner to look at using a StringBuilder.
 
With regards to the behavior. Then i would appreciate a test, and example that clearly shows how to make a federated matrix with part federated part local. This is what i thought the PR would contain, but i seem unable to find it.
 
This behaviour would be awesome to be able to use!

kev-inn:
@Baunsgaard I improved the `FederatedRCBindTest.java`, which actually didn't properly execute the federated instructions, that contains a test case for federated-federated, federated-local and local-federated bindings (both rbind and cbind). I will add a comment that clarifies that.

kev-inn:
@sebwrede could you take a look at the new r/cbind tests? I believe you used it to check transfer privacy constraints, but the fixed rbind and cbind should not transfer (except when writing result to disk), so maybe a simple write test would suffice.

sebwrede:
> @sebwrede could you take a look at the new r/cbind tests? I believe you used it to check transfer privacy constraints, but the fixed rbind and cbind should not transfer (except when writing result to disk), so maybe a simple write test would suffice.
> Edit: `FederatedWorkerHandlerTest` fails due to changes

Yes, the tests are failing right now because the arguments given are not correct anymore. As you say, they do not even transfer anything anymore, so they are not really relevant to put in a FederatedWorkerHandler test. The test was more relevant when we had the different instructions handled separately on the worker side. Furthermore, I will soon make a PR that changes the way privacy constraints are propagated for RBind and CBind. 
With this in mind, I think we could actually remove or at least ignore the privacy tests that are using CBind and RBind in FederatedWorkerHandlerTest.java.

Baunsgaard:
used closing instead of closes... in the merge...
but this is merged to master therefore i'm closing it.

PR #1063, opened 2020-09-23
[SYSTEMDS-2663, 2664] Prepare native instructions for release 2.0

corepointer:
* pulled in a few CMake tricks from gpu codegen dev branch to unify lib naming
* verify MKL and OpenBLAS working
* cleanup native lib loader class (NativeHelper)
* build script for Windows


src/main/java/org/apache/sysds/utils/NativeHelper.java
```diff
@@ -146,7 +146,20 @@ private static boolean isSupportedArchitecture() {
...
 	}
-	
+
+	/**
+	 * Note: we only support Windows and Linux at the moment.
+	 *
+	 * @return true if operating system is supported
+	 */
+	private static boolean isSupportedOS() {
+		if(SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_WINDOWS) {
```
Baunsgaard, line 156: could it not also include UNIX?
for mac support?
If not it would make sense to add a task on JIRA fro this

Baunsgaard (commented):
LGTM

src/main/java/org/apache/sysds/utils/NativeHelper.java
corepointer: We don't have Mac support anymore. We can put it on the wish list in JIRA :wink:

PR #1064, opened 2020-09-24
[WIP][SYSTEMDS 2583,2586] Reuse of compressed (dedup) DAGs

phaniarnab:
This patch extends equals and hash of lineage items to support
dedup items, which in turn allow reuse of dedup blocks.
This implementation makes deduplication transparent to the reuse
logic. And this change doesn't add any extra overhead and keep
the constant time hash calculation of lineage items intact.


phaniarnab:
Currently running experiments to confirm that 1) these changes don't add any compute or memory overhead, and 2)  reuse always works with deduplication. Till now results look promising.
In any case, this PR will be merged after 2.0 release.

PR #1066, opened 2020-09-25
[SYSTEMDS-2678] Fine-Grained Propagation of RBind and CBind

sebwrede:
This PR adds fine-grained propagation of RBind, CBind, ListAppend, and ListRemove.

PR #1074, opened 2020-09-30
Postprocess Privacy Constraints Before Lineage Cache

sebwrede:
This PR will make sure that privacy constraints are propagated to output variables before the result of the instruction is cached. This is under the assumption that the caching does not only use the CPOperand outputs, but also uses the variables stored in the execution context.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/ProgramBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/ProgramBlock.java
@@ -247,9 +247,6 @@ public abstract class ProgramBlock implements ParseInfo
 				
 				// process actual instruction
 				tmp.processInstruction(ec);
-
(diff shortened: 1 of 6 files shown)


phaniarnab:
What will happen if the result is reused from lineage cache? 
I suggest a small test to check. You can enable lineage reuse by passing `-lineage reuse_full` option.

sebwrede:
I have now added a test and put the postprocessing of privacy constraints outside the lineage cache if-block in ProgramBlock. In this way, the privacy constraints will be propagated even if the result has been read from the cache. 
I needed to add a few lines which is also a part of [another pending PR](https://github.com/apache/systemds/pull/1066) to propagate the fine-grained constraints and to release matrix input.

phaniarnab:
Thanks. Looks good. 🦾 
LGTM (if all the tests pass).

PR #1078, opened 2020-10-10
[MINOR] MKL native MM setNonZeros to number of cells

Baunsgaard:
This PR is sort of an question.

Is it okay for our MKL to assume that the output matrix is fully dense with all values !=0? 

If so then we have an potential improved performance.

I have observed up to 30% in some cases (most likely due to inter run variance) but around 10% on average, and it is because the native call to matrix multiplication has to transfer the matrix out and and then into java again and then afterwards it loops through all values to count 0's forcing another iteration through the matrix once more.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixNative.java
+++ b/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixNative.java
@@ -106,7 +106,7 @@ public class LibMatrixNative
 					Statistics.nativeLibMatrixMultTime += System.nanoTime() - start;
 					Statistics.numNativeLibMatrixMultCalls.increment();
 				}
-				ret.recomputeNonZeros();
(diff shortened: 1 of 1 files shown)


mboehm7:
The current runtime/compiler requires the non-zeros to be exact or unknown (-1) - so we cannot simply set it to fully dense. This information could be mistakenly used (e.g., when rewriting `sum(X!=0)` to `nnz(X)`)

Baunsgaard:
so would it make sense to set it to unknown?

I have the same dilemma in compression, since the counting of non zeros also is a factor there.

mboehm7:
well, for now I would recommend to properly maintain the non-zeros. It can only become a moderate overhead if the output is large compared to the compute (cubic compute and linear write vs linear counting). What usually helps is to do the nnz maintenance at the end of each multi-threaded computation task, which gives locality.

Baunsgaard:
Closing based on comments

PR #1088, opened 2020-10-26
[SYSTEMDS-2705] Federated Writer

Baunsgaard:
This commit add the ability to save a federated matrix, But only one that
has UID 0. If the UID is 0 on the remote then we know that it is a matrix,
that have just been read from a file. Therefore we can without modifying
the remote workers save our json directly, and read normally this file.

Example:

X = federated(addresses=...,ranges=...)
write(X, "X_fed.json", format="federated")

After this you can read the federated matrix, and it will allocate as
federated using the workers.

read("X_fed.json")

### DIFF
--- a/.github/workflows/functionsTests.yml
+++ b/.github/workflows/functionsTests.yml
@@ -58,6 +58,7 @@ jobs:
           dnn,
           federated.algorithms,
           federated.primitives,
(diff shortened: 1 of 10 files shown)


Baunsgaard:
merged

PR #1089, opened 2020-10-28
[MINOR] Add seed to kmeans

Baunsgaard:


scripts/builtin/kmeans.dml
```diff
@@ -86,7 +87,11 @@ m_kmeans = function(Matrix[Double] X, Integer k = 10, Integer runs = 10, Integer
 
     # Select the i-th centroid in each sample as a random sample row id with
     # probability ~ min_distances:
-    random_row = rand(rows = 1, cols = num_runs);
+    if (seed == -1){
+      random_row = rand(rows = 1, cols = num_runs);
+    } else{
+      random_row = rand(rows = 1, cols = num_rows, seed = seed);
+    }
```
phaniarnab, line 94: Only one rand call should be sufficient. If not, I suggest to replace this with a single `ifelse` call.

Baunsgaard: but i want to have default random behaviour, with not setting a seed value, and then otherwise set the seed.
Is there a way of doing that in one rand call?

PR #1092, opened 2020-10-30
[SYSTEMDS-2692, 2698] Spoof cuda initial

corepointer:
* refactored version of a previous PR.
* produces CUDA code and compiles at runtime
* supports codegen for cellwise dense operations
* split into thematic commits (so **I recommend reviewing by commit** and not by file)

* This PR adresses two sub-tasks in my [codegen epic](https://issues.apache.org/jira/browse/SYSTEMDS-2691)
* If someone merges it in, please squash to two commits (one per sub-task). Will merge it myself after three business days if no discussion is pending. 

* If your CUDA installation is not in the Linux default directory or you're running Windows, set the CUDA_PATH environment variable to point to your CUDA SDK root directory.

#### Invocation example (from the root of your cloned source tree):
``` shell
CONFIG_FILE=src/test/scripts/functions/codegen/SystemDS-config-codegen.xml   bin/systemds src/test/scripts/functions/codegen/cellwisetmpl1.dml -args output-matrix -stats -gpu

```


j143:
Thank you for, providing the self documented comments.

PR #1093, opened 2020-11-05
[SYSTEMDS-2721] Add Federated SSL

Baunsgaard:
Using a self signed SSL certificate, the communication using Netty becomes encrypted.
The initialization of the communication is not safe since anyone could construct a certificate on the server side (workers) and respond to the requests from the controller. But for an initial prototype it is fine.

To allow all communication to be valid anyway, the certificate is self signed, and trusted at the beginning. It is fine since we know that it is generated specifically for the communication we want from the federated site. And allows for encrypted answers to the controller.


Baunsgaard:
image

PR #1096, opened 2020-11-07
[SYSTEMDS-2719] Lineage exploitation in buffer pool

sergred:
This task introduces a change to object storage mechanisms. For matrices that are results of data generation instructions such as RAND, we store their lineage alongside the data. Under memory pressure, these matrices do not go to the buffer pool. We recompute them from lineage instead of reading evicted buffers from FS. This potentially results in fewer evictions and IO.


mboehm7:
LGTM - thanks @sergred for this great feature. During the merge I just made a few tweaks: added missing license header, fixed wrong formatting (tabs over spaces), changed the way we present the new statistics, fixed the recomputation by lineage, removed an unnecessary `ExecutionContext.setVariable` (which existed before), and reverted the changes in ByteBuffer as its unrelated to the given feature (the code is still in the PR and we come back once we refactor that).

PR #1097, opened 2020-11-10
[SYSTEMDS-2722] Split Testing

Baunsgaard:
Added tests (builtin and federated) for new split builtin.

I don't understand that it is working with right indexing on the federated outputs but it seems to.

### DIFF
--- a/scripts/builtin/split.dml
+++ b/scripts/builtin/split.dml
@@ -19,14 +19,15 @@
 #
 #-------------------------------------------------------------
 
-# Split input data X and y into contiguous or samples train/test sets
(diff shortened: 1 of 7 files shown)


Baunsgaard:
> Added tests (builtin and federated) for new split builtin.
> 
> I don't understand that it is working with right indexing on the federated outputs but it seems to.

oh nvm it is because it pulls the data to master

mboehm7:
Yes right, but we have @OlgaOvcharenko 's PR #1098 (federated right indexing) which would then also directly apply to this split contiguous.

PR #1102, opened 2020-11-13
[SYSTEMDS-2727-9] Federated CM, qsort & qpick

Baunsgaard:
Thanks to Olga in setting these up in  #1091.


scripts/builtin/dist.dml
```diff
@@ -23,7 +23,8 @@
 m_dist = function(Matrix[Double] X) return (Matrix[Double] Y) {
     G = X %*% t(X);
     I = matrix(1, rows = nrow(G), cols = ncol(G));
-    Y = -2 * (G) + t(I %*% diag(diag(G))) + t(diag(diag(G)) %*% I);
+    Y = -2 * (G) + (diag(G) * I) + (I * t(diag(G)));
+#   Y = -2 * (G) + t(I %*% diag(diag(G))) + t(diag(diag(G)) %*% I);
```
Baunsgaard, line 27: why change this @OlgaOvcharenko 
I think it does the same but did not verify.

Baunsgaard:
Hi, @OlgaOvcharenko

I have taken your code and merged it with master, 
there are some things we need to clean up in federationUtils (partially because i reintroduced some of our working aggregate functions), but otherwise the code looks fine.
Tomorrow either you or me have to finish this up to be ready to merge with master.

scripts/builtin/dist.dml
OlgaOvcharenko: Right, the same, just thought it would be simpler

PR #1106, opened 2020-11-15
[MINOR] Federated Type Fixes

Baunsgaard:
I hope these changes don't overlap to much with yours @mboehm7

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
@@ -372,7 +372,7 @@ public abstract class CacheableData<T extends CacheBlock> extends Data
 	}
 	
 	public boolean isFederated(FType type) {
-		return isFederated() && _fedMapping.getType() == type;
+		return isFederated() && _fedMapping.getType().isType(type);
 	}
 	
 	/**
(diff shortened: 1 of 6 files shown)


mboehm7:
no, seems fine and I'm also done for now. In order to run the test pipeline fully federated, it would be good though to extend the removeEmpty support for selection vectors, which makes the splitting vastly faster and preserves a nice balance across federated partitions.

mboehm7:
A general question on this: isn't a full block just a row or column partitioned federated matrix and hence the existing operations can directly be applied. Yes, we could implement them more efficiently (basic forwarding, no aggregation, etc) but as this is a special case I'm not sure it's worth the implementation complexity. Any thoughts?

Baunsgaard:
> A general question on this: isn't a full block just a row or column partitioned federated matrix and hence the existing operations can directly be applied. Yes, we could implement them more efficiently (basic forwarding, no aggregation, etc) but as this is a special case I'm not sure it's worth the implementation complexity. Any thoughts?

yes, but the basic issue is that when you only have one federated location, should i initialize as a row or a col?
Currently this dilemma results in scenarios where the data is aggregated from the federated sites when we have one worker.

I don't intend to optimize particularly for it, but with this addition we are able to leverage both our row and col optimized operations in this special case with not to much extra code.

mboehm7:
ok, increasing the generality of implementations is a good idea. In that case I would recommend to not probe for type FULL in individual operations, but put methods isRowPartitioned (ROW or FULL), isColumnPartitions (COL or FULL) into the enum type and then probe these methods from the instructions.

Baunsgaard:
> ok, increasing the generality of implementations is a good idea. In that case I would recommend to not probe for type FULL in individual operations, but put methods isRowPartitioned (ROW or FULL), isColumnPartitions (COL or FULL) into the enum type and then probe these methods from the instructions.

Ah good idea, I forgot java can do this with Enums

PR #1108, opened 2020-11-17
[SYSTEMDS-2735] GMM Predict for clustering new instances

Shafaq-Siddiqi:
Predict method for Gaussian Mixture Model will predict the cluster labels for new instances.


mboehm7:
LGTM - thanks for the patch @Shafaq-Siddiqi. Generally this looks very good. I just renamed the new builtin to gmmPredict for consistency, changed the formatting a bit, added the missing reset of rewrite flags in the test, and disabled the spark tests (as forced spark mode at algorithm level is a bit too much here).

PR #1110, opened 2020-11-18
[MINOR] RmVar disable for federated read data.

Baunsgaard:


Baunsgaard:
```code
Before
  2  tak+*         16.672   1200
After
  2  tak+*          6.257    600
```

mboehm7:
adding ternary aggregates is fine, but has nothing to do with the title of disabling rmvar right? Regarding the deleted wait on cleanup, please be aware that this might require federated worker changes as now they have to deal with concurrent RPC on a single execution context (not just concurrent RPC on different execution contexts that are managed in a concurrent hashmap). So it has a chance of brittle failures when these RPCs conflict on symbol table updates.

Baunsgaard:
Yes, I have realized that, i have simply gone back to having a background thread doing the cleanup.
Olga added a line at some point, forcing this to be run in the same thread, this made the rm-var a blocking operation on federated variables, unlike before.

Baunsgaard:
> adding ternary aggregates is fine, but has nothing to do with the title of disabling rmvar right? Regarding the deleted wait on cleanup, please be aware that this might require federated worker changes as now they have to deal with concurrent RPC on a single execution context (not just concurrent RPC on different execution contexts that are managed in a concurrent hashmap). So it has a chance of brittle failures when these RPCs conflict on symbol table updates.

no ternary has nothing to do with the pr.. but it was one of the operation s making L2SVM slow, so i just added it.
Unfortunally it is only relevant if y is federated as well. But in that case, the wide network becomes very slow.

PR #1111, opened 2020-11-20
Lm pipeline test with 4 workers

OlgaOvcharenko:


src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedLmPipeline.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.sysds.test.functions.federated.algorithms;
 
+import com.sun.tools.javac.util.List;
```
Baunsgaard, line 22: I think you have some automated imports that have to change priority.
Usually we don't import the com.sun libraries.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedLmPipeline.java
```diff
@@ -44,20 +46,27 @@
...
 	
 	@Test
 	public void federatedLmPipelineSampled() {
-		federatedLmPipeline(Types.ExecMode.SINGLE_NODE, false);
+		federatedLmPipeline(Types.ExecMode.SINGLE_NODE, false, TEST_NAME1);
 	}
 
-	public void federatedLmPipeline(ExecMode execMode, boolean contSplits) {
+	@Test
+	public void federatedLmPipelineSampled4Workers() { federatedLmPipeline(Types.ExecMode.SINGLE_NODE, false, TEST_NAME2); }
```
Baunsgaard, line 67: i personally like it in one line as well, but our other tests, does not follow this syntax so i would suggest multi line.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedLmPipeline.java
```diff
@@ -76,37 +85,48 @@ public void federatedLmPipeline(ExecMode execMode, boolean contSplits) {
...
 			
 			// empty script name because we don't execute any script, just start the worker
 			fullDMLScriptName = "";
 			int port1 = getRandomAvailablePort();
 			int port2 = getRandomAvailablePort();
+			int port3 = getRandomAvailablePort();
+			int port4 = getRandomAvailablePort();
 			Thread t1 = startLocalFedWorkerThread(port1, FED_WORKER_WAIT_S);
 			Thread t2 = startLocalFedWorkerThread(port2);
+			Thread t3 = startLocalFedWorkerThread(port3);
```
Baunsgaard, line 105: I have added the second argument to reduce the Federated worker startup, so we don't have to wait for a second for each of the workers, but just the last.

Baunsgaard (commented):
Hi Olga, 

thanks for the test :1st_place_medal: , The comments i have added is for next time, i will change it while merging.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedLmPipeline.java
Baunsgaard: import java.util.List;

Baunsgaard: and to create an array in oneline use:

```java
int[] x = new int[] {1, 20, 3, 40, 0}
```

PR #1112, opened 2020-11-20
[SYSTEMDS-2738] Federated rdiag, rev and uppertri instruction

OlgaOvcharenko:
This PR adds federated rdiag M2V and V2M, rev and uppertri, lowertri instructions.


src/main/java/org/apache/sysds/runtime/instructions/fed/ReorgFEDInstruction.java
```diff
@@ -50,20 +77,196 @@ public static ReorgFEDInstruction parseInstruction ( String str ) {
...
+				new long[] {mo1.getFedMapping().getID()});
+			mo1.getFedMapping().execute(getTID(), true, fr1);
+
+			//drive output federated mapping
+			MatrixObject out = ec.getMatrixObject(output);
+			out.getDataCharacteristics().set(mo1.getNumColumns(), mo1.getNumRows(), (int) mo1.getBlocksize(), mo1.getNnz());
+			out.setFedMapping(mo1.getFedMapping().copyWithNewID(fr1.getID()).transpose());
+		}
+		else if (instOpcode.equals("rdiag")) {
+			AbstractMap.SimpleEntry<FederationMap, Map<FederatedRange, int[]>> result;
```
Baunsgaard, line 100: I have never seen the AbstractMap.SimpleEntry being used.
Usually i see Map.Entry<>.
Is there a difference?

Also since the usage of this entry, is to enable  multi returns on functions calls, i would suggest making a small class inside this file, that you use to setup and return the values.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedLmPipeline.java
```diff
@@ -63,7 +63,6 @@ public void federatedLmPipelineSampled() {
 		federatedLmPipeline(Types.ExecMode.SINGLE_NODE, false, TEST_NAME1);
 	}
 
-	@Test
```
Baunsgaard, line 66: should this be added again?

if it is failing and you don't intend to fix it add an "@ Ignore" instead

src/test/java/org/apache/sysds/test/functions/federated/primitives/FederatedRdiagTest.java
```diff
@@ -0,0 +1,145 @@
...
+	public int rows;
+
+	@Parameterized.Parameter(1)
+	public int cols;
+
+	@Parameterized.Parameters
+	public static Collection<Object[]> data() {
+		return Arrays.asList(new Object[][] {
+//			{12, 12},
+			{12,1}
```
Baunsgaard, line 57: this seems very small, maybe make them alittle bigger, and if you want to test multiple things then i would suggest :
rowvector, colvector, and matrix.

src/test/java/org/apache/sysds/test/functions/federated/primitives/FederatedRdiagTest.java
```diff
@@ -0,0 +1,145 @@
...
+			"in_X4=" + TestUtils.federatedAddress(port4, input("X4")),
+			"rows=" + rows,
+			"cols=" + cols,
+			"out_S=" + output("S")};
+		runTest(null);
+
+		// compare all sums via files
+		compareResults(0.01);
+
+//		Assert.assertTrue(heavyHittersContainsString("fed_rdiag"));
```
Baunsgaard, line 133: should this not be reenabled

src/test/java/org/apache/sysds/test/functions/federated/primitives/FederatedRdiagTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+		// check that federated input files are still existing
+		Assert.assertTrue(HDFSTool.existsFileOnHDFS(input("X1")));
+		Assert.assertTrue(HDFSTool.existsFileOnHDFS(input("X2")));
```
Baunsgaard, line 137: These checks were added in some tests to check if the federated workers would accidentally delete the files, I'm not sure if it is necessary in all tests. (Lets keep it for now, but just for your information.)

src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
```diff
@@ -156,11 +156,11 @@ private void rmempty(ExecutionContext ec) {
 		MatrixObject out = ec.getMatrixObject(output);
 
 		boolean marginRow = params.get("margin").equals("rows");
-		boolean k = ((marginRow && mo.getFedMapping().getType().isColPartitioned()) ||
+		boolean isNotAligned = ((marginRow && mo.getFedMapping().getType().isColPartitioned()) ||
```
Baunsgaard, line 159: thanks, Much clearer variable name :+1:

Baunsgaard (commented):
Some verification is disabled in the tests, but otherwise LGTM :+1:

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
```diff
@@ -384,6 +384,21 @@ private static void setThreadID(long tid, FederatedRequest[]... frsets) {
 				Arrays.stream(frset).forEach(fr -> fr.setTID(tid));
 	}
 
+	public void reverseFedMap() {
```
Baunsgaard, line 387: This assumes that all federated maps are always sorted in order. I think that is not a constraint that we have.
Furthermore this would break if we have a federated mapping that is the custom variant, with ranges all over the place.

I think maybe that you actually don't want to modify the mapping at all like this, or make sure that you sort the ranges before attempting this.

src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
```diff
@@ -145,22 +149,115 @@ else if(opcode.equalsIgnoreCase("transformapply"))
 			throw new DMLRuntimeException("Unknown opcode : " + opcode);
 		}
 	}
+	private void triangle(ExecutionContext ec, String opcode) {
```
Baunsgaard, line 152: add space between methods.

src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
```diff
@@ -145,22 +149,115 @@ else if(opcode.equalsIgnoreCase("transformapply"))
...
+				mb.slice(0, mb.getNumRows()-1, _slice[0], _slice[1]-1, new MatrixBlock()) :
+				mb.slice(_slice[0], _slice[1]-1);
+
+			//triangle
+			MatrixBlock tri = soresBlock.extractTriangular(new MatrixBlock(), _lower, _diag, _values);
+			if(_rowFed) {
+				ret = new MatrixBlock(mb.getNumRows(), mb.getNumColumns(), 0.0);
+				ret.copy(0, ret.getNumRows()-1, _slice[0], _slice[1]-1, tri, false);
+				if(_slice[1] <= mb.getNumColumns()-1 && !_lower) {
+					addBlock = mb.slice(0, mb.getNumRows()-1, _slice[1], mb.getNumColumns()-1, new MatrixBlock());
```
Baunsgaard, line 222: the slice function is nice to use, but ultimately inefficient for the job you are doing here.
because the slice hash a large unnecessary overhead.
I would suggest extracting the sparse or dense block and processing that from the mb variable.

I would also suggest not to include this optimization right now in this PR, but add a todo for later.

Baunsgaard (commented):
Hi @OlgaOvcharenko, fine pr :+1: 

I will merge this shortly, but there is some comments that we need to address, maybe you want to add a sort federated ranges method in another pr?
For now lets assume that the federated ranges are sorted, since i haven't seen otherwise in any of our tests or implementations.

PR #1113, opened 2020-11-22
[SYSTEMDS-2550] Shuffle data partitioner

tobiasrieger:
This adds the shuffle data partitioner the the federated parameter server and a helpful way of running federated unit tests. See federatedBalancedAndWriteWithMTD in the AutomatedTestBase. 

The federated parameter server unit test was optimised and the scripts TwoNN.dml and CNN.dml updated.

Also .getFedMapping() of FederationMap was refactored to getFRangeFDataMap to avoid confusion with the method of MatrixObject with the same name.

**Current issues/ questions?**

- The hybrid federated parameter server unit tests fail stating "Shuffle" scheme is not supported. This is because the read somehow pulls in the matrix and presents it as a dense block. This means the parameter server is running in regular mode thus failing on the shuffle scheme.

- The federated parameter server unit test input directory is currently not cleaned up. Should this be done after every run?

- How to generate the ID of a newly created FederationMap correctly? Currently set to 1


sebwrede:
**Generally, I think this PR looks good.** 
In AutomatedTestBase:640, you create a new PrivacyConstraint. This means that it writes the privacy level to the metadata file. Is this necessary for your tests? You could just pass null to the method call and then it would not write any privacy constraints to the metadata file. 

**Comments on you questions/issues:**

> The hybrid federated parameter server unit tests fail stating "Shuffle" scheme is not supported. This is because the read somehow pulls in the matrix and presents it as a dense block. This means the parameter server is running in regular mode thus failing on the shuffle scheme.

Which "read" are you talking about here? Is it the one in ShuffleFederatedScheme or is it the one that originally reads the input? 

> The federated parameter server unit test input directory is currently not cleaned up. Should this be done after every run?

I think that tests in other packages do not clean this up. That is at least what I have observed, so I don't think you have to worry about it. 

> How to generate the ID of a newly created FederationMap correctly?

I think that you can make a call to: 
`FederationUtils.getNextFedDataID()`

tobiasrieger:
I'm talking about the read, that originally reads the input. I've already asked Sebastian B. to take a look, as I don't think the issue is with the parameter server. 

Thank you so much for your other suggestions, I've implemented them!

mboehm7:
ok, this invalid data consolidation issue is now fixed in master - please rebase.

src/test/java/org/apache/sysds/test/functions/federated/paramserv/FederatedParamservTest.java
```diff
@@ -117,62 +115,53 @@ private void federatedParamserv(ExecMode mode) {
...
+			double[][] labels = generateDummyMNISTLabels(_dataSetSize, numLabels);
+			String featuresName = "X_" + _numFederatedWorkers;
+			String labelsName = "y_" + _numFederatedWorkers;
+
+			federateBalancedAndWriteInputMatrixWithMTD(featuresName, features, _numFederatedWorkers, ports);
+			federateBalancedAndWriteInputMatrixWithMTD(labelsName, labels, _numFederatedWorkers, ports);
+
 			try {
-				Thread.sleep(1000);
+				Thread.sleep(2000);
```
sebwrede, line 135: Why does it need to sleep before continuing?

src/test/java/org/apache/sysds/test/functions/federated/paramserv/FederatedParamservTest.java
```diff
@@ -117,62 +115,53 @@ private void federatedParamserv(ExecMode mode) {
...
+					"scheme=" + _scheme,
+					"network_type=" + _networkType,
+					"channels=" + C,
+					"hin=" + Hin,
+					"win=" + Win,
+					"seed=" + 25));
+
 			programArgs = programArgsList.toArray(new String[0]);
 			LOG.debug(runTest(null));
 			Assert.assertEquals(0, Statistics.getNoOfExecutedSPInst());
```
sebwrede, line 162: It could make sense to do some assertion on the output of the execution to make sure that it executed correctly.

tobiasrieger:
Closed, because of a new pull request containing these features

sebwrede (commented):
I was about to submit the following just before this PR was closed: 

> I added two additional comments. 
> I believe that the tests will pass if this branch is rebased now. 
> Besides this, the PR LGTM.

PR #1114, opened 2020-11-23
[SYSTEMDS-2696] Overlapping relational operations

Baunsgaard:
This commit adds relational support for relational operations,
(< > <= etc) in the compressed space, for overlapping matrices.

If the relational expression returns a constant matrix, the operations
are super fast, while if the output is mixed it performs equally to our
uncompressed. Further optimizations are available, but since this
operation is not critical it is not explored yet.

nNmbers are in in ms avg.

cla     ,    0.1700  constant output
lcla    ,    0.0700  constant output
mkl     ,    1.2300  constant output
cla     ,   43.7400  Normal Perfomance
lcla    ,   43.9900  Normal Perfomance
mkl     ,   47.8400  Normal Perfomance

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -280,7 +280,7 @@
 					<reuseForks>false</reuseForks>
 					<reportFormat>brief</reportFormat>
 					<trimStackTrace>true</trimStackTrace>
-					<rerunFailingTestsCount>2</rerunFailingTestsCount>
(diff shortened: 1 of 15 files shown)


Baunsgaard:
Measurements before was wrong actual results are:
```code
cla     ,              3.1200
lcla    ,              0.4000
mkl     ,             62.3300
cla     ,              3.0800
lcla    ,              0.4300
mkl     ,             64.5300
```

PR #1124, opened 2020-12-12
[SYSTEMDS-2757] PCA Predict and Inverse

Baunsgaard:
Adds a predict function for PCA and an inverse function.

The predict is for unseen data, that the PCA was not trained for just like our other predict functions for other algorithms
The Inverse is to apply on an PCA reduced dataset, to reconstruct an approximation of the original data.

The second is added because PCA can be considered a classic Lossy compression algorithm, that gives an smaller, approximated matrix of the original data. It also allows us to calculate how well the given PCA fits to the dateset.

### DIFF
--- a/scripts/builtin/scale.dml
+++ b/scripts/builtin/scale.dml
@@ -19,29 +19,50 @@
 #
 #-------------------------------------------------------------
 
-# Scale and center individual features in the input matrix 
-# (column-wise) using z-score to scale the values.
-# -----------------------------------------------------------------------------
-# NAME   TYPE    DEFAULT  MEANING
-# -----------------------------------------------------------------------------
-# X      Matrix  ---      Input feature matrix
-# center Boolean TRUE     Indicates whether or not to center the feature matrix
-# scale  Boolean TRUE     Indicates whether or not to scale the feature matrix
-# -----------------------------------------------------------------------------
-# Y            Matrix ---  Output feature matrix with K columns
-# -----------------------------------------------------------------------------
+# Scale and center individual features in the input matrix (column wise.) using z-score to scale the values.
+# ---------------------------------------------------------------------------------------------
+# NAME         TYPE      DEFAULT  MEANING
+# ---------------------------------------------------------------------------------------------
+# X            Matrix    ---      Input feature matrix
+# Center       Boolean   TRUE     Indicates whether or not to center the feature matrix
+# Scale        Boolean   TRUE     Indicates whether or not to scale the feature matrix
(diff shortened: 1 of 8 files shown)


scripts/builtin/scale.dml
```diff
@@ -19,29 +19,48 @@
...
+
+    ScaleFactor = sqrt(colSums(X^2)/(N-1))
 
-    #scale by std-dev and replace NaNs with 0's
-    X = replace(target=X/sqrt(cvars),
-      pattern=NaN, replacement=0);
+    # Replace entries in the scale factor that are 0 with an small eps
+    # To avoid division by 0, and thereby introducing NaN to the ouput.
+    ScaleFactor = replace(target=ScaleFactor,
+      pattern=NaN, replacement=1e-16);
```
Baunsgaard, line 54: Alternatively we could replace the values with 1, that would maybe be better since that would remove the scaling of that dimension.

scripts/builtin/pcaPredict.dml
```diff
@@ -0,0 +1,49 @@
...
+# NAME          TYPE    DEFAULT       MEANING
+# ---------------------------------------------------------------------------------------------
+# X             Matrix  ---           Input feature matrix
+# Centering     Matrix  empty matrix  The column means of the PCA model, subtracted to construct the PCA
+# ScaleFactor   Matrix  empty matrix  The scaling of each dimension in the PCA model
+# ---------------------------------------------------------------------------------------------
+# Y             Matrix  ---           Output feature matrix dimensionally reduced by PCA
+# ---------------------------------------------------------------------------------------------
+
+m_pcaPredict = function(Matrix[Double] X, Matrix[Double] Clusters, Matrix[Double] Centering,  Matrix[Double] ScaleFactor)
```
Shafaq-Siddiqi: Can we reconsider the name of this built-in? As from the word predict we usually refer to some class or value prediction but here you are somehow transforming the data using pre-computed dominant vectors. I would suggest pca_tranform() it would be consistent with PCA implementation in Scikit-learn.

Baunsgaard: sure transform works as well! :+1:

Baunsgaard: but i would refrain from the underscore. 
and we have to agree on that general syntax in our interface
are we using camel case "java syntax" or "google syntax" underscore?

Shafaq-Siddiqi: I prefer the camel case. For now, I see both styles in our built-in package but the camel case is dominant. So go for this.

PR #1125, opened 2020-12-15
[DIA] Scikit-learn converter

erwinTU:
Hello!

This is our first draft for a scikit-learn converter.
After some consideration we propose two approaches.
For now we are leaning towards the second idea.

What is your opinion on this?
@Shafaq-Siddiqi


Shafaq-Siddiqi:
Hi,
I appreciate the design draft, it is a good effort. I would suggest doing the mapping of Scikit-learn algorithms to DML and vice versa. Keep it simple you only need to do the part that fits the project scope. I will suggest starting with a very simple pipeline of algorithms that are present in both DML and Scikit-learn like "scale and Kmeans" or "scale, PCA and Kmeans".  This seems quite a good contribution to me. 
@mboehm7 I would appreciate your input on this.

mboehm7:
In general, that's a good starting point. We had another use case of importing sk-learn pipelines in mind, but adding the sklearn-onnx-dml model converter is also an interesting exploratory project which we can put into staging (so the dependencies do not affect our runtime). 

For sk-learn pipelines we would have taken some example pipelines (see make_pipeline) composed of primitives like scaling, encoding, cross validation, model training and converted that into a DML script that calls the respective existing builtin functions for scale, encoding, etc. 

ONNX on the other hand is primarily used as exchange format for neural network specifications and models. So it's not a direct fit for encoding such ML pipelines. However, focusing on the exchange of a trained model and its use, many linear models can be encoded via fully connected layers and similar operations. So from my perspective, it's fine to pursue this project. I moved the existing onnx converter from our python API to staging because an incompatible API change. Fixing the onnx importer in the process would be nice, and should not be that much overhead.

erwinTU:
Hi. Sorry for the delay in answering.

I think we looked into quite a different direction at first, but after reading your comments we have now implemented a very simple proof of concept. Especially the idea with the pipeline was helpful (@mboehm7).

This POC uses a dumped (using pickle) Sklearn pipeline and produces a simple dml script:

```
Sklearn pipeline:
Pipeline(steps=[('standardscaler', StandardScaler()), ('kmeans', KMeans())])

DML Script
source("scripts/builtin/scale") as ns_scale
source("scripts/builtin/kmeans") as ns_kmeans

step_0 = ns_scale::m_scale(X, True, True)
step_1 = ns_kmeans::m_kmeans(step_0, 8, 10, 300)
```

Is this approach alright? 
@Shafaq-Siddiqi

erwinTU:
Hi.

We have implemented the sklearn-to-dml pipeline mapper, which can be used as cli or imported as python module.
There are still a few things to test and cleanup, but other than that this PR should be ready to review soon.

@Shafaq-Siddiqi could you take a look at our current implementation?
We would like to hear some feedback and how to proceed afterwards (regarding the practicals).

erwinTU:
Hi!

I have now cleaned up the files, restructured the project directory and added a short quick guide.
There is also a list of supported mapping in the `mapped_functions.rst` file.

Test-wise, we only have problems with the dml builtins for PCA, DBSCAN and GaussianMixtureModel.
For PCA and DBSCAN a `NullPointerException -- null` triggers. 
GaussianMixture triggers a `LanguageException -- ERROR: [line 0:0] -- Function get_sample_maps() is undefined.`
We have left the mappings and failing tests as there are now.

The other pipeline elements work.

Do you have any suggestions for us? 
Otherwise, we would like to submit the project as is, since the (hard) deadline is in sight.

@mboehm7 
@Shafaq-Siddiqi

mboehm7:
LGTM - overall this is a good start, and I will merge it in with minor changes in the next days. Sorry for the delay.

j143:
Hi @mboehm7 ,

we missed this one. 😅

j143:
Hi @erwinTU - I started reviewing this recently encountered issues directly running.

I will make minor changes so that it will be runnable and add usage docs.
If you would like to help, could you let me know?

Thanks for working on this! 😄

j143:
Thank you, @erwinTU @Sneaksolid and Alexander - LGTM. 👍 🎉 🚀

Your work is inside systemds staging with commit 97633c74d5fcf4d884aef433192ada6c977d465d

I have fixed the function signature mapping for gmm and impute, so that all tests pass.

If you would like to continue working on this please go ahead and give a PR. 😄

PR #1126, opened 2020-12-16
[BUGFIX] Federated LMCG Bug

sebwrede:
This PR provides a test case where rewrites are not applied in the federated execution of LMCG. 
One test case with X and y as federated data on two workers are FederatedLmCGTest.testLmMatrixDenseCPlmCG2. The execution for instance has the instruction `CP r' X.MATRIX.FP64 _mVar205.MATRIX.FP64 8`.


mboehm7:
LGTM - thanks for the test @sebwrede. I now added explicit error handling for inconsistent federated data characteristics, and fixed the test accordingly (the underlying issues of not applying the rewrites was that the fedinit specified zero columns for y).

PR #1128, opened 2020-12-18
[SYSTEMDS-2761] Transpose in place

Baunsgaard:
This commit adds a function to our Lib MatrixReorg containing a way
to transpose in place a dense matrix block. (single threaded for now)

The implementation is based on:

A  Decomposition  for  In-place  Matrix  Transposition
by Bryan Catanzaro et al. PPoPP 2014

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/data/DenseBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/data/DenseBlock.java
@@ -49,12 +49,7 @@ public abstract class DenseBlock implements Serializable
 	private double[] _reuse;
 	
 	protected DenseBlock(int[] dims) {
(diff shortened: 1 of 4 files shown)


Baunsgaard:
The in place transpose is done, but the performance is worse than our previous, therefore this PR is on standby.

After implementing this in place transpose, it is slower than our default transpose that allocate a full matrix in cases where columns != rows.
This is because for each cell in the output each value is moved 6 times, while in our allocate a new matrix we move once. The allocation of the new matrix does introduce memory overhead, that this in place operation does not.

The two instances that is much faster is the in place transpose with cols == rows, because each cell only require a single swap and instances where either rows or columns == 1, the latter is because even if the previous is a metadata operation, it still copies the vector.

PR #1129, opened 2020-12-18
[SYSTEMDS-2762] Federated rshape instruction

OlgaOvcharenko:
This PR adds federated rshape instruction and improvement for running different instructions for different workers. For now, it's only possible to perform fed rshape only if all fed parts can be reshaped.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko. During the merge I only made minor formatting changes, removed warnings, and remove the new `executeDifferentInstructions` to avoid duplication.

PR #1130, opened 2020-12-18
[SYSTEMDS-2763] Federated rowIndexMax and rowIndexMin

OlgaOvcharenko:
This PR adds the fed uarimin and uarimax instructions and also fed product of all matrix cells prod(). 
Parsing of uarimin, uarimax CP instructions is changed by adding a parameter to the instruction string to return only index or index and min/max (for fed).


Baunsgaard:
LGTM, and merged sorry for the delay!

PR #1131, opened 2020-12-18
[SYSTEMDS-2550] Balancing and data partitioning

tobiasrieger:
This PR includes the closed PR #1113 and all changes proposed in its comments. It was rebased on master and consolidated to make it easier to merge

Changes list:
- Added four new federated data partitioning schemes
	- ShuffleFederatedScheme
	- SubsampleToMinFederatedScheme
	- BalanceToAvgFederatedScheme
	- ReplicateToMaxFederatedScheme
- Added runtime balancing parameter to parameter server (has a default parameter and is optional for the federated case)
- Different runtime balancing schemes
	- RUN_MIN
	- CYCLE_AVG
	- CYCLE_MAX
- Simplified the federated control thread to accommodate the coming N-batch frequency
	- There is now only one UDF, which computes a given number of batches. If more batches than the local epoch are specified it will cycle. This one function is sufficient for batch, N-Batch or Epoch Frequencies.
- Provided a convenient way to create federated test matrices in the Automated Test Base
- .getFedMapping() of FederationMap was refactored to getFRangeFDataMap to avoid confusion with the method of MatrixObject with the same name.
- Updated and improved CNN.dml and TwoNN.dml scripts

### DIFF
--- a/src/main/java/org/apache/sysds/parser/ParameterizedBuiltinFunctionExpression.java
+++ b/src/main/java/org/apache/sysds/parser/ParameterizedBuiltinFunctionExpression.java
@@ -286,7 +286,7 @@ public class ParameterizedBuiltinFunctionExpression extends DataIdentifier
 			raiseValidateError("Should provide more arguments for function " + fname, false, LanguageErrorCodes.INVALID_PARAMETERS);
 		}
 		//check for invalid parameters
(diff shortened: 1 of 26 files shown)


mboehm7:
LGTM - thanks for the extensions @tobiasrieger. This patch is a good start for experiments on these balancing and data partitioning techniques.

During the merge, I only made a few tweaks here and there: added serial IDs to serializable classes, renamed `getFRangeFDataMap` to `getMap`, capitalized all class names, replaced the prints of non-implemented errors with `NotImplementedException`s, replaced the progress printing with log info messages (so they can be disabled if needed), reconfigured the matrix multiplies for shuffling, replication, and sampling to the available vcores (instead of 1), changed the wait in the federated paramserv test to the global fed-worker wait configuration, and made some minor formatting changes.

PR #1132, opened 2020-12-18
[SYSTEMDS-2764][WIP] DataGen operation for frames

Shafaq-Siddiqi:
The Spark implementation is incomplete.
The case for unnamed parameters is not supported yet.

### DIFF
--- a/src/main/java/org/apache/sysds/common/Types.java
+++ b/src/main/java/org/apache/sysds/common/Types.java
@@ -448,7 +448,7 @@ public class Types
 	}
 	
 	public enum OpOpDG {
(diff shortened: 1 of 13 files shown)


mboehm7:
LGTM - thanks for the initiative and the patch @Shafaq-Siddiqi. This is a great addition. I now fixed the spark instructions (seed generation, block sizes) and a few other issues (matrix redirect to reshape, warnings, formatting, frame recompilation, tests), as well as started refactoring `DataExpression` to make it more maintainable. 

In a follow-up PR (whenever you get a chance), please fix the hard-coded seed though. This requires consistent seed generation for 1k blocks on the CP side too so both CP and Spark generate exactly the same results.

PR #1135, opened 2020-12-26
Design Document for Python Script Generator

juliale-15:
@A-Postl and I created this first version of a design document for the python script generator. We would appreciate feedback if our planned approach could work like this or not.


dev/python_script_generator/Generator - Design Document.md
```diff
@@ -0,0 +1,170 @@
...
+3. Implement generator, that generates Python API functions from parsed dml function definitions
+4. Implement generator, that generates multi-line string comment for Python API functions from parsed function headers
+5. Implement generator, that generates python files with license header and can use an arbitrary amount of string parameters, in our case for python code and comments
+6. Include subprocess call for the builtin-api-generator in `src/main/python/create_python_dist.py`
+
+
+
+## Generator Folder Structure
+
+We plan on creating a folder called `generator` in `src/main/python/systemds`. In this folder we will place our `generator.py`, a `parser.py` and a folder called `scripts/builtin` where we will place the generated python scripts.
```
Baunsgaard, line 164: The folder `src/main/python/systemds` is part of the distributed pip package afterwards, i think it would be better if this generating script is not. 
Therefore making a the neww folder `generator` directly on the src/main/python path would be good.

dev/python_script_generator/Generator - Design Document.md
```diff
@@ -0,0 +1,170 @@
...
+    Performs KMeans on matrix input.
+
+    :param x: Input dataset to perform K-Means on.
+    :param k: The number of centroids to use for the algorithm.
+    :param runs: The number of concurrent instances of K-Means to run (with different initial centroids).
+    :param max_iter: The maximum number of iterations to run the K-Means algorithm for.
+    :param eps: Tolerance for the algorithm to declare convergence using WCSS change ratio.
+    :param is_verbose: Boolean flag if the algorithm should be run in a verbose manner.
+    :param avg_sample_size_per_centroid: The average number of records per centroid in the data samples.
+    :return: `OperationNode` List containing two outputs 1. the clusters, 2 the cluster ID associated with each row in x.
```
Baunsgaard, line 131: the multi return is not nice as it is, so if you have better suggestions feel free to design it better.

dev/python_script_generator/Generator - Design Document.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Generator Folder Structure
+
+We plan on creating a folder called `generator` in `src/main/python/systemds`. In this folder we will place our `generator.py`, a `parser.py` and a folder called `scripts/builtin` where we will place the generated python scripts.
+
+In summary we will create the following folder structures:
+
+1. `src/main/python/systemds/generator/generator.py`
+2. `src/main/python/systemds/generator/parser.py`
+3. `src/main/python/systemds/generator/scripts/builtin/`
```
Baunsgaard, line 170: I would like the scripts that are generated to be located in:

`systemds/operator/algorithm/builtin/kmeans.py`

Then furthermore when they are located there, make it so that the __init__.py in `systemds/operator/__init__.py` loads in all the builtin functions generated.

Baunsgaard (commented):
LGTM

there are some minor locations that i want to change, for the files but other than that, it sounds nice.

You might want to consider making it such that the thing you parse first is the DML defined function and return (Defined on line 53-56 in this document). This is what the program reacts on and not the documentation above. Then after this use that definition of function to verify that the comment above is aligned with the function definition.
This would be very useful in our test suite as well, to enforce a scheme for future function definitions, or maybe later automated correction of the documentation.

Baunsgaard:
Closed and merge, since you already have a PR for the implementation. 
the only change i did was move the file to a different folder.

PR #1138, opened 2021-01-01
Add design document for LLVM backend code generator

FraCorti:
I created the first version of the design document for the LLVM backend code generator. Any feedback and suggestions would be appreciated. Thanks in advance, Francesco.

### DIFF
--- a/dev/llvm-codegen-backend/llvm-codegen-design.md
+++ b/dev/llvm-codegen-backend/llvm-codegen-design.md
@@ -1,0 +1,55 @@
+Francesco Corti 
+
+# LLVM Code Generator - Design Document 
+
+## Introduction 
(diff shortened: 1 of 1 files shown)


mboehm7:
LGTM - thanks for the design doc @FraCorti. I moved it now to `scripts/staging`, added the license header, removed the author name (apart from the git history, we're not using author tags), and added two notes. Specifically, please try to avoid hard dependencies to LLVM by using a similar integration like our native BLAS libraries [1]. You can compile a proxy shared library, users specify the location of LLVM libraries via `-Djava.library.path`, and we try to dynamically load this library during runtime and fallback to java if unsuccessful. In maven, you can specify the scope of llvm as test.   

[1] https://github.com/apache/systemds/blob/master/src/main/java/org/apache/sysds/utils/NativeHelper.java

PR #1142, opened 2021-01-04
[SYSTEMDS-2548][SYSTEMDS-2796] Federated frame and matrix right indexing

OlgaOvcharenko:
This PR adds frame right indexing and modifies matrix right indexing by using instructions instead of UDFs and one class for both (frames and matrices).


OlgaOvcharenko:
@mboehm7 as discussed offline FederatedUnivarTest fails because the two last indices in the instruction string are literals and thus can not be found in the symbol table.

Caused by: org.apache.sysds.runtime.DMLRuntimeException: Variable '2' does not exist in the symbol table.		CP°rightIndex°0·MATRIX·FP64°1·SCALAR·INT64·true°10000·SCALAR·INT64·true°1·SCALAR·INT64·false°1·SCALAR·INT64·false°2·MATRIX·FP64

mboehm7:
LGTM, thanks for the patch @OlgaOvcharenko. As it turned out the remaining issue was still a local issue in federated indexing. In detail, the original CP instruction was:

`
CP°rightIndex°X·MATRIX·FP64°1·SCALAR·INT64·true°10000·SCALAR·INT64·true°i·SCALAR·INT64·false°i·SCALAR·INT64·false°_mVar314·MATRIX·FP64
`
 
When acquiring the index range, we obtain the value of `i` from the symbol table, but during the replacement we tried to replace the variable names only. By simply constructing new scalar literals directly we now handle this correctly. Additionally, I only removed some unnecessary imports (warnings).

PR #1149, opened 2021-01-09
builtIn outlierByArima function

juulez:
added the builtIn outlierByArima function, which focuses on the AR part


mboehm7:
LGTM. That's awesome - thanks for the patch @juulez. During the merge, I only made some minor tweaks regarding formatting and warnings, and changed `(rowMaxs(outlierFilter) == 0)` to `(outlierFilter == 0)` as this filter is always a column vector (just for readability as the compiler would anyway simplify this).

PR #1152, opened 2021-01-13
First version of Python API generator

juliale-15:
@A-Postl and I have created this first version of the python script generator.
We've divided the generator into:

1. A `parser.py` which takes the *.dml scripts and uses the function definition to create a python script. Additionally the parse parses the dml header. The information in the header is used to create a documentation for the python script as well as to check if header information matches the dml function definition.
2. A `generator.py` which takes the parsed data from the `parser.py` and generates a python file. Additionally the generator creates an `__init__.py` file where all generated scripts are added.

The scripts are placed under the folder structure `systemds/operator/algorithm/builtin`.
We have changed a set of dml files in a way that our generator accepts them (kmeans, kmeanspredict, lm, l2svm, pca, multiLogReg, multiLogRegpredict). Those are also builtin functions which already existed in the python API because we could then reuse the testcases for those functions.

Our generated scripts can be imported as before when the functions where placed in `algorthms.py`. Therefore we think that the file `algorithm.py` can be removed.

However we have 2 concerns regarding the *.dml scripts.
1. Some builtin functions such as `multiLogReg` for example require additional shape checks. 
````
    if -1 in x.shape:
        output_shape = (-1,)
    else:
        output_shape = (x.shape[1],)
````
We were wondering how our parses could use the dml file to create those checks, since they kind of depend on the function itself. 

2. In the `pca.dml` file 4 return values are defined. However for the python script we were only allowed to set the `OperationNode`like this `number_of_outputs=2, output_types=[OutputType.MATRIX, OutputType.MATRIX]`.
Why is it that the `pca.dml` defined 4 return values but the OperationNode only allowes the `number_of_outputs=2`?

We would really appreciate some feedback and maybe some suggestions on how we could handle the two problems mentioned above.


Baunsgaard:
to respond to the questions 

1. The shape requirements in the python script is only there to allow quick termination of the program, from python, therefore not requireing the dml script to crash. We can simply remove these.
2. If you look at multilogreg then you can see an example of 3 return values. The PCA have changed recently to return more values and apparently this did not break the PCA for python (which it should have). The multiReturn is currently not a good solution in general, and could benefit from some other handling like for instance a dedicated operation node type.

I also quickly went through the code and it looks nice, will come back to it with comments later.

Baunsgaard:
Thanks alot for the contribution!

juliale-15:
> hi,
> 
> I really like the functionality, great job.
> I am merging this code, but there is some design decisions i want to consider before finalizing the merge.

Thanks! :D

PR #1153, opened 2021-01-13
Gaussian Classifier

kpretterhofer:
This implementation computes a simple gaussian classifier, e.g. it outputs the respective parameters which are needed for classifying.

As input, the function basically just receives a feature matrix, and a target vector (and some small value for smoothing along the variances, to prevent numerical errors). The function computes and returns:

- prior probability
- means
- determinants
- inverse covariance matrix

per class. 
For classifying one can compute: p(C=c | x) = p(x | c) * p(c)
where p(x | c) is the (multivariate) Gaussian PDF for class c, and p(c) is the prior probability for class c. 

Please let me know if and how I can still improve the code, s.t. it fits well into SystemDS. 

One thing where I was quite unsure was the unit tests. Since calculating determinants and the inverse of the covariance matrices can lead to floating point errors,  I was not quite sure how to compare the results.  I did compare most of them, as suggested in the mailing list, with the avg. bit distance, with a quite high maxUnitsOfLeastPrecssion. 
Although the values from the inverse covariance matrices  can differ a lot (systemDS vs R), i am pretty sure that the computation  is correct, since multiplying it with the covariance matrix itself, leads to the identity (which I tested during development).


scripts/builtin/gaussianClassifier.dml
```diff
@@ -0,0 +1,127 @@
...
+  return (Matrix[Double] classPriors, Matrix[Double] classMeans,
+  List[Unknown] classInvCovariances, Matrix[Double] determinants)
+{
+  #Retrieve number of samples, classes and features
+  nSamples = nrow(D)
+  nClasses = max(C)
+  nFeats = ncol(D)
+
+  #Set varSmoothing (to prevent numerical errors)
+  varSmoothing = 1e-9
```
Baunsgaard, line 61: here you overwrite the input parameter

scripts/builtin/gaussianClassifier.dml
```diff
@@ -0,0 +1,127 @@
...
+  classVars = aggregate(target=D, groups=C, fn="variance", ngroups=as.integer(nClasses));
+  classPriors = classCounts / nSamples
+
+  smoothedVar = diag(matrix(1.0, rows=nFeats, cols=1)) * max(classVars) * varSmoothing
+
+  classInvCovariances = list()
+  determinants = matrix(0, rows=nClasses, cols=1)
+
+  #Compute determinants and inverseCovariances
+  for (class in 1:nClasses)
```
Baunsgaard, line 75: consider if it can be parallelized, if so use a parfor

scripts/builtin/gaussianClassifier.dml
```diff
@@ -0,0 +1,127 @@
...
+
+    determinants[class, 1] = det
+    classInvCovariances = append(classInvCovariances, invCovMatrix)
+  }
+
+  #Compute accuracy on the training set
+  if (verbose)
+  {
+    results = matrix(0, rows=nSamples, cols=nClasses)
+    for (class in 1:nClasses)
```
Baunsgaard, line 112: again maybe parfor

scripts/builtin/gaussianClassifier.dml
```diff
@@ -0,0 +1,127 @@
...
+        intermediate = -1/2 * log((2*pi)^nFeats * determinants[class,])
+        intermediate = intermediate - 1/2 * (meanDiff %*% as.matrix(classInvCovariances[class]) %*% t(meanDiff))
+        intermediate = log(classPriors[class,]) + intermediate
+        results[i, class] = intermediate
+      }
+    }
+    acc = sum(rowIndexMax(results) == C) / nSamples * 100
+    print("Training Accuracy (%): " + acc)
+  }
+}
```
Baunsgaard, line 127: newline in the end of the scripts to make github happy.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinGaussianClassifierTest.java
```diff
@@ -0,0 +1,142 @@
...
+		double[][] meansSYSTEMDS = TestUtils.convertHashMapToDoubleArray(meansSYSTEMDStemp);
+		double[][] determinantsR = TestUtils.convertHashMapToDoubleArray(determinantsRtemp);
+		double[][] determinantsSYSTEMDS = TestUtils.convertHashMapToDoubleArray(determinantsSYSTEMDStemp);
+		double[][] invcovsR = TestUtils.convertHashMapToDoubleArray(invcovsRtemp);
+		double[][] invcovsSYSTEMDS = TestUtils.convertHashMapToDoubleArray(invcovsSYSTEMDStemp);
+
+		TestUtils.compareMatrices(priorR, priorSYSTEMDS, Math.pow(10, -5.0), "priorR", "priorSYSTEMDS");
+		TestUtils.compareMatricesBitAvgDistance(meansR, meansSYSTEMDS, 5L,5L, this.toString());
+		TestUtils.compareMatricesBitAvgDistance(determinantsR, determinantsSYSTEMDS, (long)2E+12,(long)2E+12, this.toString());
+		TestUtils.compareMatricesBitAvgDistance(invcovsR, invcovsSYSTEMDS, (long)2E+20,(long)2E+20, this.toString());
```
Baunsgaard, line 140: i would probably add the execution of the inverse multiplied with the result,
just as you stated in the pr it gives the original.

if you add this to the tests, would it still make sense to run the R script?

src/test/scripts/functions/builtin/GaussianClassifier.R
```diff
@@ -0,0 +1,111 @@
...
+}
+
+writeMM(as(classPriors, "CsparseMatrix"), paste(args[3], "priors", sep=""));
+writeMM(as(classMeans, "CsparseMatrix"), paste(args[3], "means", sep=""));
+writeMM(as(determinants, "CsparseMatrix"), paste(args[3], "determinants", sep=""));
+writeMM(as(stackedInvCovs, "CsparseMatrix"), paste(args[3], "invcovs", sep=""));
+
+
+
+
```
Baunsgaard, line 105: to many new lines

src/test/scripts/functions/builtin/GaussianClassifier.dml
```diff
@@ -0,0 +1,37 @@
...
+invcovs = as.matrix(covs[1])
+for (i in 2:max(y))
+{
+  invcovs = cbind(invcovs, as.matrix(covs[i]))
+}
+
+write(prior, $4);
+write(means, $5);
+write(det, $6);
+write(invcovs, $7);
```
Baunsgaard, line 37: newline

Baunsgaard (commented):
minor changes needed, but otherwise LGTM.

Baunsgaard:
> One thing where I was quite unsure was the unit tests. Since calculating determinants and the inverse of the covariance matrices can lead to floating point errors, I was not quite sure how to compare the results. I did compare most of them, as suggested in the mailing list, with the avg. bit distance, with a quite high maxUnitsOfLeastPrecssion.

Maybe since the correction value is overwritten in DML and therefore not the same parameter used for the algorithm, but still as long as the difference is this small i consider it fine.

> Although the values from the inverse covariance matrices can differ a lot (systemDS vs R), i am pretty sure that the computation is correct, since multiplying it with the covariance matrix itself, leads to the identity (which I tested during development).

please do add such a test back, since it reflects "a function of the algorithm", not "similarity to R"  since in the end what we want is the first, not necessarily the second.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinGaussianClassifierTest.java
kpretterhofer: thanks for the feedback. I will commit the suggested changes as soon as possible.
Just one follow up question to this specific suggestion:
Do you mean I should check wether covariance * invCovariance is indeed the identity matrix?
If so, since the covariance matrix is not returned (but just the inverse), should I compute, 
the respective covariance matrices again in the test dml file, and then return the product, to check
wether it is indeed the identity? 
If this is not what you meant, it would be great if you could clarify. 
Thanks :)

Baunsgaard: yes covariance * invCovariance == Identity.

and if there are other interesting properties. My initial understanding was that there was a way to reconstruct the input from the output, but i also might have missed a point, and looking a covariance there does not seem to be a way of doing that?

kpretterhofer: reconstructing the input from the output is afaik not really possible - since you lose for example the number of input samples provided to the algorithm. However computing the covariance matrix again, and multiplying it with its inverse is definitely a way of proving that the inverse is indeed correct, although its R aquivalent seems to be kinda different (but as already mentioned, just because of floating point differences).

scripts/builtin/gaussianClassifier.dml
kpretterhofer: I pushed a new commit. Unfortunately parfor in this case was not possible, since I append the results to a list, which can not be parallelized.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinGaussianClassifierTest.java
kpretterhofer: For now I have not added such a test, since the cov matrix itself is not part of the output parameters. The only way to perform such a test is to calculate the cov matrices again inside the dml test file, and then multiply them with their respective inverses. Of course I can implement it like this, but I just wanted to make sure if this is the correct approach to do, since it seems a bit weird to me to recalculate the cov matrices inside the test files.

Shafaq-Siddiqi:
Hi,
Thank you for your contribution. Could you please add a test to predict some data using the classifier. We have iris dataset inside function/transform/input directory. You can write a test to predict the class labels for iris.

kpretterhofer:
> Hi,
> Thank you for your contribution. Could you please add a test to predict some data using the classifier. We have iris dataset inside function/transform/input directory. You can write a test to predict the class labels for iris.

Thanks for the feedback! 
Probably just one follow up question: Should I only write a test for predictions, or should I also implement a builtin prediction function? (The current implementation only computes and returns the parameters needed for prediction, thats why I am asking)

Shafaq-Siddiqi:
> > Hi,
> > Thank you for your contribution. Could you please add a test to predict some data using the classifier. We have iris dataset inside function/transform/input directory. You can write a test to predict the class labels for iris.
> 
> Thanks for the feedback!
> Probably just one follow up question: Should I only write a test for predictions, or should I also implement a builtin prediction function? (The current implementation only computes and returns the parameters needed for prediction, thats why I am asking)

For now, I would suggest writing a test for prediction.

kpretterhofer:
> > > Hi,
> > > Thank you for your contribution. Could you please add a test to predict some data using the classifier. We have iris dataset inside function/transform/input directory. You can write a test to predict the class labels for iris.
> > 
> > 
> > Thanks for the feedback!
> > Probably just one follow up question: Should I only write a test for predictions, or should I also implement a builtin prediction function? (The current implementation only computes and returns the parameters needed for prediction, thats why I am asking)
> 
> For now, I would suggest writing a test for prediction.

Thanks! 
I have added such a test, and took 45 samples per iris to train the algorithm, and 5 sample per iris to predict. Since this test slightly differs from the others, I needed to create a second DML script for testing. Hope that this approach is okay!

Shafaq-Siddiqi:
LGTM,
Thank you @kpretterhofer for your contribution. During the merge, I fixed some minor formatting and added a TODO for future reference.

PR #1156, opened 2021-01-14
[SYSTEMDS-2794] custom dense reader compression 

Baunsgaard:


Baunsgaard:
Going from 19235 ms to 7280 compression time on airlines on alpha, while laptop 5933 to 5458, for first compression on startup of jvm.

PR #1159, opened 2021-01-15
[WIP] Advanced CSE

thomasmo3:
Good day!

This is our  (@exokortex @thomasmo3) draft for the advanced CSE optimization.

What do you think about it? Looking forward to your feedback!
@Shafaq-Siddiqi

### DIFF
--- a/scripts/advanced_cse/design.rst
+++ b/scripts/advanced_cse/design.rst
@@ -1,0 +1,18 @@
+Advanced Common Subexpression Elimination
+==================================
+
+Common Subexpression Elimination (CSE) is a commonly used compiler optimization technique minimizing code by detecting already calculated (and therefore duplicate) expressions and replacing them with a single variable if needed, which holds the computed value. Improving CSE can drastically improve performance in terms of speed and memory requirements.
+
+Current State
+----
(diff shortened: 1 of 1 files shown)


mboehm7:
Thanks for the initial document. The individual DAGs are delineated by conditional control flow and hence, there is only limited potential for doing CSE on sub-sequences of program blocks (and thus, DAGs). Doing CSE across control flow (e.g., across branches and loops) is possible but is not the first step for extensions. 

I would recommend to focus on a single DAG with commutative operations like `A + 7` and `7 + A`, which are currently only indirectly eliminated (for plus and add, but not for `max(A,B)` vs `max(B,A)`). There are approaches: (1) either introduce canonicalization rewrites that systematically order inputs (e.g., by hop ID) or (2) extend the actual CSE rewrite for consider such permutations. I would recommend approach (1) which would be best added to  `RewriteAlgebraicSimplificationDynamic` as this canonicalization rewrite would need to check for matrix sizes too as we only support broadcasting (e.g., matrix vector elementwise operations) on the right hand side. 

The scope of operations could be: OpOpN - MIN, MAX, PLUS, as well as OpOp2 - AND, OR, XOR, COV, EQUAL, NOTEQUAL, MIN, MAX, PLUS, MULT, and potentially other comparisons but they additionally require modifications of the comparison operator on reordering.

PR #1163, opened 2021-01-18
[SYSTEMDS-2799] Lineage tracing and reuse of federated UDFs

phaniarnab:
This patch builds the foundation of lineage tracing and reuse
of federated UDFs, and adds a test which calls lower.tri in a
loop and reuse the outputs of Tri UDF in the workers.
The core idea is to treat a UDF as an instruction which always
produces the same results for same inputs.


phaniarnab:
This PR is to confirm that all the tests are passing.

PR #1164, opened 2021-01-18
[SYSTEMDS-2800] Agressive Compression Rewrite

Baunsgaard:
This commit change the COST based approach of the rewrite rules for compression to an "aggressive" compression approach.
The basic idea is to introduce many compression instructions into the computation graph, such that we have an compression command at all location that have a potential to:

1. Compress nicely from uncompressed representation
2. Restructure an already compressed matrix, into a better compression 

For the 2. case operations such as less than resulting in only 0 and 1 in the matrix have high potential for better restructuring.
But in general if the matrix inputted before the less than operation is not compressed it will try to compress anyway according to case 1, so no matter which case the instruction introduced should go the same place.

For this aggressive compression instruction to execute the object to compress have a few criteria.

1. There has to be at least 3 instructions executed on the object, that is "supported compressed operations" or the matrix is used in a loop.
2. The compressed operations that are used should not have many "supported but slow operations".
3. The matrix to compress have to be at least 1000 rows and below 10 columns, or have a ratio of 1000 to nrow/ncol.

### DIFF
--- a/src/main/java/org/apache/sysds/common/Types.java
+++ b/src/main/java/org/apache/sysds/common/Types.java
@@ -241,6 +241,7 @@ public class Types
 				case MULT2:           return "*2";
 				case NOT:             return "!";
 				case POW2:            return "^2";
+				case TYPEOF:          return "typeOf";
 				default:              return name().toLowerCase();
 			}
 		}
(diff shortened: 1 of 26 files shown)


Baunsgaard:
Further work is to introduce decompression in cases where the compressed matrix is used in more than 1 operation that would decompress.

PR #1165, opened 2021-01-19
[SYSTEMDS-2801] Rewrite rule to remove (par)for-loops over empty sequences.

pdamme:
This commit adds a new StatementBlockRewriteRule removing (par)for-loops which are known to iterate over an empty sequence, and, thus, will not perform a single iteration. This works if the loop's from, to, and increment are constants (possibly DML script arguments, and after constant folding). Furthermore, OptimizerUtils has a new flag determining whether this rule shall be applied. In ProgramRewriter's default rule sequence, this new rule is inserted directly after the removal of unnecessary branches, since they have a lot in common. Merging subsequent blocks is done once afterwards if any of removing unnecessary branches or removing
(par)for-loops shall be applied.

### DIFF
--- a/src/main/java/org/apache/sysds/hops/OptimizerUtils.java
+++ b/src/main/java/org/apache/sysds/hops/OptimizerUtils.java
@@ -124,6 +124,13 @@ public class OptimizerUtils
 	 */
 	public static boolean ALLOW_BRANCH_REMOVAL = true;
 	
+	/**
+	 * Enables the removal of (par)for-loops when from, to, and increment are constants
+	 * (original literals or results of constant folding) and lead to an empty sequence,
+	 * i.e., (par)for-loops without a single iteration.
+	 */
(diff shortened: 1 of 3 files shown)


phaniarnab:
Thanks for adding this rewrite.
One suggestion: In general, adding a test case might be a good idea. You can execute a script (local and spark) with and without the rewrite and match the results.

pdamme:
Good idea. I will add a test case soon.

mboehm7:
LGTM - welcome to SystemDS @pdamme and thanks for the patch. During the merge, I fixed a few related for/parfor issues (for correctly handling special values like NaN/Inf/-Inf), fixed the issue of my previous commit, simplified and generalized the rewrite a bit, and added the tests you already discussed. The test simply checks that after this rewrite, we do common subexpression elimination of ops before and after the loop.
```

print(sum(X));
for(i in seq(NaN,1))
  print("Iteration "+i)
print(sum(X))
```

PR #1169, opened 2021-01-21
Tokenizer API and initial algorithms

Iseratho:
This PR contains an initial implementation of a text tokenizer that serves as a design proposal. 

## Why does SystemDS benefit from an built-in tokenizer feature?

A built-in tokenizer is an important feature to allow end-to-end machine learning. At the moment SystemDS expects textual data to be provided in a tokenized form. For instance, one test case reads the 20 newsgroup dataset in bag-of-words (BoW) format. 
Currently, SystemDS supports transformations for frame-to-matrix and vice versa, as well as row-wise transformations via `map()`. However, textual data can be of arbitrary length. Thus, it does not fit the already built-in row-wise (i.e., one-to-one) mapping.

## Proposed Tokenize API

Our proposed tokenizer API is tailored towards SystemDs and provides a `tokenize()` function that enables frame to frame transformation using a JSON specification similar to the transform function. We have included an initial draft implementation (minimalistic whitespace tokenizer for BoW format), which shows how we would incorporate the tokenization feature and showcases it with test cases.
Similar to the transform functions, this should be extendable with new algorithms each with different configuration options and executable in a distributed way (Spark/Hybrid execution). 
The details are described in the [DML reference](https://github.com/apache/systemds/blob/f0d82002b85c9493940e20e240235345d1677b6e/docs/site/dml-language-reference.md#processing-frames).

## State of PR

This PR is not to be merged at the moment, but should provide a way to discuss and improve the design considerations. We plan to integrate more tokenizers and improve the options to configure the algorithms, as well as showcase the usage of the tokenizer by creating a standard natural language pipeline  (e.g., preprocess, tokenize, classify).

## Authors of PR

This design proposal was jointly authored by @Iseratho, @skogler, and @davidfroehlich.


Iseratho:
##  Update on the PR
The PR now contains a reference implementation for the tokenization API. It can be used for simple tokenization and is extensible with new algorithms.

It provides the following features:
- [x] 2 simple tokenization algorithms (i.e., whitespace and ngram)
- [x] 2 output representations (i.e., count and position)
- [x] algorithms are configurable with JSON spec

Notes on design considerations:
- [x] output representation is independent of the tokenizer algorithm
- [x] distributable function, as it does not create a dictionary for the tokens (tokens can be encoded with `transformencode`)
- [x] API similar to transform functions (e.g., using JSON spec)

There is still an open issue regarding correctly setting the DataCharacterstics for Spark execution.

Iseratho:
### Consideration before merging the PR

When representing the tokens in long-format (i.e., a transformation that expands on rows (rows: n, maxTokens: m, idCols: k) -> (m*n, k+2)), I get the message in a follow-up `transformencode`: 
> Job aborted due to stage failure: Task 0 in stage 10.0 failed 1 times, most recent failure: Lost task 0.0 in stage 10.0 (TID 18, localhost, executor driver): org.apache.sysds.runtime.DMLRuntimeException: Number of non-zeros mismatch on merge disjoint (target=1000x4, nnz target=4000, nnz source=3992)

Unfortunately, I have not been able to fix this bug since it does not occur in the `tokenize` itself.
However, I have since implemented a wide-format (i.e., a transformation that expands on columns (rows: n, maxTokens: m, idCols: k) -> (n, m+k)), where I could not reproduce the issue. The current state of the PR uses this format in the test cases and passes all checks. 

My specific questions are:
1.	Does anyone know what the issue could be or how it could be fixed?
2.	Conversely, why does the issue not occur on the wide-format? (I want to ensure that the code indeed works and not just hides the error)
3.	Should I drop the support for the long-format to circumvent the issue?

Shafaq-Siddiqi:
LGTM.
Thank you for your contribution. During the merge, I have just fixed some minor formatting and have added a TODO for the assertion tests in the test suite.

PR #1170, opened 2021-01-24
[SYSTEMDS-2747] Federated Als-CG Test

ywcb00:
Hi,
This is a PR for adding federated Als-CG as a first algorithm-level test for federated quaternary operations.
I also added a branch inside FEDInstructionUtils for rewriting BinarySP instructions into BinaryFED instructions in order to support the "fed_!=" instruction needed inside the als-CG algorithm. Since I haven't found a test for these federated operations, I added a test for MatrixScalar and MatrixMatrix federated logical operations.
I made a small fix (I hope it was a fix xD) inside BinaryMatrixMatrixFEDInstruction by changing a broadcast() to a broadcastSliced().
Hope that was not too much, otherwise just revert the last commit (bce7fdc) and the AlsCG test is using "sp_!=" for spark execution mode again.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryMatrixMatrixFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryMatrixMatrixFEDInstruction.java
@@ -39,15 +39,15 @@ public class BinaryMatrixMatrixFEDInstruction extends BinaryFEDInstruction
 	public void processInstruction(ExecutionContext ec) {
 		MatrixObject mo1 = ec.getMatrixObject(input1);
 		MatrixObject mo2 = ec.getMatrixObject(input2);
-
+		
 		//canonicalization for federated lhs
-		if( !mo1.isFederated() && mo2.isFederated()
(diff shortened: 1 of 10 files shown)


ywcb00:
I've made a new commit (8c69017) to revert the changes of BinaryMatrixMatrixFEDInstruction.java because some federated algorithm tests were failing. In addition, I ignored the MatrixMatrix tests of the new FederatedLogicalTest (not working without the previous change) while still keeping the MatrixScalar tests (still working).

ywcb00:
Found my mistake - we need to distinguish between column vector (normal broadcast) and matrix (broadcast sliced if row partitioned) in BinaryMatrixMatrixFEDInstruction.java.
Commit 8059c5e

mboehm7:
LGTM - thanks @ywcb00. I just disabled the spark tests for ALS because forced spark exec mode hurts here in terms of execution (repeated lazy evaluation as the no intermediates are materialized). 

As we discussed offline, please go ahead and do an additional cleanup over the federated binary element-wise operations to clearly differentiate between row and column partitioned federated data for matrix-row vector, matrix-column vector, and matrix-matrix. Thanks.

PR #1171, opened 2021-01-25
[SYSTEMDS-2810] Encoder and decoder serialization

OlgaOvcharenko:
This PR adds Encoder and Decoder serialization. Added PR to run all tests.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko, this is awesome and a great improvement for both, the spark and federated backends. During the merge I made some changes through to avoid unnecessary overhead and work correctly with binning: 

* Removed reflection, unnecessary type handling, and moved the object creation into the respective factory classes.
* Fixed the binning serialization/deserialization to allow a variable number of bins per encoded column, and work in scenarios where the bin boundaries are not yet allocated.
* Removed unused serialization test, and test data set.
* Fixed warnings (e.g., type unsafe iteration over hash maps, unnecessary casts)

PR #1172, opened 2021-01-25
[DIA] Builtin KNN

ywcb00:
Hi,
This is a PR for adding the knn algorithm as builtin function. There are two new builtin functions, namely "knn" and "knnbf".
In order to compare the prediction matrix, we've added a new comparison method to the TestUtils.
Note that the "knn" builtin is also based on a bruteforce approach.
The reference R script requires two packages ("FNN" and "class") which are installed inside the RScript.

Thanks for review :)


mboehm7:
LGTM - this is a great addition. Thanks @ywcb00  @ukw01 @Metka Batič. During the merge I did a few cleanups: (1) removed warnings (imports, unused variables), (2) fixed the script formatting (mix of 1/2/3/4 space indentation), (3) fixed the R script (removed R package installation, moved to `installDependencies.R`, and ignored test for now as we have to update the docker image first), and (4) vectorized `calculateDistance` function in knnbf.

Furthermore, I also folded a minor change of our github action workflows into this patch (to balance the test suites a bit).

mboehm7:
Metka Batič, please also associate the email you used in your commit with your github handle otherwise this contribution is not linked properly.

PR #1174, opened 2021-01-28
[SYSTEMDS-2812] Conv1D

sebwrede:
Initial implementation of Conv1D using the existing Conv2D implementation. 

Conv1D is implemented in DML with three functions: 
- forward, represents the forward pass (equivalent to the conv2d builtin)
- backward_data, represents the backward pass regarding the input data (equivalent to conv2d_backward_data)
- backward_filter, represents the backward pass regarding the filter (equivalent to conv2d_backward_filter)

The tests are loosely based on the Conv2D tests, but the Conv1D tests do not use R for comparison.

### DIFF
--- a/scripts/nn/layers/conv1d.dml
+++ b/scripts/nn/layers/conv1d.dml
@@ -21,7 +21,27 @@
 
 source("scripts/nn/util.dml") as util
 
-forward = function(matrix[double] input, matrix[double] filter, int pad, int stride, int numImg, int numChannels, int imgSize, int numFilters, int filterSize, matrix[double] bias)
+forward = function(matrix[double] input, matrix[double] filter, int pad, int stride, int numInput, int numChannels, int inputWidth, int numFilters, int filterSize, matrix[double] bias)
     return (matrix[double] out) {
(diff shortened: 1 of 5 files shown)


Baunsgaard:
LGTM,

do you think that a dedicated internal operation for this would be good / maybe even going to a ConvND ?

sebwrede:
> do you think that a dedicated internal operation for this would be good / maybe even going to a ConvND ?
Yes, this could be an additional feature to implement. The goal of this PR is to have a basic implementation of Conv1D. Later we can make a builtin version of it and see how efficient that is compared to the DML-version.

mboehm7:
LGTM - thanks for the extension @sebwrede, this is great. I think it's the right step to first call the existing conv2d (for which internal Java/C++/GPU operators with dedicated implementations for combinations of dense/sparse inputs exist). If we measure substantial overhead of conv1d via conv2d then we could go ahead and add the different kernels.

Btw, I see that the conv2d / pooling layers might be confusing as _builtin refers to wrappers around natively supported operators, while in scripts/builtin we have dml-bodied functions. Perhaps we should remove the script-based conv2d and pooling layers altogether.

PR #1175, opened 2021-01-31
[SYSTEMDS-2747] Federated PNMF Test

ywcb00:
Hi,
This is a PR for adding federated PNMF as an algorithm-level test for federated quaternary operations.
I also added checks for row partitioned federated data in the federated quaternary instructions.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWCeMMFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWCeMMFEDInstruction.java
@@ -67,51 +67,53 @@ public class QuaternaryWCeMMFEDInstruction extends QuaternaryFEDInstruction
 				new DoubleObject(ec.getMatrixInput(_input4.getName()).quickGetValue(0, 0));
 		}
 
-		if(!(X.isFederated(FType.ROW) && !U.isFederated() && !V.isFederated()))
+		if(X.isFederated(FType.ROW) && !U.isFederated() && !V.isFederated()) {
+			FederationMap fedMap = X.getFedMapping();
+			FederatedRequest[] fr1 = fedMap.broadcastSliced(U, false);
+			FederatedRequest fr2 = fedMap.broadcast(V);
+			FederatedRequest fr3 = null;
+			FederatedRequest frComp = null;
+
+			// broadcast the scalar epsilon if there are four inputs
+			if(eps != null) {
+				fr3 = fedMap.broadcast(eps);
+				// change the is_literal flag from true to false because when broadcasted it is no literal anymore
+				instString = instString.replace("true", "false");
+				frComp = FederationUtils.callInstruction(instString, output,
+					new CPOperand[]{input1, input2, input3, _input4},
+					new long[]{fedMap.getID(), fr1[0].getID(), fr2.getID(), fr3.getID()});
+			}
+			else {
+				frComp = FederationUtils.callInstruction(instString, output,
+				new CPOperand[]{input1, input2, input3},
+				new long[]{fedMap.getID(), fr1[0].getID(), fr2.getID()});
+			}
+
+			FederatedRequest frGet = new FederatedRequest(RequestType.GET_VAR, frComp.getID());
+			FederatedRequest frClean1 = fedMap.cleanup(getTID(), frComp.getID());
+			FederatedRequest frClean2 = fedMap.cleanup(getTID(), fr1[0].getID());
+			FederatedRequest frClean3 = fedMap.cleanup(getTID(), fr2.getID());
+
+			Future<FederatedResponse>[] response;
+			if(fr3 != null) {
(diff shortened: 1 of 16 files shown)


src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWCeMMFEDInstruction.java
```diff
@@ -59,17 +60,17 @@ public void processInstruction(ExecutionContext ec)
...
-		
+
 		if(qop.hasFourInputs()) {
 			eps = (_input4.getDataType() == DataType.SCALAR) ?
 				ec.getScalarInput(_input4) :
 				new DoubleObject(ec.getMatrixInput(_input4.getName()).quickGetValue(0, 0));
 		}
 
-		if(!(X.isFederated() && !U.isFederated() && !V.isFederated()))
+		if(!(X.isFederated(FType.ROW) && !U.isFederated() && !V.isFederated()))
```
sebwrede, line 70: Could we convert this to a positive case instead? Something more like: 
`if (X.isFederated(FType.ROW) && !U.isFederated() && !V.isFederated()){
//process instruction
}
else {
//throw exception
}`
I think this would be easier to read and leave the method open for new cases in the future, for instance to support col federation. 
This comment also applies to QuaternaryWSLossFEDInstruction and QuaternaryWSigmoidFEDInstruction.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedPNMFTest.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@BeforeClass
+	public static void init() {
+		TestUtils.clearDirectory(TEST_DATA_DIR + TEST_CLASS_DIR);
+	}
+
+	@Test
+	public void federatedPNMFSingleNode() {
+		federatedPNMF(TEST_NAME, ExecMode.SINGLE_NODE);
```
sebwrede, line 83: You don't need to pass TEST_NAME as a parameter.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedPNMFTest.java
```diff
@@ -0,0 +1,156 @@
...
+	@Parameterized.Parameter()
+	public int rows;
+	@Parameterized.Parameter(1)
+	public int cols;
+	@Parameterized.Parameter(2)
+	public int rank;
+	@Parameterized.Parameter(3)
+	public int max_iter;
+	@Parameterized.Parameter(4)
+	public double sparsity;
```
sebwrede, line 60: Where is the sparsity parameter used?

sebwrede (commented):
It looks good. I only have a few comments, mostly related to code style.

src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWCeMMFEDInstruction.java
ywcb00: I converted it to the positive case for all five federated quaternary operations.

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedPNMFTest.java
ywcb00: Careless mistake, thanks for finding it :+1: 
It's used for generating the random matrices now

ywcb00:
Thanks for the comments @sebwrede. I changed it properly :)

sebwrede:
LGTM

src/test/java/org/apache/sysds/test/functions/federated/algorithms/FederatedPNMFTest.java
```diff
@@ -0,0 +1,156 @@
...
+{
+	private final static String TEST_NAME = "FederatedPNMFTest";
+	private final static String TEST_DIR = "functions/federated/";
+	private final static String TEST_CLASS_DIR = TEST_DIR + FederatedPNMFTest.class.getSimpleName() + "/";
+
+	private final static String OUTPUT_NAME = "Z";
+	private final static double TOLERANCE = 0.2;
+	private final static int blocksize = 1024;
+
+	@Parameterized.Parameter()
```
aimiebell, line 51: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

mboehm7:
LGTM - during the merge, I just slightly modified some formatting including the license of the new test, which had tabs between the different sentences.

PR #1176, opened 2021-02-02
[SYSTEMDS-2550] Epoch Timing

tobiasrieger:
This PR adds time to epoch to info log for the parameter server.

Minor:
- Implemented the validation for the local parameter server
- Fixes an error where the local parameter server would log at the wrong time

### DIFF
--- a/conf/log4j.properties
+++ b/conf/log4j.properties
@@ -21,10 +21,6 @@
 
 log4j.rootLogger=ERROR,console
 
-# Federated Parameterserver testing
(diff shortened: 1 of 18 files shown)


tobiasrieger:
- Fixed a bug in the creation (would crash if no validation was specified - sanitization issue)
- Refactored weighing to weighting for the whole parameter server

mboehm7:
LGTM - thanks @tobiasrieger. I just added a TODO to remind us to double check if the statistics maintenance correctly works in case of multiple, potentially concurrent paramserv invocations.

PR #1178, opened 2021-02-04
[SYSTEMDS-2799] Update UDF reuse, add Fed pipeline reuse test

phaniarnab:
This patch handles reusing those UDFs that return metadata
with federated SUCCESS response (e.g. Rdiag, DiagMatrix).
Furthermore, this adds a test to reuse a full federated
pipeline having tasks such as preprocessing and
hyperparameter tuning (LM).


phaniarnab:
To run tests.

PR #1179, opened 2021-02-04
Multiplogreg compressed support again!

Baunsgaard:
The changes contained in this PR, makes it possible to execute multilogreg in compressed space!
the super cool thing in multilogreg is that it does not decompress the input if icpt is set to 0 !

except for the binary matrix matrix operations where each cell is multiplied with a value.

* This commit fixes the compression tests that sometimes break while executing parallel


Baunsgaard:
merged

PR #1182, opened 2021-02-11
[SYSTEMDS-XXXX] Federated Binary Element-wise Operations

ywcb00:
Hi,
This PR includes a cleanup over the FederatedBinaryMatrixMatrixInstruction in order to support all the edge cases and also column partitioned federated input. I've extended the FederatedLogicalTest to test all the edge cases.

Note: The test class includes also tests for full partitioned federated input (outcommented because we do not support it yet) but I would keep the functionality of testing full partitioned input for eventual future improvements where we add support for it.

Thanks for review :)

PS: There are a lot of different data configurations inside the test, most of them outcommented because of the test performance, but all of them are facing different edge cases of federated binary operations.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryFEDInstruction.java
@@ -20,6 +20,9 @@
 package org.apache.sysds.runtime.instructions.fed;
 
 import org.apache.sysds.common.Types.DataType;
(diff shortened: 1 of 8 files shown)


mboehm7:
LGTM - thanks @ywcb00. I only reduce some of the redundant tests (not every configuration needs to test all comparison operators is is done elsewhere already). Furthermore, a week ago I already fixed the handling of FULL federated matrices - so it should automatically apply for either row or col branches. Maybe if you get a chance you could double check the full handling/exceptions are still needed.

PR #1184, opened 2021-02-16
[SYSTEMDS-2863] Federated ctable instruction

OlgaOvcharenko:
This PR adds federated ctable instruction. In the general case, the output should be MatrixBlock, but for slicefinder it is federated (one-hot encoded fed input).


src/main/java/org/apache/sysds/runtime/instructions/fed/CtableFEDInstruction.java
```diff
@@ -0,0 +1,331 @@
...
+			}
+			catch(Exception e) {
+				e.printStackTrace();
+			}
+		}
+		return resultBlock;
+	}
+
+	private Long[] getOutputDimension(ExecutionContext ec, CPOperand inOp, CPOperand outOp, FederatedRange[] federatedRanges) {
+		MatrixObject in = ec.getMatrixObject(inOp);
```
sebwrede, line 264: You already loaded this input matrix in the processInstruction method. Why don't you just pass that object to this method instead of loading it again?

src/main/java/org/apache/sysds/runtime/instructions/fed/CtableFEDInstruction.java
```diff
@@ -0,0 +1,331 @@
...
+			return fedDims;
+		}
+
+		// construct new instString
+		String maxInstrString = instString.replace("ctable", "uamax");
+		String[] instParts = maxInstrString.split(Lop.OPERAND_DELIMITOR);
+		String[] maxInstParts = new String[] {instParts[0], instParts[1],
+			InstructionUtils.concatOperandParts(inOp.getName(), DataType.MATRIX.name(), (ValueType.FP64).name()),
+			InstructionUtils.concatOperandParts(outOp.getName(), DataType.SCALAR.name(), (ValueType.FP64).name()), "16"};
+		maxInstrString = String.join(Lop.OPERAND_DELIMITOR, maxInstParts);
```
sebwrede, line 284: It improves readability if something like this is put into a separate method with the name you have written in the comment. It would make this method shorter and it would be easier to see which steps the method consists of.

src/main/java/org/apache/sysds/runtime/instructions/fed/CtableFEDInstruction.java
```diff
@@ -0,0 +1,331 @@
...
+
+		// get max per worker
+		FederationMap map = in.getFedMapping();
+		FederatedRequest fr1 = FederationUtils.callInstruction(maxInstrString, outOp,
+			new CPOperand[]{inOp}, new long[]{in.getFedMapping().getID()});
+		FederatedRequest fr2 = new FederatedRequest(FederatedRequest.RequestType.GET_VAR, fr1.getID());
+		FederatedRequest fr3 = map.cleanup(getTID(), fr1.getID());
+		Future<FederatedResponse>[] tmp = map.execute(getTID(), fr1, fr2, fr3);
```
sebwrede, line 292: Same as the comment above.

src/main/java/org/apache/sysds/runtime/instructions/fed/CtableFEDInstruction.java
```diff
@@ -0,0 +1,331 @@
...
+
+		// get result
+		for(int i = 0; i < tmp.length; i ++) {
+			try {
+				fedDims[i] = ((ScalarObject) tmp[i].get().getData()[0]).getLongValue();
+			}
+			catch(Exception e) {
+				e.printStackTrace();
+			}
+		}
```
sebwrede, line 302: Same as the comment above.

src/main/java/org/apache/sysds/runtime/instructions/fed/CtableFEDInstruction.java
```diff
@@ -0,0 +1,331 @@
...
+				fr2 = FederationUtils.callInstruction(instString, output, new CPOperand[] {input1, input2, input3}, new long[] {fr1[0].getID(), fr4[0].getID(), mo1.getFedMapping().getID()});
+
+			fr3 = new FederatedRequest(FederatedRequest.RequestType.GET_VAR, fr2.getID());
+			FederatedRequest fr5 = mo1.getFedMapping().cleanup(getTID(), fr1[0].getID(), fr4[0].getID());
+			ffr = mo1.getFedMapping().execute(getTID(), true, fr1, fr2, fr3, fr4[0], fr5);
+		}
+
+		long fedSize = Collections.max(Arrays.asList(dims1), Long::compare) / ffr.length;
+
+		if(fedOutput && isFedOutput(ffr, fedSize)) {
```
sebwrede, line 159: Why do you calculate fedOutput before calling this method instead of including it in the "isFedOutput" method?

sebwrede (commented):
I think this looks good. 
I have added some comments and will add more comments when I have time to read the rest of the code.

src/test/java/org/apache/sysds/test/functions/federated/primitives/FederatedCtableTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+	@Test
+	public void federatedCtableSinglenode() {
+		runCtable(Types.ExecMode.SINGLE_NODE, false);
+	}
+
+	@Test
+	public void federatedCtableFedOutputSinglenode() {
+		runCtable(Types.ExecMode.SINGLE_NODE, true);
+	}
```
sebwrede, line 76: I think we could use more test cases. 
How about a test case where Y is federated and X is not? 
What about a third input? (especially something that triggers the reversedWeights-case in the processInstruction method)
It is also relevant with some invalid input to see if it actually throws an exception or if it computes meaningless output (for instance invalid dimensionality of the input). 
I think it is also very relevant to test with different federated range overlaps, since the slicing is an important component of the CTable instruction.

src/test/scripts/functions/federated/FederatedCtableFedOutput.dml
```diff
@@ -0,0 +1,38 @@
...
+    		list(2*$rows/4, 0), list(3*$rows/4, $cols), list(3*$rows/4, 0), list($rows, $cols)));
+
+m = nrow(X);
+n = ncol(X);
+
+# prepare offset vectors and one-hot encoded X
+fdom = colMaxs(X);
+
+foffb = t(cumsum(t(fdom))) - fdom;
+foffe = t(cumsum(t(fdom)))
```
sebwrede, line 33: Out of curiosity: what are _fdom_, _foffb_, and _foffe_ abbreviations of?

sebwrede:
As I said a couple of weeks ago, this PR looks really good. I think we need to do two things before merging: 

1. Add more test cases. I looked through the processInstruction code, but I find it difficult to convince myself that the code works if the test cases do not cover more process cases. 
2. Resolve the conflicting files.

src/test/scripts/functions/federated/FederatedCtableFedOutput.dml
OlgaOvcharenko: Copy paste from slicefinder.dml where this test case appeared, renamed now.

src/test/java/org/apache/sysds/test/functions/federated/primitives/FederatedCtableTest.java
OlgaOvcharenko: I added more tests now and fixed reversed weights.

sebwrede:
I think the test coverage is improving, but there are still some gaps that would be nice to cover. It is not because I believe there is anything wrong with the parts that are not covered, but I think it is important to cover so that we have the tests later on when we make changes to the federated instructions. 
For instance, as far as I can see, the `isFedOutput` method is never called. This is a significant part of the code which is never activated. 

Besides that, I think this PR looks good. The failing test is caused by a sparkDriver error from the master branch, so this PR seems ready to merge.

PR #1185, opened 2021-02-16
[SYSTEMDS 2582,2583] Reuse of compressed lineage DAGs

phaniarnab:
This patch extends hash and equal of lineage items to support
dedup nodes -- which also enables reuse with deduplication.
This patch brings a substantial amount of changes in deduplication
and lineage DAG comparison logic to support all the possible
cases. Even though all the basic tests are passing on this patch, we
need extensive experiments to confirm the robustness of this change.


phaniarnab:
PR to run tests.

PR #1187, opened 2021-02-21
Fixed MVIImpute mode_global crash if most values are missing

ilovemesomeramen:
MVIImpute mode_global failed if most values are missing since check was performed on null and not "null". Resulting in the string "null" being passed to Double.parseDouble in the apply step later.


phaniarnab:
Thanks for the fix.

PR #1190, opened 2021-02-24
[MINOR] null was not checked in EncoderMVImpute for Global mean

ilovemesomeramen:
There was no null check so the replacement was always "NaN" since objectToDouble returns NaN if null is given, resulting in the mean always being NaN if a value is missing.

Also added check in Test so the column is checked for NaNs


corepointer (approved):

corepointer:
LGTM - Thanks for the fixes in this PR. I merged it with minor formatting changes.

PR #1199, opened 2021-03-12
[SYSTEMDS-2604] Federated compile and federated output propagation

sebwrede:
This PR adds compilation of federated instructions when `OptimizerUtils.FEDERATED_COMPILATION` is set to true. 
Additionally, a federated output flag is added and propagated through HOPs, LOPs, and instructions. The federated output flag will later be used to change the processing of federated instructions so that the output of federated instructions are not retrieved before all subsequent federated instructions have been executed. 
The federated compilation and propagation of the federated output flag will be extended to other operations in future PRs. 

Several design decisions have been made in this PR, for instance the decision to add the federated output flag as an additional operand in the instruction string. It could also have been added as a property of the output operand, but this would have consequences on how CPOperands are handled in the system. Comments on this or new ideas are appreciated.


Baunsgaard:
did you fix some bugs since it parses all tests?

sebwrede:
> did you fix some bugs since it parses all tests?

I removed an empty file, which made one test fail since it did not have an Apache license. 
I also ran the failing tests locally, where I discovered that I needed to install some R packages and some Python packages. One of the R packages, imputeTS, failed to install when I ran our R install script, so I had to look up why this happened. I found a solution where I had to run the following: 
```
sudo apt-get install aptitude
sudo apt-get install libcurl4-openssl-dev
sudo apt-get install libxml2-dev
```
This idea came from here: https://github.com/sagemath/cloud/issues/114#issuecomment-86363054.
After installing the above, I ran the R install again and then the tests passed locally. I then committed and pushed. The only update in that commit was the removal of the empty file.

Baunsgaard:
> > did you fix some bugs since it parses all tests?
> 
> I removed an empty file, which made one test fail since it did not have an Apache license.
> I also ran the failing tests locally, where I discovered that I needed to install some R packages and some Python packages. One of the R packages, imputeTS, failed to install when I ran our R install script, so I had to look up why this happened. I found a solution where I had to run the following:
> 
> ```
> sudo apt-get install aptitude
> sudo apt-get install libcurl4-openssl-dev
> sudo apt-get install libxml2-dev
> ```
> 
> This idea came from here: [sagemath/cloud#114 (comment)](https://github.com/sagemath/cloud/issues/114#issuecomment-86363054).
> After installing the above, I ran the R install again and then the tests passed locally. I then committed and pushed. The only update in that commit was the removal of the empty file.

okay, i will try to reinstall my r because i have tests that fail in GitHub actions and not locally...

src/main/java/org/apache/sysds/hops/BinaryOp.java
```diff
@@ -496,7 +498,7 @@ else if (mbin == MMBinaryMethod.MR_BINARY_M) {
 				setOutputDimensions(binary);
 				setLineNumbers(binary);
 				setLops(binary);
-			}
+			} else throw new HopsException("Lop construction not implemented for ExecType " + et);
```
Baunsgaard, line 501: split line

src/main/java/org/apache/sysds/hops/ReorgOp.java
```diff
@@ -76,6 +77,8 @@ public ReorgOp(String l, DataType dt, ValueType vt, ReOrgOp o, ArrayList<Hop> in
 			getInput().add(i, in);
 			in.getParent().add(this);
 		}
+
+		updateETFed();
```
Baunsgaard, line 81: double updateETFed here maybe only one needed?

src/main/java/org/apache/sysds/runtime/instructions/FEDInstructionParser.java
```diff
@@ -32,8 +36,28 @@
...
+		String2FEDInstructionType.put("fedinit"  , FEDType.Init);
+		String2FEDInstructionType.put("tsmm"     , FEDType.Tsmm);
+		String2FEDInstructionType.put("ba+*"     , FEDType.AggregateBinary);
+
+		String2FEDInstructionType.put("uak+"     , FEDType.AggregateUnary);
+		String2FEDInstructionType.put( "uark+"   , FEDType.AggregateUnary);
+		String2FEDInstructionType.put( "uack+"   , FEDType.AggregateUnary);
+		String2FEDInstructionType.put( "uasqk+"  , FEDType.AggregateUnary);
+		String2FEDInstructionType.put( "uarsqk+" , FEDType.AggregateUnary);
+		String2FEDInstructionType.put( "uacsqk+" , FEDType.AggregateUnary);
```
Baunsgaard, line 48: inlining

src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryFEDInstruction.java
```diff
@@ -42,11 +47,12 @@ public static BinaryFEDInstruction parseInstruction(String str) {
 		}
 
 		String[] parts = InstructionUtils.getInstructionPartsWithValueType(str);
-		InstructionUtils.checkNumFields(parts, 3, 4);
+		InstructionUtils.checkNumFields(parts, 3, 4, 5);
 		String opcode = parts[0];
 		CPOperand in1 = new CPOperand(parts[1]);
 		CPOperand in2 = new CPOperand(parts[2]);
 		CPOperand out = new CPOperand(parts[3]);
+		boolean federatedOutput = parts.length > 5 && Boolean.parseBoolean(parts[5]);
```
Baunsgaard, line 55: does this work? should it not be index 4? or is that the number of threads?

Baunsgaard (commented):
LGTM, only minor syntax comments

src/main/java/org/apache/sysds/hops/BinaryOp.java
sebwrede: Why? It is not exceeding the line character limit.

Baunsgaard: in this pr you do it two different ways, i liked the first better with newline and brackets. I personally would prefer if there was a newline for the throw, if you want to include brackets or not is up to you.

src/main/java/org/apache/sysds/hops/ReorgOp.java
sebwrede: It is important to have both. In the first one, the execution type and fedoutput will be set during the construction of the Hops (in the order input to output). In the second call in "optFindExecType", it will set the exectype so that a once federated execution is not converted to a CP. This will happen in the order output to input (plus additional calls in certain cases).

src/main/java/org/apache/sysds/runtime/instructions/FEDInstructionParser.java
sebwrede: What do you mean?

Baunsgaard: look at the code, some of the lines you insert spaces, some you don't.

src/main/java/org/apache/sysds/runtime/instructions/fed/BinaryFEDInstruction.java
sebwrede: Number 4 is the number of threads. I assume that this has been included, otherwise we cannot parse a federated output. I try to only write federated output in the cases where it is true, but I leave room for also writing false to the instruction string. This is the reason why I check if it has been included and parse it if it has been included.

src/main/java/org/apache/sysds/runtime/instructions/FEDInstructionParser.java
sebwrede: I added a new commit with the two spaces added. If you use an editor like IntelliJ, this sort of thing is aligned even if a single space is missing.

src/main/java/org/apache/sysds/hops/BinaryOp.java
sebwrede: I have now changed the other throw so that it is also on the same line. 
I generally think that it is acceptable to have both, since it makes more sense in some cases to have it on a single line. It depends on the context. If, for instance, the other if-else parts are without brackets then it makes more sense to have the throw on a single line without brackets as well.

phaniarnab:
LGTM.
For my understanding, did you mean that this flag will help to execute a chain of federated operations before fetching the outputs to the coordinator? Is it not the case today?

sebwrede:
> For my understanding, did you mean that this flag will help to execute a chain of federated operations before fetching the outputs to the coordinator? Is it not the case today?

This PR adds a flag to compile the federated instructions during compile-time instead of simply converting the CP instructions to FED instructions during runtime. Later, the decision to compile to FED will be expanded so that it can base the decision on several scenarios, including the scenario where a chain of federated operations can be executed without sharing the data with the coordinator. Many of the FED instructions are currently implemented in a way that keeps the output federated, but this decision of keeping the output federated needs to be dependent on the flag so that we can have some kind of "FED planner" which compiles this information. 
At the moment, the federated instructions do not take the privacy constraints into account. When we have a FED planner, it could discover a privacy constraint on some input data and then compile the instructions accordingly. This may in some cases mean that it does not just blindly pick the chain of federated instructions, but instead considers the privacy constraint and makes a new plan that adheres to the constraint. This is just an example of how the FED planner will be convenient later on. 
I implemented this as a federated output flag to begin with, but this will likely be expanded later on since there are more decisions than just federated vs. local. I will deal with this in later PRs when I have a basic FED planner and support federated compilation of more operations.

phaniarnab:
> > For my understanding, did you mean that this flag will help to execute a chain of federated operations before fetching the outputs to the coordinator? Is it not the case today?
> 
> This PR adds a flag to compile the federated instructions during compile-time instead of simply converting the CP instructions to FED instructions during runtime. Later, the decision to compile to FED will be expanded so that it can base the decision on several scenarios, including the scenario where a chain of federated operations can be executed without sharing the data with the coordinator. Many of the FED instructions are currently implemented in a way that keeps the output federated, but this decision of keeping the output federated needs to be dependent on the flag so that we can have some kind of "FED planner" which compiles this information.
> At the moment, the federated instructions do not take the privacy constraints into account. When we have a FED planner, it could discover a privacy constraint on some input data and then compile the instructions accordingly. This may in some cases mean that it does not just blindly pick the chain of federated instructions, but instead considers the privacy constraint and makes a new plan that adheres to the constraint. This is just an example of how the FED planner will be convenient later on.
> I implemented this as a federated output flag to begin with, but this will likely be expanded later on since there are more decisions than just federated vs. local. I will deal with this in later PRs when I have a basic FED planner and support federated compilation of more operations.

Got it. Thanks for explaining in detail. Pushing the decision to the compiler seems to be a good idea. Regarding the chain of OPs, many today need partial output aggregation though I guess.

sebwrede:
> Regarding the chain of OPs, many today need partial output aggregation though I guess.

Yes, this is true. There will be a lot of work with implementing the processing of the federated instructions.

PR #1200, opened 2021-03-12
[MINOR] CSV parsing optimization byparsing hashfunction

Baunsgaard:


Baunsgaard:
Before these changes the parsing of MNIST took ~1.2 sec because of inefficient parsing of strings and NaN special handling.

Unfortunately this PRs implementation is not as fast as SystemML 1.0 which is able to parse MNIST
in 0.8 sec on average (i did verify with repeated executions of various kinds). I did try to use exactly the old code but in our
current system this only perform to 1 sec.


Current:
image

image

Old:
image

image

Baunsgaard:
Reading times:
Census  SystemML 3.2 sec now 3.1 sec
CovType SystemML 1.0 sec now 0.8 sec
Airlines SystemML 9.8 sec now 8.3 sec
mnist SystemML 0.78 sec now 0.81 sec
infimnist_1m SystemML 14 sec now 12.8

PR #1202, opened 2021-03-17
[SYSTEMDS-2883] CLA Specialized pointer using Offsetlist 

Baunsgaard:
This commit contain a offset list encoding interface and byte and char implementations that allow for effecient offset lists for SDC column groups.

future work include encoding in odd number bits like 2 4 and 6.


Baunsgaard:
merged

PR #1210, opened 2021-03-24
[MINOR] Update code formatter

Baunsgaard:
This commit update the code formatter settings to the latest 21 version.
And change the settings to not insert many newlines in function calls if
the parameters make the line exceed the 120 characters.

### DIFF
--- a/dev/CodeStyle_eclipse.xml
+++ b/dev/CodeStyle_eclipse.xml
@@ -16,351 +16,393 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
--->
-<profiles version="16">
-<profile kind="CodeFormatterProfile" name="SystemDS" version="16">
-<setting id="org.eclipse.jdt.core.formatter.insert_space_after_ellipsis" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_enum_declarations" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_allocation_expression" value="do not insert"/>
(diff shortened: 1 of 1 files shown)


phaniarnab:
Curious, why not break a long function call into multiple lines? In my opinion, it is much more readable than scrolling right. And sticking to 120 characters is generally a good idea provided all the different setups developers use.

Baunsgaard:
> Curious, why not break a long function call into multiple lines? In my opinion, it is much more readable than scrolling right. And sticking to 120 characters is generally a good idea provided all the different setups developers use.

It is to avoid something like this:
image

i would rather have this:
image

phaniarnab:
Okay. I would prefer something in between, 4 parameters in each line. I guess, it is up to personal preferences. 
What number of characters do you increase to for function calls in this PR?

phaniarnab:
> > Curious, why not break a long function call into multiple lines? In my opinion, it is much more readable than scrolling right. And sticking to 120 characters is generally a good idea provided all the different setups developers use.
> 
> It is to avoid something like this:
> image
> 
> i would rather have this:
> image

Even your preferred version is under 120? Was there a bug in the template?

Baunsgaard:
> Okay. I would prefer something in between, 4 parameters in each line. I guess, it is up to personal preferences.
> What number of characters do you increase to for function calls in this PR?

unlimited, until it hit the max characters of 120
i did not find a setting to enable saying a specific number of parameters per line.

phaniarnab:
> > Okay. I would prefer something in between, 4 parameters in each line. I guess, it is up to personal preferences.
> > What number of characters do you increase to for function calls in this PR?
> 
> unlimited, until it hit the max characters of 120
> i did not find a setting to enable saying a specific number of parameters per line.

I am fine as long as the line stays within 120.

PR #1212, opened 2021-03-25
[SYSTEMDS-2901] Refactor encoders

phaniarnab:
This patch refactors the transform encoder code to operate on per column basis.
This is the first step towards the multi-threaded encoders.

Note: This patch disables lineage-based reuse of federated encoders test.
Need a thorough debugging to understand what is broken in there.


phaniarnab:
A fresh to run tests

PR #1213, opened 2021-03-25
[SYSTEMDS-2914] maxpooling_backward sparse

Baunsgaard:
This PR contain a few changes.

Most notably Backward max pooling goes from 15 sec to 1 sec on MNIST, by exploiting that outputs are known to be sparse in the kernel size of the pooling. Meaning if the kernel is 2x2x2 the number of output nnz is guaranteed to be 1/8 number of cells. Before this change every invocation of max pooling allocated a dense matrix, that after execution was reallocated as sparse.

End to end performance is decreased to 140 from 188 sec.

Other items:

- Force dense matrices to encourage native calls of Conv2d, conv2d backward filter, conv2d backward data. This is done because the current implementations supporting sparse is slower than forcing the inputs into dense and then calling natively. we gain 10 seconds here
- LastRowDenseGeneric is optimized for dense aggregation, instead of using quick get and set this give 5 seconds.
- Remove examine sparsity in the Generic aggregation, since this in some executions change the aggregating matrix from dense to sparse, just to in the next invocation change back to dense.
- Remove unused statistics in DNN (no extra performance)


Baunsgaard:
Overall for two epoch Mnist with batch size 128 we go from 254 sec to 195 sec

Baunsgaard:
190 sec

PR #1214, opened 2021-03-26
[SYSTEMDS-XXXX] Federated Codegen SpoofCellwise

ywcb00:
Hi,
This is a PR for adding federated code generation for cellwise templates, and the respective tests.
Inside the test class, there are two datasets outcommented. The first one fails because we are fusing the sequence operation into the Spoof class without adjusting the grix parameter of the genexec call to the individual offsets of the federated matrix paritions. The second test fails because it doesn't create a federated instruction of the spoof operator (due to indexing, where the matrix gets copied to the coordinator).
I've implemented a fix for adjusting the grix to correctly include sequences with federated operations in a previous commit (43e9ff2), but it isn't a nice fix (adding a suffix to the instruction string whenever it is a federated spoof operation) so I removed it. Maybe we can exclude the sequence operation from code generation?

Thanks for review :)


corepointer (commented):
Thank you for submitting this PR!
* I have not checked out the code yet for a hands on test (will do). 
* Please tune down your code formatter. There are several files where I did not spot any real change but the whole file is trimmed of trailing blank spaces.
* I've been looking into your bug report. Splitting the statement blocks to prevent fusion of cbind is actually required if you want to produce a cellwise template. The wrong behavior is in producing getValue() in rowwise, which can not work as you already pointed out. This happens if there's a matrix instead of a vector to be added by cbind.

So far so good :)

ywcb00:
Sorry, I forgot to mention that:

- I introduced the new method executeMultipleSlices inside the FederationMap to support the execution with multiple broadcast-sliced inputs. So far we only support the execution with one sliced input, while executing a generated spoof class could include more than one.
- I implicitly removed the case of only one federated worker with a full federated matrix from the BinaryMatrixMatrixFEDInstruction because there is no seperate case needed anymore.

I also added the newline at eof now.

Thanks for the feedback @corepointer :)

mboehm7:
LGTM - thanks @ywcb00. During the merge, I fixed a few warnings, modified the check for federated inputs (to avoid exceptions), and reduced the tests to hybrid only (70s instead of 180s).

PR #1215, opened 2021-04-02
[SYSTEMDS-2915] 2nd follow-up fix to properly remove any references t…

corepointer:
…o GPUObjects that are to be removed


phaniarnab:
Looks good to me.

PR #1218, opened 2021-04-08
[SYSTEMDS-2922] Federated Fused Codegen Operator - Rowwise

ywcb00:
Hi,
This is a PR for adding federated code generation for rowwise templates, and the respective tests (hybrid execution mode to reduce the test time).
Note that I restricted this operation to only row partitioned federated data because column partitioned data would create some major problems within rowwise operations.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
@@ -222,8 +222,9 @@ public class FEDInstructionUtils {
 		else if(inst instanceof SpoofCPInstruction) {
 			SpoofCPInstruction instruction = (SpoofCPInstruction) inst;
 			if((instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class
-					|| instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class)
-				&& instruction.isFederated(ec)) {
+					&& instruction.isFederated(ec))
+				|| (instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class
+					&& instruction.isFederated(ec, FType.ROW))) {
 				fedinst = SpoofFEDInstruction.parseInstruction(instruction.getInstructionString());
 			}
 		}
@@ -329,8 +330,9 @@ public class FEDInstructionUtils {
 		else if(inst instanceof SpoofSPInstruction) {
 			SpoofSPInstruction instruction = (SpoofSPInstruction) inst;
 			if((instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class
-					|| instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class)
-				&& instruction.isFederated(ec)) {
+					&& instruction.isFederated(ec))
+				|| (instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class
+					&& instruction.isFederated(ec, FType.ROW))) {
 				fedinst = SpoofFEDInstruction.parseInstruction(inst.getInstructionString());
 			}
 		}
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
@@ -20,6 +20,7 @@
 package org.apache.sysds.runtime.instructions.fed;
 
 import org.apache.sysds.runtime.codegen.SpoofCellwise;
+import org.apache.sysds.runtime.codegen.SpoofRowwise;
 import org.apache.sysds.runtime.controlprogram.caching.CacheableData;
 import org.apache.sysds.runtime.controlprogram.caching.FrameObject;
 import org.apache.sysds.runtime.controlprogram.caching.MatrixObject;
@@ -220,8 +221,11 @@ public class FEDInstructionUtils {
 		}
 		else if(inst instanceof SpoofCPInstruction) {
 			SpoofCPInstruction instruction = (SpoofCPInstruction) inst;
-			if(instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class && instruction.isFederated(ec))
+			if((instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class
+					|| instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class)
+				&& instruction.isFederated(ec)) {
 				fedinst = SpoofFEDInstruction.parseInstruction(instruction.getInstructionString());
+			}
 		}
 		else if(inst instanceof CtableCPInstruction) {
 			CtableCPInstruction cinst = (CtableCPInstruction) inst;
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
```diff
@@ -220,8 +221,11 @@ else if(inst instanceof QuaternaryCPInstruction) {
 		}
 		else if(inst instanceof SpoofCPInstruction) {
 			SpoofCPInstruction instruction = (SpoofCPInstruction) inst;
-			if(instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class && instruction.isFederated(ec))
+			if((instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class
+					|| instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class)
+				&& instruction.isFederated(ec)) {
```
Baunsgaard, line 226: Do you cover cases where the federated data is split on columns?
If not then this place would be ideal to ad another verification.

src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
```diff
@@ -316,7 +320,9 @@ else if(inst instanceof QuaternarySPInstruction) {
 		}
 		else if(inst instanceof SpoofSPInstruction) {
 			SpoofSPInstruction instruction = (SpoofSPInstruction) inst;
-			if(instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class && instruction.isFederated(ec)) {
+			if((instruction.getOperatorClass().getSuperclass() == SpoofCellwise.class
+					|| instruction.getOperatorClass().getSuperclass() == SpoofRowwise.class)
```
Baunsgaard, line 324: add verification like previous commment

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedCellwiseTmplTest.java
```diff
@@ -147,8 +145,8 @@ private void testFederatedCodegen(ExecMode exec_mode) {
 
 		// generate dataset
 		// matrix handled by two federated workers
-		double[][] X1 = getRandomMatrix(fed_rows, fed_cols, 0, 1, sparsity, 3);
-		double[][] X2 = getRandomMatrix(fed_rows, fed_cols, 0, 1, sparsity, 7);
+		double[][] X1 = getRandomMatrix(fed_rows, fed_cols, 0, 1, 1, 3);
+		double[][] X2 = getRandomMatrix(fed_rows, fed_cols, 0, 1, 1, 7);
```
Baunsgaard, line 149: with this you are saying that the test does not care if the input is sparse or dense, if this is the case then to reduce the running time of these test i suggest making all the inputs sparse rather than dense, perhaps set it to 0.1 ?

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedRowwiseTmplTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+	// @Test
+	// public void federatedCodegenRowwiseSingleNode() {
+	// 	testFederatedCodegenRowwise(ExecMode.SINGLE_NODE);
+	// }
+	// 
+	// @Test
+	// public void federatedCodegenRowwiseSpark() {
+	// 	testFederatedCodegenRowwise(ExecMode.SPARK);
+	// }
```
Baunsgaard, line 102: Have you verified if these work? or just copied them?
if the latter, please try to execute them, and then enable them if it works, if they don't work add an @Ignore clause to them such that they show up as tests that currently are skipped. Or try to fix it so that they work.

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedRowwiseTmplTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+		// compare the results via files
+		HashMap<CellIndex, Double> refResults  = readDMLMatrixFromExpectedDir(OUTPUT_NAME);
+		HashMap<CellIndex, Double> fedResults = readDMLMatrixFromOutputDir(OUTPUT_NAME);
+		TestUtils.compareMatrices(fedResults, refResults, TOLERANCE, "Fed", "Ref");
+
+		TestUtils.shutdownThreads(thread1, thread2);
+
+		// check for federated operations
+		Assert.assertTrue(heavyHittersContainsSubString("fed_spoofRA"));
```
Baunsgaard, line 162: just verifying that it calls the spoof operation is fine, but i would like some verification that the result is correct as well.
If this is not possible in all your tests cases i suggest adding one where you have a known result, that you then verify calculates correctly.

Baunsgaard (commented):
I think it looks fine and probably works correctly, but there are some more verification needed before it is ready to be merged. 

My two main issues are : 
1. Verify that it works with column separated federated data, (net necessarily federated but just pulling the data to master if the implementation does not support it) and
2. Verification of results from execution in tests.

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedRowwiseTmplTest.java
Baunsgaard: nvm did not look correctly at line 157

src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
ywcb00: Column partitioned federated data is not supported within federated rowwise spoof instructions.
You're right, if we want to support column paritioned federated data as well (even without a federated instruction), we need to add a verification here, such that we don't replace the spoof instruction with a federated spoof instruction.
Thanks, I will add this :)

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedCellwiseTmplTest.java
ywcb00: That makes sense for large matrices, but since we also have matrices containing only a few elements I'm afraid of falsifying the test results by changing it. (worst case: all elements of matrix X are 0)
I'm not sure if such a case can happen, can it?

src/test/java/org/apache/sysds/test/functions/federated/codegen/FederatedRowwiseTmplTest.java
ywcb00: Both outcommented tests (SingleNode and Spark) are working, I just disabled them to preserve the test performance. We did the same for the cellwise tests, but I will enable them again.

ywcb00:
Thank you for the feedback @Baunsgaard 
I changed the sparsity to 0.1 and verified that no testcase generates an empty matrix. In addition to that, I've commented out some testcases which would test equal or very similar branches like other testcases to avoid redundant testing and further reduce the test time.
I also added some testcases for column paritioned federated data inside rowwise templates to verify that the matrices get pulled to the driver where the local operation gets executed (because column paritioned federated data is not supported by the federated rowwise spoof instruction itself).

Baunsgaard (commented):
LGTM

mboehm7:
LGTM as well - thanks @ywcb00 for the patch. I only made minor changes related to warnings and moved the partially redundant checks for federated inputs as utility functions into execution context so other instructions can share it.

PR #1225, opened 2021-04-15
[SYSTEMDS-2916] Build python dist with maven package

j143:
We are not pursuing this PR at the moment. The discussion below!

---
  * `mvn clean package -P distribution`
    also builds and outputs the python artifacts in the
    top level target folder

Closes #1225.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -361,6 +361,44 @@
 			</plugin>
 
 			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>exec-maven-plugin</artifactId>
+				<version>1.6.0</version>
+				<executions>
+					<execution>
+						<id>generate-python-package-1of2-pre-setup</id>
+						<phase>package</phase>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+						<configuration>
+							<executable>python</executable>
+							<workingDirectory>src/main/python</workingDirectory>
+							<arguments>
+								<argument>pre_setup.py</argument>
+							</arguments>
+						</configuration>
+					</execution>
+					<execution>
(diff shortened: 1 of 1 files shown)


j143:
cc @Baunsgaard

Baunsgaard:
Hi @j143 

I both like and dislike the idea. 
on one hand it makes sense to have a one place build all, but I'm worried about 2 things.

1. Increased package time... (i package the project 10 - 100 times a day) so if it increase this i would not have this change
2. Confusion about what is inside our builds, since with this we have to manage the python API in our build of the system, rather than only in connection with python.

If you really really like the feature, and intend to use it. I would make it execute under a specific -D like PythonAPI, but then again it does not hook up into anything so there is more work needed if we want to pursue this.

j143:
Hi @phaniarnab,

1. We could use this goal for python build while issuing `mvn deploy -P 'distribution'`.
2. We could disable this during the regular development testing, though. as Sebastian suggested above.

Thanks.

phaniarnab:
I agree with @Baunsgaard, especially the second point.
For deployment, which we do only during releases, I can't think of a strong reason to make it a part of the pom.
Other than that, we need to meticulously check the builds, what it contains, and the package's size if you use this proposed option. Not sure if we want to upload python packages in the Apache distribution repository and open them for voting too during releases.

j143:
In SystemML, we released python artifacts along with java ones. Here - https://dist.apache.org/repos/dist/dev/systemds/1.1.0-rc1/

after offline discussion with Arnab,

It will be a good idea to keep the python package for voting. And we could pursue this one. Also, at the moment let us hold this PR.

j143:
As per mailing list discussion,

We will build python independent of maven! ✨

PR #1227, opened 2021-04-15
[SYSTEMDS-2933] Imputation By Multi-valued Dependency

Shafaq-Siddiqi:
Implements built-in for fixing erroneous values in the group.
for example, if two employees have the same rank, their values should fall into the same bucket the built-in identifies the values that do not follow the bin/range pattern and replace them with the mix or the max value observed in the group.


j143:
Hi @Shafaq-Siddiqi , why not take this PR in?

PR #1230, opened 2021-04-16
[SYSTEMSD-2936] Run script handling execution modes and other improve…

corepointer:
…ments

* Execution mode can be set via SYSDS_EXEC_MODE env var or via -exec parameter
* Better handling of -config parameter
* Better auto search for config, log and jar file if not provided
* Default to hybrid instead of singlenode
* Check for non singlenode mode when running distributed

As much as I dislike to bash scripts for their abundance of non obvious ways of solving problems, our run script needed some attention as it does not quite work correctly in some ways. Now that I needed the exec mode setting I pushed myself to it.
Sending this one out as PR as it changes the default behavior of our run script.

### DIFF
--- a/bin/systemds
+++ b/bin/systemds
@@ -29,14 +29,19 @@
 ##############################################################
 
 #  If not set by env,  set to 1 to run spark-submit instead of local java
-#  This should be used if "-exec SPARK" or "-exec HYBRID" is required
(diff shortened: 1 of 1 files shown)


Baunsgaard:
LGTM, the error is something in python, that can be easily fixed during merge.

Baunsgaard:
merged

PR #1231, opened 2021-04-17
[DOC] add commit description guide

j143:
Add description and tags usage. This PR can be the starting point.

> Why do we need a commit style guide?

Who uses the commit style guide:
- Apache Spark team
- Google Engineering teams

This change set is inspired by https://google.github.io/eng-practices/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,19 +17,17 @@ limitations under the License.
 {% end comment %}
 -->
 
-# Contributing to SystemDS
+# Contribution Guidelines and Standards for `SystemDS`
 
-Thanks for taking the time to contribute to SystemDS!
+Thank you for contributing to SystemDS. :smile:
 
-The following are mostly guidelines, not rules. Use your best judgment, and feel free to propose changes to this document in a pull request.
-
-___
-### Contribution Guidelines and Standards
+> The following are mostly guidelines, not rules. Use your best judgement, and
+> feel free to propose changes to this doc.
 
 Before contributing a pull request for [review](https://github.com/apache/systemds/pulls),
 let's make sure the changes are consistent with the guidelines and coding style.
 
-#### General Guidelines and Philosophy for contribution
+## General Guidelines and Philosophy for contribution
 
 *   Inclusion of unit tests when contributing new features, will help
     1. prove that the code works correctly, and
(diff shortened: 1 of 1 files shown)


j143:
@corepointer - do you think this change will be helpful for us as a reference?

corepointer:
> @corepointer - do you think this change will be helpful for us as a reference?

Yes, this could definitely be helpful. This is the first list of tags I see. Is the list complete? Where did you get it? Was there a discussion about the commit style that I missed? Can there be new tags? (e.g., [CLEANUP] ?)

Thank you for this PR! I think this is a good starting point for discussion where somebody with a longer history of SystemML/SystemDS can weigh in (maybe on the mailing list?)

Best regards, Mark

corepointer:
One more question: [HOTFIX] - what is the hot branch? We practically work with one branch only, so that's always hot imho. In what way would it be different from a bugfix?

j143:
I have been with SystemML/DS team from 2016! 💖 

> Is the list complete? Where did you get it? Was there a discussion about the commit style that I missed?

1. I have analyzed the complete commit history (from 2010).
2.  And have listened to most of the discussion on the list that of the SystemML (previous) and the present team.

> Can there be new tags? (e.g., [CLEANUP] ?)

This is the complete list of the tags. But, we can adjust or create new ones with functionality.

> what is the hot branch?

There my intention is in the release context. Needs rewording. 😕 

> Hotfix vs bugfix

@mboehm7  - can you help me out here?

phaniarnab:
I generally like this commit description guide. A crisp commit header and message have many advantages.
However, I would reduce the tag options to the bare minimum to remove confusion. Attaching a jira task to a commit is always the safest choice. MINOR is okay at times. I would not recommend using any other tags.

j143:
Hi @Baunsgaard - what do you think about this guide?
If everything's ok, can you provide an LGTM.

CONTRIBUTING.md
```diff
@@ -54,7 +52,151 @@ We have provided at profile for java located in [Codestyle File ./docs/CodeStyle
...
+  > 
+  > [`f4fa5650`](https://github.com/apache/systemds/commit/f4fa565013de13270df05dd37610382ca80f7354) - `[MINOR][SYSTEMDS-43] Cleanup scale builtin function (readability)`
+  >
+
+* `[SYSTEMDS-#]`: A changeset which will a specific purpose such as a bug, Improvement, 
+   New feature, etc. The tag value is found from [SystemDS jira issue tracker](https://issues.apache.org/jira/projects/SYSTEMDS/issues).
+   
+* `[DOC]` also `[DOCS]`: Changes to the documentation
+  
+* `[HOTFIX]`: Introduce a fix in the _hot_ running system.
```
Baunsgaard, line 86: i don't like the description here, it should say that it is commits that change behavior on already released versions, rather than using slang such as hot.

CONTRIBUTING.md
```diff
@@ -54,7 +52,151 @@ We have provided at profile for java located in [Codestyle File ./docs/CodeStyle
...
+5. Include enough context for
+   - reviewers
+   - future readers to understand the Changes.
+6. Add PR number, like `Closes #1000`.
+
+The following is a commit description with all the points mentioned.
+
+[`1abe9cb`](https://github.com/apache/systemds/commit/1abe9cb79d8001992f1c79ba5e638e6b423a1382)
+
+```
```
Baunsgaard, line 133: It is good style to specify what type of code block you have in markdown, here in this instance it is txt

CONTRIBUTING.md
```diff
@@ -54,7 +52,151 @@ We have provided at profile for java located in [Codestyle File ./docs/CodeStyle
...
+cleanup of these variables (which causes crashes if the inputs are used
+in other operations of the eval call).
+The fix is simple. We avoid such inconsistent construction and
+invocation of fcall instructions by using a narrower interface and
+constructing the materialized names internally in the fcall.
+```
+
+##### Small Changeset still needs some context.
+
+Example [`7af2ae0`](https://github.com/apache/systemds/commit/7af2ae04f28ddcb36158719a25a7fa34b22d3266)
```
Baunsgaard, line 185: Only place where you write example

CONTRIBUTING.md
```diff
@@ -17,19 +17,17 @@ limitations under the License.
 {% end comment %}
 -->
 
-# Contributing to SystemDS
+## Contribution Guidelines and Standards for `SystemDS`
```
Baunsgaard, line 20: Main header should be with one '#'

CONTRIBUTING.md
```diff
@@ -17,19 +17,17 @@ limitations under the License.
...
 
-Thanks for taking the time to contribute to SystemDS!
+Thank you for contributing to SystemDS. :smile:
```
Baunsgaard, line 22: :+1:

CONTRIBUTING.md
```diff
@@ -54,7 +52,151 @@ We have provided at profile for java located in [Codestyle File ./docs/CodeStyle
...
 - [Visual Studio Code](https://stackoverflow.com/questions/46030629/need-to-import-eclipse-java-formatter-profile-in-visual-studio-code)
 
-#### License
+## Commit Style
+
+SystemDS project has linear history throughout. Rebase cleanly and verify that the commit-SHA's
+of the Apache Repo are not altered.
+A general guideline is never to change a commit inside the systemds official repo, which is
+the standard practice. If you have accidentally committed a functionality or tag, let others know.
+And create a new commit to revert the functionality of the previous commit but **not force push**.
```
Baunsgaard, line 61: do not force push

CONTRIBUTING.md
```diff
@@ -54,7 +52,151 @@ We have provided at profile for java located in [Codestyle File ./docs/CodeStyle
...
+
+1. A brief description of the problem solved.
+2. Why this is the best approach?.
+3. Shortcomings to the approach, if any (important!).
+
+Additional info
+
+4. background information
+   - bug/issue/jira numbers
+   - benchmark/test results
```
Baunsgaard, line 121: Going into to much detail, you only need two spaces to make an sub itemize in markdown.

Baunsgaard (commented):
LGTM, but there are a few things i would change, mainly the description of HOTFIX and minor syntax. But other than that thanks for the guide. :+1:

CONTRIBUTING.md
j143: Aligned to the main points, for aesthetic reason in the editor.

j143:
Thanks everyone here for the comments. 😸 

Sorry, for pinging individually. Style changes such as this one needs to have sufficient consensus.

PR #1237, opened 2021-04-23
[SYSTEMDS-2604] Federated Output for AggregateBinaryFEDInstruction

sebwrede:
This PR adds federated output flag propagation to AggregateBinaryFEDInstruction. This requires partial aggregation of federated data, which means that several new processing cases have been added to the instruction. Additionally, a new field called `_overlapNum` has been added to the FederatedRange class, which ensures that ranges of identical dimensions can be added to the federation map, and a new federation type called `FType.PART` has been added to represent partially aggregated federated data. The `readBlobFromFederated` method in MatrixObject.java has been adapted to handle partially aggregated data by aggregating the data when the method is called. 
The reviewer of this PR should especially evaluate the changes made to FederationMap and FederatedRange to support the partially aggregated data and whether this partially aggregated data is properly processed.


src/main/java/org/apache/sysds/lops/MatMultCP.java
```diff
@@ -73,15 +73,15 @@ public String toString() {
 	@Override
 	public String getInstructions(String input1, String input2, String output) {
 		if(!useTranspose) {
-			return InstructionUtils.concatOperands(getExecType().name(),
+			InstructionUtils.concatBaseOperands(getExecType().name(),
```
Baunsgaard, line 76: I don't like this change because it hides where the variable is.
also i'm not sure if it works, since strings behave differently in java than other variables (immutable),

src/main/java/org/apache/sysds/runtime/controlprogram/caching/MatrixObject.java
```diff
@@ -548,27 +550,54 @@ protected MatrixBlock readBlobFromFederated(FederationMap fedMap, long[] dims)
...
+	 * Aggregate partially aggregated data from federated workers
+	 * by adding values with the same index in different federated locations.
+	 * @param readResponses responses from federated workers containing the federated data
+	 * @return MatrixBlock of consolidated, aggregated data
+	 */
+	private MatrixBlock aggregateResponses(List<Pair<FederatedRange, Future<FederatedResponse>>> readResponses) {
+		List<Future<FederatedResponse>> dataParts = new ArrayList<>();
+		for ( Pair<FederatedRange, Future<FederatedResponse>> readResponse : readResponses )
+			dataParts.add(readResponse.getValue());
+		return FederationUtils.aggAdd(dataParts.toArray(new Future[0]));
```
Baunsgaard, line 599: I Like it, it is basically the same as I'm doing with the Overlapping in compression.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
```diff
@@ -81,6 +88,10 @@ public long getSize() {
 			size *= getSize(i);
 		return size;
 	}
+
+	public int getOverlapNum(){
+		return _overlapNum;
```
Baunsgaard, line 93: In my experience it does not matter how many overlaps you have, just have a boolean specifying if there is overlap or not.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
```diff
@@ -94,12 +105,19 @@ public int compareTo(FederatedRange o) {
 			if ( _beginDims[i] > o._beginDims[i])
 				return 1;
 		}
+		if (_overlapNum < o._overlapNum)
+			return -1;
+		if (_overlapNum > o._overlapNum)
+			return 1;
```
Baunsgaard, line 111: how is it relevant which index it is?
if you are trying to keep the overlaps in a certain order, it is really not a needed feature. 
But then again if it really is necessary to sort the federated ranges i guess this is an okay way.

if you change _overlapNum to be a boolean, you could sort on Object pointers instead in last case.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
```diff
@@ -50,6 +50,7 @@
 		ROW, // row partitioned, groups of rows
 		COL, // column partitioned, groups of columns
 		FULL, // Meaning both Row and Column indicating a single federated location and a full matrix
+		PART, // Partial aggregates in several federated locations with addition as the aggregation method
```
Baunsgaard, line 53: Call it Overlap, to not confuse (at least me) and this overlap is a concept i have in compression as well, where i call it ... overlap.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
```diff
@@ -189,9 +190,14 @@ public FederatedRequest broadcast(ScalarObject scalar) {
 		return ret;
 	}
 
+	/**
+	 * Determines if the two federation maps are aligned row/column partitions
+	 * at the same federated sites (which allows for purely federated operation)
+	 * @param that FederationMap to check alignment with
+	 * @param transposed true if that FederationMap should be transposed before checking alignment
+	 * @return true if this and that FederationMap are aligned
```
Baunsgaard, line 198: Thanks for the documentations !!!

Baunsgaard (commented):
Looks good, 

I really like the idea, but i would remove the Part keyword and replace it with Overlapping.

Other than that, it would be cool if you enabled some of the tests in the python API, since i added overlapping tests there sometime back and never got to use them, this would be a prime example of it, and should "hopefully" not give you any trouble.

src/main/java/org/apache/sysds/lops/MatMultCP.java
sebwrede: I have done this in other methods and I have not yet experienced any problems with it. The issue with the hidden variable is that it does not make any sense to return it from the concatBaseOperands method because of the internals of InstructionUtils. I think it is working with a thread-local variable and this means that for subsequent calls from this getInstructions method, it will always access the same variable.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
sebwrede: This is to ensure that I can put similar ranges in a map without replacing any values. The range needs to be unique to achieve this.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
sebwrede: See other comment about putting into map.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
sebwrede: I chose this name because it matches the other names: it is short and in contrast to "FULL".

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
Baunsgaard: i don't get your argument, if you allocate different objects for each range, then there should not be any problems other than you have to change the compare method to also in case of same values look at object pointers.

src/main/java/org/apache/sysds/runtime/controlprogram/caching/MatrixObject.java
sebwrede: I think I have to take a look at that. Where have you implemented it? Is it redundant or just the same concept?

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedRange.java
sebwrede: That would not be a good solution since it would affect all other federated ranges as well. The solution I implemented here will only affect the behavior of the overlapping ranges and will not change how the existing federated ranges behave. 
I did not really like this either when I added the _overlapNum, but it was a solution that would work for the case I added without negatively affecting anything else. 
Another option would be to generate a random ID instead of setting an overlap number, since I do not need this specific number anyway. The problem is that I also don't need a randomly generated ID, so it would be overkill to make the implementation when all I need is any integer. 
Another solution would be to add overlap as a boolean and then check if this boolean is true in compareTo, toString, equals, and hashCode. If it is true then I should include the object pointer in some way in the computation. But then the question is again if this is too much compared to just adding a unique integer from the beginning.

Baunsgaard: I don't agree,

Actually now that i think about it, you would not need a boolean nor a overlapping index, since you are using the "PART" (soon to be "OVERLAP") to mark if a given federated matrix is overlapping, therefore you should not need any of this logic to handle any incoming logic, just switch on the federated type.

again the compare part is easily solved via object compare since it contains a "random" ID in form of its object pointer that is unique for each object.

sebwrede: I would like to use the PART type as the condition for adding the object pointer to the aforementioned methods, but this is not available in FederatedRange. 

Are you sure about the name "OVERLAP"? Isn't it too long compared to the others? Additionally, it is not any kind of overlap. It is specific to the partial aggregates.

Baunsgaard: well in compression the overlap state only happens on a right matrix multiplication, but in federated i can already imagine it happening many times, if you want to convince me on another name it is fine, but i think having the same name would be nice.

I guess you chose PART because it is a "PARTial aggregate value". I chose overlap because it is matrices that could be imagined to be on top of each other and added together to produce the result. I think that the name should reflect what it is, and if we have the same concept more places it should be called the same, that said i don't know what other ppl call it when they have distributed parts that could be reduced to the actual value.

sebwrede:
I am not completely convinced that the latest commit is an improvement. 
What do you think, @Baunsgaard ?

Baunsgaard:
> I am not completely convinced that the latest commit is an improvement.
> What do you think, @Baunsgaard ?

i don't like it either.

Why not allow the two federated ranges to be equal if the arrays are the same and they are both overlapping?
then you don't have to care about the hash of this element, making the implementation cleaner.
In general why does the ranges need to implement comparable, if it is just for sorting then allowing them to be equal is fine.

also you added more of the PART, rather than rewording it to overlap.

sebwrede:
> Why not allow the two federated ranges to be equal if the arrays are the same and they are both overlapping?
> then you don't have to care about the hash of this element, making the implementation cleaner.
> In general why does the ranges need to implement comparable, if it is just for sorting then allowing them to be equal is fine.

The hashCode, equals, and compareTo are used when putting the range into a federation map. Depending on the data structure chosen for the federation map, which in most cases is a TreeMap, one or more of the methods will need to separate between the overlapping ranges. This is the point of adding some kind of uniqueness to the ranges. 
If I change one of the three methods, then I should change all of them to include this unique value. Otherwise, it will be really confusing later on if there is a difference between putting into a HashMap vs a TreeMap. 
Additionally, I add the uniqueness to the toString method so that we can see the difference in the print. 

> it would be cool if you enabled some of the tests in the python API, since i added overlapping tests there sometime back and never got to use them

Where do I find this?

Baunsgaard:
> > Why not allow the two federated ranges to be equal if the arrays are the same and they are both overlapping?
> > then you don't have to care about the hash of this element, making the implementation cleaner.
> > In general why does the ranges need to implement comparable, if it is just for sorting then allowing them to be equal is fine.
> 
> The hashCode, equals, and compareTo are used when putting the range into a federation map. Depending on the data structure chosen for the federation map, which in most cases is a TreeMap, one or more of the methods will need to separate between the overlapping ranges. This is the point of adding some kind of uniqueness to the ranges.
> If I change one of the three methods, then I should change all of them to include this unique value. Otherwise, it will be really confusing later on if there is a difference between putting into a HashMap vs a TreeMap.
> Additionally, I add the uniqueness to the toString method so that we can see the difference in the print.

Design wise would it not then make sense then to move the addresses into these ranges?
This would result in the uniqueness coming from the combination of range and address(inside same object), this would also allow us to based on the equality of objects know that we have a federated location with two intermediates that perfectly overlap on the same machine, and therefore merge these remotely.
Maybe @mboehm7  have a comment on this?

> > it would be cool if you enabled some of the tests in the python API, since i added overlapping tests there sometime back and never got to use them
> 
> Where do I find this?

In src/main/python/tests/federated/test_federated_aggregations.py and src/main/python/tests/federated/test_federated_basic.py

mboehm7:
Thanks for the patch and the detailed discussion @sebwrede and @sebwrede. I started merging it but need to finalize this tomorrow (no please no new commits on this PR).

As a high-level summary:
* The modification of the FederatedRange is actually unnecessary as we predominately iterate over the entry set - now we simply convert this map to a list of pairs and avoid any workarounds for comparison.
* In anticipation of @OlgaOvcharenko PR (currently WIP) for handling broadcasts and sliced broadcasts as federated data as well, we should clearly separate the partitioning type (ROW, COL, NONE, MIXED) and the replication type (NONE, FULL, PARTIAL) where PARTIAL might be OVERLAP.
* Currently the federatedOutput flags are conditionally compiled into the instructions. We should add a three-valued logic here (true, false, none) where true/false forced the instruction to act a certain way, while none preserves the current heuristic runtime semantics.
* We need to fix (i.e., remove) all the stateful string concatenation

PR #1244, opened 2021-04-30
[MINOR] Force ytest argument in lmPredict

Baunsgaard:
This is not the ideal solution, but just to make the test pass the LMPredict now require ytest argument,
The reason i simply because the python generators does not like default parameters on matrice arguments yet.


Baunsgaard:
merged

PR #1245, opened 2021-05-01
[SYSTEMDS-2960] Codegen: POS replacement for elementwise binary operations

ywcb00:
Hi,
I tried to correctly replace the position for calls to elementwise binary operations of "SideInputs" (b) with column vectors.
In my opinion, we need to do include the row index in the very same way as for matrices if it is an (1) elementwise operation with a (2) column vector.

Thank you for reviewing and commenting :)


mboehm7:
LGTM - thanks for catching this issue @ywcb00.

PR #1266, opened 2021-05-07
[WIP] Modified federated broadcast and broadcastSliced

OlgaOvcharenko:
This PR modifies  federated broadcast and broadcastSliced. Instead of putting the variable and then cleaning after instruction execution new FederatedData is initialized and reused after.

### DIFF
--- a/src/main/java/org/apache/sysds/common/Types.java
+++ b/src/main/java/org/apache/sysds/common/Types.java
@@ -509,7 +509,12 @@ public class Types
 			}
 		}
 	}
-	
+
+	public enum ReplicationType {
+		FULL, // full replication to all workers, broadcast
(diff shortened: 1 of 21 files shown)


mboehm7:
Thanks for getting started on this consolidation of broadcasts @OlgaOvcharenko. With the rework of the federation map, there should however, no longer be a need for separate broadcast maps, etc because every broadcast (replicated or sliced) is just handled like any other federated data. The reuse of broadcasts will then be done via simple checks for aligned federated inputs (which we will improve subsequently for all operations, in fact @ywcb00 will make a full pass over all instructions to ensure we don't miss any).

Could you maybe try to simplify the PR and let me know once you're ready, then I'll make a pass and help finalize it if necessary.

PR #1271, opened 2021-05-10
[WIP] Use SystemDS with Docker

j143:
This is working document for usage of SystemDS with Docker images.

**Motivation:**
This helps in prototyping DML or python based scripts faster and run the notebooks with jupyter support.

```sh
docker run -it --rm apache/systemds \
  /bin/bash -c "echo 'print("This is SystemDS!")' > hello.dml && systemds hello.dml"
```

**Present:**
With the existing [`docker/`](https://github.com/apache/systemds/tree/master/docker) files,
`dev` releases can be made.

**Future behavior:**

| Tag | Description |
| --- | --- |
| `latest` | The latest release of SystemDS CPU image. Default version. |
| `version` | Specify the exact version |
| `dev` | Builds for SystemDS `master` branch. Used by SystemDS developers |

With GPU:
| Tag Variants | Description |
| --- | --- |
| `tag-gpu` | The specified tag release with GPU support. |

### DIFF
--- a/docs/site/docker.md
+++ b/docs/site/docker.md
@@ -79,8 +79,18 @@ Options:
 For comprehensive guide, refer [`docker run`](https://docs.docker.com/engine/reference/run/)
 
 ```sh
-docker run [-it] [--rm] [-p hostPort:containerPort] apache/systemds[:tag} [command]
 
-# Example
-# docker run -it --rm -p 8080:8080 apache/systemds:1.2.0 ls
+docker run [-it] [--rm] [-p hostPort:containerPort] apache/systemds[:tag} [command]
 ```
+
+Examples
+
+To verify the SystemDS installation,
+
+```sh
+docker run -it --rm apache/systemds \
+  /bin/bash -c "echo 'print("This is SystemDS!")' > hello.dml && systemds hello.dml"
+```
+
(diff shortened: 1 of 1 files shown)


Baunsgaard:
i must question this, 

this documentation is not valid since it does not reflect the underlying implementation. If we want to have the system act like in these docs, we have to change some of the docker behavior.
Also just to make it clear, there is no official Apache docker management, each project have their own just like us: https://hub.docker.com/orgs/systemds 

I would suggest looking in our docker folder, i have added some docs there already for how to use the images.

j143:
> official Apache docker management, each project have their own just like us: https://hub.docker.com/orgs/systemds

Not convinced yet. I have asked around apache/airflow team. We will resolve this later. 😃

Baunsgaard:
:-1: for now.

I like the idea of separating the docker images in accordance to what you propose because it fits well with other systems separation of images.

I mainly say :-1: because this is done in the wrong order, especially because the design docs are describing something we already have. I would prefer such a thing as this to be on the dev list, as a suggestion that could be discussed, if it is only documentation, and not an implementation modifying the system. if we want to change the implementation to fit to these docs we should do it in the same process as we update the docker images.

PR #1274, opened 2021-05-14
[SYSTEMDS-2970] Initial linux instructions for GPU

j143:
* initial  linux instructions
* add GPU brand version support
* Update gpu.md
* fix format add libcudnn instructions
* add single node invocation command


j143:
for copy: @corepointer  😸

PR #1275, opened 2021-05-14
Image Manipulation Primitives

whirgod:
Adds a couple of built-in functions for manipulating images


src/test/java/org/apache/sysds/test/TestUtils.java
```diff
@@ -37,19 +37,8 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
Baunsgaard, line 40: we try to avoid wildcard imports.

src/test/java/org/apache/sysds/test/TestUtils.java
```diff
@@ -3073,4 +3059,46 @@ public static int isGPUAvailable() {
...
+	}
+
+	/**
+	 * Reads a matrix from a text file. Rows are separated by newline, Values are separated by space.
+	 * Take from https://www.tutorialspoint.com/How-to-read-a-2d-array-from-a-file-in-java
+	 * @param url URL to the resource file obtained by Class.getResource()
+	 * @param rows The expected number of rows
+	 * @param cols The expected number of columns
+	 * @return The matrix
+	 * @throws Exception
```
Baunsgaard, line 3088: would this not be an IOException.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinImageRotateTest.java
```diff
@@ -0,0 +1,102 @@
...
+				"height=" + rows, "angle=" + angle, "fill_value=" + fill_value};
+		writeInputMatrixWithMTD("A", input, true);
+		runTest(true, false, null, -1);
+
+		HashMap<MatrixValue.CellIndex, Double> dmlfile = readDMLMatrixFromOutputDir("B");
+		double[][] dml_res = TestUtils.convertMatrix(dmlfile, h, w);
+		TestUtils.compareMatrices(reference, dml_res, eps, "Pillow vs. DML");
+		// The error here seems to be in the pillow implementation. It calculates that the source coordinates of
+		// (339, 14) are (351, 87) using fixed point arithmetic, while the floating point implementation returns
+		// (351.975384 88.000514) rounded down to (351, 88). (All indices here start at 0)
```
Baunsgaard, line 100: this difference sounds large, could it be possible to test with some other framework, or write the calculation in java?

src/test/resources/org/apache/sysds/test/functions/builtin/ImageTransformInput
```diff
@@ -0,0 +1,135 @@
+0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 255 255 255 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
```
Baunsgaard, line 1: if possible it would be nice if these files were moved to the resources folder and maybe reduced in size.

Baunsgaard: also a file ending describing the format would be nice. Is this png?

Baunsgaard (commented):
In general LGTM, only very minor comments.

It would be nice if the rounding error hypothesis for rotate could be analyzed, it might be an internal bug.
if a test is added where you use double precision and the result is correct then i would call it appropriate.

The only thing really missing in this PR is documentation on the new builtins added, i would suggest taking a look in pca.dml for inspiration.

src/test/java/org/apache/sysds/test/functions/builtin/BuiltinImageRotateTest.java
whirgod: I analyzed the source code of the Pillow library. They use a fixed-point arithmetic implementation by default which for the given input returns 87 as y coordinate. If the given input does not satisfy some criteria they fall back on a floating-point arithmetic implementation. For the given input the floating-point implementation returns 88 as y coordinate which is the same as my floating-point implementation in dml.

src/test/resources/org/apache/sysds/test/functions/builtin/ImageTransformInput
whirgod: What resource folder are you referring to? It already is in src/test/resources.
For size reduction, I could change it to a binary representation, but that would make the read-in process more complex and the content would be more difficult to understand.
I suppose the file format is .csv

src/test/java/org/apache/sysds/test/TestUtils.java
whirgod: Done

src/test/java/org/apache/sysds/test/TestUtils.java
whirgod: Done

corepointer:
Hello and thank you for your effort in this PR! 
In general I agree with Sebastian, it looks good, but if I may chime in with additional questions:
Why is Pillow used? I'm sure there are image manipulation libraries for R and Java that could be used.
I guess the use of Pillow is the reason for the additional resource files and the methods in TestUtils? That (and the mentioned precision issues) could be avoided by using an R or Java library for comparison. 
jm2c ;-)
Mark

whirgod:
> Hello and thank you for your effort in this PR!
> In general I agree with Sebastian, it looks good, but if I may chime in with additional questions:
> Why is Pillow used? I'm sure there are image manipulation libraries for R and Java that could be used.
> I guess the use of Pillow is the reason for the additional resource files and the methods in TestUtils? That (and the mentioned precision issues) could be avoided by using an R or Java library for comparison.
> jm2c ;-)
> Mark

I use Pillow as a reference because one of the reasons I added the builtins is to reproduce the results of AutoAugment (https://arxiv.org/abs/1805.09501) and they use Pillow.

src/test/resources/org/apache/sysds/test/functions/builtin/ImageTransformInput
Baunsgaard: My bad, the view i had collapsed the folders.
But looking at this again it seems excessive to make all these folders. I understand why you replicated the folder structure since this could be considered standard.

The intention behind the resources folder currently is to enable sharing of common resources between multiple tests, therefore i would suggest making a folder called img. and include these files there. 'src/test/resources/img/...'

Furthermore since the file type is csv it would be nice to have a metadata file associated with each of the images, this makes it possible to read in through our standard csv parsing in systemds. If you don't want to do it i will during merge.

Baunsgaard:
> > Hello and thank you for your effort in this PR!
> > In general I agree with Sebastian, it looks good, but if I may chime in with additional questions:
> > Why is Pillow used? I'm sure there are image manipulation libraries for R and Java that could be used.
> > I guess the use of Pillow is the reason for the additional resource files and the methods in TestUtils? That (and the mentioned precision issues) could be avoided by using an R or Java library for comparison.
> > jm2c ;-)
> > Mark
> 
> I use Pillow as a reference because one of the reasons I added the builtins is to reproduce the results of AutoAugment (https://arxiv.org/abs/1805.09501) and they use Pillow.

I don't think using Pillow is an issues, since it is only used in tests and not actually added to the system, since all the results seems to be added as references. Therefore nothing will not be included as a resource in the system.

That said it would be better with the actual expected results (calculated and verified not using our system), than results where we have to consider rounding errors from floating point to uint. but i consider this out of scope of this project, since the methods are tested and verified sufficiently.

corepointer:
> > > Hello and thank you for your effort in this PR!
> > > In general I agree with Sebastian, it looks good, but if I may chime in with additional questions:
> > > Why is Pillow used? I'm sure there are image manipulation libraries for R and Java that could be used.
> > > I guess the use of Pillow is the reason for the additional resource files and the methods in TestUtils? That (and the mentioned precision issues) could be avoided by using an R or Java library for comparison.
> > > jm2c ;-)
> > > Mark
> > 
> > 
> > I use Pillow as a reference because one of the reasons I added the builtins is to reproduce the results of AutoAugment (https://arxiv.org/abs/1805.09501) and they use Pillow.

OK, wasn't aware of that.

> 
> I don't think using Pillow is an issues, since it is only used in tests and not actually added to the system, since all the results seems to be added as references. Therefore nothing will not be included as a resource in the system.

Spotted a double negation there :D Was that your intention? It actually fits well as there are resource files included.

> 
> That said it would be better with the actual expected results (calculated and verified not using our system), than results where we have to consider rounding errors from floating point to uint. but i consider this out of scope of this project, since the methods are tested and verified sufficiently.

Another suggestion that'd be out of scope: Comparing to AutoAugment is out of scope of a unit test and should go into a dedicated script that can happily use Pillow and everything that is needed to run AutoAugment. For a unit test it is sufficient to have the correct functionality of our code confirmed. 

But that is all just expressing my opinion again - no action needed on your part.

Mark

mboehm7:
LGTM - thanks for the patch and detailed tests @whirgod. During the merge I made a few modifications:

* Removed the added but redundant hashmap-matrix conversion methods
* Refactored the tests and resource files (in package pipelines, and expected files)
* Fixed the expected data for the rotation test (thanks for double checking these in detail)
* Fixed unnecessary imports and formatting issues (annotations above, not before methods)
* Started vectorizing the parfor loop in transform (but still commented as it requires more work)

PR #1277, opened 2021-05-16
[SYSTEMDS-2922] Federated Fused Codegen Operator - Multi Aggregate

ywcb00:
Hi,
This is a PR for adding federated code generation for multi aggregate templates, and the respective tests (hybrid execution mode to reduce the test time).
NOTE: All three test modes are working (SingleNode, Spark, Hybrid) for all testcases (also the outcommented tests).

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
+++ b/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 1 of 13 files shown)


Baunsgaard:
> Hi,
> This is a PR for adding federated code generation for multi aggregate templates, and the respective tests (hybrid execution mode to reduce the test time).
> NOTE: All three test modes are working (SingleNode, Spark, Hybrid) for all testcases (also the outcommented tests).
> 
> Thanks for review :)

I would suggest adding an @ignore flag on the tests that you outcommented. This makes them not run, but be part of the output so that it is easy to spot that there are some extra tests, and furthermore it should be possible to force them to run without having to change a single line of code.

ywcb00:
Thank you for the feedback @Baunsgaard :)
Great suggestion, I changed all three tests (cellwise, rowwise, multiagg) to use the @Ignore flag instead of outcommented tests

mboehm7:
LGTM - thanks for the patch @ywcb00.

PR #1280, opened 2021-05-18
[MINOR] Federated frame map for element-wise operations

OlgaOvcharenko:
This PR adds federated frame map for element-wise operations, e.g. map(X, "x -> x.toUpperCase()").


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko. I only changed few syntactic things during the merge.

PR #1283, opened 2021-05-19
[SYSTEMDS-2922] Federated Fused Codegen Operator - Outer Product

ywcb00:
Hi,
This is a PR for adding federated code generation for outer product templates, and the respective tests (hybrid execution mode to reduce the test time).
I created some nested classes in _SpoofFEDInstruction.java_ for the different spoof template types to better distinguish between them (e.g., cellwise, rowwise, multiagg, outerproduct). Each class has its own _setOutput()_ method, whereas only the class of the outerproduct template has its own _needsBroadcastSliced()_ method, because the other classes can share it.

The federated spoof instruction for outerproduct templates is determining if an input matrix gets broadcast sliced based on its dimensions. This could create some problems, for example when computing
`X * log(U %*% t(V))`
with federated matrix _X_ in dimensions _2000x2000_,  _U_ in _2000x10_, and _V_ in _2000x10_, it broadcasts both, _U_ and _V_, in a sliced manner because the row number matches with the dimension of the federated matrix. Obviously, depending on the federation type of _X_, either _U_ or _V_ should've been broadcast sliced. However, when _U_ and _V_ differ in the dimensions, everything is working how it should. How can we handle this case of similar dimensions?

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
+++ b/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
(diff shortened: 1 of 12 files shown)


ywcb00:
I changed the _needsBroadcastSliced()_ method to check whether it is the side input _U_ or _V_ based on the index of the inputs.
If _X_ is row partitioned, _V_ is not getting broadcast sliced - and vice versa.

mboehm7:
LGTM - thanks for the patch @ywcb00. Great to see the federated codegen support completed.

PR #1284, opened 2021-05-23
[SYSTEMDS-2978] Federated frame tokenize

OlgaOvcharenko:
This PR adds federated frame tokenize.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko. During the merge, I cleaned up some surrounding redundancy, and fixed two minor things in the tokenization (exception handling, unnecessary symbol table changes at coordinator).

Furthermore, note that JSON does not need double quotes around attribute names, so we can avoid the escaping `\"attribute\":value` at DML script level.

PR #1286, opened 2021-05-27
[SYSTEMDS-2889] Add initial lineage debugger for CP instructions

bnyra:
This patch adds a lineage based debugger which keeps track of special values in matrixblocks in CP instructions.

Therefore, the following design decisions are made and implemented:

- `LINEAGE_DEBUGGER` flag as DMLScript option to enable the debugger (default: false)
- Added `class BooleanArray32` for storing boolean information in an memory efficient way
- This class is used as an attribute `_specialValueBits` in `LineageItem` to store the special value information
- Check of special values is done in the `postprocessInstruction()`
- Added serialization and deserialization of `_specialValueBits` with surrounding brackets
- Added helper function to find the first occurrence of a special value given a `LineageItem`.

Additional info: 

- Only CP instructions are supported
- Debugger keeps track of `Double.NaN`, `Double.POSITIVE_INFINITY` and `Double.NEGATIVE_INFINITY`
- Fixed missing default post-process call in `CPInstruction` and `GPUInstruction`


phaniarnab:
Thanks for the patch @bnyra. Before merging, I moved the debugging-related maintenance from Instruction to CPInstruction as calling acquireRead can be problematic for federated and GPU instructions, and removed serialization of the debug byte if debugger is disabled.

PR #1287, opened 2021-05-27
[SYSTEMDS-2982] Federated codegen w/ multiple aligned matrices

ywcb00:
Hi,
This PR adds support for federated **spoof** instructions with multiple **aligned** federated matrices.
It was necessary to create an additional _isAligned()_ method in order to support partial alignments.
The _SpoofFEDInstruction_ needed some rewrites to not change the order of inputs, because we need the correct order to distinguish between the side inputs _U_ and _V_ for the outerproduct template.
I added comparision of the end dimensions of federated ranges in _FederatedRange.compareTo()_, hope this does not create any problems with other functionalities.
I also added the method _heavyHittersContains(Sub)String()_ with an additional parameter _minCallCount_ which allows to check for a minimum number of calls for a specific heavy hitter.
There are a lot of outcommented tests because I tried to bundle multiple testcases into a single test to reduce the time for test preparation.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
+++ b/src/main/java/org/apache/sysds/runtime/codegen/SpoofCellwise.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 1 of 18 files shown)


mboehm7:
LGTM - thanks @ywcb00 for the patch. During the merge, I made a few cleanups (code simplification) and left a FIXME. Could you please try to consolidate (e.g., in the task on general aligned operations) the now more complex `isAligned` in the SpoofCP and SpoofSP instructions to avoid redundancy? Thanks.

PR #1294, opened 2021-06-03
[SYSTEMDS-3009] Federated MSVM Test

OlgaOvcharenko:
This PR adds a federated MSVM test.


mboehm7:
LGTM - thanks for the additional test @OlgaOvcharenko. During the merge, I only changed (unrelated to this PR but the test showed it) the log level of notifications regarding pulling federated data into the coordinator because this is very common given that many instructions keep their outputs federated.

PR #1295, opened 2021-06-05
Copyright year update

j143:
as per:
https://github.com/apache/systemds/search?p=1&q=2020

This PR is to pin all the files where year update is required.

Note:
If you would like to add any more files, update directly in https://github.com/j143/systemds/tree/copyright-year (maintainer access!)


Baunsgaard:
LGTM.
but instead of us committing to your forks-branch just merge it in when you find them (preferably in bulk in one commit)

PR #1297, opened 2021-06-07
[SYSTEMDS-206] HDF5 Data Format

fathollahzadeh:
This commit is an implementation for HDF5 data format files. The HDF5 The file format specification is available from the HDF Group [here](https://en.wikipedia.org/wiki/Hierarchical_Data_Format) 

Currently, this patch is targeting the HDF5 format with supporting some features of the HDF5. The supported features:
 1. Two Dimension Data (Matrix)
 2. Matrix with FP64 (double) Data Type
3. Single Dataset
4. Single Group 
5. Contiguous Dataset
6. Read/Write and SystemDS Parallel Feature


phaniarnab:
Thanks for the PR @fathollahzadeh. Can you please rebase your changes on top of Master and resolve the conflicts, so that the tests run? During that process, also please clean the history if needed (I see a few merge-branches).

fathollahzadeh:
Thanks for the comments @adnaribo. I'll fix them and create a new PR.

PR #1302, opened 2021-06-08
[SYSTEMDS-2570] Remove mentions of caffe in license

Baunsgaard:
All mentions of Caffe is removed from the license since caffe is
no longer used in systemds.


phaniarnab:
LGTM.
Updating the License file shouldn't impact generating extra jar as extra.xml includes the whole license.

PR #1305, opened 2021-06-08
[SYSTEMDS-2992+2993] Workload costing and Cost-based Compression Planning

Baunsgaard:
[SYSTEMDS-2992+2993] Workload costing and Cost-based Compression Planning

This commit contains a new package in the compression framework, that
allow selection of cost models for compression.
One classical cost model is memory size, in this commit there are
multiple more, most notably

MEMORY, W_TREE, HYBRID, DISTINCT.

- Memory is the memory based optimization,
- W_TREE is based on the extracted WTree of instructions
- HYBRID is a combination of memory and W_TREE,
- DISTINCT is based on making the dictionaries as small as possible.
- LEFT_MATRIX_MULT is based on optimizing only for Left Matrix Multiplications
- Decompression is ... for decompression time
- TSMM for fast transpose self matrix multiplication

Additionally to these additions all CoCoding thechniques now optimize
towards the given type of cost model, making it posible to make for
for instance BIN_PACKING DISTINCT optimization, or HYBRID Brute Force,
etc.


Baunsgaard:
Closing because of continued work in PR : #1320

PR #1309, opened 2021-06-09
[MINOR] Transformencode minor performance improvement and bugfix

ilovemesomeramen:
This PR removes some redundant code that caused a significant performance decrease in large datasets. Furthermore it fixes a bug in which DC produced the wrong results in a single threaded environment.


phaniarnab:
Thanks for the bug fixes @ilovemesomeramen.

PR #1313, opened 2021-06-11
[SYSTEMDS-3018] Federated Execution Rewriter

sebwrede:
This PR adds a federated execution rewriter. 
The rewriter currently loads the privacy constraints of federated data input from the federated workers and propagates these privacy constraints through the Hop DAG. The rewriter will be extended in a later PR with an execution plan builder, which will represent the different federated execution plans and evaluate the utility of each plan.
This PR additionally refactors the PrivacyPropagator to ensure that the Hop and instruction privacy propagation goes through the same core propagation.


mboehm7:
LGTM - thanks for the patch @sebwrede. I applied the same rewrite also predicates now (as they might similarly tough protected datasets), fixed some warnings and imports, and would recommend for the future to use the MetaDataAll for handling mtd file as it tries to limit the surface of dealing with specific JSON APIs that we want to replace in the near future.

PR #1314, opened 2021-06-11
[SYSTEMDS-2982] Federated instructions w/ multiple aligned matrices

ywcb00:
Hi,
This PR adds support for federated **binary, aggregate ternary, aggregate binary, ctable, and mmchain** instructions with multiple **aligned** federated matrices.
I introduced a new enum _AType_ and the method _isAligned(AType)_ inside the FederationMap to simplify the alignment checks (to check for multiple possible partial alignments with a single call).
To avoid redundant code, I created a new static method _isFederated()_ in _SpoofFEDInstruction_, which is only a consolidation of the _isFederated()_ method of _SpoofCPInstruction_ and _SpoofSPInstruction_.
In addition, I changed several tests in order to include testcases with multiple federated inputs.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
@@ -99,13 +99,13 @@ public class FederationMap {
 	}
 
 	// Alignment Check Type
-	public enum AType {
-		FULL,
-		ROW,
-		COL,
-		FULL_T,
-		ROW_T,
-		COL_T;
+	public enum AlignType {
+		FULL, // exact matching dimensions of partitions on the same federated worker
+		ROW, // matching rows of partitions on the same federated worker
+		COL, // matching columns of partitions on the same federated worker
+		FULL_T, // matching dimensions with transposed dimensions of partitions on the same federated worker
+		ROW_T, // matching rows with columns of partitions on the same federated worker
+		COL_T; // matching columns with rows of partitions on the same federated worker
 
 		public boolean isTransposed() {
 			return (this == FULL_T || this == ROW_T || this == COL_T);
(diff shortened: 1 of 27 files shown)


src/main/java/org/apache/sysds/runtime/instructions/cp/SpoofCPInstruction.java
```diff
@@ -138,34 +136,6 @@ public boolean isFederated(ExecutionContext ec) {
 	}
 	
 	public boolean isFederated(ExecutionContext ec, FType type) {
```
sebwrede, line 138: It seems a bit counter-intuitive that a CP instruction has an `isFederated` method. Could we move this to `FedInstructionUtils.checkAndReplaceCP` which I believe is the only place where a call to `isFederated` is needed. Now that you cut the internals of the method to a single line this should be easy.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
```diff
@@ -98,6 +98,29 @@ public boolean isType(FType t) {
 		}
 	}
 
+	// Alignment Check Type
+	public enum AType {
```
sebwrede, line 102: Suggestion:
Could this be called "AlignType" or something else that is a bit more specific. I read "AType" as "a type" which is a bit too generic :smiley:

sebwrede (commented):
I only added a few comments. 
Otherwise, LGTM.

src/main/java/org/apache/sysds/runtime/instructions/cp/SpoofCPInstruction.java
ywcb00: You're right :)
I changed it to directly call the static method _isFederated()_ from _SpoofFEDInstruction_ and introduced the getter method _getInputs()_ in _SpoofCPInstruction_ and _SpoofSPInstruction_ to get the inputs for that call.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
ywcb00: It was so tempting to name it _AType_ alongside the other enum named _FType_ :rofl: 
I changed the name to _AlignType_ now :+1:

ywcb00:
Thank you for the feedback @sebwrede :)
I changed the code according to your suggestions.

PR #1316, opened 2021-06-14
[SYSTEMDS-3021] Python Source reuse fix

Baunsgaard:
This commit fix a bug in python source where if the source is used in multiple compute statements it breaks.
This is partially because the source command worked differently than other nodes. This commit now unifies this,
to a single way, and fixes the bug.

We should include this in the release.


phaniarnab:
LGTM.
Thanks for the fix. I agree that this should be included in the release.

PR #1320, opened 2021-06-20
[SYSTEMDS-3037] CLA Spark support

Baunsgaard:
This PR adds spark support back for CLA, with various tests for verifying that no crashes is experienced. Further work is to allow workload aware SP compression, but currently the support covers default compression instructions.

Currently the main limitation is left multiplications where if the compressed side of a multiplication is smaller (in terms of cells) it is re blocked to be broadcasted, in that process the all blocks in that matrix RDD is decompressed.


Baunsgaard:
Closing Because of continued work in #1322

PR #1321, opened 2021-06-21
[SYSTEMDS-3039] Federated Statistics

ywcb00:
Hi,
This PR adds the functionality of collecting, aggregating, and displaying statistics of the federated workers.
The following statistics are considered:
* all the cache statistics
* garbage collection statistics
* heavy hitters
* JIT times

I introduced a new dml option _fedStats_ (similar to the existing _stats_ option) to enable the creation of federated statistics.
To better seperate the federated statistics from the local statistics, I created a new class _FederatedStatistics_.

Thanks for review :)


mboehm7:
LGTM - thanks for the patch @ywcb00. During the merge I converted print-based error handling into proper exception handling, and added the fedStats flag to at least one test (the existing federated statistics test was ignored and now serves both federated statistics and the other coordinator statistics related to federated transfers)

PR #1323, opened 2021-06-23
[WIP][SYSTEMDS-2835] Python end-to-end tutorial

Gandagorn:
In this PR we implement a Python End-to-end tutorial for a standard ML classification task using the adult dataset.
The preprocessing of the data includes 
- one-hot-encoding
- missing value imputation
- outlier removal
- scaling
- binning
- recoding

Classification is done using logistic regression and a neural network. 
The results on the test set are evaluated using various metrics and a confusion matrix.


src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,243 @@
...
+            self.dataset_path_train,
+            schema=SCHEMA
+        )
+        F2 = self.sds.read(
+            self.dataset_path_test,
+            schema=SCHEMA
+        )
+        jspec = self.sds.read(self.dataset_jspec, data_type="scalar", value_type="string")
+        #scaling does not have effect yet. We need to replace labels in test set with the same string as in train dataset
+        X1, M1 = F1.rbind(F2).transform_encode(spec=jspec).compute()
```
Baunsgaard, line 192: you should not need to compute here.
use the output, X1 and call our builtin split

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,243 @@
...
+        col_length = len(X1[0])
+        X = X1[0:train_count, 0:col_length-1]
+        Y = X1[0:train_count, col_length-1:col_length].flatten()-1
+        # Test data
+        Xt = X1[train_count:train_count+test_count, 0:col_length-1]
+        Yt = X1[train_count:train_count+test_count, col_length-1:col_length].flatten()-1
+
+        _ , mean , sigma = scale(self.sds.from_numpy(X), True, True).compute()
```
Baunsgaard, line 200: here also we should not need to compute, but simply get the values out, and call apply on the test data.

The first result from this call is the scaled and shifted X, therefore the following second scaling of x is not needed.

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,243 @@
...
+        for x in range(0,105):
+            if not x in numerical_cols:
+                mean_copy[0][x] = 0
+                sigma_copy[0][x] = 1
+
+        mean = self.sds.from_numpy(mean_copy)
+        sigma = self.sds.from_numpy(sigma_copy)
+        X = self.sds.from_numpy(X)
+        Xt = self.sds.from_numpy(Xt)
+        X = scaleApply(winsorize(X, True), mean, sigma)
```
Baunsgaard, line 221: second scaling not needed. if you use the output from the first.

Baunsgaard: also, maybe apply winsorize before getting the scaling? (i'm not sure what the best practice here is)

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+        network = FFN_package.train(X, self.sds.from_numpy(Y), 1, 16, 0.01, 1)
+
+        self.assertTrue(type(network) is not None) # sourcing and training seems to works
+
+        FFN_package.save_model(network, '"model/python_FFN/"').compute(verbose=True)
+
+        # TODO This does not work yet, not sure what the problem is
+        # FFN_package.eval(Yt, Yt).compute()"""
```
Baunsgaard, line 238: i will make this work.

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,243 @@
...
+
+        network = FFN_package.train(X, Y, 1, 16, 0.01, 1)
+
+        self.assertTrue(type(network) is not None) # sourcing and training seems to works
+
+        FFN_package.save_model(network, '"model/python_FFN/"').compute(verbose=True)
+
+        # TODO This does not work yet, not sure what the problem is
+        #probs = FFN_package.predict(Xt, network).compute(True)
+        # FFN_package.eval(Yt, Yt).compute()
```
Baunsgaard, line 171: i will make this work.

Baunsgaard (commented):
Hi all, the PR looks to be developing in the right direction,
i would suggest focusing on the tutorial text for now so that there is a guide until a trained model. Make some text and accompanied by the different code parts that you already have working in the tests.

There is some compute calls in the tests that could be avoided to improve performance in the preprocessing, since every time you call compute the results are transferred into python and the execution therefore is slower than it needs to be.

I will address the eval call and fix that for the system.

Best regards
Sebastian

j143:
@Baunsgaard - Is it possible to keep datasets in the website instead of committing to this repo.

Baunsgaard:
> @Baunsgaard - Is it possible to keep datasets in the website instead of committing to this repo.

Yes. Will do when we get to merging.

src/main/python/tests/examples/tutorials/test_adult.py
Gandagorn: @Baunsgaard Hi I was just wondering if there has been some progress with this yet?

Baunsgaard: not yet.
i did not run it , will do next wreek

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -386,51 +386,35 @@ def test_level2(self):
...
-        Y = X1[0:train_count, col_length - 1:col_length].flatten()
-        # Test data
-        Xt = X1[train_count:train_count + test_count, 0:col_length - 1]
-        Yt = X1[train_count:train_count + test_count, col_length - 1:col_length].flatten()
+        PREPROCESS_package = self.sds.source(self.preprocess_src_path, "preprocess", print_imported_methods=True)
 
+        X = PREPROCESS_package.get_X(X1, train_count)
+        Y = PREPROCESS_package.get_Y(X1, train_count)
+        #We lose the column count information after using the Preprocess Package. This triggers an error on multilogregpredict. Otherwise its working
+        Xt = self.sds.from_numpy(np.array(PREPROCESS_package.get_Xt(X1, train_count).compute()))
```
codeyeeter, line 401: We lose the column count information after splitting the matrix in a sourced dml file. Is there a way around this issue without relying on this pretty bad workaround?

codeyeeter: @Baunsgaard

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -387,6 +387,11 @@ def test_level2(self):
         ################################################################################################################
         X1, M1 = X1.transform_encode(spec=jspec)
 
+        # better alternative for encoding
+        # X1, M = F1.transform_encode(spec=jspec)
+        # X2 = F2.transform_apply(spec=jspec, meta=M)
+        # testX2 = X2.compute(True)
```
Gandagorn: @Baunsgaard
Hi, we tried to implement a better version of the encoding using transform_apply, because otherwise we would calculate statistics for imputation on the whole data instead of just the training data.
Unfortunately we ran into the problem that the labels in the train data are slightly different than the labels in the test data ("<=50K" != "<=50K."), which hinders us in using the encoding M for the test data. We tried different methods for correcting the labels in the test data before encoding, however the main problem is that we have not found a good way to apply changes in a column of a frame, and we are also unable to use frames as arguments for a dml script function (it seems to be not supported yet with python?).
The simplest solution would be to remove the "." at the end of the test labels in the file itself. Other possible workarounds using just systemds functions seem to be quite complex and would probably miss the main goal of this tutorial.
How should we proceed?

martinhofwe:
Hi, 
we added the lvl1 and lvl2 rst code. Should we split the different levels into different files or should they all be in one file?  If they are all in the same file should we only comment on the new code bits in each progressive level or should every level be self-explanatory with the comments?
BR

src/main/python/tests/examples/tutorials/test_adult.py
Baunsgaard: if i understand correctly.

after you make anything into a matrix in the system like X and Y, you don't know how many columns and rows there is.
this is correct, since materializing the column and row count in the python API would require us to do processing, that we only evaluate after compute.
once you have the result back from compute you should have the correct number of columns and rows in numpy, but all intermediates you should not know.

src/main/python/tests/examples/tutorials/test_adult.py
Baunsgaard: two things here.

1. Yes, if there is a different number of labels in the training vs the test data, then the transform_apply does not work, currently there is no way around this. Do you really have 50k different classes in one of the features? I think you might be using a wrong encoding scheme for some of the columns.
2. Frames should be supported, but they are very new so there are bound to be bugs, the function definitions should specify frame if the input type is frame otherwise you should not call a function with frames, could you tell me which function you are trying to call, then i can try to fix it if there is a bug?

Baunsgaard:
> Hi @Baunsgaard,
> we added the lvl1 and lvl2 rst code. Should we split the different levels into different files or should they all be in one file? If they are all in the same file should we only comment on the new code bits in each progressive level or should every level be self-explanatory with the comments?
> BR

Your choice, do what makes the most sense to you.

src/main/python/tests/examples/tutorials/test_adult.py
Gandagorn: 1. There are only 2 different labels to predict, if a person makes "<=50K" a year or ">50K" a year. The 2 labels in the training set are denoted correctly as  either "<=50K" and ">50K", however in the test set they are called "<=50K."  and ">50K.". The dot at the end of the test set labels hinders us to use the transform_apply function correctly.
2. The dml function we want to use should be able to replace a string in the target column of the frame, similar to 

```
replace_target_frame = function(String replacement, String to_replace, Frame[Unknown] X)
  return(Frame[Unknown] X)
{
  for (i in 1:nrow(X)) {
    if (as.scalar(X[i, ncol(X)]) == to_replace) {
      X[i, ncol(X)] = replacement;
    }
  }
}
```

However when trying to load the function into python, we get the error 
> "NotImplementedError: Not Implemented type parsing for function def: Frame[Unknown]X"

Baunsgaard: 1. Okay now i understand the problem... a classic ... "someone made an error when making the dataset"...
2. Since you have this issue i just extended the frame to support the replace operation simply use

replace(target=X, pattern="<=50K.", replacement="<=50K")

Gandagorn: Thank you very much! Still I am confused how to use the replace function in the python environment? Because we load the data in python into a frame, and when trying to use `sds.source` with a dml file that includes a function with a frame as an argument, it throws the above error.

Baunsgaard: well i guess, i did not add it to the python API... will do.

Baunsgaard: Should be there now ... if you have a matrix or a frame simply call .replace("target","pattern")

Gandagorn: Thank you!
One problem still remains: The replace works for ">50K.", however it does not for "<=50K.". I played around a bit, and my guess is that the "<=" is interpreted as some kind of operator, because when I execute
`F2 = F2.replace("\<=50K.", "<=50K")`
I get this error

> Exception       : An error occurred while calling o0.prepareScript.
: org.apache.sysds.parser.ParseException: 
The following 4 parse issues were encountered:
#1 [line 2:32] [Syntax error] -> V1=replace(target=V0,pattern='\<=50K.',replacement='<=50K');
   no viable alternative at input 'V1=replace(target=V0,pattern==' ([@25,226:226='=',<7>,2:32])
#2 [line 2:32] [Syntax error] -> V1=replace(target=V0,pattern='\<=50K.',replacement='<=50K');
   extraneous input '=' expecting {'(', '[', '-', '+', '!', 'TRUE', 'FALSE', ID, INT, DOUBLE, COMMANDLINE_NAMED_ID, COMMANDLINE_POSITION_ID, STRING} ([@25,226:226='=',<7>,2:32])
#3 [line 2:32] [Validation error] -> V1=replace(target=V0,pattern='\<=50K.',replacement='<=50K');
   cannot parse the int value: '50K'
#4 [line 2:35] [Syntax error] -> V1=replace(target=V0,pattern='\<=50K.',replacement='<=50K');
   extraneous input 'K' expecting {')', ',', '^', '-', '+', '%*%', '%/%', '%%', '*', '/', '>', '>=', '<', '<=', '==', '!=', '&', '&&', '|', '||'} ([@27,229:229='K',<60>,2:35])

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+
+Note that nothing has been calculated yet. In SystemDS the calculation is executed once compute() is called.
+E.g. betas_res = betas.compute().
+
+We can now use the trained model to make predictions on the test data.
+
+.. code-block:: python
+
+    from systemds.operator.algorithm import multiLogRegPredict
+    [_, y_pred, acc] = multiLogRegPredict(Xt, betas, Yt)
```
Baunsgaard, line 101: i don't think the braces for [_, y_pred, acc] is needed here.

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+A confusion matrix is a useful tool to analyze the performance of the model and to obtain a better understanding
+which classes the model has difficulties separating.
+The confusionMatrix function takes the predicted labels and the true labels. It then returns the confusion matrix
+for the predictions and the confusion matrix averages of each true class.
+
+.. code-block:: python
+
+    from systemds.operator.algorithm import confusionMatrix
+    confusion_matrix_abs, _ = confusionMatrix(y_pred, Yt).compute()
+    print(confusion_matrix_abs)
```
Baunsgaard, line 120: maybe include what it prints in the guide.

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+In the full script, some steps are combined to reduce the overall script.
+
+.. code-block:: python
+
+    import numpy as np
+    from systemds.context import SystemDSContext
+    from systemds.examples.tutorials.adult import DataManager
+    from systemds.operator.algorithm import multiLogReg, multiLogRegPredict, confusionMatrix
+
+    sds = SystemDSContext()
```
Baunsgaard, line 134: here again 

with SystemDSContext() as sds:

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+
+This part of the tutorial shows an overview of the preprocessing capabilities that SystemDS has to offer.
+We will take an unprocessed dataset using the csv format and read it with SystemDS. Then do the heavy lifting for the preprocessing with SystemDS.
+As mentioned before, we want to use the Adult dataset for this task.
+If one wants to skip the explanation, then the full script is available at the end of this level.
+
+Step 1: Metadata and reading
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First of all, we need to download the dataset and create a mtd-file for specifying different metadata about the dataset.
```
Baunsgaard, line 170: specifying the metadata of the dataset

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+    dataset_jspec = "adult/jspec.json"
+    jspec = sds.read(dataset_jspec, data_type="scalar", value_type="string")
+
+Finally, we need to define a custom dml file to split the features from the labels and replace certain values, which we will need later.
+We will call this file "preprocess.dml":
+
+.. code-block::
+
+    get_X = function(matrix[double] X, int start, int stop)
+    return (matrix[double] returnVal) {
```
Baunsgaard, line 300: i would indent this line and the next

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+    returnVal = X[start:stop,1:ncol(X)-1]
+    }
+    get_Y = function(matrix[double] X, int start, int stop)
+    return (matrix[double] returnVal) {
+    returnVal = X[start:stop,ncol(X):ncol(X)]
```
Baunsgaard, line 305: indent

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+    }
+    replace_value = function(matrix[double] X, double pattern , double replacement)
+    return (matrix[double] returnVal) {
+    returnVal = replace(target=X, pattern=pattern, replacement=replacement)
```
Baunsgaard, line 309: indent

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+
+The print_imported_methods flag can be used to verify whether every method has been parsed correctly.
+
+Step 3: Applying the preprocessing steps
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Generally speaking, we would use the transform_encode function on the train dataset and with the returned encoding call the transform_apply function on the test dataset.
+In the case of the Adult dataset, we have inconsistent label names inside the test dataset and the train dataset, which is why we will show how we can deal with that using SystemDS.
+First of all, we combine the train and the test dataset by using the rbind() function. This function simply appends the Frame F2 at the end of Frame F1.
+This is necessary to ensure that the encoding is identical between train and test dataset.
```
Baunsgaard, line 329: the encoding should be extracted from the training data, and then applied to the test data, this is why we have two different outputs from transform_encode. you can also argue that you need to save the M, to be able to produce predictionson new data, since you need to apply the same preprocessing on unseen data.

if i remember correctly the code is something like:

X, M = X.transform_encode(spec=jspec)
X_test = X_test.transform_apply(M)

src/main/python/docs/source/guide/python_end_to_end_tut.rst
Baunsgaard: also replace should be callable directly on the matrix/frame in python.

src/main/python/docs/source/guide/python_end_to_end_tut.rst
Baunsgaard: ah i see, you make this function to enable slicing columns out.

I will add a slice function later, than remove the need of using this.

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+    PREPROCESS_package = sds.source(preprocess_src_path, "preprocess", print_imported_methods=True)
+
+    X1 = F1.rbind(F2)
+    X1, M1 = X1.transform_encode(spec=jspec)
+
+    X = PREPROCESS_package.get_X(X1, 1, train_count)
+    Y = PREPROCESS_package.get_Y(X1, 1, train_count)
+
+    Xt = PREPROCESS_package.get_X(X1, train_count, train_count+test_count)
+    Yt = PREPROCESS_package.get_Y(X1, train_count, train_count+test_count)
```
Baunsgaard, line 409: It would be nicer if we use the split function to construct the train and test datasets,

Baunsgaard: then we would not need to use source here in level 2, and make it cleaner i now understand why you commented on this in the emails

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+First, we need to source the dml file for neural networks.
+This file includes all the necessary functions for training, evaluating, and storing the model.
+The returned object of the source call is further used for calling the functions.
+The file can be found here:
+
+    - :doc:tests/examples/tutorials/neural_net_source.dml
+
+.. code-block:: python
+
+    FFN_package = sds.source(neural_net_src_path, "fnn", print_imported_methods=True))
```
Baunsgaard, line 480: i would not print the included methods in the source call.

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+Then we need to set the hyperparameters of the model.
+We choose to train for 1 epoch with a batch size of 16 and a learning rate of 0.01, which are common parameters for neural networks.
+The seed argument ensures that running the code again yields the same results.
+
+.. code-block:: python
+
+    epochs = 1
+    batch_size = 16
+    learning_rate = 0.01
+    seed = 42
```
Baunsgaard, line 496: 42 nice

src/main/python/systemds/examples/tutorials/adult.py
```diff
@@ -0,0 +1,162 @@
...
+                                    {'relationship': ['Wife', 'Own-child', 'Husband', 'Not-in-family', 'Other-relative', 'Unmarried']},
+                                    {'race': ['White', 'Asian-Pac-Islander', 'Amer-Indian-Eskimo', 'Other', 'Black']},
+                                    {'sex': ['Female', 'Male']},
+                                    {'native-country': ['United-States', 'Cambodia', 'England', 'Puerto-Rico', 'Canada', 'Germany', 'Outlying-US(Guam-USVI-etc)', 'India', 'Japan', 'Greece', 'South', 'China', 'Cuba', 'Iran', 'Honduras', 'Philippines', 'Italy', 'Poland', 'Jamaica', 'Vietnam', 'Mexico', 'Portugal', 'Ireland', 'France', 'Dominican-Republic', 'Laos', 'Ecuador', 'Taiwan', 'Haiti', 'Columbia', 'Hungary', 'Guatemala', 'Nicaragua', 'Scotland', 'Thailand', 'Yugoslavia', 'El-Salvador', 'Trinadad&Tobago', 'Peru', 'Hong', 'Holand-Netherlands']},
+                                    {'income': ['>50K', '<=50K']}]
+
+
+
```
Baunsgaard, line 70: some formatting needed

src/main/python/tests/examples/tutorials/neural_net_source.dml
```diff
@@ -0,0 +1,217 @@
...
+  # Use paramserv function
+  model_trained = paramserv(model=model_list, features=X, labels=y, 
+    val_features=matrix(0, rows=0, cols=0), val_labels=matrix(0, rows=0, cols=0), 
+    upd="./network/TwoNN.dml::gradients", agg="./network/TwoNN.dml::aggregation",
+    mode=mode, utype=utype, freq=freq, epochs=epochs, batchsize=batch_size,
+    k=workers, hyperparams=params, checkpointing="NONE")
+
+}
+
+save_model = function (list[unknown] model, String baseFolder){
```
Baunsgaard, line 203: remove the save model

src/main/python/tests/examples/tutorials/neural_net_source.dml
```diff
@@ -0,0 +1,217 @@
...
+  b2  = as.matrix(model[5])
+  b3  = as.matrix(model[6])
+
+  write(W1, (baseFolder + "/W1.data"), format="binary")
+  write(W2, (baseFolder + "/W2.data"), format="binary")
+  write(W3, (baseFolder + "/W3.data"), format="binary")
+  write(b1, (baseFolder + "/b1.data"), format="binary")
+  write(b2, (baseFolder + "/b2.data") , format="binary")
+  write(b3, (baseFolder + "/b3.data") , format="binary")
+}
```
Baunsgaard, line 217: add newline in the end of the script

src/main/python/tests/examples/tutorials/neural_net_source.dml
```diff
@@ -0,0 +1,217 @@
...
+      b2 = as.matrix(model_updated[5])
+      b3 = as.matrix(model_updated[6])
+
+    }
+  }
+
+  model_trained = list(W1, W2, W3, b1, b2, b3)
+}
+
+train_paramserv = function(matrix[Double] X, matrix[Double] y,
```
Baunsgaard, line 179: maybe remove the paramserv

src/main/python/docs/source/guide/python_end_to_end_tut.rst
```diff
@@ -0,0 +1,561 @@
...
+    FFN_package = sds.source(neural_net_src_path, "fnn", print_imported_methods=True)
+
+    epochs = 1
+    batch_size = 16
+    learning_rate = 0.01
+    seed = 42
+
+    network = FFN_package.train(X, Y, epochs, batch_size, learning_rate, seed)
+
+    FFN_package.save_model(network, '"model/python_FFN/"').compute(verbose=True)
```
Baunsgaard, line 561: we can now save lists directly. therefore just simply call write, also when reading in the model again later, simply call read.

src/main/python/tests/examples/tutorials/preprocess.dml
```diff
@@ -0,0 +1,46 @@
...
+  returnVal = X[start:stop,ncol(X):ncol(X)]
+}
+
+replace_value = function(matrix[double] X,
+                 double pattern , double replacement)
+    return (matrix[double] returnVal) {
+  returnVal = replace(target=X, pattern=pattern, replacement=replacement)
+}
+
+#replace_target_frame = function(String replacement, String to_replace, Frame[Unknown] X)
```
Baunsgaard, line 38: remove commented function

src/main/python/tests/examples/tutorials/preprocess.dml
```diff
@@ -0,0 +1,46 @@
...
+#  return(Frame[Unknown] X)
+#{
+#  for (i in 1:nrow(X)) {
+#    if (as.scalar(X[i, ncol(X)]) == to_replace) {
+#      X[i, ncol(X)] = replacement;
+#    }
+#  }
+#}
```
Baunsgaard, line 46: add newline

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,324 @@
...
+        X = PREPROCESS_package.get_X(X1, 1, train_count)
+        Y = PREPROCESS_package.get_Y(X1, 1, train_count)
+
+        Xt = PREPROCESS_package.get_X(X1, train_count, train_count+test_count)
+        Yt = PREPROCESS_package.get_Y(X1, train_count, train_count+test_count)
+
+        Yt = PREPROCESS_package.replace_value(Yt, 3.0, 1.0)
+        Yt = PREPROCESS_package.replace_value(Yt, 4.0, 2.0)
+
+        # better alternative for encoding. This was intended, but it does not work
```
Baunsgaard, line 205: still the case?

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,324 @@
...
+        #F2 = F2.replace("<=50K.", "<=50K")
+        #F2 = F2.replace(">50K.", ">50K")
```
Baunsgaard, line 207: remove the commented code unless it has to do with the comment, if it does then somehow clearly mark that the comment is associated with this code

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,324 @@
...
+
+        epochs = 1
+        batch_size = 16
+        learning_rate = 0.01
+        seed = 42
+
+        network = FFN_package.train(X, Y, epochs, batch_size, learning_rate, seed)
+
+        """
+        If more ressources are available, one can also choose to train the model using a parameter server.
```
Baunsgaard, line 289: in general it is actually a bit faster with the parameter server.
the use of the parameter server architecture is not necessarily multiple machines

src/main/python/tests/examples/tutorials/test_adult.py
```diff
@@ -0,0 +1,324 @@
...
+        FFN_package.save_model(network, '"model/python_FFN/"').compute(verbose=True)
+
+        """
+        Next we evaluate our network on the test set which was not used for training.
+        The predict function with the test features and our trained network returns a matrix of class probabilities.
+        This matrix contains for each test sample the probabilities for each class.
+        For predicting the most likely class of a sample, we choose the class with the highest probability.
+        """
+        ################################################################################################################
+        #probs = FFN_package.predict(Xt, network)
```
Baunsgaard, line 311: this was not working, correct?

src/main/python/tests/manual_tests/multi_log_reg_adult.py
```diff
@@ -0,0 +1,27 @@
...
+# under the License.
+#
+# -------------------------------------------------------------
+
+from systemds.context import SystemDSContext
+from systemds.operator.algorithm import multiLogReg, multiLogRegPredict
+from systemds.examples.tutorials.adult import DataManager
+
+d = DataManager()
+
```
Baunsgaard, line 27: does this file actually test?

src/test/resources/datasets/adult/jspec.json
```diff
@@ -0,0 +1,24 @@
...
+     ,{ "name": "hours-per-week", "method": "global_mean" }
+     ,{ "name": "native-country"        , "method": "global_mode" }
+    ],
+    "bin":
+    [ { "name": "age"  , "method": "equi-width", "numbins": 3 }],
+   "dummycode": ["age", "workclass", "education", "marital-status", "occupation", "relationship", "race", "sex", "native-country"],
+   "recode": ["income"]
+
+
+}
```
Baunsgaard, line 24: formatting

Baunsgaard (commented):
LGTM, the comments i have left are something we can fix while merging.
i have a few questions in the comments that i would like some answers to.

src/main/python/tests/examples/tutorials/test_adult.py
martinhofwe: Yes, correct (predict seems to be the problem).

src/main/python/tests/examples/tutorials/test_adult.py
Gandagorn: Yes. The replace function you made for us to clean the target labels works for one label correctly, however not for the other (See our discussion that started on 5.7.).
The code that is commented out is how it should be done, using the replace function and the transform_apply function. The not-commented code is a temporary, working solution, but it is not optimal.

src/main/python/tests/manual_tests/multi_log_reg_adult.py
codeyeeter: No, we initially wanted to add some testing functionality in that file. Seems like we missed deleting the file.

codeyeeter:
@Baunsgaard 
We went through the comments and tried to answer all of your questions, based on your initial comment can we assume that you will do the code and formatting changes while merging? Or is there still anything expected from us?
Best regards,
Jörg Rainer

Baunsgaard:
> @Baunsgaard
> We went through the comments and tried to answer all of your questions, based on your initial comment can we assume that you will do the code and formatting changes while merging? Or is there still anything expected from us?
> Best regards,
> Jörg Rainer

I will do the rest.Thanks!

Baunsgaard:
Thanks for the PR, i have now merged it.

While merging i have moved the data to our webpage, and changed some of the tests.
Thanks for the contribution.

PR #1325, opened 2021-06-26
Pypi publish script

j143:
**Usage instruction:**

```sh
$ systemds > ./dev/release/pypi_upload.sh
```

minimize variables, add further documentation and check for the artifacts.

Published test artifacts at https://test.pypi.org/project/systemds/2.1.0rc6/

### DIFF
--- a/dev/release/pypi_upload.sh
+++ b/dev/release/pypi_upload.sh
@@ -20,12 +20,15 @@
 #
 #-------------------------------------------------------------
 
+SELF=$(cd $(dirname $0) && pwd)
+# . "$SELF/release-utils.sh"
+
 #BASE_DIR=$(pwd)
 #BASE_DIR="/c/virtual\ D/SystemDS/systemds"
-BASE_DIR="../.." #points to systemds directory
-RELEASE_WORK_DIR=$BASE_DIR/target/release2
-RELEASE_VERSION=2.0.0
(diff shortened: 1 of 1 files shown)


phaniarnab:
I have reservations for use of pushd and popd. I'd avoid using sub-shell unless absolutely necessary.

j143:
Should we use `cd`?

phaniarnab:
> Should we use `cd`?
> 
> The `pushd` and `popd` may work like, which is a built-in function.
> 
> ```shell
> pushd ()
> {
>     dirname=$1
>     DIR_STACK="$dirname ${DIR_STACK:-$PWD' '}"
>     cd ${dirname:?"missing directory name."}
>     echo "$DIR_STACK"
> }
>    
> popd ()
> {
>     DIR_STACK=${DIR_STACK#* }
>     cd ${DIR_STACK%% *}
>     echo "$PWD"
> }
> ```
> 
> source: https://www.oreilly.com/library/view/learning-the-bash/1565923472/ch04s05.html
> 
> Does this spawn a subshell?

Yes. Directory stack makes sense when working with multiple directories. 
For our use case, you can save the `pwd` in a local variable, use `cd` to enter your target folder, and later apply something like `cd $pwd` to get back to your original location if needed.

PR #1326, opened 2021-06-26
[AMLS] EfficientNet to Builtin

Tsebeb:
Implemented a minified version of Efficient-Net B0 with an additional mini test script on the MNist data set.
This includes a few new layer implementations:

- SILU nonlinearity Layer
- Adaptive Global Avg Pooling Layer
- Inverted Residual Mobile Layer (MBConv)

We only used a single MBConv layer instead of the total of 16. 
The top and stem part are however identical to Efficient-Net B0.


mboehm7:
LGTM - thanks for the patch @Tsebeb. This is a good start and we'll promote this to a builtin function once the necessary mechanics are in place. During the merge I simplified some instances of list construction (via list range indexing), added TODOs for tests and scaling up the network architecture, and fixed all formatting issues (mostly related to indentation).

PR #1329, opened 2021-06-29
[MINOR] Sync python project_version with pom.

phaniarnab:


j143:
We can update the UpdateAPI.sh in this PR itself, so that the documentation will also be updated to pom version.

PR #1335, opened 2021-07-03
[SYSTEMDS-2982] Federated quaternary instructions w/ aligned matrices

ywcb00:
Hi,
This PR adds support for federated *quaternary* instructions with multiple *aligned* federated matrices.
I also changed the tests so that we are testing the different possibilities of alignments.

I commented out the test _federatedWeightedUnaryMatrixMultPow2Spark_ because it is producing _NaNs_, and hence it is failing. I haven't found the reason for this yet, but somehow it depends on the sparsity of the input matrix.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
@@ -21,7 +21,6 @@ package org.apache.sysds.runtime.instructions.fed;
 
 import java.util.concurrent.Future;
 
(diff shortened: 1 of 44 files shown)


mboehm7:
LGTM - thanks @ywcb00. During the merge I only fixed a few warnings and reverted the exception handling as it was failing the tests. I'll also take care of the spark issue.

mboehm7:
@ywcb00: After a more detailed look, I am a bit confused why WSigmoid and WuMM consolidate the output locally at the coordinator via rbind/cbind. Shouldn't they retrain the output as a federated matrix if there is no aggregation? Could you please have a look when you get a chance?

PR #1345, opened 2021-07-13
Fix wrong brackets in the document

XorSum:
```
An Error Occured : 
        LanguageException -- Unsupported Parameters : ERROR:  ........ Expecting scalar parameter for function SEQ
```


mboehm7:
LGTM - thanks for catching this issue @XorSum.

PR #1350, opened 2021-07-26
[SYSTEMDS] Basic Auto-Differentiation for Affine Layer

Shafaq-Siddiqi:
- The autoDiff() parameterizedBuiltin takes the X, the last layer's output, and the list of weights and bias as inputs and returns their 
     derivatives.
  - It also takes lineage as input for future extensions.


Shafaq-Siddiqi:
@mboehm7 as discussed, the second commit has a modified implementation where all the needed inputs are computed from lineage instead of being passed into the function call. 
Thank you.

src/main/java/org/apache/sysds/runtime/util/AutoDiff.java
```diff
@@ -0,0 +1,135 @@
...
+				String lin = lineage.get(i).toString();
+				// get rid of foo flag
+				lin = lin.replace("foo", "");
+
+				// get lineage of X, Y, W and dout for affine backward
+				String[] linInst = lin.split("\n");
+				// assumption first line in the lineage is X and second is weight
+				MatrixObject dout = (MatrixObject)LineageRecomputeUtils.parseNComputeLineageTrace(lin, null);
+				MatrixObject X =  (MatrixObject)LineageRecomputeUtils.parseNComputeLineageTrace(linInst[0], null);
+				MatrixObject W =  (MatrixObject)LineageRecomputeUtils.parseNComputeLineageTrace(linInst[1], null);
```
phaniarnab, line 125: Does this mean recomputing the forward passes again?

src/main/java/org/apache/sysds/runtime/util/AutoDiff.java
```diff
@@ -0,0 +1,135 @@
...
+			if(layer.equals("affine"))
+			{
+				// get the lineage
+				String lin = lineage.get(i).toString();
+				// get rid of foo flag
+				lin = lin.replace("foo", "");
+
+				// get lineage of X, Y, W and dout for affine backward
+				String[] linInst = lin.split("\n");
+				// assumption first line in the lineage is X and second is weight
```
phaniarnab, line 122: This can be a risky assumption. Any small instruction in the middle can lead to wrong results.

src/main/java/org/apache/sysds/runtime/util/AutoDiff.java
Shafaq-Siddiqi: Yes, For now, I am recomputing it but we can change it and pass it in as a parameter.

Shafaq-Siddiqi:
@mboehm7 @phaniarnab  This third commit contains a slightly different implementation of autoDiff in which
the derivations are computed while creating the HOP DAG for lineage instructions.
Now, after parsing the lineage foreach lineage instruction instead of creating the HOP DAG for
instruction the HOP DAG for its derivation is being created.
This code is only created and tested for affine layer and it needs extensions for networks.

phaniarnab:
Thanks for the patch @Shafaq-Siddiqi. Can you please clean up the unnecessary methods not related to the last commit and make sure that the tests are passing. I will then check in the changes.

Shafaq-Siddiqi:
> Thanks for the patch @Shafaq-Siddiqi. Can you please clean up the unnecessary methods not related to the last commit and make sure that the tests are passing. I will then check in the changes.

Thank you @phaniarnab, I have pushed a cleanup commit.

phaniarnab:
Thanks again @Shafaq-Siddiqi.
I fixed a few syntax issues, revert files with unnecessary/incomplete changes, added debug prints for those constructed plans under a debug flag, and added a case for `Literal` in the Hop construction logic.

PR #1351, opened 2021-07-26
[SYSTEMDS-2972] Dependency Task execution for transform encode

ilovemesomeramen:
This PR adds a Framework for dependent Task execution consisting of the `DependencyTask` and `DependencyThreadPool` as well as the implementation of build and apply in said Framework.

It is now possible for different encoders to provide tasks which best fit the given data characteristics by overriding the `getApplyTasks` and `getBuildTasks` functions. These functions can return a DAG of Tasks which will be integrated into the DAG of the whole Matrix/Frame. Subsequently this DAG is then executed in the `DependencyThreadPool`. This approach enables the possibility of trying different execution orders without the need to make major changes to the code.

At the moment only the Dummycoders implement a different apply method compared to the default. This will be changed in future PRs.

Currently there are two approaches implemented staged and non staged.
During staged execution the building and applying steps take place one after another. Which means the building will be done for the whole Matrix/Frame and then the apply takes place.
Non staged execution on the other hand constructs one large DAG for building and applying.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 		<version>18</version>
 	</parent>
 	<groupId>org.apache.systemds</groupId>
-	<version>2.2.0-SNAPSHOT</version>
+	<version>2.2.0-SNAPSHOT-dev</version>
 	<artifactId>systemds</artifactId>
 	<packaging>jar</packaging>
 	<name>SystemDS</name>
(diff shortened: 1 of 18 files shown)


phaniarnab:
Thanks for the patch, Lukas. I fixed a few typos and added SuppressWarnings to resolve the warning for unused methods.

PR #1354, opened 2021-07-29
builtin function stratified statistics

juulez:


mboehm7:
LGTM - thanks for the patch @juulez. During the merge I only made minor changes related to formatting and added a TODO for vectorizing the `fStat_tailprob` function (which first requires extended the support of density distribution functions for matrix inputs).

PR #1355, opened 2021-07-30
[SYSTEMDS-2972] Dependency Task execution for transform encode

phaniarnab:
This patch adds a Framework for dependent Task execution consisting of
the DependencyTask and DependencyThreadPool, as well as the
implementation of build and apply in said Framework.
It is now possible for different encoders to provide tasks which best
fit the given data characteristics by overriding the getApplyTasks and
getBuildTasks functions. These functions can return a DAG of Tasks which
will be integrated into the DAG of the whole Matrix/Frame. Subsequently
this DAG is then executed in the DependencyThreadPool. This approach
enables the possibility of trying different execution orders without the
need to make major changes to the code.


phaniarnab:
PR to run tests.

PR #1358, opened 2021-08-02
[SYSTEMDS-3067] Model Averaging in paramserver

atefeh-asayesh:
in this PR we use an optional Boolean parameter "modelAvg" to select the type of updating or averaging the model in param server. In this PR, it is implemented for both federated and local param server. Moreover, two java test files have been added.

"AvgModelFederatedParamservTest.java" is created to test in federated mode.
"ParamservLocalNNAveragingTest.java" is created to test in local mode.


mboehm7:
LGTM - thanks for finalizing this PR @atefeh-asayesh. I fixed the remaining warnings, unnecessary imports (of unintended packages) and made a few simplifications of the core update logic.

PR #1360, opened 2021-08-07
[SYSTEMDS-831] Builtin t-SNE algorithm

AlexanderErtl:
Testing was done using visual inspection and a reference R implementation (see comments in Java Test file), since it is not possible to directly use an R Test, as the random number Generation is done differently. The CP test runs fine (3s at 1000 iterations), the Spark test takes extremely long (30s at 10 iterations) and therefore never was run to completion.


mboehm7:
LGTM - thanks for creating the builtin function and test @tim-sagaster. Also thanks for the original algorithm @iyounus. 

During the merge I made a few tweaks though, which I mention here just for completeness. First, I fixed the test to use a relatively large epsilon so it passes (so far it compared equivalence of doubles which even for tiny discrepancies would fail; it seems ok, but there are a few values where the absolute error goes up to 10% of the value range; we will follow up to understand the algorithm-level differences; also, for iterative algorithms it's not useful to force all operations into distributed Spark ops so nothing to worry there). Second, I left a few TODOs to consolidate the distance computation with the respective builtin and compare against R instead of inlining expected outputs. Finally, I also fixed some formatting and stdout issues and unnecessary imports in the test.

PR #1361, opened 2021-08-09
[SYSTEMDS-3086] Federated WDivMM with Federated Output

ywcb00:
Hi,
This PR adds support to keep the output of the federated WDivMM instruction on the federated sites instead of binding it on the coordinator.

Thanks for review :)


mboehm7:
LGTM - thanks @ywcb00 for catching and fixing this issue.

PR #1363, opened 2021-08-12
[SYSTEMDS-3093] Check and replace SP instructions with federated instructions

OlgaOvcharenko:
This PR adds federated reblock, cast, cumulative offset instructions, and extends FEDInstructionUtils.checkAndReplaceSP.
Thanks for the review.


mboehm7:
LGTM - thanks for this major extension @OlgaOvcharenko. During the merge, I fixed the merge conflicts (in the future please rebase if possible), fixed the warnings and a few minor issues. As a next step, we should think about consolidating the now replicated federated matrix multiplication (in fed mapmm). Maybe we get away with dedicated instruction string handling and otherwise can channel everything through a single matrix multiplication (because there are of course many more spark matrix multiplications such as cpmm and rmm).

PR #1365, opened 2021-08-16
Nbatches parameter

atefeh-asayesh:
In this PR we add freq=NBATCHES along with the optional parameter nbatches to extend  paramserv builtin for NBATCH updates.
In this step, It has been implemented for both local (computeNBatches(dataSize, batchIter)) and federated parameter servers (computeWithNBatchUpdates in FederatedPSControlThread.java).

### DIFF
--- a/src/main/java/org/apache/sysds/parser/ParameterizedBuiltinFunctionExpression.java
+++ b/src/main/java/org/apache/sysds/parser/ParameterizedBuiltinFunctionExpression.java
@@ -112,7 +112,7 @@ public class ParameterizedBuiltinFunctionExpression extends DataIdentifier
 	}
 
 	public ParameterizedBuiltinFunctionExpression(Builtins op,
-			LinkedHashMap<String, Expression> varParams, ParseInfo parseInfo) {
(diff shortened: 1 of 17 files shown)


atefeh-asayesh:
Draft- Some small changes would be done.

mboehm7:
LGTM - thanks @atefeh-asayesh  for the extension of the parameter server. During the merge I mode the following changes:

* Fixed the local paramserv worker which pulled and pushed at the same iteration (for all but the first), although it needs to push the iteration before it pulls a new model and then work with that for nbatches.
* Fixed the federated paramserv implementation which did not update the model correctly under modelAvg (last model update missing, and outside `computeGradientsForNBatches` the returned model was assumed as gradients in an additional model update). The related changes also simplified some duplicated code.
* Fixed the computation of `numSetsPerEpocNbatches` which due to integer division might have lost a few batches.
* Removed unused nbatch member variables and test methods, fixed test methods which used the constructor name
* Fixed the federated tests to use a longer wait time after the last worker to ensure everything is ready before the experiments start. 
* Fixed some minor formatting issues and warnings

mboehm7:
Let's close this PR as it was merged with #1382

PR #1366, opened 2021-08-17
[SYSTEMDS-3097] Fix simple-aggregation-example

mar-volk:
- See Jira ticket;

- Without the change in line 85 the tutorial will not run.
- The change in line 57 makes the script return the expected result.


Baunsgaard:
I don't think it is needed to add the header to the file csv file. But the federated line change indeed needs to happen. Thanks for the  fix.

Baunsgaard:
> I don't think it is needed to add the header to the file csv file. But the federated line change indeed needs to happen. Thanks for the fix.

Oh i see in the Jira task you found a bug. For reading the matrix when federated. We will have to address this specifically in another commit.

Baunsgaard:
Thanks for pointing out an error in the CSV parsing when federated and the error in the docs.

PR #1368, opened 2021-08-21
[SYSTEMDS][IOGEN] Automatically Generate Reader

fathollahzadeh:
This PR adds a new Reader feature. The primary goal of this work is to extract the file format properties of the file from a sample. Then, generate a reader based on extracted properties. This is a draft version of the inferred section. 

Tested file formats:
1. Full CSV file, CSV file with missed values, Missed values filled with arbitrary NA string
2. LibSVM file with an arbitrary order of stored index and value.
3. MtrixMarket with General, symmetric, and skew-symmetric.

The tests were tested with unique and duplicated values.


fathollahzadeh:
Wrong in changed files

PR #1369, opened 2021-08-21
[SYSTEMDS][IOGEN] Auto Generate Reader

fathollahzadeh:
This PR adds a new Reader feature. The primary goal of this work is to extract the file format properties of the file from a sample. Then, generate a reader based on extracted properties. This is a draft version of the inferred section.

Tested file formats:
1. Full CSV file, CSV file with missed values, Missed values filled with arbitrary NA string
2. LibSVM file with an arbitrary order of stored index and value.
3. MtrixMarket with General, symmetric, and skew-symmetric.

I tested the tests with unique and duplicate values.


mboehm7:
Thanks @fathollahzadeh - that's a good first step and we can bring it in as experimental after some additional cleanups (dependencies, imports). As next steps we should run some experiments for better understanding the runtime breakdown, and then generalize the hard-coded readers to code-generated reads that utilize sub-templates for key concepts.

fathollahzadeh:
> Thanks @fathollahzadeh - that's a good first step and we can bring it in as experimental after some additional cleanups (dependencies, imports). As next steps we should run some experiments for better understanding the runtime breakdown, and then generalize the hard-coded readers to code-generated reads that utilize sub-templates for key concepts.

Thanks @mboehm7 for the first review and your supports. The new updates are supporting auto-generate readers for Matrix and Frame. In matrix readers, we are supporting general, symmetric (upper, lower triangular), and skew-symmetric. At this PR we do not support pattern matching. For Frame readers, we are supporting three value types (STRING, Numeric, and BOOLEAN). Unknown value types in Frame are our future work. 
For the test side, we don't have any DML scripts. We are calling pure java public methods for generating readers. The new DML script is also our future work. 

Thanks for the review.

mboehm7:
LGTM - thanks @fathollahzadeh  for finalizing this initial version of the IOGEN framework. So far this implementation uses generalized readers for types of structured inputs - it's fine to postpone the actual code generation until we come to structure that requires generation for performance.

During the merge I made a number of minor changes:
* Included the iogen tests into our github test workflow
* Fixed two failing tests (set to ignore) which missed some input file. Similarly other tests throw exceptions due to missing files but succeed. Please do a separate PR for fixing these tests including better checks for expected results.
* Fixed a number of warnings of unused variables, and simplified some verbose code snippets
* Fixed the parsing of ints via the fast string tokenizer (instead of `(int)Double.parseDouble()` we should use `Integer.parseInt()`)
* In the future please avoid wildcard imports and reformatting files (e.g., `DataConverter` here) where no changes have been made.

PR #1380, opened 2021-09-02
[SYSTEMDS-3101] Federated Spoof Instruction with Federated Output

ywcb00:
Hi,
This PR adds support to keep the output of the federated spoof instruction on the federated sites instead of binding it on the coordinator.
I've always split the method _setOutput_ into _setFedOutput_ and _aggResult_ to clearly differentiate between creating a federated output and aggregating the results locally.

Thanks for review :)


mboehm7:
LGTM - thanks @ywcb00  for the cleanup and fixing this issue. During the merge, I only removed warnings and made minor formatting changes.

PR #1367, opened 2021-08-18
[SYSTEMDS-3018] Basic Federated Planner

sebwrede:
This PR contains various additions necessary for building a federated planner, which will generate and select execution plans for federated executions. The cost estimation currently added is not connected to the decisions made by the federated planner. This will be done in later PRs, where the data structure for describing the relation between Hops with different FOUT/LOUT/NONE values and the corresponding costs will be implemented.

PR #1381, opened 2021-09-02
[SYSTEMDS-3086] Federated WDivMM - Federated Output

ywcb00:
Hi,
This PR adds the overwritten changes from the commit [5e3863842](https://github.com/apache/systemds/commit/5e386384296a781cf6d3adf1ffe52105e4356407) again.
It was overwritten by [6649bcd5b](https://github.com/apache/systemds/commit/6649bcd5bc8c32abae76aaf54315b48d79ff5397).

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWDivMMFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWDivMMFEDInstruction.java
@@ -161,7 +161,7 @@ public class QuaternaryWDivMMFEDInstruction extends QuaternaryFEDInstruction
 
 			FederatedRequest frC = null;
 			if((wdivmm_type.isLeft() && X.isFederated(FType.ROW))
-				|| (wdivmm_type.isRight() && X.isFederated(FType.COL))) { // output aggregated locally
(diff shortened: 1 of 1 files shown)


sebwrede:
Do you know if it was overwritten on purpose?

sebwrede:
We can merge this again, but I think that if the change could be overwritten without any tests failing then it is a hint that we should add some tests that actually needs this additional processing case. This could be done in another PR or in the current PR.

ywcb00:
It wasn't overwritten on purpose.

The additional methods are already taken from the current test cases, but the tests do not verify that the output is kept federated.
We should definitely change the tests so that they fail if the output is not federated when it should be.
Since this is the case with several instructions, I will create a new jira task for this and add it with another PR.

Btw, do you know any better possibility to verify the federated output than performing an additional (unary) instruction on it (e.g., _fed_!=_) and verifying that this additional instruction is performed federated?

sebwrede:
Creating a new Jira task sounds like a good idea. 
Computing an additional instruction and looking at the heavy-hitters is also the way I test federated outputs. You can see that in my current PR #1367.

mboehm7:
LGTM - thanks for fixing this @ywcb00. The overwrite was my fault, it was unintentional and happened during a massive merge of ~40 commits with many conflicts.

PR #1388, opened 2021-09-08
[SYSTEMDS-2972] Transformencode sparse improvements

phaniarnab:
This PR introduces the sparse implementations for the transform encoders.
Furthermore, this adds support for row partitioning, statistics, and
debug logging. Now we can enable multithreaded transorm via a flag
in the config file, "parallel.encode".


phaniarnab:
To run the tests.

PR #1389, opened 2021-09-08
[SYSTEMDS-3114] Support for replace on frames CP

OlgaOvcharenko:
Modification of the existing frames replace, now replace is supported not only for string, but also integer, double and boolean columns.


mboehm7:
Thanks for starting the work on exposing the frame replace and making it more robust. For use at script level, please introduce copy on write semantics with shallow copy of unmodified columns. Right now the implementation does a shallow copy for all columns and then modifies these columns in place, which changes both the output and input and thus, yields incorrect results (e.g., B=replace(A) also modifies A). Additional, please double check if the `UtilFunctions.objectToObject` and similar functions could be used to avoid conversions to string.

mboehm7:
sorry, you're right. I didn't realize that we already use this in a basic form, and the `new FrameBlock(this)` does an internal deep copy. Maybe try to shallow copy unaffected columns, but other than that it should be fine, but we need to investigate the failing tests.

PR #1390, opened 2021-09-09
[WIP][SYSTEMDS-3124] Federated Binary Aggregation and (Map)MM - Federated Output

ywcb00:
Hi,
This PR contains a small fix for the MV case to keep the output federated. The problem was that the branch to create a federated output was only taken if it was forced federated. I have now changed it so that the partial output is only set if it is forced federated, and the federated output is set if it is not forced local.
The other changes inside AggregateBinaryFEDInstruction should not change the functional behaviour of the instruction. I only tried to cleanup some code redundancies that could be merged.

This PR also includes an implicit change to the FederatedTokenizeTest, where I removed the creation of the fourth thread because it is never used.

Thanks for review :)


ywcb00:
Since I now know that we should not create a federated output in the MV case (unless it is forced federated) for reasons of performance, I'm taking over the refactoring of the two instructions (AggregateBinaryFEDInstruction and MMFEDInstruction) to SYSTEMDS-3206 and continue there with the fix of the endless loop mentioned above.

PR #1393, opened 2021-09-19
[SYSTEMDS-3098] Add synchronization to async. broadcast

phaniarnab:
This patch wraps the creation of partitioned broadcast handle
code inside a synchronized block to remove redundant partitioning
by the CP or the new early-broadcast thread.
Moreover, this patch fixes a bug in broadcast count stat collection.


phaniarnab:
Running regression tests.

PR #1407, opened 2021-10-01
[MINOR] Fixes in cleaning pipelines

Shafaq-Siddiqi:
- fix schema issues in getDirtyScore() method
  - fix issues in utils::gridSearch when optimizing for non-binary data
  - fix a minor bug in regular expression of detectSchemaFromRow as the existing
    RE was identifying "123-456" as a double value


Shafaq-Siddiqi:
Merged into Master

PR #1409, opened 2021-10-04
[SYSTEMDS-3148] Federated ALS Performance Tests

ywcb00:
Hi,
This PR adds the first federated performance test, ALS-CG.
The script generates the data, creates a specific amount (in our case 5) of federated worker processes on localhost, splits the data into partitions, creates a federated object from the different partitions and workers, runs the ALS-CG algorithm with the federated object as data input, and kills the woker processes again.
I tried to make the shell scripts for starting and killing the federated workers, and also the dml script for splitting the data and creating a federated object, as general as possible so that the same scripts can be used the further federated performance tests.

Thanks for review :)


scripts/perftest/fed/runALSFed.sh
```diff
@@ -0,0 +1,58 @@
...
+  # split the generated data into paritions and create a federated object
+  ${CMD} -f ${BASEPATH}/data/splitAndMakeFederated.dml \
+    --config ${BASEPATH}/../conf/SystemDS-config.xml \
+    --nvargs data=${DATADIR}/X${d} nSplit=$NUMFED transposed=FALSE \
+      target=${DATADIR}/X${d}_fed.json hosts=${DATADIR}/workers/hosts fmt="csv"
+
+  echo "-- Running ALS-CG with federated data ("$d") on "$NUMFED" federated workers" >> results/times.txt
+
+  # run the als algorithm on the federated object
+  ${BASEPATH}/../runALS.sh ${DATADIR}/X${d}_fed.json $MAXITR $DATADIR systemds;
```
ywcb00, line 55: What do you think about the decision of calling the main algorithm script from the outside and passing the federated object as input?

Baunsgaard: in general i like it the best if it is from the outside, since then the script is the same if you use federated input or normal local input.

sebwrede:
> Hi, This PR adds the first federated performance test, ALS-CG. The script generates the data, creates a specific amount (in our case 5) of federated worker processes on localhost, splits the data into partitions, creates a federated object from the different partitions and workers, runs the ALS-CG algorithm with the federated object as data input, and kills the woker processes again. I tried to make the shell scripts for starting and killing the federated workers, and also the dml script for splitting the data and creating a federated object, as general as possible so that the same scripts can be used the further federated performance tests.
> 
> Thanks for review :)

Thanks for this PR!

I ran the performance test and have two comments: 

1. I get the following message continuously during the execution: `TRACE nio.NioEventLoop: instrumented a special java.util.Set into: sun.nio.ch.EPollSelectorImpl@`. Do you have any idea why this happens and how to fix it? 
2. The script prints `Max iteration achieved but not converged!`. Could we increase the maximum number of iterations or change something else to make sure that it reaches its target? Otherwise it is difficult to determine from the prints whether the algorithm actually succeeded or if it was even close to succeeding.

sebwrede:
> Thank you for the comments @sebwrede :)
> 
> I set the log level of io.netty to INFO to avoid the trace logs. I also set the maximum number of iterations to 100 and the threshold to 0.001 so that the algorithm converges.

LGTM. I think this is ready to merge, but I don't know if we should wait until after the next release. What do you think, @phaniarnab?

Another point: I noticed that when I run `runAll.sh` then I get the error:
`Error in ./runAllMultinomial.sh on line 53`
`Error in ./runAllRegression.sh on line 51`
This is not related to this PR. It also happens on master. I don't know if it is something with my setup, but it does not happen with the call to `runAllFed.sh`, so that is good :smiley:.

PR #1419, opened 2021-10-21
[SYSTEMDS-3125] Verify Federated Output

ywcb00:
Hi,
This PR adds a few checks to the primitve function tests to ensure that the output is federated.
I always verified it by simply adding an additional instruction (uak+) - which is then executed federated - and checking that the heavy hitters contain this specific federated instruction (fed_uak+).

Thanks for review :)


mboehm7:
LGTM - thanks for the patch @ywcb00. During merge I only made a minor modification in the `FederatedWeightedDivMatrixMultTest` test to use our util `CollectionUtils.asSet`.

PR #1421, opened 2021-10-25
[SYSTEMDS-3185] Federated Lookup Table

ywcb00:
Hi,
This PR introduces the Federated Lookup Table as a first step for supporting Federated Multi Tenancy.
The Federated Lookup Table maps a Coordinator Identifier (FedUniqueCoordID) to an individual ExecutionContextMap at the Federated Worker. Each Federated Worker creates a FederatedLookupTable, and selects the corresponding ExecutionContextMap per FederatedRequest depending on the coordinator host (obtained from the netty channel) and the pid added to the FederatedRequest. Therefore, it allows the local sequential variable ID creation at the coordinator without collisions with any other coordinators, since every different coordinator has its own ExecutionContextMap.
I also added some shell scripts to test the Setup with multiple coordinators locally, although I will most likely change the tests again in a later PR, since I'm not happy with them myself. :D

Thanks for review :)


mboehm7:
Thanks @ywcb00 - that's a great start, as just discussed offline let's move the tests into our java test suite and create individual worker and coordinator process accordingly.

ywcb00:
Thank you for the advice of adding the multi tenant test cases into our test suite @mboehm7. Now I can withdraw my statement that I am not satisfied with the tests :smiley: 
I added the junit class FederatedMultiTenantTest, which includes two different test setups:

1. SameWorkers: all the different coordinators share the same federated sites with the same data and partitioning
2. SharedWorkers: each coordinator shares some federated sites with the prior coordinator and some with the subsequent coordinator, with different data/partitions on the federated sites

mboehm7:
LGTM - this is a great start toward multi-tenant federated learning where multiple data scientists (coordinators) are properly isolated at the federated sites.

Initially the tests did not work for me because the wait for process completion ran into a blocking state due to filled up queues. After reimplementing how we obtain the output before going into wait state this issues was resolved. However, there still seems to be an issue:

```
org.apache.spark.SparkException: A master URL must be set in your configuration
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:380)
```

because the `DMLScript.USE_LOCAL_SPARK_CONFIG` is still set to false in the spawned coordinator processes. Could you have a look - maybe a new SystemDS-config property could do the trick?

PR #1426, opened 2021-10-30
[SYSTEMDS-3190] Do not run workflow for empty checks

j143:
```yml
on:
  push:
    paths-ignore:
      - 'docs/**'
      - '*.md'
      - '*.html'
      - 'src/main/python/docs/**'
      - 'dev/**'
```

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -23,12 +23,21 @@ name: Documentation
 
 on:
   push:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+      - '*.html'
+      - 'src/main/python/docs/**'
+      - 'dev/**'
     branches:
       - master
   pull_request:
     paths-ignore:
       - 'docs/**'
       - '*.md'
+      - '*.html'
+      - 'src/main/python/docs/**'
+      - 'dev/**'
     branches:
       - master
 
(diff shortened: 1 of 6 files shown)


.github/workflows/documentation.yml
```diff
@@ -23,12 +23,21 @@ name: Documentation
...
+      - 'dev/**'
     branches:
       - master
   pull_request:
     paths-ignore:
       - 'docs/**'
       - '*.md'
+      - '*.html'
+      - 'src/main/python/docs/**'
+      - 'dev/**'
```
Baunsgaard, line 40: This is the one tests i would run on changes in docs, since it is testing if the docs build.

Baunsgaard (commented):
LGTM except the documentation test.

.github/workflows/documentation.yml
j143: Is this the change?

```diff
  pull_request:
    paths-ignore:
      - 'docs/**'
      - '*.md'
      - '*.html'
-      - 'src/main/python/docs/**'
      - 'dev/**'
```

or should not ignore any paths here?


Thanks

Baunsgaard: 'docs/**'
 'src/main/python/docs/**'
 and
 '*.md'
should not be ignored when we test building the docs.

PR #1428, opened 2021-10-30
[DOCS][MINOR] Update API docs and python version

j143:
Changes after running the `updateAPI.sh`, also the python version is updated to `2.3.0-SNAPSHOT`

@phaniarnab - Is this version the right choice?

> 📝side note: This PR verifies 30062f4a78e9e6f7a17c98ef8ff6e3c017291a66


Baunsgaard:
if you only changed the HTML files through the docs/updateAIP.sh
then the changes should be fine

Baunsgaard:
> if you only changed the HTML files through the docs/updateAPI.sh then the changes should be fine

found a way to double check it, and it is only touching the correct files.
LGTM

PR #1432, opened 2021-11-02
[MINOR] Python `SYSTEMDS_BIN` update for all the version types

j143:
```python
>>> regex1 = fnmatch.translate('systemds-*-bin.zip')
>>> regex1
'(?s:systemds\\-.*\\-bin\\.zip)\\Z'
>>> resobj1 = re.compile(regex1)
>>> resobj1.match('systemds-2.3.0-SNAPSHOT-bin.zip')
<re.Match object; span=(0, 31), match='systemds-2.3.0-SNAPSHOT-bin.zip'>
>>> resobj1.match('systemds-2.3.0-bin.zip')
<re.Match object; span=(0, 22), match='systemds-2.3.0-bin.zip'>
```

Resolves bug with 8df051fedfd60c9613c6aa4158af0c1b89968f48

References:
[1] https://docs.python.org/3/library/fnmatch.html
[2] https://docs.python.org/3/using/cmdline.html#command-line

### DIFF
--- a/src/main/python/pre_setup.py
+++ b/src/main/python/pre_setup.py
@@ -39,7 +39,7 @@ if os.path.exists(TMP_DIR):
     shutil.rmtree(TMP_DIR, True)
 os.mkdir(TMP_DIR)
 
-SYSTEMDS_BIN = 'systemds-*-SNAPSHOT-bin.zip'
+SYSTEMDS_BIN = 'systemds-*-bin.zip'
 for file in os.listdir(os.path.join(root_dir, 'target')):
     if fnmatch.fnmatch(file, SYSTEMDS_BIN):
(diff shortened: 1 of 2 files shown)


src/main/python/pre_setup.py
```diff
@@ -39,7 +39,7 @@
     shutil.rmtree(TMP_DIR, True)
 os.mkdir(TMP_DIR)
 
-SYSTEMDS_BIN = 'systemds-*-SNAPSHOT-bin.zip'
+SYSTEMDS_BIN = 'systemds-*-bin.zip'
```
phaniarnab, line 42: This change leads to test failure. That is why I suggested using `sed `to do it only during release.

PR #1434, opened 2021-11-02
[SYSTEMDS-3194] Builtins for detecting and correcting value swaps

Shafaq-Siddiqi:
1. fixInvalidLengths() to detect and correct value swaps in columns of
      same schema using string length information
  2. valueSwaps() - to detect and correct value swaps in columns of different
     schema using detetcSchema() and distance function.


Shafaq-Siddiqi:
Merged in to main.

PR #1438, opened 2021-11-05
[SYSTEMDS-3202] Python slice selection of rows and cols

Baunsgaard:
This commit adds slice arguments to python to allow slicing a
selection of rows or columns:

X = sds.from_numpy(a)
getCols 1 and 2:
b = X[:,[1,2]]
getRows 3,5
b = X[[3,5]]

This fits to the python numpy specification.


Baunsgaard:
To add full support we need to be able to call removeEmpty on frames.
https://issues.apache.org/jira/browse/SYSTEMDS-3201

PR #1439, opened 2021-11-05
[SYSTEMDS-2941] Docker images on apache

Baunsgaard:
This commit update all docker images to java 11, and moves the images
to apache. All images are pushed as nightly since the version is slightly
newer than the official release, we can add the official release as an
image if wanted with the same docker building image.

Also changed in this commit is our tests for functional and application
to now run on adoptopenJDK 11.


Baunsgaard:
@j143  Once this is merged, we can make the automatic builds happen.

j143:
Thanks @Baunsgaard - I am building a workflow for automatic push on my fork.

PR #1440, opened 2021-11-06
[SYSTEMDS-3204] Frame map instruction margin parameter

OlgaOvcharenko:
This PR modifies frames map instruction by adding a margin parameter (similar to R apply()). Margin is optional and indicates how to apply the lambda expression (0-per cell, 1-row, 2-col). If margin is not given it is set to 0.
map was changed to the ternary instruction since there are 3  inputs now.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko. During the merge I resolved a number of merge conflicts, reverted the changed html java files (because the file names and classes were inconsistent), removed warnings (e.g., unnecessary imports), and fixed a special case of shape inference for map operations on vectors.

PR #1441, opened 2021-11-06
Action for building docker images automatically

j143:
**Tasks**

- [ ] 1. Run nightly build as cron job
- [x] 2. run test build for each commit
- [ ] 3. version build for release version

**How to Review?**

Published artifact at https://hub.docker.com/r/return01/test-ghactions-systemds

Usage test:

```console
janardhan@cloudshell:~ (ai-experiments-001)$ docker run -it --rm return01/test-ghactions-systemds
Hello from SystemDS
SystemDS Statistics:
Total execution time:           0.020 sec.
```

---
Secrets: `DOCKERHUB_USER` and `DOCKERHUB_TOKEN`
These two variables are already set by INFRA. No need to set from our end.

https://issues.apache.org/jira/browse/INFRA-22430


Baunsgaard:
I think we can split the images a bit and reduce builds. 

1. Build max once a day. 
2. Only build the test image if we change R dependencies or the pom.xml
3. Only build the other images if there is changes in src. 

Thanks for the help!

.github/workflows/docker-pub.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+  
+    - name: Set up QEMU
```
Baunsgaard, line 20: QEMU is for ?

docker/sysds.Dockerfile
```diff
@@ -60,14 +60,16 @@ RUN apt-get update -qq \
 	rm -r target/hadoop-test && \
 	rm -r target/maven-archiver && \
 	rm -r target/systemds-** && \
-	rm -r docker && \
+	# rm -r docker && \
```
Baunsgaard, line 63: you cannot just comment out a line in this multi-line instruction, then everything after it does not execute.

docker/sysds.Dockerfile
```diff
@@ -60,14 +60,16 @@ RUN apt-get update -qq \
...
 	rm -r src && \
 	rm -r /usr/lib/mvn && \
 	rm -r CONTRIBUTING.md && \
 	rm -r pom.xml && \ 
 	rm -r ~/.m2
 
-COPY docker/mountFolder/main.dml /input/main.dml
+RUN mkdir /input && echo 'print("Hello from SystemDS")' > /input/main.dml
+
+# COPY docker/mountFolder/main.dml /input/main.dml
```
Baunsgaard, line 73: i'm all for running it but don't remove the copy, so that the only change you make it  to add the RUN line.

j143: I am having error with running `COPY`

.github/workflows/docker-pub.yml
j143: It is an add-on, incase we want to build multi arch images. Optional though!

docker/sysds.Dockerfile
j143:
```suggestion
```

docker/sysds.Dockerfile
j143: This one

image

.github/workflows/docker-pub.yml
Baunsgaard: hmm, then for now lets not include it if it is not needed

docker/sysds.Dockerfile
Baunsgaard: Not sure what you mean?
just for information the reason why i have this line is to remove the docker folder from the internal copied systemds repository, it does not make a difference for the building of the docker image except it makes the end result slightly smaller.

so therefore, don't out comment it :)

docker/sysds.Dockerfile
Baunsgaard: this is happening because of your out commented line mentioned above.

Baunsgaard: this one:

```
# rm -r docker && \
```

Baunsgaard:
I think it is fine if we just add the on commit part of the docker for now, 
If it is still working, we should start merging it in.

Baunsgaard:
@j143 
Is the only thing missing not just a Infra ticket to add keys to push the images?

j143:
> Is the only thing missing not just a Infra ticket to add keys to push the images?

There are already secrets available.

the problem seems to be `COPY docker/mountFolder/main.dml /input/main.dml` statement in the Dockerfile.

PR #1444, opened 2021-11-08
[MINOR] Set language level to 11

Baunsgaard:
This commit fully change our project to java 11, on the compiler and target level.


Baunsgaard:
For all, 
We have to agree when we make this switch since it then require use of JDK 11.

PR #1446, opened 2021-11-08
[SYSTEMDS-3206] Federated Execution w/ Misaligned Matrices

ywcb00:
Hi,
This PR fixes two issues with the _AggregateBinaryFEDInstruction_ and the _MMFEDInstruction_.
1. It removes the seperate alignment check inside the two instructions where there is no verification that the range-aligned federated data partitions are located on the same federated worker. Since the case of having two aligned matrices is already covered by the very first if branch, I removed the whole case containing the local alignment check.
2. Since the federated spark instructions do not automatically create the output variables yet, we need to create the output matrices by an additional PUT_VAR request to the federated worker. However, this PR adds some of those PUT_VAR requests that were missing inside the _MMFEDInstruction_.

In addition, this PR addresses the problem of creating an endless loop between the _acquireReadIntern()_ and the _readBlobFromHDFS()_ method (mentioned in #1390) by adding the constraint to read the file from federated even if it is broadcast but does not exist on hdfs.

Thanks for review :)


ywcb00:
When I extended the tests with misaligned matrices (FederatedMisalignedTest), I discovered two more issues with the _BinaryMatrixMatrixFEDInstruction_ and the _AppendFEDInstruction_.
The BinaryMatrixMatrixFEDInstruction was throwing an Exception when called with two federated matrices which are not aligned. I changed it to not throwing an exception and pulling one of the federated matrices to the driver where we can then handle it like a instruction with one federated matrix.
I've added support for the case of having two aligned federated matrices to the AppendFEDInstruction.

I think this PR is now ready for review :rocket:

Baunsgaard (commented):
LGTM will merge shortly,
I like the changes / simplifications in AggregateBinaryFEDInstruction.java especially.

Some changed files are just formatting and slight changes in calls for cleanup,
It would make it easier / more likely to review quickly if these changes were not part of the PR and maybe separated in other PRs named "clanup x" or you can merge such changes in [MINOR] commits yourself.
example: src/main/java/org/apache/sysds/runtime/instructions/fed/QuaternaryWDivMMFEDInstruction.java

(but i am doing the same error in my PRs so i can hardly complain)

PR #1448, opened 2021-11-08
[SYSTEMDS-3209] Builtin for random under-sampling

Shafaq-Siddiqi:
- the builtin accepts matrix data with last column as labels and a ratio parameter
    and randomly remove the tuples from the majority class


Shafaq-Siddiqi:
merged into main

PR #1453, opened 2021-11-10
[SYSTEMDS-3215] Federated Worker deriving SparkExecutionMode

ywcb00:
Hi,
This PR adds a workaround to the problem of running into a ClassCastException because of casting an ExecutionContext into a SparkExecutionContext.
It simply checks if there is a need for having a SparkExecutionContext instead of a casual ExecutionContext at every get call to the ExecutionContextMap, and derives a new SparkExecutionContext from the actual ExecutionContext if this is the case.
The PR also includes the change of setting the global dml execution mode of the federated worker to HYBRID if there is an incoming SPARK Exec_Inst request and the worker is on SINGLE_NODE execution mode.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/context/ExecutionContextFactory.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/context/ExecutionContextFactory.java
@@ -58,8 +58,9 @@ public class ExecutionContextFactory
 	}
 
 	public static boolean needsSparkEC(Program prog) {
-		boolean retVal = OptimizerUtils.getDefaultExecutionMode()==ExecMode.HYBRID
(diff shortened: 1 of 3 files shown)


ywcb00:
Since this PR resolves at least the ClassCastExceptions of the federated primitive tests in our test system, we might want to add this for the release as well. :thinking: :thought_balloon: 
What do you think?

ywcb00:
I made a small mistake when checking whether we need a spark execution context, that's why the tests were failing.
It works perfectly now :smiley:

mboehm7:
LGTM - thanks for the fix @ywcb00. I changed the implementation a bit to make it easier to understand and avoid unnecessary indirections. Furthermore, I left a TODO as we should rework who controls the execution of spark instructions at the federated site.

PR #1454, opened 2021-11-11
[SYSTEMDS-3217] PythonAPI working with SYSTEMDS_ROOT set

Baunsgaard:
This commit modify the PythonAPI at startup to work with our release
binaries. While modifying this the building of the python release
artifact have been modified to not include the test and generator code.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -27,7 +27,8 @@ on:
       - 'docs/**'
       - '*.md'
       - '*.html'
-      - 'src/main/python/docs/**'
+      - 'src/test/**'
+      - 'src/assembly/**'
       - 'dev/**'
     branches:
       - main
@@ -36,7 +37,7 @@ on:
       - 'docs/**'
       - '*.md'
       - '*.html'
-      - 'src/main/python/docs/**'
+      - 'src/test/**'
       - 'dev/**'
     branches:
       - main
@@ -94,7 +95,7 @@ jobs:
         architecture: 'x64'
   
     - name: Install pip Dependencies
-      run: pip install numpy py4j wheel scipy sklearn requests pandas
+      run: pip install numpy py4j wheel scipy sklearn requests pandas unittest-parallel
 
     - name: Build Python Package
       run: |
@@ -107,15 +108,16 @@ jobs:
         export PATH=$SYSTEMDS_ROOT/bin:$PATH
         export SYSDS_QUIET=1
         cd src/main/python
-        python -m unittest discover -s tests -p 'test_*.py'
+        unittest-parallel -t . -s tests --module-fixtures
+        # python -m unittest discover -s tests -p 'test_*.py'
         echo "Exit Status: " $?
     
-    # TODO debug and fix JDK11 environment 
-    #- name: Run all python tests no environment
-    #  run: |
-    #    cd src/main/python
-    #    python -m unittest discover -s tests -p 'test_*.py'
-    #    echo "Exit Status: " $?
+    - name: Run all python tests no environment
+      run: |
+        cd src/main/python
+        unittest-parallel -t . -s tests --module-fixtures
+        # python -m unittest discover -s tests -p 'test_*.py'
+        echo "Exit Status: " $?
 
     - name: Run Federated Python Tests
       run: |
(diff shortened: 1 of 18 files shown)


src/main/python/pre_setup.py
```diff
@@ -26,40 +26,56 @@
...
         unzipped_dir_name = file.rsplit('.', 1)[0]
-        shutil.copytree(os.path.join(TMP_DIR, unzipped_dir_name), java_dir_full_path)
-        if os.path.exists(TMP_DIR):
-            shutil.rmtree(TMP_DIR, True)
+        shutil.copytree(os.path.join(TMP_DIR, unzipped_dir_name, 'lib'), LIB_DIR)
+
+# Take hadoop binaries.
+HADOOP_DIR_SRC = os.path.join(SYSTEMDS_ROOT, 'target', 'lib', 'hadoop')
+if os.path.exists(HADOOP_DIR_SRC):
+    shutil.copytree(HADOOP_DIR_SRC, os.path.join(LIB_DIR,"hadoop"))
```
phaniarnab, line 63: Please document why do we need the hadoop and winutils dll/libs?

src/main/python/pre_setup.py
```diff
@@ -26,40 +26,56 @@
...
+shutil.copy(os.path.join(SYSTEMDS_ROOT,'conf', 'SystemDS-config-defaults.xml'), os.path.join(this_path, PYTHON_DIR, 'conf', 'SystemDS-config-defaults.xml'))
 
-# delete old build and dist path
+SYSTEMDS_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.getcwd())))
+shutil.copyfile(os.path.join(SYSTEMDS_ROOT, 'LICENSE'), 'LICENSE')
+shutil.copyfile(os.path.join(SYSTEMDS_ROOT, 'NOTICE'), 'NOTICE')
+
+# Remove old build and dist path
+if os.path.exists(TMP_DIR):
+    shutil.rmtree(TMP_DIR, True)
```
phaniarnab, line 78: Changes in the release-related files need testing @j143  and myself, as they can break the release automation.
I would recommend sticking to necessary changes and leaving the improvements for future releases.

phaniarnab (commented):
General comment: Please stick to only bug fixes/absolutely necessary changes for the patch release. Code changes in a new RC should only contain the fixes missing in the last RC.

src/assembly/bin.xml
```diff
@@ -91,23 +92,28 @@
...
 				<include>*:log4j*</include>
 				<include>*:netty*</include>
 				<include>*:protobuf-java*</include>
 				<include>*:py4j*</include>
+				<include>*:re2j*</include>
 				<include>*:slf4j-api*</include>
 				<include>*:slf4j-log4j*</include>
 				<include>*:spark-core*</include>
+				<include>*:stax2-api*</include>
+				<include>*:woodstox*</include>
```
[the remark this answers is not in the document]
Baunsgaard: These are not for the patch fixes of the old release, but they are missing if you make a new release now.

src/main/python/pre_setup.py
Baunsgaard: This is for consistancy between release bin version and the python release version.
i will include this in the patch release

src/main/python/pre_setup.py
Baunsgaard: yes, that is why i have added the test for the python build back.
i am also testing it for the release artifact.

.gitignore
```diff
@@ -41,19 +41,6 @@ buildNumber.properties
 .jekyll-cache/
 _site/
 
-# Python artifacts
-src/main/python/systemds/systemds-java/
```
j143, line 45: I think it is better only to have one gitignore file for the project.

unless the root .gitignore file is ignoring unintended files in src/main/python!

j143:
Thank you for the cleanup.
Also, we need to create a PR to `branch-2.2.0`!, with specific changes so we can test for spark, java version.

.gitignore
Baunsgaard: i would tend to agree, but in this case the python part is sort of a separate system, that could be in its own repository, and when i develop in the python i teat it as such, therefore i don't have access to the root .gitignore in the IDE when developing the python.

there are examples of doing both designs in Apache.

Hadoop only does one .gitignore
while Spark has one in each language
https://github.com/apache/spark

Baunsgaard:
> Thank you for the cleanup. Also, we need to create a PR to `branch-2.2.0`!, with specific changes so we can test for spark, java version.

yes that is the next task, i will take the appropriate parts from this, and make specific targeted changes for the release branch

PR #1455, opened 2021-11-11
[SYSTEMDS-3203] Remove empty columns frame out of bounds

OlgaOvcharenko:
rmempty instruction parameter select can be either row or column vector now, independent of the margin (before for margin cols - row vector,  margin rows - column vector).
Additionally, tests failed because of the incorrect pd.DataFrame column slices.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/matrix/data/FrameBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/matrix/data/FrameBlock.java
@@ -30,7 +30,14 @@ import java.lang.ref.SoftReference;
 import java.lang.reflect.InvocationTargetException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -48,7 +55,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.io.Writable;
 import org.apache.sysds.api.DMLException;
-import org.apache.sysds.common.Types;
 import org.apache.sysds.common.Types.ValueType;
 import org.apache.sysds.runtime.DMLRuntimeException;
 import org.apache.sysds.runtime.codegen.CodegenUtils;
@@ -66,6 +72,7 @@ import org.apache.sysds.runtime.meta.MatrixCharacteristics;
 import org.apache.sysds.runtime.transform.encode.ColumnEncoderRecode;
 import org.apache.sysds.runtime.util.CommonThreadPool;
 import org.apache.sysds.runtime.util.DMVUtils;
+import org.apache.sysds.runtime.util.DataConverter;
 import org.apache.sysds.runtime.util.EMAUtils;
 import org.apache.sysds.runtime.util.IndexRange;
 import org.apache.sysds.runtime.util.UtilFunctions;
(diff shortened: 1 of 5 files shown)


src/test/java/org/apache/sysds/test/component/frame/FrameRemoveEmptyTest.java
```diff
@@ -46,25 +46,25 @@ public void setUp() {
...
 	}
 
 	@Test
-	public void testRemoveEmptyRowsSparseCP() {
+	public void testRemoveEmptyColsCP() {
 		runTestRemoveEmpty(TEST_NAME1, "cols", Types.ExecType.CP, true);
 	}
 
 	@Test
 	@Ignore
```
Baunsgaard, line 59: should these still be ignored?

Baunsgaard (commented):
some work left.

src/main/java/org/apache/sysds/runtime/matrix/data/FrameBlock.java
```diff
@@ -2533,7 +2539,8 @@ private FrameBlock removeEmptyRows(MatrixBlock select, boolean emptyReturn) {
 				isEmpty = isEmpty && (ArrayUtils.contains(new double[]{0.0, Double.NaN}, UtilFunctions.objectToDoubleSafe(type, colData.get(i))));
 			}
 
-			if((!isEmpty && select == null) || (select != null && select.getValue(i, 0) == 1)) {
+			double selectValue = select != null ? !selectColVector ? select.getValue(i, 0) : select.getValue( 0, i) : 0.0;
+			if(selectValue == 1.0 || (!isEmpty && select == null)) {
```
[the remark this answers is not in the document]
OlgaOvcharenko: I modified the implementation, now the indices are obtained and we iterate this list.

src/main/java/org/apache/sysds/runtime/matrix/data/FrameBlock.java
```diff
@@ -2520,25 +2527,35 @@ public  FrameBlock removeEmptyOperations(boolean rows, boolean emptyReturn, Matr
...
-			for(int j = 0; j < getNumColumns(); j++) {
-				Array colData = _coldata[j].clone();
-				row[j] = colData.get(i);
-				ValueType type = _schema[j];
-				isEmpty = isEmpty && (ArrayUtils.contains(new double[]{0.0, Double.NaN}, UtilFunctions.objectToDoubleSafe(type, colData.get(i))));
+				if(!isEmpty)
+					ret.appendRow(row);
 			}
+		} else {
+			List<Integer> indices = DataConverter.convertVectorToIndicesList(select);
```
Baunsgaard, line 2548: better with int[]

src/main/java/org/apache/sysds/runtime/util/DataConverter.java
```diff
@@ -307,6 +307,34 @@ public static MatrixBlock readMatrixFromHDFS(ReadProperties prop)
...
+		if( mb.getNonZeros() > 0 )
+		{
+			if( mb.isInSparseFormat() )
+			{
+				Iterator<IJV> iter = mb.getSparseBlockIterator();
+				while( iter.hasNext() ) {
+					IJV cell = iter.next();
+					if( cell.getV() != 0.0)
+						indices.add(cell.getI()*cols+cell.getJ());
+				}
```
Baunsgaard, line 325: on a sparse block call:

```java
SparseBlock sb = mb.getSparseBlock()
int[] aix = sb.indexes(0);
return aix;
```

This is assuming that you have only one row of values.

src/main/java/org/apache/sysds/runtime/util/DataConverter.java
```diff
@@ -307,6 +307,34 @@ public static MatrixBlock readMatrixFromHDFS(ReadProperties prop)
...
+			}
+			else {
+				for( int i=0, cix=0; i<rows; i++ )
+					for( int j=0; j<cols; j++, cix++)
+						if( mb.getValueDenseUnsafe(i, j) != 0.0 )
+							indices.add(cix);
```
Baunsgaard, line 331: don't add to the ArrayList (in general don't use an ArrayList here)

preallocate a single int[] based on nnz in the input.
then put the values into that.

src/main/java/org/apache/sysds/runtime/util/DataConverter.java
```diff
@@ -307,6 +307,34 @@ public static MatrixBlock readMatrixFromHDFS(ReadProperties prop)
 		return ret;
 	}
 
+	public static List<Integer> convertVectorToIndicesList(MatrixBlock mb)
```
Baunsgaard, line 310: return int []

src/main/java/org/apache/sysds/runtime/util/DataConverter.java
```diff
@@ -307,6 +307,34 @@ public static MatrixBlock readMatrixFromHDFS(ReadProperties prop)
...
+	{
+		int rows = mb.getNumRows();
+		int cols = mb.getNumColumns();
+		ArrayList<Integer> indices = new ArrayList<>();
+
+		if( mb.getNonZeros() > 0 )
```
Baunsgaard, line 316: first if check for empty. (return null)
second check for sparse (return already allocated the sparse allocated indexes )
third (default) iterate dense. (allocate based on nnz)

src/test/java/org/apache/sysds/test/component/frame/FrameRemoveEmptyTest.java
```diff
@@ -35,7 +35,7 @@
 	private final static String TEST_DIR = "functions/frame/";
 	private static final String TEST_CLASS_DIR = TEST_DIR + RemoveEmptyTest.class.getSimpleName() + "/";
 
-	private final static int _rows = 10;
+	private final static int _rows = 100;
```
Baunsgaard, line 38: add tests for selecting for selecting 0, 1, half, all-but-one, all.
in case of selecting all, simply return a deep copy.

100 does not really change performance measurements.
while debugging try to increase it, same goes for columns.

then create measurements for selecting 1, half, all-but-one elements from the input.
and find out if the new version is faster or slower in each case.

In the PR message argue the performance difference say something like 2x faster etc.

Baunsgaard (commented):
Progress, but to make it perfect i have a few suggestions.

mboehm7:
LGTM - thanks for the improvements @OlgaOvcharenko and @Baunsgaard. During the merge I made the following changes for performance / correctness:
* rmEmpty rows (select all) returns shallow copy of frame block (due to copy on write semantics)
* rmEmpty rows allocates the temporary object array just once outside the loops
* rmEmpty cols use shallow copies of columns instead of cloned columns (due to copy on write semantics)
* `DataConverter.convertVectorToIndexList` now copies a range of the physical array in order to guarantee correctness (the physical index/value arrays might be larger than the number of indexes, in which case we would have mistakenly appended column 0 one or multiple times)

mboehm7:
Something seems to be off with the python tests (which failed on main) - for now I'm disabling them, but would appreciate if you could have a look into that.

PR #1456, opened 2021-11-11
[MINOR] Run script bug fixes:

corepointer:
* Move the setting of SYSTEMDS_ROOT before its first use
* Search for a config xml file name that is actually included in the bin release


Baunsgaard (commented):
LGTM

PR #1459, opened 2021-11-16
[WIP][SYSTEMDS-3221] Federated DataGENInstruction

OlgaOvcharenko:
This PR adds a simple matrix rand(..) case. Currently, federated rand(..) is called only for the corresponding test, later this should be modified.


src/main/java/org/apache/sysds/api/DMLScript.java
```diff
@@ -277,6 +280,7 @@ public static boolean executeScript( Configuration conf, String[] args )
 			//reset runtime platform and visualize flag
 			setGlobalExecMode(oldrtplatform);
 			EXPLAIN = oldexplain;
+			DMLScript.FED_WORKER_PORTS.clear();
```
Baunsgaard, line 283: I would like if this was not a variable in DMLScript, since it would not be available when executing in python.

maybe put it in: 
src/main/java/org/apache/sysds/runtime/controlprogram/context/ExecutionContext.java
or 
not have it.

Maybe the best design is to add another argument to rand, that allows you to specify a federated range of worker addresses.

src/main/java/org/apache/sysds/runtime/instructions/fed/DataGenFEDInstruction.java
```diff
@@ -0,0 +1,366 @@
...
+		Arrays.fill(instStrings, 0, rowsPerWorker.length, instString);
+
+		InetAddress addr = null;
+		try {
+			addr = InetAddress.getLocalHost();
+		}
+		catch(UnknownHostException e) {
+			e.printStackTrace();
+		}
+		String host = addr.getHostName();
```
Baunsgaard, line 286: With only here is another argument for why we need to specify a list of federated addresses, in the arguments for the instruction since we need to specify other hosts than localhost.

src/main/java/org/apache/sysds/runtime/instructions/fed/DataGenFEDInstruction.java
```diff
@@ -0,0 +1,366 @@
...
+
+		// generate seeds and rows
+		long[] lSeeds = randomSeedsGenerator(nworkers);
+
+		// new sizes
+		int size = (int) Math.round(getRows() / (double) nworkers);
+		int addedRows = size * (nworkers -1);
+		long[] rowsPerWorker = new long[nworkers];
+		Arrays.fill(rowsPerWorker, 0, nworkers-1, size);
+		rowsPerWorker[nworkers-1] = getRows() - addedRows;
```
Baunsgaard, line 262: we should specify the rows as well in the instruction, using the federated addresses, since these give the dimensions of each federated site.

src/main/java/org/apache/sysds/runtime/instructions/fed/FEDInstructionUtils.java
```diff
@@ -145,6 +150,11 @@ else if (inst instanceof UnaryCPInstruction && ! (inst instanceof IndexingCPInst
 					&& mo.isFederatedExcept(FType.BROADCAST) )
 					fedinst = ReorgFEDInstruction.parseInstruction(
 						InstructionUtils.concatOperands(rinst.getInstructionString(),FederatedOutput.NONE.name()));
+			} else if(inst instanceof DataGenCPInstruction) {
+				DataGenCPInstruction dinst = (DataGenCPInstruction) inst;
```
Baunsgaard, line 154: maybe make a DataGenFedInstruction that contain the extra arguments you need.

src/test/java/org/apache/sysds/test/functions/federated/datagen/FederatedRandTest.java
```diff
@@ -0,0 +1,156 @@
...
+		programArgs = new String[] {"-nvargs", "X1=" + TestUtils.federatedAddress(port1, input("X1")),
+			"X2=" + TestUtils.federatedAddress(port2, input("X2")),
+			"r=" + rows, "c=" + cols, "min=" + min, "max=" + max, "sp=" + sparsity, "seed=" + seed, "Z=" + output("Z")};
+		runTest(null);
+
+		HashMap<MatrixValue.CellIndex, Double> output = TestUtils.readDMLMatrixFromHDFS(outputDirectories[0]);
+		int rowOut = output.keySet().stream().max(Comparator.comparingInt(e -> e.row)).get().row;
+		int colOut = output.keySet().stream().max(Comparator.comparingInt(e -> e.column)).get().column;
+		double minOut = output.values().stream().min(Comparator.comparingDouble(Double::doubleValue)).get();
+		double maxOut = output.values().stream().max(Comparator.comparingDouble(Double::doubleValue)).get();
```
Baunsgaard, line 143: this is not a nice comparison, since it only veifies max and min, and also steam through all cells
just make use of a simple compare of the outputs.

Baunsgaard (commented):
The design with a global list of federated addresses is not good, 
i would extend the rand to include federated addresses,

Then a next task would be to make a rewrite instruction to rewrite data gen to a federated data gen based on subsequent operations.

With this design it will nicely fit into the framework.

PR #1468, opened 2021-11-28
[DOC] Add instructions for download R dependencies

j143:
.

### DIFF
--- a/docs/site/install.md
+++ b/docs/site/install.md
@@ -83,6 +83,13 @@ sudo add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu focal-
 sudo apt install r-base
 ```
 
+Additionally, you need to install the R depedencies for integration tests, like this:
+(use `sudo` mode if the script couldn't write to local R library)
+
+```bash
+Rscript ./src/test/scripts/installDependencies.R
+```
+
 See [Build the project](#Build%20the%20project) to compile the code from here.
 
 ---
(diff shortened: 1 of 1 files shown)


phaniarnab:
Why an user needs to install R?

docs/site/install.md
```diff
@@ -83,6 +83,13 @@ sudo add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu focal-
 sudo apt install r-base
 ```
 
+Additionally, you need to install the R depedencies for integration tests, like this:
```
j143, line 86:
```suggestion
Optionally, you need to install the R depedencies for integration tests, like this:
```

docs/site/install.md
```diff
@@ -127,6 +134,13 @@ export JAVA_HOME=`/usr/libexec/java_home -v 1.8`
 
 For running all tests [r-base](https://cran.r-project.org/bin/macosx/) has to be installed as well since this is used as a secondary system to verify the correctness of our code, but it is not a requirement to enable building the project.
 
+Additionally, you need to install the R depedencies for integration tests, like this:
```
j143, line 137:
```suggestion
Optionally, you need to install the R depedencies for integration tests, like this:
```

j143:
> Why an user needs to install R?

This doc is also, helpful for developers. (intent of this is for systemds developers too!)

phaniarnab:
If I am not wrong, the current doc doesn't tell how to build the test classes. And that is because the primary target is the users who want to build SystemDS from the source.

Baunsgaard:
> If I am not wrong, the current doc doesn't tell how to build the test classes. And that is because the primary target is the users who want to build SystemDS from the source.

I agree with this statement, for the devs we should do an extended guide.

PR #1469, opened 2021-11-29
[MINOR][PYTHON-DOC] Update Multiple Federated Environments Example Result

fathollahzadeh:
Minor update in python document, Multiple Federated Environments Example Result.


j143:
Thanks @fathollahzadeh 🚀 🎉

PR #1476, opened 2021-12-07
[SYSTEMDS-3018] Edit Federated Planner to Handle Double Hop Output Cases

sebwrede:
This PR: 

- Creates a separate class for HopRel memo table to avoid code duplication and to have a cleaner interface to the data structure. 
- Changes the final setting of fedouts in the federated planner to avoid visiting the same hops more than once and to cover the case where a single hop is used as input to two different hops. In the latter case, the hop needs to be set to FOUT and then be prefetched before being used as input to a locally executed operation. 
- Adds prefetch of hops and lops depending on a field set in the federated planner and the ExecType of the operation.

### DIFF
--- a/src/main/java/org/apache/sysds/hops/Hop.java
+++ b/src/main/java/org/apache/sysds/hops/Hop.java
@@ -93,6 +93,13 @@ public abstract class Hop implements ParseInfo {
 	 */
 	protected FederatedOutput _federatedOutput = FederatedOutput.NONE;
 	protected FederatedCost _federatedCost = new FederatedCost();
+	/**
+	 * Field defining if prefetch should be activated for operation.
+	 * When prefetch is activated, the output will be transferred from
+	 * remote federated sites to local before one of the subsequent
+	 * local operations.
+	 */
+	protected boolean activatePrefetch;
 	
 	// Estimated size for the output produced from this Hop in bytes
 	protected double _outputMemEstimate = OptimizerUtils.INVALID_SIZE;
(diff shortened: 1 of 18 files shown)


mboehm7:
Let me know once you're happy with it (no longer in draft) - then I'll have a look and merge it with minor changes.

sebwrede:
I will add the new way of testing the double hop case in another PR (it is not ready yet). 
This PR is ready to be merged.

phaniarnab:
Don't you need to extend the Prefetch instruction? The current instruction is implemented specifically for pending RDD operations? Otherwise, the logic to add prefetch lop looks good to me.

sebwrede:
> Don't you need to extend the Prefetch instruction? The current instruction is implemented specifically for pending RDD operations? 

I didn't think about that. In that case, this PR is not ready for merge. It will not break anything, except if federated compilation is activated when running a script and this specific double hop case occurs (which is very rare). 
I will have to look at it later.

mboehm7:
Interestingly, by calling `acquireRead` it should work for both an RDD-backed MatrixObject and a Federated-Matrix-based MatrixObject. And for both still keep the original representation in tact as well.

phaniarnab:
> Interestingly, by calling `acquireRead` it should work for both an RDD-backed MatrixObject and a Federated-Matrix-based MatrixObject. And for both still keep the original representation in tact as well.

That is true. However, we check if any RDD operations are pending before calling `acquireRead`. That condition needs to be extended.

mboehm7:
ok makes sense - do you wanna change it or should I during the merge?

phaniarnab:
I will do it now, and add a commit to this PR.

phaniarnab:
Done. 
I noticed the new tests don't generate a prefetch instruction currently. I think @sebwrede will fix that in another PR as part of the new way of testing.

mboehm7:
LGTM - thanks the both of you. During the merge, I only slightly changed some formatting, and the logic in `updateETFed` in order to make it specific to properties of individual inputs (instead of someFederatedInput and somePrefetch), which gives us more flexibility if some input is known to be broadcast. Furthermore, I added a TODO for refactoring as we might want to generalize and reuse the codegen memo table here too.

PR #1477, opened 2021-12-07
[SYSTEMDS-3241] Federated quantile

OlgaOvcharenko:
This PR adds federated quantile for row federated matrices.

1. qsort creates a copy of the input MatrixObject without sorting it.
2. qpick recursively creates histogram between min and max of the MatrixObject.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko, this is a great extension and very useful for plenty of pre-processing primitives. During the merge, I only made a few minor tweaks in the qpick instruction, and re-enabled the tests for column-partitioned data again (which were commented, probably for local testing).

Once you get a chance, could you please then follow up and improve the performance of the histogram construction (e.g., the `GetValuesInRange` UDF) to use an array of pre-materialized bin boundaries and binary search similar to transformencode-binning?

PR #1478, opened 2021-12-08
[SYSTEMDS-3187] Add documentation for the release scripts

j143:
- Add software and credential requirements


j143:
git2jira sync works fine. https://issues.apache.org/jira/browse/INFRA-22587

PR #1479, opened 2021-12-08
[SYSTEMDS-3183] Builtins for set operations on vectors

m2key1:
Added unique, union, difference and symmetricDifference operators. 
We also changed the intersect builtin to work with negative and decimal numbers.


src/test/java/org/apache/sysds/test/functions/builtin/setoperations/BuiltinSymmetricDifferenceTest.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.sysds.test.functions.builtin.setoperations;
```
j143, line 1: missing license header!

j143:
Hi @m2key1 @JoachimDunkel , thanks a lot for the contribution. 😸 💖 

**Few house keeping comments:**
1. the jitify submodule seems to be removed. We shouldn't remove that!
2. you have kept the licenses good in all the folders except two.
3. How about adding these two functions in https://github.com/apache/systemds/blob/main/docs/site/builtins-reference.md

j143:
Hi thanks for addressing some review comments, I will make a deep pass review tomorrow. correct me if I am wrong, I am assuming this should work similar to the `R` implementation.

Shafaq-Siddiqi:
> Hi thanks for addressing some review comments, I will make a deep pass review tomorrow. correct me if I am wrong, I am assuming this should work similar to the `R` implementation.

Yes, these functions are expected to be comparable with R's set operations and have tested with R scripts.

m2key1:
Hi, we have reverted the commit where the jitify submodule has been removed. Furthermore we have added the missing licenses and added entries to "builtins-reference.md" for the union, unique, setdiff and symmetricDifference operations.

m2key1:
> 

And about the literature used for the implementation: We have figured out the implementations ourselves (trail and error ;) ). So we didn't use any real literature, except to look at the definitions of the individual set operators.

j143:
> 
> And about the literature used for the implementation: We have figured out the implementations ourselves (trial and error ;) ). So we didn't use any real literature, except to look at the definitions of the individual set operators.

[MATLAB has nice references with examples](https://in.mathworks.com/help/matlab/ref/double.union.html), to see how it works on different combinations.

j143:
Hi @m2key1 - Just want to let you know that I am actively reviewing this PR. This might take 2 days (today, tomorrow).

1. for 1d the operation is straight forward, I am looking at matrices etc. We are just row appending for unique  - I am not sure that is how we should handle dimensionality (matrices with different features).
2. Strings - these are also important
3. I will see why the spark tests are failing. Of course, you can debug them! :)

docs/site/builtins-reference.md
```diff
@@ -1823,6 +1827,36 @@ scale=TRUE;
...
+| Type   | Description |
+| :----- | :---------- |
+| Matrix[Double] | values of X that are not in Y.|
+
+### Example
+
+```r
+X = matrix("1 2 3 4", rows = 4, cols = 1)
+Y = matrix("2 3", rows = 2, cols = 1)
+R = setdiff(X = X, Y = Y)
```
j143, line 1858: Thanks for the examples, these are really helpful.

j143 (approved):
Thank you, @m2key1 @JoachimDunkel  - LGTM. 👍 
🎉 🚀


For the scope of this PR, only the operations over doubles would be sufficient.
Will merge this PR shortly.
I will take care of spark instructions.

j143:
Hi @m2key1 , can you associate your college mail with github so that proper credit will available in the contributor section!

m2key1:
> Hi @m2key1 , can you [associate your college mail with github](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/adding-an-email-address-to-your-github-account) so that proper credit will available in the contributor section!

Thanks for pointing this out! Should be done now :)

PR #1480, opened 2021-12-08
[SYSTEMDS-3243] Consistent allocation of MatrixBlock for MM

Baunsgaard:
This commit change the Matrix Multiply inside MatrixBlock to
  consistently allocate a sparse or dense output block.
  Previously there was a inconsistent check for sparsity in the MatrixBlock
  code that these changes address.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixNative.java
+++ b/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixNative.java
@@ -64,86 +64,83 @@ public class LibMatrixNative
 	 * @param m2 rhs matrix block
 	 * @param ret output matrix block
 	 * @param k number of threads
+	 * @return the ret matrixBlock if allocated otherwise a new matrixBlock.
 	 */
-	public static void matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k) {
-		matrixMult(m1, m2, ret, k, true);
-	}
+	public static MatrixBlock matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k) {
+		
+		if(NativeHelper.isNativeLibraryLoaded()){
+			// Sanity check:
+			k = k <= 0 ? NativeHelper.getMaxNumThreads() : k;
+			
+			// check inputs / outputs
+			if (m1.isEmptyBlock(false) || m2.isEmptyBlock(false))
+				return LibMatrixMult.emptyMatrixMult(m1,m2, ret);
+			
+			boolean isValidForNative = !isMatMultMemoryBound(m1.rlen, m1.clen, m2.clen) 
+				&& !m1.isInSparseFormat() && !m2.isInSparseFormat()
+				&& (m1.getDenseBlock().isContiguous() || !isSinglePrecision())
+				&& m2.getDenseBlock().isContiguous() //contiguous but not allocated
+				&& 8L * ret.getLength() < Integer.MAX_VALUE;
 	
-	public static void matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k, boolean examSparsity) {
-		// Sanity check:
-		k = k <= 0 ? NativeHelper.getMaxNumThreads() : k;
-		
-		// check inputs / outputs
-		if (m1.isEmptyBlock(false) || m2.isEmptyBlock(false)){
-			ret.setNonZeros(0);
-			if(examSparsity)
-				ret.examSparsity(); // turn empty dense into sparse
-			return;
-		}
-		
-		boolean isValidForNative = !isMatMultMemoryBound(m1.rlen, m1.clen, m2.clen) 
-			&& !m1.isInSparseFormat() && !m2.isInSparseFormat()
-			&& (m1.getDenseBlock().isContiguous() || !isSinglePrecision())
-			&& m2.getDenseBlock().isContiguous() //contiguous but not allocated
-			&& 8L * ret.getLength() < Integer.MAX_VALUE;
-
-		if( NativeHelper.isNativeLibraryLoaded() && isValidForNative ) 
-		{
-			ret.sparse = false;
-			ret.allocateDenseBlock();
-			long start = DMLScript.STATISTICS ? System.nanoTime() : 0;
-			long nnz = 0;
-			if( isSinglePrecision() ) {
-				FloatBuffer fin1 = toFloatBuffer(m1.getDenseBlockValues(), inBuff, true);
-				FloatBuffer fin2 = toFloatBuffer(m2.getDenseBlockValues(), filterBuff, true);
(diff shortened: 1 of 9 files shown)


Baunsgaard:
Note Lineage unstable?:
https://github.com/apache/systemds/runs/4510306348?check_suite_focus=true

Baunsgaard:
unstable:
https://github.com/apache/systemds/runs/4511370731?check_suite_focus=true

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
```diff
@@ -5096,34 +5081,52 @@ public final MatrixBlock aggregateBinaryOperations(MatrixBlock m1, MatrixBlock m
...
-		SparsityEstimate sp = estimateSparsityOnAggBinary(m1, m2, op);
-		
-		//create output matrix block
-		if( ret==null )
-			ret = new MatrixBlock(rl, cl, sp.sparse, sp.estimatedNonZeros);
+
+		final boolean m1Perm = m1.isSparsePermutationMatrix();
+		final boolean ultraSparse = LibMatrixMult.isUltraSparseMatrixMult(m1, m2, m1Perm);
+		final boolean sparse = !m1Perm && !ultraSparse && LibMatrixMult.isSparseOutputMatrixMult(m1, m2);
+		final boolean sparseRet = ultraSparse | sparse;
```
Baunsgaard, line 5114: copied sparsity check from matrix mult lib

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
```diff
@@ -5096,34 +5081,52 @@ public final MatrixBlock aggregateBinaryOperations(MatrixBlock m1, MatrixBlock m
...
+		ret.allocateBlock();
+
+		if(!sparseRet && NativeHelper.isNativeLibraryLoaded())
+			LibMatrixNative.matrixMult(m1, m2, ret, k);
+		else if(k > 1)
+			LibMatrixMult.matrixMult(m1, m2, ret, k, m1Perm, ultraSparse, sparse);
 		else
-			LibMatrixMult.matrixMult(m1, m2, ret);
-		
+			LibMatrixMult.matrixMult(m1, m2, ret, m1Perm, ultraSparse, sparse);
```
Baunsgaard, line 5128: all calls to kernel does skips some of the steps previously duplicated.

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
```diff
@@ -5096,34 +5081,52 @@ public final MatrixBlock aggregateBinaryOperations(MatrixBlock m1, MatrixBlock m
...
+			throw new DMLRuntimeException("Invalid aggregateBinaryOperatio: one of either input should be this");
+		return matrixMult(m1, m2, ret, op.getNumThreads());
+	}
+
+	public MatrixBlock matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k){
+
+		final int rl = m1.rlen;
+		final int cl = m2.clen;
+
+		if(m1.isEmptyBlock(false) || m2.isEmptyBlock(false)) {
```
Baunsgaard, line 5100: empty return is checked once in this method, and skipped in all kernel calls.

Baunsgaard:
Review requested!

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
```diff
@@ -5096,34 +5081,52 @@ public final MatrixBlock aggregateBinaryOperations(MatrixBlock m1, MatrixBlock m
...
 		else
-			ret.reset(rl, cl, sp.sparse, sp.estimatedNonZeros);
-		
-		//compute matrix multiplication (only supported binary aggregate operation)
-		if( NativeHelper.isNativeLibraryLoaded() )
-			LibMatrixNative.matrixMult(m1, m2, ret, op.getNumThreads());
-		else if( op.getNumThreads() > 1 )
-			LibMatrixMult.matrixMult(m1, m2, ret, op.getNumThreads());
+			ret.reset(rl, cl, sparseRet);
+		ret.allocateBlock();
```
mboehm7, line 5121: Do not allocate the block here, there are still cases in ultra-sparse matrix multiplications where the output representation is decided in a data-driven manner (e.g., permutation matrices).

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
```diff
@@ -5096,34 +5081,52 @@ public final MatrixBlock aggregateBinaryOperations(MatrixBlock m1, MatrixBlock m
...
+
+		if(!sparseRet && NativeHelper.isNativeLibraryLoaded())
```
mboehm7, line 5123: Why all this splitting of core functionality just for this check? For sparse inputs it would still enter the native lib, and then come back. Couldn't you instead just check if neither lhs nor rhs are sparse or even make the checks in LibMatrixNative a method and call it here (which would cover the common case)?

mboehm7 (commented):
Thanks @Baunsgaard. Overall, the motivation and goals of this PR are well-intended, but there are some issues. Could you please try to reduce the attack surface (number of entry points) and avoid splitting functionality across multiple places? The separation into aggregateBinaryOperations and matrixMult is fine, and a cleaner implementation seems possible.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixNative.java
```diff
@@ -66,18 +66,12 @@ public static boolean isMatMultMemoryBound(int m1Rlen, int m1Clen, int m2Clen) {
 	 * @param k number of threads
 	 */
 	public static void matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k) {
-		matrixMult(m1, m2, ret, k, true);
-	}
-	
-	public static void matrixMult(MatrixBlock m1, MatrixBlock m2, MatrixBlock ret, int k, boolean examSparsity) {
```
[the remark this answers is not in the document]
Baunsgaard: FixedRet is not implemented for NativeMM, and the argument was always set to true for all calls.
the one place the argument is used is for forcing allocation in fallback to normal matrix multiplication, and i have changed this call to do the same as before.

src/main/java/org/apache/sysds/runtime/matrix/data/MatrixBlock.java
Baunsgaard: resolved by moving it to general method call, but unfortunately since we currently need two boolean returns it is using a tuple adding a Object wrapper for the tuple and for each of the booleans, a future task is to refine it and possibly join these booleans to one, to allow simple static methods with boolean returns.

PR #1489, opened 2021-12-21
[SYSTEMDS-3167] Alternative Implementations of Eigen Decomposition

Lampisan:
[SYSTEMDS-3167] Alternative Implementations of Eigen Decomposition

This commit adds implementations of the following matrix algorithms in LibCommonsMath: Lanczos algorithm, QR decomposition, QR algorithm and Householder transformation (unused). Furthermore tests for the Lanczos and the QR algorithm are added.
Lanczos algorithm is numerically unstable and therefore sometimes produces incorrect results.
QR decomposition is working as intended.
QR algorithm works for symmetric matrices. While it should theoretically also work for non-symmetric matrices, it currently does not produce all correct eigenvectors in some cases. Furthermore a runtime-exception is thrown if the matrix has complex eigenvalues or the algorithm hasn't converged.
We are under the impression that our implementations of these algorithms are really slow compared to the existing implementation. Although we only tested this with a single thread this implementation is orders of magnitudes slower.

Co-authored-by: Nikolaus Grogger <nikolaus.grogger@student.tugraz.at>


src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -76,11 +79,17 @@ else if (opcode.equals("cholesky"))
...
 		else if (opcode.equals("lu"))
 			return computeLU(in);
 		else if (opcode.equals("eigen"))
 			return computeEigen(in);
-		else if ( opcode.equals("svd"))
+		else if (opcode.equals("eigen_lanczos"))
+			return computeEigenLanczos(in);
+		else if (opcode.equals("eigen_qr"))
+			return computeEigenQR(in);
+		else if (opcode.equals("svd"))
```
j143, line 92: Hi @Baunsgaard 

how about we document these functions and mention they are experimental (in the doc)?

Baunsgaard: for now no, these operations are only available sideways calling directly into LibCommonsMath

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -33,6 +33,9 @@
 import org.apache.commons.math3.linear.SingularValueDecomposition;
 import org.apache.sysds.runtime.DMLRuntimeException;
 import org.apache.sysds.runtime.data.DenseBlock;
+import org.apache.sysds.runtime.functionobjects.*;
```
Baunsgaard, line 36: don't do wildcard imports, only import things that are in use.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+	 * @param in matrix object
+	 * @return array of matrix blocks
+	 */
+	private static MatrixBlock[] computeEigenLanczos(MatrixBlock in) {
+		if(in.getNumRows() != in.getNumColumns()) {
+			throw new DMLRuntimeException(
+				"Lanczos algorithm and Eigen Decomposition can only be done on a square matrix. "
+					+ "Input matrix is rectangular (rows=" + in.getNumRows() + ", cols=" + in.getNumColumns() + ")");
+		}
+		if(!isSym(in)) {
```
Baunsgaard, line 307: this isSym check is taking much of your time, 
In the computeEigen we have skipped this check, i think it is fair to do the same here.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+			throw new DMLRuntimeException("Lanczos algorithm can only be done on a symmetric matrix.");
+		}
+
+		int num_Threads = 1;
```
Baunsgaard, line 311: give the number of threads as argument to the method.

and add tests with 1 and more threads.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+		int m = in.getNumRows();
+		MatrixBlock v0 = new MatrixBlock(m, 1, 0.0);
+		MatrixBlock v1 = MatrixBlock.randOperations(m, 1, 1.0, 0.0, 1.0, "UNIFORM", 0xC0FFEE);
```
Baunsgaard, line 315: i like the Coffee but allow the method to have a seed argument (you can always add coffee to the seed if you want to have a little easteregg)!

i would like it if you isolate the entire allocation of v1 into another method, this method should return the normalized v1.
This allows us to consider if we can find better ways of generating this, and profilers will make it obvious if this is a bottleneck.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+		int num_Threads = 1;
+		int m = in.rlen;
+
+		MatrixBlock A_n = new MatrixBlock(m, m, 0.0);
+		A_n.copy(in);
+
+		MatrixBlock Q_n = new MatrixBlock(m, m, 0.0);
+		for(int i = 0; i < m; i++) {
+			Q_n.setValue(i, i, 1.0);
```
Baunsgaard, line 384: you can allocate the matrix with a specific value in one of the other constructors.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+	 * @param in matrix object
+	 * @return array of matrix blocks [Q, R]
+	 */
+	private static MatrixBlock[] computeQR2(MatrixBlock in) {
+		if(in.getNumRows() != in.getNumColumns()) {
+			throw new DMLRuntimeException("QR2 Decomposition can only be done on a square matrix. "
+				+ "Input matrix is rectangular (rows=" + in.getNumRows() + ", cols=" + in.getNumColumns() + ")");
+		}
+
+		int num_Threads = 1;
```
Baunsgaard, line 376: add threads argument

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+		MatrixBlock Q_n = new MatrixBlock(m, m, 0.0);
+		for(int i = 0; i < m; i++) {
+			Q_n.setValue(i, i, 1.0);
+		}
+
+		ReorgOperator op_t = new ReorgOperator(SwapIndex.getSwapIndexFnObject(), num_Threads);
+		AggregateBinaryOperator op_mul_agg = InstructionUtils.getMatMultOperator(num_Threads);
+		BinaryOperator op_sub = InstructionUtils.parseExtendedBinaryOperator("-");
+		RightScalarOperator op_div_scalar = new RightScalarOperator(Divide.getDivideFnObject(), 1, num_Threads);
```
Baunsgaard, line 390: use higher abstraction as the field type (ScalarOperator)

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+		int num_Threads = 1;
+
+		int m = in.getNumRows();
+		MatrixBlock v0 = new MatrixBlock(m, 1, 0.0);
+		MatrixBlock v1 = MatrixBlock.randOperations(m, 1, 1.0, 0.0, 1.0, "UNIFORM", 0xC0FFEE);
+
+		// normalize v1
+		double v1_sum = v1.sum();
+		RightScalarOperator op_div_scalar = new RightScalarOperator(Divide.getDivideFnObject(), v1_sum, num_Threads);
```
Baunsgaard, line 319: Use higher abstraction for the field type

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+			w1 = in.aggregateBinaryOperations(in, v1, op_mul_agg);
+			MatrixBlock w1_t = w1.reorgOperations(op_t, new MatrixBlock(), 0, 0, m);
+			MatrixBlock alpha = w1_t.aggregateBinaryOperations(w1_t, v1, op_mul_agg);
+			if(i < m - 1) {
+				w1 = w1.ternaryOperations(op_minus_mul, v1, alpha, new MatrixBlock());
+				w1 = w1.ternaryOperations(op_minus_mul, v0, beta, new MatrixBlock());
+				beta.setValue(0, 0, Math.sqrt(w1.sumSq()));
+				v0.copy(v1);
+				op_div_scalar = (RightScalarOperator) op_div_scalar.setConstant(beta.getDouble(0, 0));
+				v1 = w1.scalarOperations(op_div_scalar, new MatrixBlock());
```
Baunsgaard, line 350: this can be done in place on v1

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+		MatrixBlock beta = new MatrixBlock(1, 1, 0.0);
+		for(int i = 0; i < m; i++) {
+			if(i == 0)
+				TV.copy(v1);
+			else
+				TV = TV.append(v1, new MatrixBlock(), true);
+
+			w1 = in.aggregateBinaryOperations(in, v1, op_mul_agg);
+			MatrixBlock w1_t = w1.reorgOperations(op_t, new MatrixBlock(), 0, 0, m);
```
Baunsgaard, line 342: i think we have a in place transpose of vectors. make sure you use it here to avoid allocation

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+
+			MatrixBlock vkt = vk.reorgOperations(op_t, new MatrixBlock(), 0, 0, m);
+			MatrixBlock vkvkt = vk.aggregateBinaryOperations(vk, vkt, op_mul_agg);
+			MatrixBlock vkvkt2 = vkvkt.scalarOperations(op_mult_2, new MatrixBlock());
+
+			A_n = A_n.binaryOperations(op_sub, A_n.aggregateBinaryOperations(vkvkt2, A_n, op_mul_agg));
+			Q_n = Q_n.binaryOperations(op_sub, Q_n.aggregateBinaryOperations(Q_n, vkvkt2, op_mul_agg));
+		}
+		// Q_n= Q
+		// A_n = R
```
Baunsgaard, line 412: remove commented code

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+	private static MatrixBlock[] computeEigenQR(MatrixBlock in) {
+		return computeEigenQR(in, 300);
+	}
+
+	private static MatrixBlock[] computeEigenQR(MatrixBlock in, int num_iterations) {
+		if(in.getNumRows() != in.getNumColumns()) {
+			throw new DMLRuntimeException("Eigen Decomposition (QR) can only be done on a square matrix. "
+				+ "Input matrix is rectangular (rows=" + in.getNumRows() + ", cols=" + in.getNumColumns() + ")");
+		}
+		int num_Threads = 1;
```
Baunsgaard, line 433: same, threads in argument

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+		double tol = 1e-10;
```
Baunsgaard, line 434: tolerance in argument

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+		int m = in.rlen;
+
+		AggregateBinaryOperator op_mul_agg = InstructionUtils.getMatMultOperator(num_Threads);
+
+		MatrixBlock Q_prod = new MatrixBlock(m, m, 0.0);
```
Baunsgaard, line 439: use other constructor

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+		for(int i = 0; i < m; i++) {
+			Q_prod.setValue(i, i, 1.0);
+		}
+
+		for(int i = 0; i < num_iterations; i++) {
```
Baunsgaard, line 444: add todo or make this loop parallel.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+			MatrixBlock[] QR = computeQR2(in);
+			Q_prod = Q_prod.aggregateBinaryOperations(Q_prod, QR[0], op_mul_agg);
+			in = in.aggregateBinaryOperations(QR[1], QR[0], op_mul_agg);
+		}
+
+		for(int i = 0; i < m; i++) {
+			for(int j = 0; j < m; j++) {
+				if(i != j) {
+					if(Math.abs(in.getValue(i, j)) > tol)
```
Baunsgaard, line 453: call getDenseBlockValues to allow you to iterate through a double[] array directly instead.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+							+ " (position i = " + i + ", j = " + j + ")");
+				}
+			}
+		}
+
+		// If complex evals then A is in real Schur-form
+		// 2x2 blocks on diagonal of A correspond to a matrix that has the same complex evals
+		double[] eval = new double[m];
+		for(int i = 0; i < m; i++) {
+			eval[i] = in.getValue(i, i);
```
Baunsgaard, line 464: here also leverage the double array from the comment above.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+		}
+
+		double[][] evec = DataConverter.convertToArray2DRowRealMatrix(Q_prod).getData();
```
Baunsgaard, line 467: if we can avoid getting the values out as 2dRowRealMatrix, then it would be nice.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+	}
+
+	/**
+	 * Sort the eigen values (and vectors) in increasing order (to be compatible w/ LAPACK.DSYEVR())
+	 *
+	 * @param eValues  Eigenvalues
+	 * @param eVectors Eigenvectors
+	 * @return array of matrix blocks
+	 */
+	private static MatrixBlock[] sortEVs(double[] eValues, double[][] eVectors) {
```
Baunsgaard, line 526: consider if you can modify this method to not use the double [][] of eVectors, but instead use a the already materialized double[] underlying the MatrixBlock.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
```diff
@@ -295,11 +281,283 @@ private static MatrixBlock computeMatrixInverse(Array2DRowRealMatrix in) {
...
+				}
+			}
+		}
+
+		MatrixBlock eval = DataConverter.convertToMatrixBlock(eValues, true);
+		MatrixBlock evec = DataConverter.convertToMatrixBlock(eVectors);
+		return new MatrixBlock[] {eval, evec};
+	}
+
+	private static boolean isSym(MatrixBlock in) {
```
Baunsgaard, line 553: 1. consider if you want to use it the one place it is in use i commented that you probably should not. if you decide to throw it away ignore 2.
2. The convertToArray2DRowReadlMatrix call makes this exceedingly slow, instead use the getDenseDoubleValues() to get a row major double[] that you can compare in.

src/test/java/org/apache/sysds/test/TestUtils.java
```diff
@@ -1802,6 +1802,16 @@ public static MatrixBlock generateTestMatrixBlock(int rows, int cols, double min
 		return MatrixBlock.randOperations(rows, cols, sparsity, min, max, "Uniform", seed);
 	}
 
+	public static MatrixBlock generateTestMatrixBlockSym(int rows, int cols, double min, double max, double sparsity, long seed){
```
Baunsgaard, line 1805: add java doc comment here, to make it easy to find this method,(add keywords or something) we have a bunch of tests using symmetric matrices and all implement their own, and it would be nice if everyone use the same generator (this could be it!).

src/test/java/org/apache/sysds/test/TestUtils.java
```diff
@@ -1802,6 +1802,16 @@ public static MatrixBlock generateTestMatrixBlock(int rows, int cols, double min
...
+		MatrixBlock m = MatrixBlock.randOperations(rows, cols, sparsity, min, max, "Uniform", seed);
+		for(int i = 0; i < rows; i++) {
```
Baunsgaard, line 1807: consider starting from i = 1

src/test/java/org/apache/sysds/test/component/matrix/EigenDecompTest.java
```diff
@@ -0,0 +1,112 @@
...
+				       2, 1, -2, -1};
+		in.init(a, 4, 4);
+		MatrixBlock[] m1 = LibCommonsMath.multiReturnOperations(in, "eigen");
+		MatrixBlock[] m2 = LibCommonsMath.multiReturnOperations(in, "eigen_lanczos");
+
+		TestUtils.compareMatrices(m1[0], m2[0], tol, "Result of eigenvalues of new eigen_lanczos function wrong");
+		testEvecValues(m1[1], m2[1], tol);
+	}
+
+	@Test public void testLanczosRandom() {
```
Baunsgaard, line 47: put Test keyword on the line above.

src/test/java/org/apache/sysds/test/component/matrix/EigenDecompTest.java
```diff
@@ -0,0 +1,112 @@
...
+public class EigenDecompTest {
+
+	@Test public void testLanczosSimple() {
+		double tol = 1e-4;
+
+		MatrixBlock in = new MatrixBlock(4, 4, false);
+		double[] a = { 4, 1, -2,  2,
+					   1, 2,  0,  1,
+		              -2, 0,  3, -2,
+				       2, 1, -2, -1};
```
Baunsgaard, line 38: format

src/test/java/org/apache/sysds/test/component/matrix/EigenDecompTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.sysds.runtime.matrix.data.LibCommonsMath;
+import org.apache.sysds.runtime.matrix.data.MatrixBlock;
+import org.apache.sysds.runtime.util.DataConverter;
+import org.apache.sysds.test.TestUtils;
+import org.junit.Test;
+import org.junit.Assert;
+
+public class EigenDecompTest {
+
+	@Test public void testLanczosSimple() {
```
Baunsgaard, line 31: keyword line above

src/test/java/org/apache/sysds/test/component/matrix/EigenDecompTest.java
```diff
@@ -0,0 +1,112 @@
...
+		testEvecValues(m1[1], m2[1], tol);
+	}
+
+	@Test public void testQREigenRandom() {
+		double tol = 1e-4;
+
+		MatrixBlock in = TestUtils.generateTestMatrixBlockSym(10, 10, 0.0, 1.0, 1.0, 1);
+		// MatrixBlock in = TestUtils.generateTestMatrixBlock(5, 5, 0.0, 1.0, 1.0, 5); // fails evals correct evecs wrong (evec corresponding to largest eval correct)
+		// MatrixBlock in = TestUtils.generateTestMatrixBlock(10, 10, 0.0, 1.0, 1.0, 2); // fails complex EVs
+		// MatrixBlock in = TestUtils.generateTestMatrixBlockSym(50, 50, 0.0, 1.0, 1.0, 1); // fails not converged
```
Baunsgaard, line 87: instead of leaving this outcommented, add more tests, each calling another method that execute the test with the given argument MatrixBlock "in"

Baunsgaard (commented):
Hi, @Lampisan and Nikolaus, (i did not find a github handle for you Nikolaus.)

I like the PR, there is some things we can do to improve the performance, let me know if you need help with any of them or you are running out of time.

1.
I suggest you try out using a profiler to find out where the time is used, one such tool is called visualvm. then when executing your tests you can modify them to run on bigger datasets or wrap them in loops to profile the methods.

2.
The testing as it is now is not sufficient, i would really like to see you spit the test from the input to then enable tests with more varied inputs.
ex:

```
public void testSmallEigen(){
... gen data
}

public void testEighen(MatrixBlock in){
... run test
}
```
3.
The implementation could benefit from considering what operations could be done in place --- without allocating intermediates--, and for the for loops we could consider allocating intermediates before the loops and using these for the results.

4.
Multi threading, you already have the argument and it should be simple to add the tests to include it, 

If you have any issues just reach out,

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Lampisan: We are not sure how to create an identity matrix using the existing MatrixBlock constructors.
We could create a double[] and use that with MatrixBlock.init(), but this might not be the solution you have in mind.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Lampisan: In the following calculations both w1 and w1_t are needed.
We tried to do it inplace (by double transposing), but that did not work.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Lampisan: We are not sure if we can parallelize that loop, since it is a iterative algorithm.
Or is there another way of making it parallel?

src/test/java/org/apache/sysds/test/TestUtils.java
Lampisan: If we use i=1 we ignore the first row/col completely.
We changed it and it did not work, maybe we misunderstand

Lampisan:
Hi @Baunsgaard,
Thanks for your suggestions, we added a few questions to your replies.
We will push the changes after resolving those.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Baunsgaard: Oh, i read it wrongly, i quickly saw it as a matrix you filled with a specific value.
for now lets keep it as it is, unless you observe that there is some bottleneck related to Q_n in the first iteration of the algorithm.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Baunsgaard: My comment is hard to understand and not saying what i intended.
You only use this transposed form on the next line for a matrix multiplication,

this can be seen as:

alpha = t(w1) %*% v1

you could consider using (this gives a equivalent output):

alpha  = t(t(v1) %*% W1)

since the v1 is a vector it will not be allocated when transposing (the underlying dense double[] is the same),
making it a metadata operation that cost nothing, unlike the transpose of W1 that is potentially costly.

The challenging part is to write the calls in such a way that they do not unnecessarily allocate alpha, and v1 more than once.

src/main/java/org/apache/sysds/runtime/matrix/data/LibCommonsMath.java
Baunsgaard: unclear comment from my side again,
i was referring to using the num_threads argument and make the instructions parallel.
this would entail adding the num_threads to the computeQR2 call.

Baunsgaard:
> Hi @Baunsgaard and @foebe (Nikolaus), Thanks for your suggestions, we added a few questions to your replies. We will push the changes after resolving those.

Thanks for the GitHub handle.

The comments you have responded on are indeed insufficiently described,
hopefully my new answers cover your questions.

Baunsgaard:
Thanks for the contribution, i will merge shortly after minor cleanup!
I close the PR to add the modifications in another ( #1561 ).

PR #1494, opened 2021-12-30
[MINOR] Removing dimension reduction algorithms from pipelines

Shafaq-Siddiqi:
and setting output buffering


Shafaq-Siddiqi:
Merged.

PR #1495, opened 2021-12-30
[SYSTEMDS-????] Equi-height binning in transformencode

phaniarnab:
This patch introduces single- and multi-threaded equi-height
binning. Currently we use partition-wise sorting and heap-based
merging of sorted blocks.


phaniarnab:
NOTE:
This PR is not to be merged now. I overlooked that there is another group of contributors is implementing this feature.

PR #1496, opened 2022-01-04
[SYSTEMDS-3260] Builtin for Matthews Correlation Coefficient

lederb:
This feature implements the Matthews Correlation Coefficient (MCC) calculation as a metric for binary classification evaluation.

The selected builtin function name  is "mcc" and its signature allows three modes of calculating the MCC:

* one vector for predictions and one vector for labels, entries should be binary but are tested. If they are not binary, values are interpreted as lower value -> 0, higher value -> 1

* alternatively a 2x2 confusion matrix can be used containing TN, FP, FN, TP counts in row major order

* option three is the calculation based on using named params TN, FP, FN, TP  as parameters for the mcc call directly.

Currently it contains a recursive call so checks on the confusion matrix are performed even if input were two vectors for predictions and labels, can be easily removed if unwanted. 

There are also some tests included, where some of those are expected to fail due to invalid input, the ones not expected to fail are compared against the result of a R mcc implementation.

### DIFF
--- a/scripts/builtin/mcc.dml
+++ b/scripts/builtin/mcc.dml
@@ -76,16 +76,21 @@ m_mcc = function(Matrix[Double] predictions = matrix(0,0,0), Matrix[Double] labe
         if (sum(confusionM) == 0) {
           stop("@mcc Input Error: confusion matrix contains only zeros")
         }
-        mode = 0
+        mattCC = s_mcc(TN=as.scalar(confusionM[1,1]), FP=as.scalar(confusionM[1,2]), 
+                    FN=as.scalar(confusionM[2,1]), TP=as.scalar(confusionM[2,2]))
     } else {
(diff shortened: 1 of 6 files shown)


lederb:
there is a typo in the comment at the moment (* instead of +), the actual computation is correct  

https://github.com/apache/systemds/blob/0cb9d586e586460de1b64de9f93fbaac8919716a/scripts/builtin/mcc.dml#L106

to avoid rerunning all the tests on the system again I'll wait fix a new push till there is some feedback in case there are other changes required.

Shafaq-Siddiqi:
LGTM,
Thank you @lederb for your contribution. During the merge, I fixed the formatting of the scripts as in DML we use two space formatting. I have also simplified the input parameters by only accepting the actual and prediction matrices and computing the confusion matrix internally.

lederb:
ohh okay. thank you very much for the refactoring 😃

PR #1497, opened 2022-01-05
[SYSTEMDS-3150] Outlier Detection via DBSCAN

ratschillerp:
[SYSTEMDS-3150] Outlier Detection via DBSCAN

This feature implements the outlier detection/prediction using a DBScan model.

The existing dbscan builtin is extended by returning a model which contains only the cluster members as the points coordinates.
The builtin function is called outlierByDB and takes an Xtest matrix, a model which is returned from the existing dbscan builtin and eps as parameters.
It is used to find outliers from newly generated data inside an existing cluster model.
Points are considered outliers if they are not within the eps distance of at least one member in a cluster.


j143:
Hi @ratschillerp - thanks for the feature.

A doc about this function https://github.com/apache/systemds/blob/main/docs/site/builtins-reference.md#outlier-function , would be great!

ratschillerp:
I added the documentation. 
The tests still have some problems, that I am unsure of. I sometimes find a few additional points considered inside the cluster that the R implementation doesn't have. I am still looking into it.

ratschillerp:
I now manually checked the Matrices, the issue seems to be that in DML the decimal places of my clustermembers are rounded to 3. If i use this representation in R, I also find those additional points.
Is there a way to use more precision (6 decimal places in this case) in DML?

corepointer:
> I now manually checked the Matrices, the issue seems to be that in DML the decimal places of my clustermembers are rounded to 3. If i use this representation in R, I also find those additional points. Is there a way to use more precision (6 decimal places in this case) in DML?

Hi! 
Could it be that you're referring to the default setting "decimal=3" used in DML's toString() function? See table 16 in [1]. An equivalent setting in R is print's "digits" setting (see [2]).

hth, Mark

[1] https://apache.github.io/systemds/site/dml-language-reference
[2] https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/print

ratschillerp:
Oh this could help a lot, thank you @corepointer

src/test/java/org/apache/sysds/test/functions/builtin/part2/BuiltinOutlierByDBTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Map;
+import java.util.Collection;
+
+public class BuiltinOutlierByDBTest extends AutomatedTestBase
+{
+	private final static String TEST_NAME = "outlierByDB";
+	private final static String TEST_DIR = "functions/builtin/";
+	private static final String TEST_CLASS_DIR = TEST_DIR + BuiltinOutlierByDBTest.class.getSimpleName() + "/";
+
+	private final static double eps = 1e-3;
```
j143, line 42: aim for a good `eps`, perhaps `1e-9`. 🎯

ratschillerp: The issue is the tests eps is either 0.0 or 1.0, since I return a yes/no mapping of outliers. 
Whats confusing me is that I get the exact same points, with some seemingly random ones added.

j143: I will take a closer look now. :)

ratschillerp: I figured something out: if I do some preprocessing in R. Removing the outliers and calling dbscan again, all the tests work 🤔

ratschillerp:
This should be done now.

PR #1502, opened 2022-01-12
[SYSTEMDS-3271] Publish docker test image when requested

Baunsgaard:
Adds a new docker-testImage.yml to allow manual initialization of building
our test image. This allows us to update the image used in our github
actions if for instance a new R package is added.

This commit also slightly modify the docker action to also build
the python image, when the default night cpu build is made.


j143 (commented):
Thanks for extending for better automation. 😄

PR #1503, opened 2022-01-12
[MINOR] Minor changes in logical enumeration

Shafaq-Siddiqi:
- This commit introduce the default parameter values for the cleaning primitives
  - The logical enumeration now only use the default parameter values to evaluate the
     primitives and iteratively add the categories to logical pipelines until converge.
  - The output is a single best logical pipeline


Shafaq-Siddiqi:
merged.

PR #1504, opened 2022-01-12
[SYSTEMDS-3272] setSchema built-in to set the schema of frame from DML

Shafaq-Siddiqi:
- This built-in will allow changing the schema of a frame inside DML
  - Function call is placed inside  detectSchema test

### DIFF
--- a/src/main/java/org/apache/sysds/common/Builtins.java
+++ b/src/main/java/org/apache/sysds/common/Builtins.java
@@ -109,6 +109,7 @@ public enum Builtins {
 	DECOMPRESS("decompress", false),
 	DEEPWALK("deepWalk", true),
 	DETECTSCHEMA("detectSchema", false),
+	SETSCHEMA("setSchema", false),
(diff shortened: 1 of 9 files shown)


phaniarnab:
Thanks for this extension @Shafaq-Siddiqi.
This crashes because FP32 and INT32 are missing in the exact serialized size calculation logic (`FrameBlock.java::getExactSerializedSize()`).
I will investigate it further.

Baunsgaard:
> Thanks for this extension @Shafaq-Siddiqi. This crashes because FP32 and INT32 are missing in the exact serialized size calculation logic (`FrameBlock.java::getExactSerializedSize()`). I will investigate it further.

This sounds related to the Bug just fixed in de3b9703d689d7877ccd8c7746ba10e42941fafa

Baunsgaard:
Is this PR no longer relevant?

phaniarnab:
Nope. It is a duplicate. Safe to close.

PR #1505, opened 2022-01-14
[SYSTEMDS-2837] Improve DAG linearization order

jhaberl:
Improve linearization order of LOPs in the DAG to reduce memory consumption of intermediates.

The existing solution is a depth-first algorithm. We modify this to execute LOPs that result in smaller intermediates first in order to reduce maximum memory consumption. This is achieved using a bottom-up approach considering memory estimates (OptimizerUtils.estimateSizeExactSparsity) of parent LOPs.

**Results**
We've compared the implementations using this script:
```
totalResult = matrix(0, rows=1000, cols=1)
for (i in 1:100) {
    A = rand(rows=1000, cols=1000)
    b = rand(rows=1000, cols=1000)
    C = rand(rows=1000, cols=1000)
    d = rand(rows=1000, cols=1000)
    result = (A + b) %*% rowMaxs(C + d)
    totalResult = totalResult + result
}
print(toString(totalResult))
```
In the existing implementations to calculate the result `A + b` is  executed first and results in a 1000x1000 matrix as an intermediate while calculating `rowMaxs`. In our solution, `rowMaxs` is executed first, which only leads to an intermediate vector of 1000x1.

For the following data, we've forced the garbage collector to run after every instruction.

When enabling JMLC_MEM_STATISTICS we get these results:
**Max size of live objects:**
Depth-first (existing): 30.526 MB (5 total objects)
Breadth-first: 38.155 MB (6 total objects)
Min-intermediate (ours): 22.904 MB (5 total objects)

This is a graph of memory consumption over time for the test script:
mem_consumption

For a more detailed view, the same graph but only the first 45 instructions. This includes two loop iterations.
mem_consumption_parts

### DIFF
--- a/src/main/java/org/apache/sysds/lops/compile/Dag.java
+++ b/src/main/java/org/apache/sysds/lops/compile/Dag.java
@@ -190,7 +190,6 @@ public class Dag<N extends Lop>
 		// create ordering of lops (for MR, we sort by level, while for all
 		// other exec types we use a two-level sorting of )
 		List<Lop> node_v = 
-			//doTopologicalSortStrictOrder(nodes) :
 			//doTopologicalSortTwoLevelOrder(nodes);
 			//doBreadthFirstSort(nodes);
 			doMinIntermediateSort(nodes);
@@ -254,7 +253,7 @@ public class Dag<N extends Lop>
 			v.stream().filter(l -> !l.getOutputs().isEmpty()).sorted(Comparator.comparing(l -> l.getID())),
 			v.stream().filter(l -> l.getOutputs().isEmpty())).collect(Collectors.toList());
 
-		printStats(nodes);
+		//printStats(nodes);
 
 		//NOTE: in contrast to hadoop execution modes, we avoid computing the transitive
 		//closure here to ensure linear time complexity because its unnecessary for CP and Spark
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/sysds/lops/compile/Dag.java
```diff
@@ -65,13 +65,7 @@
...
 import org.apache.sysds.runtime.meta.MatrixCharacteristics;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
Baunsgaard, line 68: no wildcard imports, i think perhaps that your editor does this.

Baunsgaard (commented):
I like the implementation, and your PR message!

We should probably add a config to set this, and some basic tests
please do as follows:

1. Go to src/main/java/org/apache/sysds/conf/DMLConfig.java
2. add a new setting (Perhaps INSTRUCTION_ORDER ?)
3. Make a enum type with the different versions of execution in this Dag fileto switch on.
4. Set a default value to the previous version
5. Add some tests of your new approach vs old. I would suggest to test it by comparing the results of the different approaches on some algorithms, including your example in the comment on the PR. If the results match then the tests should parse.

One example test where i compare results with different config is: src/test/java/org/apache/sysds/test/functions/compress/CompressScale.java

jhaberl:
We've now made your requested changes.

The setting we have added is "DAG_LINEARIZATION" (sysds.compile.linearization) with options "topological" and "min_intermediate". Default is "topological".

Our test cases include our initial example and two other existing tests (CsplineCG, LinearRegression) we have adapted to compare outputs using topological vs min_intermediate linearization. These existing test were helpful to us during debugging.

mboehm7:
This is a great patch (but I leave the review to @Baunsgaard as he already started) - during the merge, please rename the 'topological' to depth-first (as in the past topological sort referred to breadth-first but here it refers to our current depth-first), and potentially add the breadth-first as an enum option too (would be good to experiment with that externally).

Baunsgaard (commented):
Thanks for the commit, I like the implementation while merging i just moved some of the things around:

- Removed the new entry in the xml template, since introducing a default vaule there makes it harder for us to change default in the future. A much cleaner design is to set nothing, thereby using the system default setting than then can change as we please, and a user can set it if they know what they are doing.
- Changed the name Topological to depth_first
- Moved all the linearization algorithms to an interface file.
- Modified the test to test all linearizations + depth-first
- Moved tests to function test, and added them to github actions.

Baunsgaard:
Closing because imminent merge #1539

PR #1506, opened 2022-01-14
[SYSTEMDS-3149] Implement Decision Tree Prediction

mgd-hin:
A builtin/decisionTreePredict functionality is added that implements one of the three strategies of the Hummingbird paper (https://www.usenix.org/system/files/osdi20-nakandala.pdf). As discussed, only Tree Traversal strategy is implemented thus far. Further strategies (GEMM or PTT) will be added if there is still time after review.


scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+#
+# Builtin script implementing prediction based on classification trees with scale features using prediction methods of the
+# Hummingbird paper (https://www.usenix.org/system/files/osdi20-nakandala.pdf).
+#
+# INPUT   PARAMETERS:
+# ---------------------------------------------------------------------------------------------
+# NAME    TYPE             DEFAULT  MEANING
+# ---------------------------------------------------------------------------------------------
+# Decision tree in format:
+  # Matrix M where each column corresponds to a node in the learned tree and each row contains the following information:
```
Baunsgaard, line 31: consider linking to the training method for the description of the M input.

Also format it such that the comment sign '#' is the first character.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+  #  M[3,j]: Feature index of the feature (scale feature id if the feature is scale or categorical feature id if the feature is categorical)
+  #    that node j looks at if j is an internal node, otherwise 0
+  #  M[4,j]: Type of the feature that node j looks at if j is an internal node: holds the same information as R input vector
+  #  M[5,j]: If j is an internal node: 1 if the feature chosen for j is scale, otherwise the size of the subset of values
+  #    stored in rows 6,7,... if j is categorical
+  #    If j is a leaf node: number of misclassified samples reaching at node j
+  #  M[6:,j]: If j is an internal node: Threshold the example's feature value is compared to is stored at M[6,j] if the feature chosen for j is scale,
+  #     otherwise if the feature chosen for j is categorical rows 6,7,... depict the value subset chosen for j
+  #     If j is a leaf node 1 if j is impure and the number of samples at j > threshold, otherwise 0
+#
```
Baunsgaard, line 43: You are missing docs for X

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+#-------------------------------------------------------------
+
+#
+# Builtin script implementing prediction based on classification trees with scale features using prediction methods of the
+# Hummingbird paper (https://www.usenix.org/system/files/osdi20-nakandala.pdf).
+#
+# INPUT   PARAMETERS:
+# ---------------------------------------------------------------------------------------------
+# NAME    TYPE             DEFAULT  MEANING
```
Baunsgaard, line 28: The docs should be aligned with this header such that the thing you descripes type is starting with the TYPE etc.

See for instance scripts/builtin/msvm.dml

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+  #    stored in rows 6,7,... if j is categorical
+  #    If j is a leaf node: number of misclassified samples reaching at node j
+  #  M[6:,j]: If j is an internal node: Threshold the example's feature value is compared to is stored at M[6,j] if the feature chosen for j is scale,
+  #     otherwise if the feature chosen for j is categorical rows 6,7,... depict the value subset chosen for j
+  #     If j is a leaf node 1 if j is impure and the number of samples at j > threshold, otherwise 0
+#
+# strategy String  Prediction strategy, can be one of ["GEMM", "TT", "PTT"], referring to "Generic matrix multiplication", 
+#                     "Tree traversal", and "Perfect tree traversal", respectively
+# -------------------------------------------------------------------------------------------
+# OUTPUT   PARAMETERS:
```
Baunsgaard, line 47: only write
```
# OUTPUT:
```
on this line

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+# ---------------------------------------------------------------------------------------------
+# NAME    TYPE             DEFAULT  MEANING
+# ---------------------------------------------------------------------------------------------
+#	1- Matrix Y containing the predicted labels for X 
+# ---------------------------------------------------------------------------------------------
+
+m_decisionTreePredict = function(
+    Matrix[Double] M,
```
Baunsgaard, line 55: please format to not split the arguments in newlines.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+  } 
+  else {
+    print ("No such strategy" + strategy)
+    Y = matrix("0", rows=0, cols=0)
+  }
+}
+
+predict_TT = function (
+    Matrix[Double] M,
+    Matrix[Double] X
```
Baunsgaard, line 71: remove new lines

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+    Y[1, k] = class
+  }
+}
+
+createNodeTensors = function(
+  Matrix[Double] M
+  ) return (
+    Matrix[Double] N_L,
+    Matrix[Double] N_R,
+    Matrix[Double] N_F,
```
Baunsgaard, line 110: remove some newlines.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+      for (k in 1:ncol(C)){ 
+        # add node if it maps to a class
+        if (as.scalar(M[4, i]) == as.scalar(C[1, k])){           
+          N_C[i, k] = 1
+        }
+      }
+    }
+  }
+
+}
```
Baunsgaard, line 216: add newline in the end of the file.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+      C = unique_classes[1,]
+    }
+
+  N_C = matrix(0, rows=length(N), cols= length(C))
+
+  for (i in 1:ncol(N)){
+    
+    # check if node is leaf node   
+    if (as.scalar(I[1, i])==0){ 
+
```
Baunsgaard, line 206: remove these empty lines in the beginning or end of if statements.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+    Matrix[Double] N_L,
+    Matrix[Double] N_R,
+    Matrix[Double] N_F,
+    Matrix[Double] N_T,
+    Matrix[Double] N_C
+    ){
+      N = M[1,] # all tree nodes
+      I = M[2,] # list of node offsets to their left children
+
+      N_L = matrix(0, rows=1, cols=0)
```
Baunsgaard, line 117: i am worried about performance of this implementation of a empty matrix you cbind into.
Since N is known from the start you could allocate the correct size from the beginning.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+      N_R = matrix(0, rows=1, cols=0)
+      N_F = matrix(0, rows=1, cols=0)
+      N_T = matrix(0, rows=1, cols=0)
+
+      for (i in 1:ncol(N)){
```
Baunsgaard, line 122: you have 4 for loops in a row that goes to N, could these be joined?
also, if they can. consider if it is possible to use parfor instead if you already have removed the cbinds. (parfor is simply parallelizing each loop iteration.)

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+        # if the node is an internal node, add the threshold of the feature it evaluates
+        if (as.scalar(M[6,i]) != 0){
+          N_T = cbind(N_T, M[6,i])
+        
+        } else {
+          N_T = cbind(N_T, as.matrix(0))
+        }
+      }
+
+    unique_classes = matrix(0, rows=1, cols=1)
```
Baunsgaard, line 170: your indentation suddenly change here, please make the indentaton two spaces inside the function as a base.

scripts/builtin/decisionTreePredict.dml
```diff
@@ -0,0 +1,216 @@
...
+    predicted_labels = M[4,]
+    zero = FALSE
+
+    for (i in 1: ncol(predicted_labels)){
```
Baunsgaard, line 174: please consider if you could split this up into two parses,
1, find number of unique.
2. add prediction labels

Baunsgaard (commented):
I like the PR, 

But the current implementation seems a bit slow because of much use of cbind.
each call to cbind allocate a new double[], making this current implementation very expensive.
I would suggest to try to follow the comments to improve the performance.
I would also suggest for you to establish a baseline with this implementation using your tests, or calling a script, to see if your changes give improvements.

scripts/builtin/decisionTreePredict.dml
mgd-hin: I have removed the creation of the N_C vector altogether. In the paper that the implementation is based on, an N_C vector is created to hold the class label each leaf node predicts. However, since the input tree matrix M is already in such a format that the predicted label can be obtained by the node index, I think it's unnecessary to leave this in.

scripts/builtin/decisionTreePredict.dml
mgd-hin: The loops are now joined, parfor is used, and the matrices are pre-allocated (eliminating use of cbind). I averaged the test time for `testDecisionTreePredictSP()` over several runs and it appears faster compared to the previous implementation, although not by much (~5% on average).

Baunsgaard: how big is the input?
if the input is small then the different will be small.
if possible try something like 100k elements.

Baunsgaard: also mark comments as resolved if you have addressed them, thanks!

mgd-hin: Using larger input, the test time is reduced to about 30% of the previous implementation. I will therefore resolve this comment as well.

mgd-hin:
All issues were now addressed and resolved.

Baunsgaard:
LGTM, while merging i only changed a couple tabs and spaces.

PR #1507, opened 2022-01-14
[SYSTEMDS-3152] Min and Max support in aggregate() builtin

burimvrella:
This implementation adds support for the group "min" and "max"  operations in the existing aggregate() function. ([Reference](https://issues.apache.org/jira/browse/SYSTEMDS-3152))

What do you think about it? 
@Shafaq-Siddiqi


reichli:
We ran the test locally (test reports from IntelliJ are attached if you need them). Some of the Lineage-tests (FedUDFReuseTest, LineageTraceParforTest) seem to be flaky so we had to run them a second time. But in the end all tests passed.
[Test Results.zip](https://github.com/apache/systemds/files/7889784/Test.Results.zip)

mboehm7:
LGTM - thanks for this patch. I only did some minor cleanups of the tests, and left a TODO to separate function objects in the future.

PR #1509, opened 2022-01-18
[SYSTEMDS-3274] Disable docker login on pull_request event

j143:
Ideally, the docker credentials should never be used in the pull_request. If the testing images are outdated, it is a good practice to build changed Dockerfile's via github actions and merge into the main and then push to Dockerhub.


j143:
image

We could trigger docker cd on pull_request event at the moment.

PR #1513, opened 2022-01-18
[SYSTEMDS-3228] Constructing knn graph

losmib:
This patch adds a built-in function for constructing knn graph matrix. The algorithm first calculates all the distances between neighbors and constructs the distance matrix using the built-in function dist. Then for each row in this matrix, it finds the kth smallest distance for reference using quickselect and then puts 0 in the graph for all distances greater than this reference and 1 for distances smaller than the reference. This approach has the smallest expected asymptotic runtime O(N^2).


mboehm7:
@Shafaq-Siddiqi could you please have a look?

Shafaq-Siddiqi:
Thank you @losmib and @krautgortna for the contribution. While merging in, I fixed the formatting issues and removed the redundant operations. I would appreciate if anytime later you could add the documentation for this builtin.

PR #1514, opened 2022-01-20
[SYSTEMDS-3175] - Builtin for date processing

itsthestranger:
These commits add the builtin for date processing, which identifies the date column by looking at the data (via regular expressions) and enables transformations of this date column according to the input parameters. The function name for the builtin is "dateProcessing" and its signature allows 4 parameters:

- convertToNumber -> allows the conversion of the date column to numbers (i.e. timestamps (in milliseconds))
- convertToDominant -> allows the conversion of all dates in the date column to the dominating pattern in the date column
- valToAdd -> the value to be added/subtracted from the dates in the date column (returns dates in date-time format)
- timeformatToAdd -> the specific time-format of the value to be added/subtracted (i.e. if there should be seconds/hours/days/etc. added/subtracted)

### DIFF
--- a/scripts/builtin/dateProcessing.dml
+++ b/scripts/builtin/dateProcessing.dml
@@ -1,0 +1,29 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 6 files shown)


mboehm7:
@Shafaq-Siddiqi could you please have a look?

Shafaq-Siddiqi:
Thank you @itsthestranger for your contribution. During the merge, I have removed the code redundancy and have supported the methods via multiple map() calls.

itsthestranger:
Ah ok. Thank you very much.

PR #1515, opened 2022-01-20
[SYSTEMDS-3211] added date functions

craftbyte:
Added an overflow to toMillis that accepts a predetermined date format, introduced dateFormat() that takes either a string (one version tries to parse date, other allows inputting a date format) or a long containing a date and outputs in a specified format.


j143:
Hi @craftbyte  - thanks a lot for the contribution. 😀

How about adding usage instructions in [builtins-reference.md](https://github.com/apache/systemds/blob/main/docs/site/builtins-reference.md).

mboehm7:
LGTM - sorry for the delay and thanks for the patch @craftbyte. I merged it in with minor cleanups.

PR #1517, opened 2022-01-20
[SYSTEMDS-3227] Equi-height binning in transformencode/apply

Hiebl:
This patch extends the built-in functions transformencode and transformapply by equi-height binning on the local runtime. For this purpose the selected column gets sorted, then the bin boundaries are calulated as quantiles which then are converted to inout indicies and applied as bin boundaries analogous to already existing equi-width binning. The bining problem for non divisible bin numbers has been adressed with spillover. This has the advantage that all but the last bin a gurantued to be equi-height. The drawback is that the last bin has equi-height + inputlength % bin number height.


mboehm7:
LGTM - thanks for the patch @Hiebl. During the merge, I only did some minor code cleanups (avoid materialization of bin positions for quantiles), and left a FIXME for us to handle NaNs correctly (currently NaNs in the input would introduce zeros into the quantile computation).

@phaniarnab could you please address the NaN handling while finalizing your inputs to binning.
@OlgaOvcharenko once the federated quantile is ready, could you please extend the federated transform for equi-height as well?

Thanks everybody - this is a great addition.

PR #1518, opened 2022-01-20
[SYSTEMDS-2831] Foundation for new DeltaDDC Column Group

Muh-Osa:
This commit builds a foundation for a new DeltaEncoding ColGroup as specified in the corresponding JIRA task.
The main contributions are as follows:

- New ColGroupDeltaDDC class which builds upon the existing ColGroupDDC but with delta encoded column values as discussed. Currently the implementation handles scalar operations and compress/decompress to dense block as was outlined in the task.
- New DeltaDictionary for handling internal operations on the delta encoded data and corresponding tests
- New ColGroupDeltaDDCTest to testing compress and decompress with simple matrixes. (Would preferably be integrated into the existing CompressedMatrixTests later on, but for now serves as a standalone test for the new ColGroup)
- Integration into JolEstimateTests

Please have a look @Baunsgaard

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/compress/colgroup/AColGroup.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/colgroup/AColGroup.java
@@ -48,7 +48,7 @@ public abstract class AColGroup implements Serializable {
 
 	/** Public super types of compression ColGroups supported */
 	public enum CompressionType {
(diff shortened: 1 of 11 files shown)


Baunsgaard:
Hi @Muh-Osa 

I like the contribution, will merge shortly with some minor changes.
Then we will see in the future if we can find applicability of it.

PR #1487, opened 2021-12-18
[SYSTEMDS-3185] Federated Lookup Table Statistics

ywcb00:
Hi,
this PR (1) adds statistics regarding the federated lookup table to the federated statistics, (2) splits the federated PUT request count into separate counts for the individual Object types (e.g., Matrix, Frame, Scalar, and MatrixCharacteristics) and adds a respective byte count for matrices and frames, and (3) refactors the whole Statistics class and moves related statistics into their own statistics class (Codegen, HOP, Native, ParamServ, ParFor, Spark, Transform).

Thanks for review :)

PR #1519, opened 2022-01-21
[SYSTEMDS-3185] Refactor Statistics module

phaniarnab:
This patch refactors the Statistics class by moving related
statistics into their own classes in a new package, utils/stats.
Moreover, this adds statistics for federated lookup table access
and splits the federated PUT request count into separate counts
for the individual object types.


phaniarnab:
Running tests before merging PR #1487

PR #1520, opened 2022-01-21
[SYSTEMDS-3184] Builtin for computing information gain using entropy and gini

morf1us:
This builtin computes the measure of impurity for the given dataset based on the passed method. The current version expects the target vector to contain only 0 or 1 values and categorical data to be positive integers. Additionally, the builtin expects a row vector R to denote which features are continuous and which are categorical. In case of continuous features, the current implementation applies equal width binning.
It returns a row vector with gini gain or information gain for each feature. In both cases, the higher the gain, the better the split.

### DIFF
--- a/docs/site/builtins-reference.md
+++ b/docs/site/builtins-reference.md
@@ -50,6 +50,7 @@ limitations under the License.
     * [`img_brightness`-Function](#img_brightness-function)
     * [`img_crop`-Function](#img_crop-function)
     * [`img_mirror`-Function](#img_mirror-function)
+    * [`impurityMeasures`-Function](#impurityMeasures-function)
     * [`imputeByFD`-Function](#imputeByFD-function)
     * [`intersect`-Function](#intersect-function)
     * [`KMeans`-Function](#KMeans-function)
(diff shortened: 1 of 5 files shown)


j143:
Hi @morf1us - thanks a lot for the contribution. 😸

1. How about adding usage instructions in https://github.com/apache/systemds/blob/main/docs/site/builtins-reference.md
2. testing seems fine.

Keep working on finalizing `bins` related changes.

morf1us:
Hi @j143 Thanks for reviewing!

I added usage instructions and also added some more tests.

j143:
Hi @morf1us - Just curious. Did you notice the impurity measures logic in decisionTree.dml? The implementation in this PR and that one is same?

Docs are here: https://apache.github.io/systemds/site/algorithms-classification.html#decision-trees

https://github.com/apache/systemds/blob/b8d4897b2d46c5ae34a6d69772a66ad009feee54/scripts/builtin/decisionTree.dml#L264-L348

morf1us:
Hi @j143 Thanks for taking the time. Yes, it is quite similar. I was aware of both the [decisionTree](https://github.com/apache/systemds/blob/main/scripts/builtin/decisionTree.dml) and [randomForest](https://github.com/apache/systemds/blob/main/scripts/builtin/randomForest.dml) implementations before starting.

j143:
> Yes, it is quite similar. I was aware of both the [decisionTree](https://github.com/apache/systemds/blob/main/scripts/builtin/decisionTree.dml) and [randomForest](https://github.com/apache/systemds/blob/main/scripts/builtin/randomForest.dml) implementations before starting.


Yes, eventually we need to use Impurity measures inside the the scripts.

j143:
Thank you, @morf1us - LGTM. 👍 
🎉 🚀

We can work on using these measures functions inside the decisionTree scripts.

PR #1521, opened 2022-01-22
[SYSTEMDS-3185] Federated Read Cache

ywcb00:
Hi,
this PR introduces the _FederatedReadCache_ for the federated worker. The _FederatedReadCache_ is basically a map that maps from filename to a data object. In order to allow and optimize for concurrent READ_VAR requests on the federated site, the data object is wrapped inside a placeholder (_ReadCacheEntry_) so that only one thread reads the object from fs while the other threads wait for the data to be placed inside the placeholder.
In addition, this PR enables multithreaded processing of requests at the federated worker with a maximal threadpool size of 4 times the number of cpu processors.

Thanks for review :)


mboehm7:
LGTM - during the merge I fixed the following issues (which were only partially related to the change):
* On the multi-tenant tests, I got a `CreateProcess error=206, The filename or extension is too long` and hence modified these tests to pass the file name just once and then concatenate within dml
* Fixed the javadoc issue (missing read cache parameter)
* Fixed federated unary and binary aggregates (forced federated output was not waiting to completion; with the multi-threaded event loop the sequence of operations was no longer guaranteed)

PR #1522, opened 2022-01-24
[SYSTEMDS-3185] Lineage-based Reuse of Federated Reads

ywcb00:
Hi,
this PR replaces the _FederatedReadCache_ with a solution for reusing the reads based on the lineage.
With the lineage based solution, the reads are only reused with lineage reuse enabled on the federated worker.
I added two new methods to the LineageCache for reusing and putting a read object.

Thanks for review :)


phaniarnab:
Thanks @ywcb00. Looks good to me. I will merge this PR tonight with a few minor fixes.
I thought we want to keep both the read cache and the lineage-based read logic so that we can fall back to the read cache when lineage reuse is globally disabled.

ywcb00:
Thank you for the feedback @phaniarnab :)
As just discussed offline, I brought the read cache back again as a fall back whenever it is not possible to reuse it by lineage.
In addition, I moved the actual read logic into a separate method (_readDataNoReuse()_) to make the code more readable, and extended the tests to test both, reuse by lineage and the read cache.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -246,74 +246,43 @@ private FederatedResponse readData(String filename, DataType dataType,
...
-		if(cd == null) {
+		ExecutionContext ec = ecm.get(tid);
+		LineageItem linItem = new LineageItem(filename);
+
+		CacheableData<?> cd = null;
+
+		if(!LineageCache.reuseRead(Long.toString(id), dataType, linItem, ec)) {
+			if(ReuseCacheType.isNone() || dataType != DataType.MATRIX)
+				// using the read cache if lineage reuse is disabled or the data is not a matrix
+				cd = _frc.get(filename);
```
phaniarnab, line 257: Correct me if I am wrong. It is possible that lineage-based reuse is enabled after the worker is initiated. In that case, the read cache may not be empty. Hence, we may want to lookup both the caches. However, the condition above `_frc.get(filename)` restricts readcache lookup if lineage reuse is enabled.
I tried to remove the condition but that led to a test failure probably due to an uninitialized readcache.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -246,74 +246,43 @@ private FederatedResponse readData(String filename, DataType dataType,
...
 
-				// put meta data object in symbol table, read on first operation
-				cd.setMetaData(new MetaDataFormat(mc, fmt));
-				if(fmt == FileFormat.CSV)
-					cd.setFileFormatProperties(new FileFormatPropertiesCSV(header, delim,
-						DataExpression.DEFAULT_DELIM_SPARSE));
-				cd.enableCleanup(false); // guard against deletion
+					if(!ReuseCacheType.isNone() && dataType == DataType.MATRIX)
+						// put the object into the lineage cache
+						LineageCache.putReadObject(cd, linItem, ec, t1 - t0);
```
phaniarnab, line 268: Note: As we lazily read the actual data, this `computetime `only records the metadata read. A small computetime wrongly dictates the cache eviction logic to remove this entry early in case the cache is full. The `putReadObject` will trigger the actually read and place the MatrixBlock in the lineage cache.

phaniarnab:
Thanks, @ywcb00. I merged the changes in the main with minor fixes.
I left two FIXMEs in the code and commented the same in this PR. Please have a look.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
ywcb00: Thanks for pointing this out @phaniarnab, you are totally right. I will move this into the _putFedReadObject_ method to obtain the needed time for the actual read of the matrix. :)

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
ywcb00: I don't think that we allow for a dynamic change of the reuse flags on the federated worker yet. However, since I'm not 100% sure about that and we might want to add the possibility for changing the reuse types during runtime in the future, I will add the respective solution to the next PR.

PR #1523, opened 2022-01-25
[GIO][SINGLE-ROW] New Implementation of IOGEN

fathollahzadeh:
This PR is a new implementation of the GIO (generating readers for custom datasets). In the new implementation, removed all hard-coded implementations for flat datasets and replaced them with code gen. One of the primary goals of GIO is to support single and multi-row representations of tuples in source datasets. This PR is for supporting single row representation and the multi-row feature will be in a separate PR.   
An outline of the PR are:
1. It supports Matrix and Frame
2. It supports nested and flat datasets 
3. It supports non-standard datasets, i.e., an incomplete number of cols in a CSV file


Baunsgaard:
Hi @fathollahzadeh 

Is this PR still in use, or did everything move to the new PR ?

fathollahzadeh:
merged with a new implementation

PR #1524, opened 2022-01-25
[MINOR] Update Antler generated files to 4.8 

Baunsgaard:
Reason for change:
to avoid warning from Spark using different version and therefore warning on every execution.

Further suggestions: 
- I would like to suggest that we remove the files:
DmlLexer, DmlListener, DmlParser, and DMLBaseListener from GitHub, and instead git ignore these.
I know the decision was previously made to keep them but the current solution
for handling them specified in our .gitignore is (in my opinion) insufficient. 
  - The downside is we have to build them on first pull of systemds by running 'mvn package' once.
- I have removed the DML.tokens and DMLLexer.tokens files since these have changed in v 4.8 and are generated differently.  df231d6c1bffc25dc8477d7a2e3a71d59ee509f9 speficied these to be included, but i would suggest we try to git ignore them to see if they really are needed.


mboehm7:
As discussed offline, please keep the generated parser in the repository to keep the setup simple for new contributors (without the need for a maven build that creates the parser). Also please verify (if you didn't do it yet) that it still runs with Spark 2.4 (which also has an ANTLR dependency - getting warnings on Spark 3 is better than no warnings on Spark 3 but errors on Spark 2.4)

Baunsgaard:
Merged

PR #1525, opened 2022-01-26
[SYSTEMDS-3280] WIP: Add homomorphic encryption to Federated Parameter Server

flo-l:
This is WIP.

this adds homomorphic encryption functionality to Parameter Server. It allows a federated param server to encrypt the data of each client and train the network using this encrypted data and homomorphic operations. The data never leaves the clients without encryption.

TODO:

- [x] PrivacyConstraint handling
- [x] ~~need a Jira ticket ID~~ JIRA ticket added: https://issues.apache.org/jira/browse/SYSTEMDS-3280
- [x] proper commit message
- [x] license on all added files
- [ ] windows libsystemds cross-compiled binaries

@sebwrede


src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -417,9 +417,11 @@ private FederatedResponse execUDF(FederatedRequest request, ExecutionContextMap
 		ExecutionContext ec = ecm.get(request.getTID());
 
 		// get function and input parameters
+		// FIXME: handle HE privacy constraints
```
flo-l, line 420: @sebwrede I was unsure how to implement privacy constraint handling, so I disabled it while developing the rest of this patch. You told me you know this part of the code base very well. Do you maybe have suggestions for a good way forward?

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/homomorphicEncryption/SEALServer.java
```diff
@@ -0,0 +1,84 @@
...
+    public MatrixObject average(CiphertextMatrix encrypted_sum, PlaintextMatrix[] partial_plaintexts) {
+        double[] raw_result = NativeHelper.average(ctx, encrypted_sum.getData(), extractRawData(partial_plaintexts));
+        int[] dims = encrypted_sum.getDims();
+        int result_len = Arrays.stream(dims).reduce(1, (x,y) -> x*y);
+        DataCharacteristics dc = encrypted_sum.getDataCharacteristics();
+
+        DenseBlock new_dense_block = DenseBlockFactory.createDenseBlock(Arrays.copyOf(raw_result, result_len), dims);
+        MatrixBlock new_matrix_block = new MatrixBlock((int)dc.getRows(), (int)dc.getCols(), new_dense_block);
+        MatrixObject new_mo = new MatrixObject(Types.ValueType.FP64, OptimizerUtils.getUniqueTempFileName(), new MetaDataFormat(dc, Types.FileFormat.BINARY), new_matrix_block);
+        new_mo.exportData(); // write data, otherwise it might get evicted and thus get lost
```
flo-l, line 64: @mboehm7 I initially tried to keep these matrices in memory only, but I ran into problems. The data got evicted sometimes and then reloading it from a non-existent file naturally failed. My solution was to write them to the HDFS. As far as I understood that's the only way to use MatrixObject in general. A memory-only MatrixObject doesn't exist. Please correct me if I'm wrong.

mboehm7: Great progress, for now I just quickly answer the question before doing a full code review in the next days.

The constructor that takes a MatrixBlock bypassed our default eviction logic. Please, create a MatrixObject new_mo only with the meta data, and then call new_mo.acquireModify(new_matrix_block) which correctly hands it over to the buffer pool,  where it is evicted and restored as needed. An acquireRead then pins the object in memory, and release() unpins it, while acquireReadAndRelease is a shorthand to get the MatrixBlock. Conceptually a MatrixObject is an abstraction for hiding local vs distributed vs federated data, evictions, and multi-device replication; while a MatrixBlock is a plain old java object for in-memory matrices of different sparse/dense/compressed formats.

mboehm7: Btw, the license check (aka Rat check) fails because all Java and script files need to have the ASF license on top.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
sebwrede: When the `-federatedCompilation` flag is added to the execution of the DML script, the rewriter `RewriteFederatedExecution.java` will be called where the privacy constraints will be retrieved from the federated workers for all federated data inputs in the script. You need to add a check for when the privacy level "Private" is set for an input to the parameter server so that you can switch to the encrypted version of the federated parameter server. 
In the FederatedWorkerHandler, the call to the PrivacyMonitor needs to check if the UDF received is the UDF from the encrypted version of the parameter server. If this is the case, the PrivacyMonitor should not block data inputs with the privacy level "Private". For all other UDFs or instructions, the PrivacyMonitor should still block requests for the "Private" data elements, since the data elements will be transferred unencrypted to the coordinator otherwise.
The idea behind the PrivacyMonitor check in the FederatedWorkerHandler is that the UDF has some input data and the privacy constraints of that input is checked. It is the responsibility of the SystemDS developer to ensure that no other data is exposed when running the UDF since it is possible to access other data in the UDF execution without checking the privacy level of those data elements.  
Regarding the propagation of the privacy constraints, this happens in the `PrivacyPropagator.java` class. During runtime, CPInstructions will, before execution, propagate the privacy constraints from the inputs to the outputs. In your case, I don't think that you execute any CPInstructions since you only send a UDF. In the UDF, you need to handle the propagation if you leave any output on the federated workers. You can add such propagation logic to the PrivacyPropagator class and reuse the logic in the corePropagation method by calling that method. If you do not leave any output on the federated workers then you do not have to think about the propagation, but you still have to think about which data elements you send back to the coordinator. If the data has the privacy level "Private" then you should only transfer the encrypted version of it. The encrypted data does not need the privacy constraint unless the decryption key is transferred also (which it is not because only the federated workers can fully decrypt it). When the federated workers decrypt the aggregated data sent from the coordinator, you need to add a privacy constraint to that output so that it is not sent back to the coordinator without being encrypted again. The output should again have the privacy level "Private". 

This is now a long comment :smile: . Please ask if anything is unclear.

flo-l: haha thank you for this extensive write-up! OK this clears things up for me. I've already implemented the automatic switch to homomorphic encryption. As I'm only accessing the features and labels in one UDF, I will just load these without going through the PrivacyMonitor. I've made the relevant changes in commit 9ac7f62d1088ff43827d8fa1b5a056d88061fe86.

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/homomorphicEncryption/SEALServer.java
flo-l: Thank you for the explanation. This is fixed in 13a81b77b3e401fa5e1d1cd2f2ea9ae257a6a0fb. I think the constructor that takes a MatrixBlock is faulty then, because this is undocumented. Maybe it was meant to be private?

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/FederatedPSControlThread.java
```diff
@@ -324,6 +401,7 @@ protected ListObject pullModel() {
 	}
 
 	protected void weightAndPushGradients(ListObject gradients) {
+		assert (!(_weighting && _use_homomorphic_encryption)) : "weights and homomorphic encryption are not supported together";
```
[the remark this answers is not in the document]
flo-l: good catch! I moved this check into ParamservBuiltinCPInstruction.java#runFederated. There is a similar check already there which makes sure we're using modelAvg. I'll leave the assert as a sanity check. fixed in 3f9bbdb64e771343576c6ab50588e605a49acf5e.

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/FederatedPSControlThread.java
```diff
@@ -354,11 +432,12 @@ protected void computeWithBatchUpdates() {
 				int localStartBatchNum = getNextLocalBatchNum(currentLocalBatchNumber++, _possibleBatchesPerLocalEpoch);
 				ListObject model = pullModel();
 				ListObject gradients = computeGradientsForNBatches(model, 1, localStartBatchNum);
-				if (_modelAvg)
+				if (_modelAvg && !_use_homomorphic_encryption)
+					// we can't call the agg fn if we use HE, because it is implemented homomorphically in SEALServer::aggregateCiphertexts
```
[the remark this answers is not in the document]
flo-l: I don't think so, because this code also runs in non-encrypted mode. In this case we need to call the agg fn.

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/FederatedPSControlThread.java
```diff
@@ -160,21 +161,45 @@ public void setup(double weightingFactor) {
...
+									_ps.getAggInst().getFunctionName(), _ec.getListObject("hyperparams"),
+									_modelVarID, _nbatches, a)));
+		} else {
+			udfResponse = _featuresData.executeFederatedOperation(
+					new FederatedRequest(RequestType.EXEC_UDF, _featuresData.getVarID(),
+							new SetupFederatedWorker(_batchSize, dataSize, _possibleBatchesPerLocalEpoch,
+									programSerialized, _inst.getNamespace(), _inst.getFunctionName(),
+									_ps.getAggInst().getFunctionName(), _ec.getListObject("hyperparams"),
+									_modelVarID, _nbatches, _modelAvg)));
+		}
```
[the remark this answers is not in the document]
flo-l: nice catch and excellent idea! fixed in 90018118a.

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/homomorphicEncryption/SEALServer.java
flo-l: licenses fixed in 3a12c9f6d

sebwrede (commented):
I think I am missing something related to SEAL in my setup. When I run build.sh, I get: 
`CMake Error at CMakeLists.txt:84 (find_package):
  By not providing "FindSEAL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SEAL", but
  CMake did not find one.` 
  This in turn means that I cannot run `homomorphicEncryption/InOutTest`.
  Can you explain how to get this to run? 
  At some point, we also need to add your homomorphicEncryption test package to the GitHub workflow, so that it is executed whenever a push is made to the repository. This can be set up by adding the package to `.github/workflows/functionsTests.yml` somewhere between line 52 and 68.

flo-l:
>  Can you explain how to get this to run?

You have to install SEAL and it's development headers on your system. This can be done by compiling from source like described [here](https://github.com/microsoft/SEAL#installing-microsoft-seal). It's easy to do if you already have cmake on your system.

The alternative if you're on Linux: use the precompiled lib I added to this PR (i.e. don't run build.sh).

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/homomorphicEncryption/SEALServer.java
```diff
@@ -0,0 +1,104 @@
...
+
+    // accumulates the given ciphertext blocks into a sum ciphertext and returns it
+    // stores c0 of the sum to be used in averageBlocks()
+    public CiphertextMatrix accumulateCiphertexts(CiphertextMatrix[] ciphertexts) {
+        return new CiphertextMatrix(ciphertexts[0].getDims(), ciphertexts[0].getDataCharacteristics(), NativeHelper.accumulateCiphertexts(ctx, extractRawData(ciphertexts)));
+    }
+
+    // averages the partial decryptions and stores the result in old_mo
+    // encrypted_sum is the result of accumulateCiphertexts() and partial_plaintexts is the result of partiallyDecryptBlock
+    // of each ciphertext fed into accumulateCiphertexts
```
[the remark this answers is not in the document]
flo-l: fixed in 1d738f1766c8369884ce0b3ed793fe2bc7d0d1c1

sebwrede:
> The alternative if you're on Linux: use the precompiled lib I added to this PR (i.e. don't run build.sh).

I am on Linux. Before I executed build.sh, I still got an error message. Maybe it is not related to SEAL, actually. 
I get:
`java.lang.UnsatisfiedLinkError: /tmp/libsystemds_openblas-Linux-x86_64.so6067933: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/libsystemds_openblas-Linux-x86_64.so60679330)`

flo-l:
>  java.lang.UnsatisfiedLinkError: /tmp/libsystemds_openblas-Linux-x86_64.so6067933: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.32' not found (required by /tmp/libsystemds_openblas-Linux-x86_64.so60679330)`

oh this sounds like a version mismatch between my libc and yours. I'm on fedora so I always have quite recent packages. But idk what's the policy here regarding libc linking. Or how to fix this tbh.

On a sidenote: I just checked and SEAL is linked statically, so you don't have to install it unless you want to run build.sh.

sebwrede:
> > java.lang.UnsatisfiedLinkError: /tmp/libsystemds_openblas-Linux-x86_64.so6067933: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.32' not found (required by /tmp/libsystemds_openblas-Linux-x86_64.so60679330)`
> 
> oh this sounds like a version mismatch between my libc and yours. I'm on fedora so I always have quite recent packages. But idk what's the policy here regarding libc linking. Or how to fix this tbh.
> 
> On a sidenote: I just checked and SEAL is linked statically, so you don't have to install it unless you want to run build.sh.

I think you are right. I actually don't know if we have a standard for libc versioning and linking.

sebwrede:
**Just to give an update on the merging of this PR:**
I have been trying for the last couple of days to set up SEAL in the Docker test image so that the HE tests can be run by GitHub actions. This is still unsuccessful. It seems like I can set up SEAL sucessfully in the Docker container, but something goes wrong with the linking so when I run the tests, SystemDS cannot use SEAL and therefore the tests fail. 
I don't know what to do about this yet, but I am planning to disable the HE tests for this PR and merge it with the tests disabled and then add the Docker details in a later PR where the tests are activated. 

**An unrelated comment:** It looks like the licenses are still missing from this PR. We need a license in the top of every code file added. See [this link](https://github.com/apache/systemds/blob/main/CONTRIBUTING.md) under the headline "License" how the licenses need to look for the different programming languages.

Baunsgaard:
> **Just to give an update on the merging of this PR:** I have been trying for the last couple of days to set up SEAL in the Docker test image so that the HE tests can be run by GitHub actions. This is still unsuccessful. It seems like I can set up SEAL sucessfully in the Docker container, but something goes wrong with the linking so when I run the tests, SystemDS cannot use SEAL and therefore the tests fail. I don't know what to do about this yet, but I am planning to disable the HE tests for this PR and merge it with the tests disabled and then add the Docker details in a later PR where the tests are activated.
> 
> **An unrelated comment:** It looks like the licenses are still missing from this PR. We need a license in the top of every code file added. See [this link](https://github.com/apache/systemds/blob/main/CONTRIBUTING.md) under the headline "License" how the licenses need to look for the different programming languages.

I would suggest that you open a new PR where you modify the Docker container.
Then once we are happy with the changes we merge it and let git update the container.

sebwrede:
> this current implementation means that we built in a dependency to mkl/blas and now SEAL in the same resource.

I think this is why it does not work in Docker currently. SEAL would only work if we use mkl/blas as well and this is not something we do in the Docker tests at the moment. 

> Preferably i think it would be better if we separate these two dependencies such that:
> 
>     1. We make a new header file for the SEAL dependency and build into new different .so /.dll filles for SEAL bindings in src/main/cpp/lib, perhaps called libsystemds_seal-Linux-x86_64.so and libsystemds_seal-Windows-AMD64.so.
> 
>     2. We make a new Java file that binds to it (instead of extending src/main/java/org/apache/sysds/utils/NativeHelper.java)

I think suggestion 1 is a good idea, but suggestion 2 depends on how we define the purpose of NativeHelper. Currently, it is the only interface to the libraries, which means that we will always know where to look for the native calls. I see that until now the native helper has only been used for the mkl/blas calls, but I don't know if this has been a design decision or a coincidence since it was the only libraries added in the beginning.

flo-l:
Thank you for your comments :)

I'm with @sebwrede. It really depends how we're defining the purpose of NativeHelper. The name suggests it's a general binding mechanism to all native code.

I want to make a strong case in favor of building only one native lib and then linking all java code through only one class (NativeHelper) with it. The problem with these JNI bindings is complexity. The system is already quite complex and it was hard for me as a new contributor to the project to find my way around. Keeping linkage and native Java methods in one place each really is the simplest solution. Once we start linking several dylibs complexity only grows, with no big gain in code clarity imho. I would even go so far and say that a multi-lib solution would be way harder to understand, debug and get running.

flo-l:
@sebwrede some of my commits got lost in the course of squashing. I've restored those and pushed the proper result. sorry for the inconvenience!

sebwrede:
Could you add your additional commits so that I can merge this branch?

flo-l:
@sebwrede I've made it link to SEAL dynamically, recompiled the linux artifacts (this time on a proper docker image with a sufficiently old glibc) and forced pushed everything. the libraries are now considerably smaller (only 200 kB instead of 10 MB)

flo-l:
Dear @sebwrede !

I've updated this PR. SEAL is now compiled into its own library, which gets loaded if HE is used in the code. I've moved the native method definitions in Java to their own class.

sebwrede:
> Dear @sebwrede !
> 
> I've updated this PR. SEAL is now compiled into its own library, which gets loaded if HE is used in the code. I've moved the native method definitions in Java to their own class.

Thanks for this update, @flo-l !

Do you know why there are still changes to the mkl and openblas ".so" files? It does not look like you have changed any files that could cause this, so I am not sure what has been changed in those files. Maybe it is just because you have recompiled them on your computer and I can just exclude the changes from the merge commit.

sebwrede:
I have rebased the branch on main, but since there are a lot of conflicts, it is not easy to get it right.
I think there is something wrong with my merge of FederatedWorker.java, but I cannot see what is wrong. All the federated tests seem to be failing because of it. Can you see if anything looks wrong in this rebased version of FederatedWorker.java?: https://github.com/sebwrede/systemds/blob/1f8f6ed84ef3cd5c996e443005820e33bb43c1c8/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorker.java

Another problem is running the InOutTest. It leads to: 

`java.lang.UnsatisfiedLinkError: /tmp/libhe-Linux-x86_64.so6291387220729597515: libseal.so.3.7: cannot open shared object file: No such file or directory`

So I understand that this is because SEAL has to be included, but I think we need to add a readme or something to define exactly how to include it, now that it is not included statically. 

Another problem is running the build.sh. I get the following error:

`/systemds/src/main/cpp/he/libhe.cpp:23:10: fatal error: gsl/span: No such file or directory
   23 | #include <gsl/span>
      |          ^~~~~~~~~~
compilation terminated.
`

So one idea could be to include a check for whether gsl/span is there in the build.sh before trying to compile and then install before the compilation. Or maybe you have some better idea for a solution.

flo-l:
One thing: in [this line](https://github.com/sebwrede/systemds/blob/1f8f6ed84ef3cd5c996e443005820e33bb43c1c8/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorker.java#L116), I think it should be `FederatedResponseEncoder`. Still, it fails locally with:
```
org.apache.sysds.runtime.DMLRuntimeException: Failed sending federated operation
	at org.apache.sysds.runtime.controlprogram.federated.FederatedData.executeFederatedOperation(FederatedData.java:206)
	at org.apache.sysds.runtime.controlprogram.federated.FederatedData.clearFederatedWorkers(FederatedData.java:236)
	at org.apache.sysds.api.DMLScript.cleanupHadoopExecution(DMLScript.java:548)
	at org.apache.sysds.api.DMLScript.execute(DMLScript.java:465)
	at org.apache.sysds.api.DMLScript.executeScript(DMLScript.java:304)
	at org.apache.sysds.api.DMLScript.main(DMLScript.java:198)
Caused by: java.lang.IllegalStateException: group set already
	at io.netty.bootstrap.AbstractBootstrap.group(AbstractBootstrap.java:92)
	at org.apache.sysds.runtime.controlprogram.federated.FederatedData.executeFederatedOperation(FederatedData.java:177)
	... 5 more
```

I don't know what's wrong here, but it seems unrelated to my changes.

Regarding the dynamic linking: Yes, it would be nice to document this, but where?

Both errors you're getting stem from SEAL not being installed on the system. In the first case the dynamic library is missing and the build error comes from the SEAL header files not being installed (SEAL bundles gsl).

flo-l:
Installing SEAL is simple:
```
git clone https://github.com/microsoft/SEAL.git
cd SEAL
git checkout v3.7.3
cmake -S . -B build -DBUILD_SHARED_LIBS=ON
cmake --build build
sudo cmake --install build
```

sebwrede:
> Installing SEAL is simple:
> 
> ```
> git clone https://github.com/microsoft/SEAL.git
> cd SEAL
> git checkout v3.7.3
> cmake -S . -B build -DBUILD_SHARED_LIBS=ON
> cmake --build build
> sudo cmake --install build
> ```

I already installed SEAL and I just tried again with your instructions, but it still gives me the same error message when running the test. Am I missing some other step? 

Regarding the other exception for the federated tests, I guess something went wrong during rebase. Maybe I should try again.

PR #1535, opened 2022-02-05
[SYSTEMDS-3185] Worker-specific Multithreading of Operations

ywcb00:
Hi,
This PR adds the _MultiThreadedOperator_ class as a superclass for all operators supporting multithreading. This allows us to replace the number of threads of the operator with a worker-specific amount after parsing the received instruction from the EXEC_INST request.
With this PR, we are setting the number of threads to the amount of processors available to the JVM of the federated worker whenever the operator allows for multithreading (i.e. a subclass of _MultiThreadedOperator_).

Thanks for review :)


mboehm7:
LGTM - thanks for the patch @ywcb00. This is a nice, generic way of handling degree of parallelism at the federated workers. Before merging it in, I'd like to understand why the Python tests are now failing in this and other PRs but not on main.

Baunsgaard:
@mboehm7 

Python test failed because when reading in a file for some reason we to import Apache commons-compress and that is not available in our bin release.

The only place where we have automated the testing of our bin release is in Python. We should add a test of our normal system that use our bin release.

See  336ac8b4a0c698094d6f3300cb79d08e5ccff06e for fix.

mboehm7:
ok thanks for the clarification.

mboehm7:
LGTM again, and sorry for the delay.

PR #1537, opened 2022-02-06
[MINOR] Include Apache commons compress tar in release

Baunsgaard:


Shafaq-Siddiqi:
Thanks for the fix @Baunsgaard

PR #1540, opened 2022-02-07
[SYSTEMDS-3185] Lineage-based reuse of federated reads

phaniarnab:
This patch adds lineage-based reuse of federated reads on the workers.
We fall back to the read cache if lineage-based reuse is globally disabled.


phaniarnab:
Running tests.

PR #1542, opened 2022-02-08
[SYSTEMDS-3185] Federated Read Reuse - fix computetime and robustness

ywcb00:
Hi,
This PR addresses the two issues mentioned in #1522, namely (1) obtaining the wrong computation time from a lazy read, and (2) allowing to change the lineage reuse type on the worker during runtime.

Thanks for review :)


phaniarnab:
Thanks for the fixes @ywcb00. I just checked in the changes.

PR #1558, opened 2022-03-11
[SYSTEMDS-3305] Federated qpick binning modification

OlgaOvcharenko:
This PR modifies federated qpick instruction binning. Binning is implemented similar to transform encode, bin index = Math.ceil((val - _min) / _range).


mboehm7:
Thanks for the patch @OlgaOvcharenko - this is great, but the test failure is reproducible also in my local environment. Could you please have a look?

OlgaOvcharenko:
Sure, it seems like it's a flaky test.

OlgaOvcharenko:
@mboehm7 federated quantile is fixed now, could you please have a look? 
Also, I modified federated variance aggregation (tests were failing before).

mboehm7:
LGTM - thanks @OlgaOvcharenko this works like a charm, and on a 3x100,000,000 vector the federated quantile is even faster than local (67s vs 87s) because local qsort is still single-threaded (if I force a multi-threaded qsort if comes to 20.5s).

PR #1559, opened 2022-03-13
[SYSTEMDS-3185] Federated Lineage Trace Collection

ywcb00:
Hi,
This PR introduces the MatrixLineagePair as a wrapper around a _MatrixObject_ with its corresponding _LineageItem_ in order to nicely couple the information of the lineage with a matrix until we broadcast it to the federated worker. The lineage trace is then included inside the PUT request of the matrix and transferred to the federated worker. To do this, this PR includes also a refactoring of the federated instructions which changes the MatrixObjects into MatrixLineagePairs whenever they are broadcast.

Thanks for review :)


phaniarnab:
Thanks for the patch @ywcb00. And sorry for the delay.

PR #1562, opened 2022-03-17
[SYSTEMDS-3328] Federated transform for equi-height

OlgaOvcharenko:
This PR extends federated transform for equi-height.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko and sorry for the delay. During the merge, I resolve the conflicts that have been piled up, and done some minor cleanups. As a follow-up cleanup it would be great to move the federated-specific things out of the location-agnostic encoders.

PR #1565, opened 2022-03-18
Compression Speedup

Baunsgaard:
This PR adds speedups to the compression by adding:
-  a direct path compression,
- remove nnz count  bug
-  sparse combine optimization in cocoding
- greedy cocode first combine parallel.


Baunsgaard:
Close Because of #1589

PR #1566, opened 2022-03-20
[SYSTEMDS-3334] Code-gen rewrite ROWMAXS_VECTMULT

corepointer:
This patch adds a rewrite to avoid vector intermediates in the generated row template of connected components by doing the elementwise multiplication, row_maxs and max in one pass.

Works only for a certain small input size. Then other rewrites seem to get to modify the DAG before this pattern can be tested. E.g., tested with 1000 nodes -> works, tested with 100000 nodes -> pattern not applied because DAG is already different.

This is for java codegen only. The CUDA version works analogously and will be pushed with other pending CUDA codegen changes (and after the input size issue is fixed)

### DIFF
--- a/src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
+++ b/src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
@@ -38,6 +38,7 @@ import org.apache.sysds.hops.Hop;
 import org.apache.sysds.hops.LiteralOp;
 import org.apache.sysds.hops.OptimizerUtils;
 import org.apache.sysds.hops.codegen.cplan.CNode;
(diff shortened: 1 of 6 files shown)


corepointer:
The rewrite was tuned down from max(rowMaxs(vec_mult())) to only rowMaxs(vect_mult()). This way the rewrite would apply more easily. The subsequent max() was fused anyway. 
Nevertheless, the rewrite still has issues with memory estimates when running hybrid and/or GPU exec modes. 
And last but not least this is an over-specialized rewrite that could be generalized into agg+vect_op. 
Will close with initial working implementation for Java + CUDA (leaving the mentioned issues as future work)

PR #1567, opened 2022-03-21
[SYSTEMDS-3024] Improve performance by batching data descriptor trans…

corepointer:
…fers

The spoof cuda operators do several little cudaMemcpy() invocations per operator execution. By transferring all data in one go the overhead can be reduced. In addition, using asynchronous copies can further improve things and are a first step towards using more asynchronicity in the GPU operations.


corepointer:
LGTMS - CUDA codegen tests ran through just fine.

PR #1568, opened 2022-03-21
[MINOR] Cleaning Pipelines: improvements in resource allocation 

Shafaq-Siddiqi:
- This commit used weights for computing the number of resources for each bucket
   the pipelines in buckets are sorted in order of accuracy where the last bucket keeps the most
   performant pipelines. If the initial resource value is 50 then instead of assigning 50 resources
   to each bucket, this commit computes the weight vector and then assign the resources accordingly ensuring the
   the last bucket gets max resource value.


Shafaq-Siddiqi:
A new PR is opened.

PR #1569, opened 2022-03-21
[MINOR] Fed Statistics: Transferred Objects

ywcb00:
Hi,
This PR extends the federated count and bytes statistics for put objects to include the inputs of a UDF.


phaniarnab:
Thanks for the fix @ywcb00.

PR #1574, opened 2022-03-30
[SYSTEMDS-3185] Federated Broadcast Slices Lineage

ywcb00:
Hi,
This PR addresses the problem of differentiating between different slices from the same broadcast data object. At the moment we can only identify broadcast slices by its original data object, not by its slice indices. This leads to reuse collisions when broadcasting two different slices from the same original file (i.e., different slice indices). However, this PR resolves the problem by appending the slice indices in form of a RightIndex lineage item to the transferred lineage trace. Then, we can identify the slice range on the federatet worker, and hence differentiate between the different data slices.

Thanks for review :)

PS: Note that github shows the added test _FederatedReuseSlicesTest_ as if the _FederatedReadCacheTest_ would have been renamed. That is actually not the case. The _FederatedReadCacheTest_ was actually replaced by the _FederatedReuseReadTest_ when we added the lineage based reuse of read objects. However, somehow this test managed to get again into the system by a mistaken rebase, I am sorry for that. :bowtie:


src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
```diff
@@ -226,11 +228,27 @@ public FederatedRequest broadcast(ScalarObject scalar) {
...
+		LineageItem li = null;
+		if(objLi != null) {
+			final String scalarCPOSuffix = InstructionUtils.concatOperands(
+				DataType.SCALAR.name(), ValueType.INT64.name(), String.valueOf(true));
+			// manually create a lineage item for indexing to complete the lineage trace for slicing
+			li = new LineageItem(RightIndex.OPCODE, new LineageItem[]{objLi,
+				new LineageItem(InstructionUtils.concatOperandParts(String.valueOf(ix[0] + 1), scalarCPOSuffix)),
+				new LineageItem(InstructionUtils.concatOperandParts(String.valueOf(ix[1] + 1), scalarCPOSuffix)),
+				new LineageItem(InstructionUtils.concatOperandParts(String.valueOf(ix[2] + 1), scalarCPOSuffix)),
+				new LineageItem(InstructionUtils.concatOperandParts(String.valueOf(ix[3] + 1), scalarCPOSuffix))});
```
phaniarnab, line 245: Please follow how we create lineage items in ParameterizedBuiltinCPInstruction. You simply need to create a CPOperand for each input and pass those to the getLineage function.

ywcb00: Thank you for the advice @phaniarnab. I modified it to create CPOperand for each index and derive the LineageItem from the CPOperands. However, using the getLineage() method to obtain the LineageItems is not that easy, as we would need the ExecutionContext for that.

phaniarnab: You probably need to fix some javadoc warnings @ywcb00. Please have a look at why the tests failed.

PR #1575, opened 2022-04-04
[SYSTEMDS-3343] Boolean Arguments in GridSearch

Baunsgaard:
Boolean arguments in lists evaluate to false behind eval calls.

### DIFF
--- a/src/test/java/org/apache/sysds/test/functions/mlcontext/MLContextTest.java
+++ b/src/test/java/org/apache/sysds/test/functions/mlcontext/MLContextTest.java
@@ -114,21 +114,59 @@ public class MLContextTest extends MLContextTestBase {
 
 	@Test
 	public void testExecuteEvalBuiltinTest() {
-		LOG.debug("MLContextTest - eval builtin test");
-		Script script = dmlFromFile(baseDirectory + File.separator + "eval3-builtin-test.dml");
-		ml.setExplain(true);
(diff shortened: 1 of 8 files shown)


Baunsgaard:
This PR contains some tests that currently incorrectly map boolean arguments to the functions through eval calls.
I hypothesize that it has to do with conversion to scalars somewhere for our boolean arguments.
The evaluation is correct if the boolean arguments are part of the grid, but not if they are set to default values.

Baunsgaard:
@mboehm7 
It seems like your fix to boolean arguments ( a0987e536a2be71d16d64ac64e9873206083e49b ) with defaults did not fix the issue.

Baunsgaard:
Tests are in 
src/test/java/org/apache/sysds/test/functions/mlcontext/MLContextTest.java

- testExecuteEvalGridSearchNoDefault
- testExecuteEvalGridSearchWithDefault
- testExecuteEvalGridSearchWithTwoDefault

mboehm7:
I did not try out this PR here yet, but there was no issue with boolean defaults. My recent changes first introduced that we support defaults in eval calls at all. In the past, gridSearch simply fixed verbose=FALSE but I added tests which now check that gridSearch works correctly for both verbose configurations.

Baunsgaard:
My tests here, might be different since i here call methods that have no default value,
but instead my default value is set inside the lists.

PR #1580, opened 2022-04-10
[SYSTEMDS-3346] Multi-threaded buffer pool evictions

phaniarnab:
This patch implements a simple technique to exploit multithreading
while evicting multiple buffer pool entries. Here we distribute
the files to be evicted among threads. More sophisticated strategies
include using our parallel reader/writes, multithreaded write to an
output stream, however, they have more implementation overhead and
left as future work. This simple strategy is already beneficial for
unified memory as UMM tends to evict multiple objects in each
eviction call while making space for worst-case output sizes.


phaniarnab:
This work needs more testing. I will keep it open for the next release.

PR #1581, opened 2022-04-10
[SYSTEMDS-3347] Flatten the nested loop for parallel pipelines execution

Shafaq-Siddiqi:
- This commit flattens the nested parfor loop in run_with_hyperparam()
    to increase the degress of parallelism by materializing the pipelines and
    resource hyper-parameters together in one frame and then execute them in a single parfor
   instead of parfor over pipelines and then nesting for hyper-parameters
 - This commit also introduce a frame builtin freplicate() that takes two parameters
   one multi-row frame and one single row frame and replicate the single row frame to multi-row.


Shafaq-Siddiqi:
merged in.

PR #1583, opened 2022-04-12
[SYSTEMDS-3349] Base monitoring backend structure to handle REST requests

MKehayov:
This commit implements a simple base MVC framework, with dynamic parsing of REST requests, which will serve as a backend to the monitoring tool for coordinators and workers. Specifically, it starts a server that parses the URI dynamically and reroutes the request to the right controller (provided the controller is registered) for further request processing and returns the constructed response.


Baunsgaard (commented):
LGTM, based on my experience with netty it seems correct,
but it would be nice with some tests, that verify behavior.
I will merge as is, but could you please make a followup PR that adds tests.

PR #1584, opened 2022-04-13
[SYSTEMDS-3303] WIP: NN Builtin: Attention Layer

SteveineiterTU:
This is WIP. 

This implements the attention layer. This allows to train RNNs with long term sequences easier using the attention mechanism. 

A quick reference to what an attention layer is: https://analyticsindiamag.com/a-beginners-guide-to-using-attention-layer-in-neural-networks/

Tasks:
- [ ] Add attention.dml to scripts/nn/layers/ with the implementation of the layer.
- [ ] Add an AttentionExample.dml file to scripts/nn/examples where the attention layer is used with an lstm-layer to demonstrate that an attention model can be trained and used for predictions.


scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,64 @@
...
+
+calculate_scores = function(matrix[double] query, matrix[double] key)
+    return (matrix[double] scores) {
+    /*
+     * dim(query) = AxB
+     * dim(key) = CxB
+     * dim(scores) = AxC
+     */
+    scores = query %*% t(key)
+}
```
SteveineiterTU, line 34: Here we are unsure how many parameters / choices a user should have. 

E.g: 
- masking
- scaling
- different calculation variants (or is dot enough)

scripts/nn/examples/attention_layer.dml
```diff
@@ -0,0 +1,82 @@
...
+# expected = q * k = 1.1 * 1.6 = 1.76
+expected = matrix(1.76, rows=1, cols=1)
+assert(toString(calculated) == toString(expected))
+
+# ============== Example ======================================
+# 1. input eg text data
+# 2. embedding layer
+# 3. lstm layer
+# 4. attention layer
+# 5. dense layer
```
SteveineiterTU, line 55: We are unsure if it is fine to use simple examples as in lines 25-34 or 38-48, or if we should construct something more relevant/practical for example a text classification with an architecture similar as described in lines 51-55?

scripts/nn/layers/attention.dml
Baunsgaard: Start with the simplest thing you can think of, then we can add arguments later that add the different parts.

scripts/nn/examples/attention_layer.dml
Baunsgaard: you choose, personally i would prefer a relevant/practical solution, that we can plug in that actually works nicely :)
But if you find it to complicated you can start with a simpler model, and then make perhaps a boolean argument chose a simple or more complicated model.

Baunsgaard:
@SteveineiterTU  and @A-Postl : Are we closing this PR because of the other one being more finished?

PR #1590, opened 2022-04-19
[MINOR] Fix federated rmempty

OlgaOvcharenko:
This PR fixes tests and one line in federated rmempty instruction for sparse matrices.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
@@ -906,8 +906,8 @@ public class ParameterizedBuiltinFEDInstruction extends ComputationFEDInstructio
 		@Override
 		public FederatedResponse execute(ExecutionContext ec, Data... data) {
 			MatrixBlock mb = ((MatrixObject) data[0]).acquireReadAndRelease();
-			int r = mb.getDenseBlockValues() != null ? mb.getNumRows() : 0;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/sysds/runtime/instructions/fed/ParameterizedBuiltinFEDInstruction.java
```diff
@@ -906,8 +906,8 @@ public GetMatrixCharacteristics(long varID) {
 		@Override
 		public FederatedResponse execute(ExecutionContext ec, Data... data) {
 			MatrixBlock mb = ((MatrixObject) data[0]).acquireReadAndRelease();
-			int r = mb.getDenseBlockValues() != null ? mb.getNumRows() : 0;
-			int c = mb.getDenseBlockValues() != null ? mb.getNumColumns() : 0;
+			int r = mb.getDenseBlockValues() != null || mb.isInSparseFormat() ? mb.getNumRows() : 0;
+			int c = mb.getDenseBlockValues() != null || mb.isInSparseFormat() ? mb.getNumColumns() : 0;
```
Baunsgaard, line 910: Cool, thanks for the fix.
But i think we can improve it to work with any matrix type, since you could use mb.isEmpty()
also, minor thing, in this code you use two calls, and you could make do with one, if you split the code in if else.

```
if(mb.isEmpty())
  return new .... SUCESS, new int[] {0,0});
return new ... SUCESS, new int[] {mb.getNumRows(), mb.getNumColumns()};
```

Then everything works with compressed matrices as well :)

Baunsgaard:
LGTM will merge shortly with the commented change.

PR #1593, opened 2022-04-21
[SYSTEMDS-3355] MatrixBlock size using CSR when allowed

Baunsgaard:
This PR change/fix the size estimation to correctly use CSR size if that is allowed.
I have artificially limited it to only cases with more than one column,
since we have code paths that exploit the fact that we do not have any sparse single columns.

I personally would like this merged in, but i can see that it potentially reduce performance in some cases.


mboehm7:
I'll have a look in the next days - although I would prefer bringing this in after the release, I think with four eyes and the perftest experiments we should find any potential regression.

mboehm7:
LGTM - thanks for the patch @Baunsgaard. This is a good improvement of `examSparsity` while preserving the worst-case characteristic of other memory estimates.

PR #1601, opened 2022-05-01
[SYSTEMDS-3363] Netty Encoding - Buffer Allocation Size

ywcb00:
Hi,
this PR replaces the ObjectEncoders for encoding federated communication messages (i.e., FederatedRequest and FederatedResponse) with a respective FederatedRequestEncoder and FederatedResponseEncoder. These two new classes extent from ObjectEncoder and override the allocateBuffer method in order to allocate the buffer with an initial capacity corresponding to the FederatedRequest/FederatedResponse instead of the default value of 256.
I've tried estimating the entire serialized size of the requests and responses to allocate the exact size needed, but haven't found a good solution for it because of the many cases to consider. Hence, the initial capacity in this PR only results from taking into consideration a static offset (FederatedRequest: 512, FederatedResponse: 312), the size of the CacheBlock parameters, and the length of the lineage trace (only for FederatedRequests).
In my experiments with a matrix of dimensions 600.000x200, it reduces the encoding time from 50 seconds to 1.9 seconds. :tada: 

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
@@ -19,6 +19,7 @@
 
 package org.apache.sysds.runtime.controlprogram.federated;
 
+import java.io.Serializable;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


Baunsgaard:
Nice! 
I will take a look tomorrow.

Baunsgaard:
Merged

Baunsgaard:
I like the change, I am a bit worried for the static offsets but they do not currently seem to fail in any of our cases.

PR #1608, opened 2022-05-08
[SYSTEMDS-3371] Federated monitoring backend communication with worker nodes

MKehayov:
Implements worker registration in the backend, currently with derby as a local database, but the DB solution can be easily changed, just by implementing the IRepository interface. Once the worker is registered, statistics can be requested, and the FedStatsCollectFunction is used, sending FederatedUDF to the worker. To start the monitoring backend process, a solution similar to the starting of worker nodes is implemented with the "-fedMonitor" flag.


daphne-eu:
Thanks for the patch @MKehayov. As just discussed offline, this looks already good - please double check the fasterxml dependency, tabs vs spaces, exception handling, check of existing schema, use of prepared statements, and a more compact style of assertion checking in the tests.

mboehm7:
LGTM - thanks for the patch and the fixes @MKehayov. During the merge I only fixed some minor warnings and formatting.

PR #1609, opened 2022-05-09
[SYSTEMDS-3374] Federated primitive for transferring a local data object to a federated representation

OlgaOvcharenko:
This PR adds a federated primitive for transferring a local data object (matrices and frames) to a federated representation, e.g. `federated(local_matrix=A1, addresses=list($in_X1, $in_X2, $in_X3, $in_X4),
    ranges=list(list(0, 0), list($rows/4, $cols), list($rows/4, 0), list(2*$rows/4, $cols),
    list(2*$rows/4, 0), list(3*$rows/4, $cols), list(3*$rows/4, 0), list($rows, $cols)))`.


mboehm7:
LGTM - thanks for the patch @OlgaOvcharenko. I will merge this in with minor changes as a number of merge conflicts already accumulated. However, there is quite some redundancy in `InitFEDInstruction` - could you please (in a follow-up PR) eliminate this redundancy, construct a federatedMap, and subsequently simply call broadcastSliced as an existing primitive for transferring slices of a local data object to individual federated workers. That way, we're reuse the well-tested and compact runtime primitive without need for changing the worker handler.

PR #1611, opened 2022-05-11
[SYSTEMDS-3185] Federated Response Serialization Caching

ywcb00:
Hi,
This PR introduces the reuse of serialized federated responses. In order to avoid redundant serialization, we put the serialized byte array of federated responses containing a CacheBlock into the lineage cache and mark it with a separate lineage item for serialization derived from the lineage item of the underlying data. The next time the same data is requested, the serialization process for the federated response is skipped and the cached bytes are loaded directly into the output buffer.
The implementation overrides the encode method of netty's ObjectEncoder to check whether we can reuse the serialization prior to the actual encoding, and to obtain and store the serialized bytes after the encoding.

Thanks for review :)


mboehm7:
LGTM - thanks @ywcb00 for the patch. During the merge, I fixed a few unnecessary imports, formatting, and the stringbuilder handling in federated statistics.

PR #1617, opened 2022-05-17
[WIP] Perftest update and cleanup before release

Baunsgaard:


scripts/perftest/runAll.sh
```diff
@@ -26,50 +26,94 @@ then
   exit 1;
 fi
 
-# Optional argument that can be a folder name for where generated data is stored
```
j143, line 29: https://github.com/apache/systemds/blob/f59b906469f8e9fa54870223bb4efdf91c29fdf2/dev/release/do-release.sh#L28

it will be helpful to be able run the scripts from the root folder!

Baunsgaard: I don't understand your code reference, but i think i get your point.
I disagree since the perftest ideally should be run as a self contained test suite on released artifacts.
This currently is possible by changing folder from root, because our shell script for executing systemds defaults to use our
src if you run it from somewhere that looks like the root of our project overwriting any specific use of artifacts.

PR #1620, opened 2022-05-21
generate proto file

j143:
discussion points:
1. the generated java file is not taking the license header
2. Content seems to be a bit different from the older version of generated java file. (is there a way to test it)
3. Is this generation step going to happen every time we run `mvn package`

Solves #1578


Baunsgaard:
overall i think this seems nice.
We can just add that license is not needed in this file.

PR #1622, opened 2022-05-22
[SYSTEMDS-3382] Federated Left Indexing with Scalar

ywcb00:
Hi,
this PR adds support for the federated left indexing instruction with a scalar rhs, including the respective tests.
When running the FederatedLeftIndexTest, it throws an exception because of invalid matrix dimensions (-1, -1) at the first test run. The second test run passes. Somehow this exception is only thrown if the two frame tests (_testLeftIndexFullDenseFrameCP_, _testLeftIndexFullDenseFrameSP_) are executed before the failing scalar test (_testLeftIndexScalarCP_). I tried to put the empty matrix for the output to the federated workers with the correct sliced dimensions instead of [-1, -1], but then we get a "file does not exist on hdfs" exception. Any idea or suspicion why this could be the case?

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
@@ -351,6 +351,11 @@ public class FederationMap {
 		return ret.toArray(new Future[0]);
 	}
 
+	public Future<FederatedResponse>[] execute(long tid, boolean wait, FederatedRange[] fedRange1,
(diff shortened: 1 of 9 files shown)


mboehm7:
LGTM - thanks for the patch @ywcb00 as well as catching and reporting this issue.

After spending way too much time on debugging this issue, I found the reason and fixed it. A simple `FederationUtils.resetFedDataID()` at the beginning of the test mitigated the described problem, which however pointed to a more severe issue. In detail, after running the two other tests, we got an output ID 13 for the federated left indexing. The scalar literal was also 13 which lead to incorrect replacements when preparing the individual instructions sent to the workers. By rearranging the order to input/output replacements, we now generally prevent such problems. Thanks.

PR #1624, opened 2022-05-30
[SYSTEMDS-3383] Extend monitoring backend with worker statistics

MKehayov:
This change extends the already existing federated worker monitoring logic, by having a separate entity for statistics gathering and aggregation. It does this in a periodic way (currently every 3rd second) from every registered worker.
It also extends the CRUD request logic for the worker entity by adding the additional "remove" and "update" functionalities, as well as the same for the coordinator entity.


mboehm7:
LGTM - thanks for the patch @MKehayov. During the merge I reverted the original small worker startup time, resolved the merge conflict, and fixed minor warnings and formatting issues.

PR #1625, opened 2022-06-02
[SYSTEMDS-3303] NN Builtin: Attention Layer

A-Postl:
This PR implements an attention layer using scaled dot-product as the alignment score function,
as well as an example showing the usage of the attention layer (as self-attention) in combination with an LSTM layer on
a text classification problem on real data.


scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+source("nn/layers/softmax.dml") as softmax
+
+
+forward = function(matrix[double] query, matrix[double] key, matrix[double] value, integer K)
```
A-Postl, line 25: Should we specify the key as optional and add a boolean to check if the user wants to use the key or the value?

scripts/nn/examples/AttentionExample.dml
```diff
@@ -0,0 +1,442 @@
...
+
+source("nn/layers/attention.dml") as attention
+source("nn/layers/affine.dml") as affine
+source("nn/layers/lstm.dml") as lstm
+source("nn/layers/relu.dml") as relu
+source("nn/layers/sigmoid.dml") as sigmoid
+source("nn/optim/adam.dml") as adam
+source("nn/layers/log_loss.dml") as log_loss
+
+
```
A-Postl, line 38: Is it necessary to write a script to fetch the data?

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+    key_norm_n = matrix(key_norm[n],rows=K, cols=M)
+    value_n = matrix(value[n], rows=K, cols=L)
+    scores = query_n %*% t(key_norm_n)
+    #column wise softmax
+    probs = t(softmax::forward(t(scores)))
+    attention_n = probs %*% value_n
+    attention[n] = matrix(attention_n, rows=1, cols=K*L)
+  }
+}
+
```
A-Postl, line 56: Should we add more alignment score variants?
In this [guide](https://analyticsindiamag.com/a-beginners-guide-to-using-attention-layer-in-neural-networks/) there are 6 variants listed.

mboehm7:
thanks for the revised patch - @sebwrede and/or @Baunsgaard could you please have a look as you helped reviewing #1584.

j143:
I will review this tonight, @A-Postl .

scripts/nn/examples/AttentionExample.dml
Baunsgaard: it would be nice to have yes,
simply write something like `wget https://super.nice.webpage/dataset.csv` 
inside a download_attentionExample.sh script.
and make sure that you add a git ignore for it.

If you cannot make such a link then we can put the data set on our Apache webpage and download from there.

scripts/nn/examples/AttentionExample.dml
```diff
@@ -0,0 +1,442 @@
...
+# 1 get data
+data_loc = "scripts/nn/examples/data/"
+tableschema = "string,int"
+N=32000 # Samples of whole dataset
+n=8000  # Samples to use for training
+max_length = 32 # maximum sequence length
+epochs = 30
+batch_size = n/200
```
Baunsgaard, line 46: this is a questionable way of selecting batch size.
Should you not simply set it to 40?

scripts/nn/examples/AttentionExample.dml
```diff
@@ -0,0 +1,442 @@
...
+data = read(data_loc + "clickbait_data.csv", format="csv", header=TRUE, sep=",", data_type="frame", schema=tableschema, cols=2, rows=N)
+
+
+[x_train, y_train, vocab_size] = preprocess(data, max_length, N)
+
+x_train = x_train[1:n]
+y_train = y_train[1:n]
+
+# train network
+# TODO fix: get error when batch size is not always equal
```
Baunsgaard, line 58: hopefully not a problem anymore.

scripts/nn/layers/attention.dml
Baunsgaard: Yes key should be optional and i suggest set as third parameter.

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+forward = function(matrix[double] query, matrix[double] key, matrix[double] value, integer K)
+    return (matrix[double] attention) {
+  /*
+   * Computes the forward pass for the attention layer.
+   *
+   * Inputs:
+   * - query: Input querys of shape (N,K*M).
+   * - key: Key(s) for value(s) of shape (N,K*M).
+   * - value: Value(s) for key(s) of shape (N,K*L).
```
Baunsgaard, line 33: These dimensions does not align with TensorFlow's specification
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Attention

Key and value is same dimensions

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+  L = ncol(value) / K
+  norm = 1/M^0.5
+  key_norm = key * norm
+  attention = matrix(0, rows=N, cols=K*L)
+  for (n in 1:N)
+  {
+    query_n = matrix(query[n], rows=K, cols=M)
+    key_norm_n = matrix(key_norm[n],rows=K, cols=M)
+    value_n = matrix(value[n], rows=K, cols=L)
+    scores = query_n %*% t(key_norm_n)
```
Baunsgaard, line 49: This is inefficient, you slice out a single row in each of the inputs and then matrix multiply.
instead you should simply multiply the entire input once to get scores.
with this change you can remove the entire for loop and process on matrices instead.

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+   * - K: Sequence length / number of timesteps.
+   * Outputs:
+   * - attention: Attention on value(s) for given query(s), of shape (N,K*L).
+   */
+  N = nrow(key)
+  M = ncol(query) / K
+  L = ncol(value) / K
+  norm = 1/M^0.5
+  key_norm = key * norm
+  attention = matrix(0, rows=N, cols=K*L)
```
Baunsgaard, line 43: I would like it if this attention matrix is a part of the arguments.
Then in the beginning of the attention step simply assign it fully to zero when calling this method.

The interface then become: query, value, key, K, attMatrix

With this we can avoid allocating the matrix every loop.

scripts/nn/layers/attention.dml
Baunsgaard: we could, but for now softmax is fine, if we want more then we add an optional argument specifying the the type wanted.

scripts/nn/layers/attention.dml
Baunsgaard: we can make a check inside the function to say, if key  is empty matrix (that you set as default argument) then set key to be equal to value.

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+
+    dvalue_n = t(probs) %*% dattention_n
+    dprobs = t(value_n %*% t(dattention_n))
+    dscore = t(softmax::backward(t(dprobs), t(scores)))
+    dquery_n = dscore %*% key_norm_n
+    dkey_n = t(dscore) %*% query_n * norm
+    dquery[n] = matrix(dquery_n, rows=1, cols=K*M)
+    dkey[n] = matrix(dkey_n, rows=1, cols=K*M)
+    dvalue[n] = matrix(dvalue_n, rows=1, cols=K*L)
+  }
```
Baunsgaard, line 107: similar to forward i think we can avoid this for loop.

scripts/nn/layers/attention.dml
```diff
@@ -0,0 +1,108 @@
...
+  N = nrow(key)
+  M = ncol(query) / K
+  L = ncol(value) / K
+
+  norm = 1 / M^0.5
+  key_norm = key * norm
+
+  dquery = matrix(0, rows=N, cols=K*M)
+  dkey = matrix(0, rows=N, cols=K*M)
+  dvalue = matrix(0, rows=N, cols=K*L)
```
Baunsgaard, line 87: Similar to forward we can add these as arguments to avoid having to re allocate them, and on call simply set all values to zero.

src/test/scripts/applications/nn/grad_check.dml
```diff
@@ -2537,3 +2538,78 @@ elu = function() {
      }
    }
 }
+
+attention = function() {
```
Baunsgaard, line 2542: I do not understand what this method is testing, maybe a comment would help.

Baunsgaard (commented):
Overall i like this PR,
There are some things that needs to be addressed:

- For performance (removing for loop)
- For consistency add the arguments for the attention matrix and gradients for forward and backwards call to make the functions like the other operations.
- The example does not verify if the attention behaves correctly in case of different dimensions for query and value, and it would be nice if we could have this. 
- I do not understand the test, and how that verifies if the method works but i think some comments in the tests would help solve that.

best regards
Sebastian

scripts/nn/examples/AttentionExample.dml
```diff
@@ -0,0 +1,442 @@
...
+  [out1, c_out, cache_out_out, cache_c_out, cache_ifog_out]=
+    lstm::forward(emb, W_0, b_0, max_sequence_length, embedding_size, TRUE, out0, c0)
+  out2 = attention::forward(out1, out1, out1, max_sequence_length)
+  out3 = affine::forward(out2, W_1, b_1)
+  out4 = relu::forward(out3)
+  out5 = affine::forward(out4, W_2, b_2)
+  y_hat = sigmoid::forward(out5)
+}
+
+fetch_embeddings = function(matrix[double] indexes, matrix[double] W_E,
```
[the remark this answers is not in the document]
Baunsgaard: not sure either, but we could limit the example to not include embedding, to make the project simpler :) since it goes out of scope of attention models anyway.

src/test/scripts/applications/nn/grad_check.dml
A-Postl: This test is to verify the gradient of the backward pass numerically.
The comments are just copied and modified from the other test cases in this file.

A-Postl:
@Baunsgaard Thanks for the feedback. We'll address the feedback shortly.

scripts/nn/layers/attention.dml
A-Postl: We're having a little trouble with this... 
Since each line is a vectorized matrix, if we reshape and matrix multiply we get a lot of unused values.
The only values we need are then as blocks on the diagonal.

Baunsgaard: we have a diagonal command to extract the diagonal from a matrix.

But if it is the case that you only need the diagonal,
then there should still be some smarter way of looking at it,
to still avoid this slicing of the matrix,
but it can also be that i was wrong in my analysis of the script.

scripts/nn/examples/AttentionExample.dml
j143: It is better to upload files to our website: https://github.com/apache/systemds-website#how-to-add-new-dataset

Then download it from there. Ideally, we are not supposed to keep datasets in the project repo.

SteveineiterTU: Please review https://github.com/apache/systemds-website/pull/110.

A-Postl:
@Baunsgaard  i think we addressed all issues, with the exception of the for loop in the layer.

scripts/nn/layers/attention.dml
Baunsgaard: okay, looking closer, this is hard because of us treating the entire input as one matrix. If we had it slightly different we could have solved it with element wise multiplications and rowSums, but this is not feasible currently i will think about it a bit more since the current way (and the only way i can tell currently) still is inefficient.

src/test/scripts/applications/nn/grad_check.dml
Baunsgaard: The intention of me commenting on it was, that you add a comment in code. When you add it here only i get the help from you.

Baunsgaard (commented):
Thanks for addressing my comments,
I will take it from here.

Baunsgaard:
Closing for merging.

PR #1627, opened 2022-06-03
[SYSTEMDS-3185] Multi-tenant Federated Backend - Documentation and Cleanup

ywcb00:
Hi,
This PR
- adds a paragraph about multi-tenant federated learning in SystemDS and the respective limitations of a multi-tenant federated infrastructure to the python federated documentation.
- changes the local variable map back to a basic HashMap instead of a ConcurrentHashMap, since we already handle the accesses in a safe way.
- fixes the FederatedReuseSlicesTest by considering the lineage cache writes coming from response-serialization-caching.

Thanks for review :)

### DIFF
--- a/docs/api/python/sources/guide/federated.rst.txt
+++ b/docs/api/python/sources/guide/federated.rst.txt
@@ -99,3 +99,27 @@ The print should look like
   that you have:
   
   a csv file, mtd file, and SystemDS Environment is set correctly.
+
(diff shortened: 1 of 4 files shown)


mboehm7:
LGTM - thanks for the patch @ywcb00. These docs and cleanups are very nice, and I only added a small sentence on tenant isolation to make it more explicit.

PR #1628, opened 2022-06-07
[SYSTEMDS-3386] Refactor replacement of CP instructions

kev-inn:
This reworks our runtime replacement of CP instructions with FED versions when appropriate. 
SP replacement will be done in another PR.

This implementation was my second approach, the first one consisted of static methods in the FED instructions. I much prefer this one as it is quite clear which instructions are replaced, which constraints the CP imposes and we don't have to lead every check with an `instanceof` check. The downside of this approach is that it can get a bit confusing with inheritance, but it is---and will probably stay---manageable.

Note that not everything is replaced yet, I will work on it while the general approach can already be reviewed.

During the testing I found myself requiring to check if actual FED instructions are executed, is there a way to ensure this in the test framework, or is this already ensured in some way via privacy? Tests checking the heavy hitters do it implicitly.

NOTE: I deleted a fix in `QuantilePickCPInstruction` (bf19244c909ae8354bc8b3da0985eaf448aa5f0c), as the common layout is no longer necessary and null is cleaner. This is intentional and the fix is no longer necessary.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/cp/QuantilePickCPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/cp/QuantilePickCPInstruction.java
@@ -65,7 +65,7 @@ public class QuantilePickCPInstruction extends BinaryCPInstruction {
 			CPOperand out = new CPOperand(parts[2]);
 			OperationTypes ptype = OperationTypes.valueOf(parts[3]);
 			boolean inmem = Boolean.parseBoolean(parts[4]);
-			return new QuantilePickCPInstruction(null, in1, new CPOperand(), out, ptype, inmem, opcode, str);
+			return new QuantilePickCPInstruction(null, in1, out, ptype, inmem, opcode, str);
 		}
 		else if( parts.length == 6 ) {
 			CPOperand in1 = new CPOperand(parts[1]);
@@ -126,5 +126,13 @@ public class QuantilePickCPInstruction extends BinaryCPInstruction {
 			default:
 				throw new DMLRuntimeException("Unsupported qpick operation type: "+_type);
 		}
+	}
+
+	public OperationTypes getOperationType() {
+		return _type;
+	}
+
+	public boolean isInMem() {
+		return _inmem;
 	}
 }
--- a/src/main/java/org/apache/sysds/runtime/instructions/cp/QuantileSortCPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/cp/QuantileSortCPInstruction.java
@@ -119,4 +119,8 @@ public class QuantileSortCPInstruction extends UnaryCPInstruction {
 		//set and release output
 		ec.setMatrixOutput(output.getName(), resultBlock);
 	}
+
+	public int getNumThreads() {
+		return _numThreads;
+	}
 }
--- a/src/main/java/org/apache/sysds/runtime/instructions/cp/UnaryScalarCPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/cp/UnaryScalarCPInstruction.java
@@ -26,10 +26,10 @@ import org.apache.sysds.runtime.controlprogram.context.ExecutionContext;
 import org.apache.sysds.runtime.matrix.operators.Operator;
 import org.apache.sysds.runtime.matrix.operators.UnaryOperator;
 
-public class UnaryScalarCPInstruction extends UnaryMatrixCPInstruction {
+public class UnaryScalarCPInstruction extends UnaryCPInstruction {
 
 	protected UnaryScalarCPInstruction(Operator op, CPOperand in, CPOperand out, String opcode, String instr) {
-		super(op, in, out, opcode, instr);
+		super(CPType.Unary, op, in, out, opcode, instr);
 	}
 
 	@Override 
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
@@ -36,7 +36,9 @@ import org.apache.sysds.runtime.controlprogram.federated.FederationMap;
 import org.apache.sysds.runtime.controlprogram.federated.FederationUtils;
 import org.apache.sysds.runtime.controlprogram.federated.MatrixLineagePair;
 import org.apache.sysds.runtime.instructions.InstructionUtils;
+import org.apache.sysds.runtime.instructions.cp.AggregateBinaryCPInstruction;
 import org.apache.sysds.runtime.instructions.cp.CPOperand;
+import org.apache.sysds.runtime.instructions.spark.AggregateBinarySPInstruction;
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 import org.apache.sysds.runtime.matrix.operators.Operator;
 
@@ -52,7 +54,17 @@ public class AggregateBinaryFEDInstruction extends BinaryFEDInstruction {
 		String opcode, String istr, FederatedOutput fedOut) {
 		super(FEDType.AggregateBinary, op, in1, in2, out, opcode, istr, fedOut);
 	}
-	
+
+	public static AggregateBinaryFEDInstruction parseInstruction(AggregateBinaryCPInstruction instr) {
+		return new AggregateBinaryFEDInstruction(instr.getOperator(), instr.input1, instr.input2, instr.output,
+			instr.getOpcode(), instr.getInstructionString(), FederatedOutput.NONE);
+	}
+
+	public static AggregateBinaryFEDInstruction parseInstruction(AggregateBinarySPInstruction instr) {
+		return new AggregateBinaryFEDInstruction(instr.getOperator(), instr.input1, instr.input2, instr.output,
+				instr.getOpcode(), instr.getInstructionString(), FederatedOutput.NONE);
+	}
+
 	public static AggregateBinaryFEDInstruction parseInstruction(String str) {
 		String[] parts = InstructionUtils.getInstructionPartsWithValueType(str);
 		String opcode = parts[0];
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/MMChainFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/MMChainFEDInstruction.java
@@ -31,6 +31,7 @@ import org.apache.sysds.runtime.controlprogram.federated.FederatedRequest.Reques
 import org.apache.sysds.runtime.controlprogram.federated.MatrixLineagePair;
 import org.apache.sysds.runtime.instructions.InstructionUtils;
 import org.apache.sysds.runtime.instructions.cp.CPOperand;
+import org.apache.sysds.runtime.instructions.cp.MMChainCPInstruction;
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 
 import java.util.concurrent.Future;
@@ -47,6 +48,11 @@ public class MMChainFEDInstruction extends UnaryFEDInstruction {
 
 	public ChainType getMMChainType() {
 		return _type;
+	}
+
+	public static MMChainFEDInstruction parseInstruction(MMChainCPInstruction instr) {
+		return new MMChainFEDInstruction(instr.input1, instr.input2, instr.input3, instr.output, instr.getMMChainType(),
+			instr.getNumThreads(), instr.getOpcode(), instr.getInstructionString());
 	}
 
 	public static MMChainFEDInstruction parseInstruction ( String str ) {
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/TsmmFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/TsmmFEDInstruction.java
@@ -33,6 +33,7 @@ import org.apache.sysds.runtime.controlprogram.federated.FederationMap;
 import org.apache.sysds.runtime.controlprogram.federated.FederationUtils;
 import org.apache.sysds.runtime.instructions.InstructionUtils;
 import org.apache.sysds.runtime.instructions.cp.CPOperand;
+import org.apache.sysds.runtime.instructions.cp.MMTSJCPInstruction;
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 
 public class TsmmFEDInstruction extends BinaryFEDInstruction {
@@ -49,7 +50,12 @@ public class TsmmFEDInstruction extends BinaryFEDInstruction {
 	public TsmmFEDInstruction(CPOperand in, CPOperand out, MMTSJType type, int k, String opcode, String istr) {
 		this(in, out, type, k, opcode, istr, FederatedOutput.NONE);
 	}
-	
+
+	public static TsmmFEDInstruction parseInstruction(MMTSJCPInstruction instr) {
+		return new TsmmFEDInstruction(instr.input1, instr.getOutput(), instr.getMMTSJType(), instr.getNumThreads(),
+			instr.getOpcode(), instr.getInstructionString());
+	}
+
 	public static TsmmFEDInstruction parseInstruction(String str) {
 		String[] parts = InstructionUtils.getInstructionPartsWithValueType(str);
 		String opcode = parts[0];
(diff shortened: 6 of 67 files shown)


src/main/java/org/apache/sysds/runtime/instructions/cp/CovarianceCPInstruction.java
```diff
@@ -82,4 +85,28 @@ public void processInstruction(ExecutionContext ec)
...
+				...
+			else
+				...
+		
+		This allowed `input1(Broadcast & Row) & input2(!Broadcast & ...)`, I am not sure if that is intentional.
+		 */
+		if((input1.isMatrix() && ec.getMatrixObject(input1).isFederated(FTypes.FType.ROW) &&
+			ec.getMatrixObject(input1).isFederatedExcept(FTypes.FType.BROADCAST)) ||
+			(input2.isMatrix() && ec.getMatrixObject(input2).isFederated(FTypes.FType.ROW) &&
+				ec.getMatrixObject(input2).isFederatedExcept(FTypes.FType.BROADCAST)))
```
kev-inn, line 108: I am not sure about this, no testcase covers the case explained at the bottom of the comment I think.

src/main/java/org/apache/sysds/runtime/instructions/cp/QuantilePickCPInstruction.java
```diff
@@ -65,7 +68,7 @@ else if( parts.length == 5 ) {
 			CPOperand out = new CPOperand(parts[2]);
 			OperationTypes ptype = OperationTypes.valueOf(parts[3]);
 			boolean inmem = Boolean.parseBoolean(parts[4]);
-			return new QuantilePickCPInstruction(null, in1, new CPOperand(), out, ptype, inmem, opcode, str);
+			return new QuantilePickCPInstruction(null, in1, out, ptype, inmem, opcode, str);
```
kev-inn, line 71: Our new approach allows us to remove this fix/hack.

src/main/java/org/apache/sysds/runtime/instructions/cp/QuantilePickCPInstruction.java
```diff
@@ -127,4 +130,22 @@ public void processInstruction(ExecutionContext ec) {
...
+
+	public boolean isInMem() {
+		return _inmem;
+	}
+
+	@Override
+	protected FEDInstruction tryReplaceWithFederated(ExecutionContext ec) {
+		// Should we federate if the quantiles are given as a federated matrix? (input2)
+		// This might be useful for privacy and for performance just in extreme corner cases.
+		// The current implementation does not support it at the moment.
```
kev-inn, line 146: This comment could be removed as we anyway only replace instructions where equivalent FED implementations exist, but it is worth discussing/remembering what to do about this case.

src/main/java/org/apache/sysds/runtime/instructions/cp/QuantileSortCPInstruction.java
```diff
@@ -119,4 +123,23 @@ public void processInstruction(ExecutionContext ec) {
...
+			return null;
+
+		MatrixObject mo1 = ec.getMatrixObject(input1);
+		if(!mo1.isFederatedExcept(FType.BROADCAST))
+			return null;
+
+		if(mo1.isFederated(FType.ROW) ||
+			(mo1.getFedMapping().getFederatedRanges().length == 1 && mo1.isFederated(FType.COL)))
+			return new QuantileSortFEDInstruction(this);
+		return null;
```
kev-inn, line 143: As you can notice in all implementations I opted to use early return checks to reduce nesting. The last if on the other hand checks if replacement is appropriate. This is inconsistent to the checks before, which checked for failure, but is much more readable and therefore maintainable than confusing inverted conditions in my experience.

src/main/java/org/apache/sysds/runtime/instructions/cp/UnaryScalarCPInstruction.java
```diff
@@ -26,10 +26,10 @@
 import org.apache.sysds.runtime.matrix.operators.Operator;
 import org.apache.sysds.runtime.matrix.operators.UnaryOperator;
 
-public class UnaryScalarCPInstruction extends UnaryMatrixCPInstruction {
+public class UnaryScalarCPInstruction extends UnaryCPInstruction {
```
kev-inn, line 29: This is a problem due to inheritance. We replace `UnaryMatrixCPInstruction`s, but don't replace `UnaryScalarCPInstruction`s, with FED instructions. I think this inheritance made no sense and probably was a bug.
If it *wasn't* we can also overload the `tryReplaceWithFederated()` here.

src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
```diff
@@ -53,6 +56,12 @@ public AggregateBinaryFEDInstruction(Operator op, CPOperand in1, CPOperand in2,
 		super(FEDType.AggregateBinary, op, in1, in2, out, opcode, istr, fedOut);
 	}
 	
+	public AggregateBinaryFEDInstruction(AggregateBinaryCPInstruction instruction) {
+		// TODO @KEVIN: should we update the instruction string?
+		this(instruction.getOperator(), instruction.input1, instruction.input2, instruction.output,
+			instruction.getOpcode(), instruction.getInstructionString(), FederatedOutput.NONE);
```
kev-inn, line 62: I always keep the instruction string without changes, but we might want to replace the leading `CP` with `FED` and add the type of `FederatedOutput` (always `NONE`, as it is decided at runtime).

src/main/java/org/apache/sysds/runtime/instructions/fed/AggregateBinaryFEDInstruction.java
```diff
@@ -69,6 +78,22 @@ public static AggregateBinaryFEDInstruction parseInstruction(String str) {
...
+			return null;
+		}
+		MatrixObject mo1 = ec.getMatrixObject(instruction.input1);
+		MatrixObject mo2 = ec.getMatrixObject(instruction.input2);
+		if (!(mo1.isFederated(FType.ROW) || mo2.isFederated(FType.ROW) || mo1.isFederated(FType.COL))) {
+			return null;
+		}
+		return new AggregateBinaryFEDInstruction(instruction);
+	}
+
```
kev-inn, line 96: This is outdated from my first attempt, will be removed.

src/main/java/org/apache/sysds/runtime/instructions/fed/MMChainFEDInstruction.java
```diff
@@ -43,6 +44,12 @@ public MMChainFEDInstruction(CPOperand in1, CPOperand in2, CPOperand in3,
 		_type = type;
 	}
 	
+	public MMChainFEDInstruction(MMChainCPInstruction instruction) {
+		this(instruction.input1, instruction.input2, instruction.input3, instruction.output,
+			instruction.getMMChainType(), -1 /* ignore numThreads */, instruction.getOpcode(),
+			instruction.getInstructionString());
```
kev-inn, line 50: I will update this to use the same `numThreads`, but I don't think this has any effect for the federated implementation.

src/main/java/org/apache/sysds/runtime/instructions/fed/TsmmFEDInstruction.java
```diff
@@ -50,6 +52,11 @@ public TsmmFEDInstruction(CPOperand in, CPOperand out, MMTSJType type, int k, St
 		this(in, out, type, k, opcode, istr, FederatedOutput.NONE);
 	}
 	
+	public TsmmFEDInstruction(MMTSJCPInstruction instruction) {
+		this(instruction.getInputs()[0], instruction.getOutput(), instruction.getMMTSJType(), -1 /*ignore numThreads*/,
```
kev-inn, line 56: Again, `numThreads` without any impact I think.

kev-inn (commented):
Some comments to explain myself :)

kev-inn:
As a side note, when I ran all the tests the following testcases fail, but work when executed in isolation:
- `FederatedWorkerIntegrationCRUDTest.testWorkerAddedForMonitoring`
- `FederatedColAggregateTest.testTernaryColSumDenseMatrix`
- `FederatedCtableTest.federatedCtableSinglenode` (configuration [3])
- `FederatedFullAggregateTest.testTernarySumDenseMatrixCP` (configuration [0] and [1])
- `FederatedWeightedCrossEntropy` (configuration [0] and [1])
  - `federatedWeightedCrossEntropySpark`
  - `federatedWeightedCrossEntropySparkEpsilon`
  - `federatedWeightedCrossEntropySparkNode`
  - `federatedWeightedCrossEntropySparkNodeEpsilon`
- `FederatedWeightedDivMatrixMultTest` all or almost all
- `FederatedWeightedSigmoidTest` all
- `FederatedWeightedSquaredLossTest` all
- `FederatedWeightedUnaryMatrixMultTest` all (both configurations)

I remember that this is a known problem.

`FederatedReuseSlicesTest.testElementWisePlusCP` fails, because `checkForReuses()` expects `"LinCache writes (Mem/FS/Del):	8/"`, but the actual value is `"LinCache writes (Mem/FS/Del):	12/"`. I am not sure what the reason could be here.

Baunsgaard:
> As a side note, when I ran all the tests the following testcases fail, but work when executed in isolation:
> 
>     * `FederatedWorkerIntegrationCRUDTest.testWorkerAddedForMonitoring`
> 
>     * `FederatedColAggregateTest.testTernaryColSumDenseMatrix`
> 
>     * `FederatedCtableTest.federatedCtableSinglenode` (configuration [3])
> 
>     * `FederatedFullAggregateTest.testTernarySumDenseMatrixCP` (configuration [0] and [1])
> 
>     * `FederatedWeightedCrossEntropy` (configuration [0] and [1])
>       
>       * `federatedWeightedCrossEntropySpark`
>       * `federatedWeightedCrossEntropySparkEpsilon`
>       * `federatedWeightedCrossEntropySparkNode`
>       * `federatedWeightedCrossEntropySparkNodeEpsilon`
> 
>     * `FederatedWeightedDivMatrixMultTest` all or almost all
> 
>     * `FederatedWeightedSigmoidTest` all
> 
>     * `FederatedWeightedSquaredLossTest` all
> 
>     * `FederatedWeightedUnaryMatrixMultTest` all (both configurations)
> 
> 
> I remember that this is a known problem. Edit: I was behind a few commits, so this information might be outdated.

Yes this is known.

You can run your tests such that they repeat if failing, there you will detect these "flaky" tests.
If you use IntelliJ, you can also add a new execution mode where you repeat failing tests

mvn clean compile test -Dmaven.test.skip=false -Drerun.failingtests.count=1 -D automatedtestbase.outputbuffering=true -Dtest=org.apache.sysds.test.$@

Baunsgaard:
> This implementation was my second approach, the first one consisted of static methods in the FED instructions. I much prefer this one as it is quite clear which instructions are replaced, which constraints the CP imposes and we don't have to lead every check with an instanceof check. The downside of this approach is that it can get a bit confusing with inheritance, but it is---and will probably stay---manageable.

I am not sure if this is the best location to edit.
Would it not make more sense to change it when we construct the CP vs SP vs GPU instructions from the string versions of instructions themselves?
This way if we for instance in the future add FPGA instructions it naturally also support changing these into federated instructions, and we do not have to maintain both all instances for each federated instruction which would lead to excessive testing requirements where we need to hit each instruction as either cases.

kev-inn:
You have a point there, I also realized this recently when trying to also replace SP Instructions without duplicating code. I will weigh the options, if I come up with more, but will most likely opt to replace it already during parsing.

kev-inn:
Sure, that is the number one priority of this rework, that is why I also would like to remove the dependency for the string representations of those instructions to be similar/matching.
In the future we want to send only CP instructions to the workers and locally choose CP, SP or GPU (maybe even FED?). Keeping this in mind, we could add conversions for all kinds of instructions to CP (and require constructors like the ones I added for FED) or add interfaces with getters for specialized instructions like `BinaryAggregationInstruction`.
The approach, which should be the nicest to work with, would be a complete rework of the hierarchy of instructions, adding the type of execution only at the leaf specializations. Instead of `ComputationCPInstruction` and `ComputationSPInstruction` we would have `ComputationInstruction. This would require a very intrusive change of a lot of core implementations and seems infeasable.

kev-inn:
After a lot of consideration I propose the following:
1. Switch to the first implementation approach. The refactor makes the code more maintainable, because of less indentation and implementation in the corresponding `FEDInstruction`. The static methods perform checks for both `CPInstruction` and `SPInstruction`, which will necessarily have large overlaps. A constructor with the `SPInstruction` as input (like the ones I already added with `CPInstruction`) will be added where necessary.
2. Add a method to most `SPInstruction`s to convert it to a `CPInstruction`. This will be used to allow the federated worker to choose which execution type he wants to use. Most `SPInstruction` can be converted to `CPInstruction` (and vice versa) by changing the execution type in the instruction string before parsing, those we could simply tag with an interface. This would reduce code quite a bit, but is harder to maintain. Opinions are appreciated.
3. Some `SPInstruction`s like `ReblockSPInstruction` don't have an equivalent `CPInstruction`, those will be sent to the worker as is. A problem arises when the federated worker opts to use the `CPInstruction` for inputs instead, this would require us to ignore the `SPInstruction`. Not sure if checking for RDD handles on the inputs is enough(?).

mboehm7:
Thanks for getting started on this issue @kev-inn. Unfortunately, I think the PR goes a bit in the wrong direction. We first want to do a basic cleanup of the federated instructions, without modifying the CP or Spark instructions. The replace in `FEDInstructionUtils` should stay mostly as it is, and all federated instructions should have overloaded `parseInstruction` methods to consume either CP or Spark instructions (objects) directly and construct the FED instruction like [1] without replicating the string parsing logic. For the parseInstruction with strings, we can call the parse on the respective CP instruction (for consistency) and use again the above methods. 

In a second step, we would remove the replicated instructions like `MMFEDInstruction` (redundant to `AggregateBinaryFEDInstruction`) and let all instruction send CP instruction strings to the federated workers. 

In a third step, we then extend the federated workers to (under certain conditions) create a hop that represents the CP instruction and generate lops and instructions as needed (e.g., CP, Spark, GPU).

Sorry, if I was not clear during our offline discussions.

[1] https://github.com/apache/systemds/blob/main/src/main/java/org/apache/sysds/runtime/instructions/fed/CentralMomentFEDInstruction.java#L77

Baunsgaard:
> After a lot of consideration I propose the following:
> 
>     1. Switch...

>     2. Add a method to most `SPInstruction`s ... 

>     3. Some `SPInstruction`s like `ReblockSPInstruction` ...

I suggest that you make the methods inside CP and SP instructions part of  a interface that allows any instruction to change to FED, and then make the footprint of these functions as small as possible inside each of the CP and SP instructions. All of these instructions should call into the fitting FEDInstructions class that then handle the logic of converting the instructions to FEDInstructions though static constructors that can return null.
With this we simply try to call this interface function on the instructions to change them into FEDInstructions when they can before execution. In the static construction it is easy to change the SP instructions (or GPU etc) to CP if possible, otherwise the static constructor would construct SP type.

This design addresses all 3 points with (i think) the least amount of code.

kev-inn:
> Thanks for getting started on this issue @kev-inn. Unfortunately, I think the PR goes a bit in the wrong direction. We first want to do a basic cleanup of the federated instructions, without modifying the CP or Spark instructions. The replace in `FEDInstructionUtils` should stay mostly as it is, and all federated instructions should have overloaded `parseInstruction` methods to consume either CP or Spark instructions (objects) directly and construct the FED instruction like [1] without replicating the string parsing logic. For the parseInstruction with strings, we can call the parse on the respective CP instruction (for consistency) and use again the above methods.
> 
> In a second step, we would remove the replicated instructions like `MMFEDInstruction` (redundant to `AggregateBinaryFEDInstruction`) and let all instruction send CP instruction strings to the federated workers.
> 
> In a third step, we then extend the federated workers to (under certain conditions) create a hop that represents the CP instruction and generate lops and instructions as needed (e.g., CP, Spark, GPU).
> 
> Sorry, if I was not clear during our offline discussions.
> 
> [1] https://github.com/apache/systemds/blob/main/src/main/java/org/apache/sysds/runtime/instructions/fed/CentralMomentFEDInstruction.java#L77

No worries, as @Baunsgaard stated, I think the majority of this PR is still a nice addition and it fits your first step nicely. I would like to have 3 methods:
1. Interface method named `toFedInstruction` for CP and SP Instructions
2. Static "constructor" method for `FEDInstructions` `tryReplace()`
3. FED constructors taking `CPInstruction` and `SPInstruction`. I propose a constructor instead of `parseInstruction()`, because it always returns an instance (the checks happen in `tryReplace()`) and `parseInstruction()` can be kept for actual parsing logic.

The second step I might combine with this PR, but the third one will definitely be seperate.

kev-inn:
I am sorry that this PR got so large. I believe it is better to do it in one go, but I can also break it down so the review is faster.

I recommend to open the old `FEDInstructionUtils.java` file on one screen and on the other screen go through all added `toFedInstruction()` methods. This will allows to easily match the old conditions with the new, more compact and (hopefully) easier to read conditions.
There aren't many changes to the logic, except that I changed (or added, see `AppendSPInstruction`) some super-classes.

kev-inn:
Not ready for review, found some problems with the spark instructions.

PR #1634, opened 2022-06-09
[SYSTEMDS-2830] Functional Compression

wedenigt:
This commit introduces a new column group called `ColGroupLinearFunctional`. The column group represents each column as a linear function (i.e., intercept & slope). Each linear function is computed such that it minimizes the quadratic distance to the data points in the respective column.

`ColGroupLinearFunctional` implements
- Efficient transposed self matrix multiplication `tsmm` (which runs in `O(numCols^2)`)
- Multiplication with other `ColGroups` (which is especially efficient if another transposed `ColGroupLinearFunctional` is multiplied from the left) and with `MatrixBlock` (from the right)
- `max`, `min`, `sum`, `sumSq`, `colProduct`, `colSums`, `colSumsSq`, `rowProduct`, `rowSums`, `decompressToDenseBlock`, `decompressToSparseBlock`

The implementations of `ColGroupLinearFunctional` and of the line-fitting module `LinearRegression` are accompanied by test cases.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
@@ -42,6 +42,8 @@ public class ColGroupLinearFunctional extends AColGroupCompressed {
 
 	protected double[][] _coefficents;
 
+	protected int _numRows;
+
 	/** Constructor for serialization */
 	protected ColGroupLinearFunctional() {
 		super();
@@ -53,30 +55,29 @@ public class ColGroupLinearFunctional extends AColGroupCompressed {
 	 * @param colIndices  The Colum indexes for the column group.
 	 * @param coefficents The dictionary containing one tuple for the entire compression.
 	 */
-	private ColGroupLinearFunctional(int[] colIndices, double[][] coefficents) {
+	private ColGroupLinearFunctional(int[] colIndices, double[][] coefficents, int numRows) {
 		super(colIndices);
 		this._coefficents = coefficents;
+		this._numRows = numRows;
 	}
 
 	/**
 	 * Generate a constant column group.
 	 *
-	 * @param colIndices   The specific column indexes that is contained in this constant group.
-	 * @param coefficents The coefficents vector
-	 * @return A Constant column group.
+	 * @param colIndices   The specific column indexes that is contained in this column group.
+	 * @param coefficents  The coefficents vector
+	 * @return A LinearFunctional column group.
 	 */
-	public static AColGroup create(int[] colIndices, double[][] coefficents) {
+	public static AColGroup create(int[] colIndices, double[][] coefficents, int numRows) {
 		if(coefficents.length != 2 || colIndices.length != coefficents[0].length)
 			throw new DMLCompressionException("Invalid size of values compared to columns");
-		return new ColGroupLinearFunctional(colIndices, coefficents);
+		return new ColGroupLinearFunctional(colIndices, coefficents, numRows);
 	}
 
 
 	@Override
 	protected void computeRowMxx(double[] c, Builtin builtin, int rl, int ru, double[] preAgg) {
-		double v = preAgg[0];
-		for(int i = rl; i < ru; i++)
-			c[i] = builtin.execute(c[i], v);
+		throw new NotImplementedException();
 	}
 
 	@Override
@@ -87,6 +88,42 @@ public class ColGroupLinearFunctional extends AColGroupCompressed {
 	@Override
 	public ColGroupType getColGroupType() {
 		return ColGroupType.LinearFunctional;
+	}
+
+	@Override
+	public double getMin() {
+		double min = Double.POSITIVE_INFINITY;
+
+		for(int colIndex : _colIndexes) {
+			double intercept = _coefficents[0][colIndex];
+			double slope = _coefficents[1][colIndex];
+			if(slope >= 0 && (intercept + slope) < min) {
+				min = intercept + slope;
+			}
+			else if(slope < 0 && (intercept + _numRows * slope) < min) {
+				min = intercept + _numRows * slope;
+			}
+		}
+
+		return min;
+	}
+
+	@Override
+	public double getMax() {
+		double max = Double.NEGATIVE_INFINITY;
+
+		for(int colIndex : _colIndexes) {
+			double intercept = _coefficents[0][colIndex];
+			double slope = _coefficents[1][colIndex];
+			if(slope >= 0 && (intercept + _numRows * slope) > max) {
+				max = intercept + _numRows * slope;
+			}
+			else if(slope < 0 && (intercept + slope) > max) {
+				max = intercept + slope;
+			}
+		}
+
+		return max;
 	}
 
 	@Override
--- a/src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
@@ -26,6 +26,7 @@ import org.apache.sysds.runtime.data.DenseBlock;
 import org.apache.sysds.runtime.data.SparseBlock;
 import org.apache.sysds.runtime.functionobjects.Builtin;
 import org.apache.sysds.runtime.instructions.cp.CM_COV_Object;
+import org.apache.sysds.runtime.matrix.data.LibMatrixMult;
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 import org.apache.sysds.runtime.matrix.operators.BinaryOperator;
 import org.apache.sysds.runtime.matrix.operators.CMOperator;
@@ -50,9 +51,9 @@ public class ColGroupLinearFunctional extends AColGroupCompressed {
 	}
 
 	/**
-	 * Constructs an Constant Colum Group, that contains only one tuple, with the given value.
+	 * Constructs a Column Group that compresses its content using a linear functional.
 	 *
-	 * @param colIndices  The Colum indexes for the column group.
+	 * @param colIndices  The Column indexes for the column group.
 	 * @param coefficents The dictionary containing one tuple for the entire compression.
 	 */
 	private ColGroupLinearFunctional(int[] colIndices, double[][] coefficents, int numRows) {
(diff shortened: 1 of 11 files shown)


src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+			// We simply compute a matrix multiplication chain in coefficient space, i.e.,
+			// 					t(L) %*% R = t(coeff(L)) %*% W %*% coeff(R)
+			// where W is a weight matrix capturing the size of the shared dimension (weightMatrix above)
+			// and coeff(X) denotes the 2 x n matrix of the m x n matrix X.
+			MatrixBlock tmp = new MatrixBlock(lhs.getNumCols(), 2, false);
+			LibMatrixMult.matrixMult(coefficientMatrixLhs, weightMatrix, tmp);
+			LibMatrixMult.matrixMult(tmp, coefficientMatrixRhs, tmpRet);
+		} else if(lhs instanceof APreAgg) {
+			// TODO: implement
+			throw new NotImplementedException();
```
Baunsgaard, line 343: i think the best design for this would be some kind of iterator for the functional group that gives double[] line by line that can then be aggregated.

You do not need (i suggest not to) to implement this now

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+			coefficientMatrixLhs = LibMatrixReorg.transposeInPlace(coefficientMatrixLhs,
+				InfrastructureAnalyzer.getLocalParallelism());
+
+			// We simply compute a matrix multiplication chain in coefficient space, i.e.,
+			// 					t(L) %*% R = t(coeff(L)) %*% W %*% coeff(R)
+			// where W is a weight matrix capturing the size of the shared dimension (weightMatrix above)
+			// and coeff(X) denotes the 2 x n matrix of the m x n matrix X.
+			MatrixBlock tmp = new MatrixBlock(lhs.getNumCols(), 2, false);
+			LibMatrixMult.matrixMult(coefficientMatrixLhs, weightMatrix, tmp);
+			LibMatrixMult.matrixMult(tmp, coefficientMatrixRhs, tmpRet);
```
Baunsgaard, line 340: next level would be to not use a tempRet but instead directly output the result into the output matrix.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+		else {
+			double[] tmpRetV = tmpRet.getDenseBlockValues();
+			for(int row = 0; row < lhs.getNumCols(); row++) {
+				final int offRes = lhs._colIndexes[row] * result.getNumColumns();
+				final int offTmp = row * getNumCols();
+				for(int col = 0; col < getNumCols(); col++) {
+					resV[offRes + _colIndexes[col]] += tmpRetV[offTmp + col];
+				}
+			}
+		}
```
Baunsgaard, line 374: this code looks suspiciously much like the code inside ColGroupUncompressed,
i suggest that we combine the two into a util function in ColGroupUtil.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+	}
+
+	@Override
+	protected AColGroup sliceMultiColumns(int idStart, int idEnd, int[] outputCols) {
+		throw new NotImplementedException();
+	}
+
+	@Override
+	public AColGroup copy() {
+		throw new NotImplementedException();
```
Baunsgaard, line 394: if you are a bit bold, simply return this.
since all your values should be treated as immutable.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+
+	@Override
+	public void write(DataOutput out) throws IOException {
+		throw new NotImplementedException();
+	}
+
+	@Override
+	public long getExactSizeOnDisk() {
+		long ret = super.getExactSizeOnDisk();
+		ret += _coefficents[0].length * _coefficents.length * 8L;
```
Baunsgaard, line 425: here similarly to the ColGroupSizes function the calculation is wrong.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+		ret += 4L; // _numRows
+		return ret;
+	}
+
+	@Override
+	protected void computeProduct(double[] c, int nRows) {
+		c[0] = 1;
```
Baunsgaard, line 432: do not set it to 1, since you have to consider the input c as already allocated correctly.
Instead you can add an early abort `if 0 return`

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+			for(int i = 0; i < nRows; i++) {
+				c[0] *= intercept + slope * (i + 1);
+			}
+		}
+	}
+
+	@Override
+	protected void computeRowProduct(double[] c, int rl, int ru, double[] preAgg) {
+		for(int rix = rl; rix < ru; rix++) {
+			c[rix] = 1;
```
Baunsgaard, line 445: similar here. don't set it to 1

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
Baunsgaard: furthermore you could add a speedup, by adding a containsZero method that detects if any of the values contained in the function is zero return zero

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+				double slope = _coefficents[1][colIndex];
+				c[rix] *= intercept + slope * (rix + 1);
+			}
+		}
+	}
+
+	@Override
+	protected void computeColProduct(double[] c, int nRows) {
+		for(int colIndex : _colIndexes) {
+			c[colIndex] = 1;
```
Baunsgaard, line 457: Do not set it

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+			double slope = _coefficents[1][colIndex];
+			for(int i = 0; i < nRows; i++) {
+				c[colIndex] *= intercept + slope * (i + 1);
+			}
+		}
+	}
+
+	@Override
+	protected double[] preAggSumRows() {
+		throw new NotImplementedException();
```
Baunsgaard, line 468: use these methods to allocate what you want to use in the SumRows call.

for instance you currently have:

```
	@Override
	protected void computeRowSums(double[] c, int rl, int ru, double[] preAgg) {
	        double intercept_sum = 0;
		for(double intercept : _coefficents[0])
			intercept_sum += intercept;

		double slope_sum = 0;
		for(double slope : _coefficents[1])
			slope_sum += slope;

		for(int rix = rl; rix < ru; rix++)
			c[rix] += intercept_sum + slope_sum * (rix + 1);
        }
```

instead of processing all the coefficients in all calls to all rl to ru,
preAggregate the first two for loops into the double[] array to contain two values `double[]{intercept_sum, slope_sum}`
this makes the parallelization of different row ranges trivial, and much much much faster.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+	@Override
+	public AColGroup rexpandCols(int max, boolean ignore, boolean cast, int nRows) {
+		throw new NotImplementedException();
+	}
+
+	@Override
+	public double getCost(ComputationCostEstimator e, int nRows) {
+		final int nCols = getNumCols();
+		// We store 2 tuples in this column group, namely intercepts and slopes
+		return e.getCost(nRows, nRows, nCols, 2, 1.0);
```
Baunsgaard, line 505: this is going to be tricky to calculate, since this does not correctly reflect the calculation.
i suggest you add a 
`LOG.warn("Not precise cost calculation for LinearFunctional ColGroup");`

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupLinearFunctional.java
```diff
@@ -0,0 +1,516 @@
...
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append(super.toString());
+		sb.append(String.format("\n%15s", " Intercepts: \t" + Arrays.toString(_coefficents[0])));
```
Baunsgaard, line 512: if you format it, don't use `\t` since we want it to align based on the `%15s` not what tab length it fit after.

Baunsgaard (commented):
Hi @wedenigt 

I really like this PR, it nicely extends into the column groups.
The only major thing (that is very minor) is the memory estimates that are wrong because of java object overhead.
I suggest that you try to change it from double[][] to double[] to not only improve the memory overhead, but also the performance.

With what you have i would consider the AMLS project done,
but if you want to i suggest you go though my comments. (if you address any of them feel free to close them on the PR)

best regards
Sebastian

src/test/java/org/apache/sysds/test/component/compress/colgroup/ColGroupLinearFunctionalTest.java
```diff
@@ -0,0 +1,547 @@
...
+import org.apache.sysds.runtime.matrix.operators.AggregateOperator;
+import org.apache.sysds.runtime.matrix.operators.AggregateUnaryOperator;
+import org.apache.sysds.runtime.util.DataConverter;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.EnumSet;
+import java.util.Random;
+
+public class ColGroupLinearFunctionalTest {
```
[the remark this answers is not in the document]
Baunsgaard: if you detect that your line suddenly is a constant or zero, i suggest you make the constructor return a ColGroupConst or ColGroupZero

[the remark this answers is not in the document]
Baunsgaard: yes, it does the checks when constructing a new column group from uncompressed input, the checks are mainly for when you perform operations on the functional compression, and for some reason it ends up as a constant or zero. This makes you construct a functional group rather than the much cheaper const or empty groups if naively implemented.
Hence making a constructor that is able to return these other types is beneficial.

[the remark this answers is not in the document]
wedenigt: Alright, got it. Thanks for the explanation.

PR #1636, opened 2022-06-13
[MINOR] Minor fixes i.e., validation checks, formatting e.t.c.

Shafaq-Siddiqi:


j143:
Hi @Shafaq-Siddiqi , can you clarify whether we need to add `commons-compiler-3.0.16` to `lib` or this PR would be enough?

Shafaq-Siddiqi:
> Hi @Shafaq-Siddiqi , can you clarify whether we need to add `commons-compiler-3.0.16` to `lib` or this PR would be enough?

We need to add that library too along with this PR.

phaniarnab:
We are currently not packing commons-compiler in lib for the binary. `assembly/bin.xml` dictates the list of libraries to be included in the binary. Even for the local experiments, we usually point to the all-included lib directory, which hides the problem. @mboehm7

mboehm7:
thanks for clarifying @phaniarnab - this makes a lot of sense.

Baunsgaard:
Currently our automated tests only verify that the packaged lib (from bin.xml) works in python. We just need more tests that verify execution with more varied scripts to automate this verification of packaging the minimum jar files needed.

We should also add some application test via scripts that we execute with the release without python.

https://github.com/apache/systemds/blob/fd5be669fe7a03bd804415dcf5fe11e8dbd74645/.github/workflows/python.yml#L115

PR #1643, opened 2022-06-20
[SYSTEMDS-3393] Implement SIMD usage for basic dense dense MM

kev-inn:
# `DoubleVector` replacement for matrix multiply
JDK 17 adds `Vector` classes to use SIMD instructions. This PR replaces the basic dense dense matrix multiply with an equivalent `DoubleVector` implementation. It is necessary to use JDK 17, therefore we should not merge this yet, but keep it in staging for future reference.

As an experiment we check a simple matrix multiply:
`Z = X %*% Y`, $X\in \mathbb{R}^{n\times k}, Y\in \mathbb{R}^{k\times m}$

The experiment script performs 10 matrix multiplications and saves the time of the last 5 to give the JVM some time to optimize.

### Alpha Node
| $k = 1000$ | $k = 10000$ |
| --------------- | --------------- |
| plot_alpha_n_1000 | plot_alpha_n_10000 |

### Lima Node
| $k = 1000$ | $k = 10000$ |
| --------------- | --------------- |
| plot_lima_n_1000 | plot_lima_n_10000 |

### Alpha Node
| $k = 1000$ | $k = 10000$ |
| --------------- | --------------- |
| plot_alpha_m_1000 | plot_alpha_m_10000 |

### Lima Node
| $k = 1000$ | $k = 10000$ |
| --------------- | --------------- |
| plot_lima_m_1000 | plot_lima_m_10000 |

## Conclusion
The implementation seems to boost the performance in most cases. The case where we vary the number of columns n on the alpha node needs some more exploration, but it seems we are never worse than the current implementation.

## Experiment Script

```R
X = read($Xfname);
Y = read($Yfname);

lim = 10;
R = matrix(0, rows=lim, cols=1);
for (i in 1:lim) {
  t1 = time();
  Z = X %*% Y;
  t2 = time();
  R[i,1] = (t2-t1)/1000000;
}

print(sum(Z));
res = R[5:lim,];
write(res, $fname, format="csv", sep="\t");
```

### DIFF
--- a/bin/systemds
+++ b/bin/systemds
@@ -463,6 +463,7 @@ elif [ $SYSDS_DISTRIBUTED == 0 ]; then
   java $SYSTEMDS_STANDALONE_OPTS \
   -cp $CLASSPATH \
   $LOG4JPROP \
+  --add-modules=jdk.incubator.vector \
   org.apache.sysds.api.DMLScript \
   -f $SCRIPT_FILE \
   -exec $SYSDS_EXEC_MODE \
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
 		<slf4j.version>1.7.36</slf4j.version>
 		<log4j.version>2.17.2</log4j.version>
 		<!-- Set java compile level via argument, ex: 1.8 1.9 10 11-->
-		<java.level>11</java.level>
+		<java.level>17</java.level>
 		<!-->Testing settings<!-->
 		<maven.test.skip>true</maven.test.skip>
 		<test-parallel>classes</test-parallel>
@@ -245,6 +245,9 @@
 				<configuration>
 					<source>${java.level}</source>
 					<target>${java.level}</target>
+					<compilerArgs>
+						<arg>--add-modules=jdk.incubator.vector</arg>
+					</compilerArgs>
 				</configuration>
 			</plugin>
 
@@ -307,6 +310,7 @@
 					<reportFormat>brief</reportFormat>
 					<trimStackTrace>true</trimStackTrace>
 					<rerunFailingTestsCount>${rerun.failing.tests.count}</rerunFailingTestsCount>
+					<argLine>--add-modules=jdk.incubator.vector</argLine>
 				</configuration>
 			</plugin>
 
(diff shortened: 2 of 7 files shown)


phaniarnab:
This looks pretty good. I think Alpha has wider SIMD registers, which explains why most configurations perform better in Alpha.
Is there any JVM flag that you needed to enable? If so, can you please mention those as well for documentation purposes? @kev-inn

Baunsgaard:
What if you use some of the other experimental JVM that should have better support ?

kev-inn:
> What if you use some of the other experimental JVM that should have better support ?

Which ones do you have in mind, and what kind of support do you expect (which the current does not support)? To clarify, do you expect better performance or that we can remove the `--add-modules=jdk.incubator.vector` flag?

Baunsgaard:
> > What if you use some of the other experimental JVM that should have better support ?
> 
> Which ones do you have in mind, and what kind of support do you expect (which the current does not support)? To clarify, do you expect better performance or that we can remove the `--add-modules=jdk.incubator.vector` flag?

Project Panama: https://openjdk.java.net/projects/panama/
And JDK 19 have the official full support for vectorizing; https://openjdk.org/jeps/426

kev-inn:
# More experiments
Dump of more experiments and updated plots.

All in all the results look promising, but we can also clearly some of the weak spots.

## Alpha
### Variable columns
Alpha_1_1000_variable
Alpha_10_1000_variable
Alpha_1000_1_variable
Alpha_1000_10_variable
Alpha_1000_1000_variable
Alpha_1000_10000_variable
### Variable rows
Alpha_variable_1_1000
Alpha_variable_10_1000
Alpha_variable_1000_1
Alpha_variable_1000_10
Alpha_variable_1000_1000
Alpha_variable_10000_1000

## Lima
### Variable columns
Lima_1_1000_variable
Lima_10_1000_variable
Lima_1000_1_variable
Lima_1000_10_variable
Lima_1000_1000_variable
Lima_1000_10000_variable

### Variable rows
Lima_variable_1_1000
Lima_variable_10_1000
Lima_variable_1000_1
Lima_variable_1000_10
Lima_variable_1000_1000
Lima_variable_10000_1000

kev-inn:
Closed by 9bf0a9fd9a64264301b0fc9310d6d9f976780870 (messed up the commit message)

PR #1647, opened 2022-06-28
[SYSTEMDS-3396] LocalVarMap Concurrency in Federated Execution

ywcb00:
Hi,
This PR changes the local variable map back to a _ConcurrentHashMap_ in order to allow for simultaneously modification and iteration of the map.
Without a _ConcurrentHashMap_, some algorithms might throw a _ConcurrentModificationException_ because the coordinator does not wait for the worker's result and triggers a new federated instruction in the meanwhile. Therefore, the worker iterates the local variable map when performing an '_rmvar_' instruction, and creates new entries in the local variable map at the same time, resulting in a _ConcurrentModificationException_. Thanks to @kev-inn for finding this issue.
Another approach to resolve this problem would be to create a queue of federated requests for every coordinator+threadid pair and process the requests of the queue sequentially.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/LocalVariableMap.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/LocalVariableMap.java
@@ -20,6 +20,7 @@
 package org.apache.sysds.runtime.controlprogram;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


Baunsgaard:
Thanks, and merged!

PR #1648, opened 2022-06-28
[SYSTEMDS-3397] Python NN testExample

Baunsgaard:
This pr contains a simple python example of neural network training and preprocessing.

two different scenarios are tested. 
1. if we train and measure accuracy in one go
2. If we train a model then save it, to then load and predict.

Both cases use preprocessing transform encode giving full end to end pipeline from frames to trained models.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
 	</licenses>
 
 	<properties>
-		<hadoop.version>3.3.2</hadoop.version>
+		<hadoop.version>3.3.3</hadoop.version>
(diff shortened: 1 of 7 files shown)


Baunsgaard:
comments appreciated!
especially on neural network code, or python for evaluating the neural network for making it easier/better to use.

phaniarnab:
This is a good test case. The python implementation looks pretty easy on the eyes to me.
Albeit I am very curious about the performance on larger datasets (compared to full dml-based implementations).

PR #1649, opened 2022-06-30
[SYSTEMDS-3018] Dynamic recompilation of functions with Federated Planner

kev-inn:
WIP: Currently hard-coded to use the cost-based planner, will be extended to choose based on configuration.

During compilation, if activated, we choose the best plan for a federated execution. This PR extends this by reevaluating/recomputing the plan, if a function is called multiple times with different (federated) arguments.


src/main/java/org/apache/sysds/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
```diff
@@ -2398,7 +2399,9 @@ else if( bop.getOp() == OpOp2.MINUS && left.getDataType()==DataType.MATRIX
 			}
 			
 			//rewire parent-child operators if rewrite applied
-			if( ternop != null ) { 
+			if( ternop != null ) {
+				if (right.getForcedExecType() == Types.ExecType.FED)
+					ternop.setForcedExecType(Types.ExecType.FED);
```
kev-inn, line 2404: This is necessary, because our planner might assume the output to be federated. The fused ternary instruction is generated to execute in CP mode. We could either stop those rewrites from the beginning or be careful where we have to add the fed execution mode.

src/main/java/org/apache/sysds/hops/fedplanner/FederatedPlannerCostbased.java
```diff
@@ -53,23 +54,29 @@
...
-	private final static Set<Long> hopRelUpdatedFinal = new HashSet<>();
+	private final Set<Long> hopRelUpdatedFinal = new HashSet<>();
 	/**
 	 * Terminal hops in DML program given to this rewriter.
 	 */
-	private final static List<Hop> terminalHops = new ArrayList<>();
-	private final static Map<String, Hop> transientWrites = new HashMap<>();
+	private final List<Hop> terminalHops = new ArrayList<>();
+	private final Map<String, Hop> transientWrites = new HashMap<>();
+	private LocalVariableMap localVariableMap = new LocalVariableMap();
```
kev-inn, line 79: I think the `localVariableMap` can be combined into `transientWrites`. I don't think we necessarily require Hops, but the implementation needs some more changes in that case, so I added it like this for now.

sebwrede:
Is this still WIP or should it be merged?

PR #1650, opened 2022-07-03
[SYSTEMDS-3390] Replace slicing for countDistinctApprox() row/col aggregation with direct ops

BACtaki:
JIRA: https://issues.apache.org/jira/browse/SYSTEMDS-3390

This patch improves the performance of `countDistinctApprox()` row/col aggregation by replacing matrix slicing with direct ops on the input matrix. This has the most impact in local CP execution mode, as some simple experiments show:

(numbers represent average over 3 runs)
1. row aggregation
    (A) dense: 10000x1000 with sparsity=0.9
    1.198s with slicing, 0.874s without slicing - a 27% improvement

    (B) sparse: 10000x1000 with sparsity=0.1
    0.528s with slicing, 0.512s without slicing - a 3% improvement

As expected, the larger and the more dense the input matrix, the larger the performance improvement.

2. col aggregation
    (A) dense: 1000x10000 with sparsity=0.9
    1.186s with slicing, 1.036s without slicing - a 13% improvement

    (B) sparse: 1000x10000 with sparsity=0.1
    1.272s with slicing, 0.647s without slicing - a 49% improvement

In this case, the sparser the input matrix, the larger the performance improvement. This phenomenon is a result of employing a hash map M in the implementation: as the RxC input matrix becomes denser, M's keyset size approaches C, and the performance approaches the baseline, which uses slicing.


Baunsgaard:
Sorry, missed this PR, will take a look tomorrow!

phaniarnab:
Thanks for the PR @BACtaki. I had an initial look at the code changes today. Before I get into detailed comments, I'd like to clarify a few things.
- Are the optimizations only improve the naive cases where the input dimension is less than 1024 for the given direction (row/col)? 
- I see that you are now iterating the dense and sparse inputs in a more cache-conscious manner (reducing CPU cache misses). Are there any other optimizations you are employing (e.g. reducing the number of intermediates)?

Sorry for the delay. I will add my comments and suggestions tomorrow.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixCountDistinct.java
```diff
@@ -102,66 +114,255 @@ else if(in.getNonZeros() < minimumSize) {
...
+				}
 			}
-			else {
-				for(int i = 0; i < in.getNumRows(); i++) {
-					if(!sb.isEmpty(i)) {
-						data = in.sparseBlock.values(i);
+
+			blkOut.setValue(0, 0, distinctCount);
+		} else if (op.getDirection().isRow()) {
+			blkOut = blkIn.slice(0, blkIn.getNumRows() - 1, 0, 0);
```
phaniarnab, line 170: Why deep copying (slice) the first column is required? You are anyway going to set individual values. Won't just allocating a dense column vector work? Something like below.
`blkOut = new MatrixBlock(blkIn.getNumRows(), 1, false, blkIn.getNumRows());
 blkOut.allocateBlock();`

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixCountDistinct.java
```diff
@@ -102,66 +114,255 @@ else if(in.getNonZeros() < minimumSize) {
...
 				}
 			}
-		}
-		else if(in.denseBlock != null) {
-			DenseBlock db = in.denseBlock;
-			for(int i = 0; i <= db.numBlocks(); i++) {
-				data = db.valuesAt(i);
-				countDistinctValuesNaive(data, distinct);
+		} else {  // Col aggregation
+			blkOut = blkIn.slice(0, 0, 0, blkIn.getNumColumns() - 1);
```
phaniarnab, line 253: Similar to above.

BACtaki:
Thanks for your review @phaniarnab! Comments inline:

> Are the optimizations only improve the naive cases where the input dimension is less than 1024 for the given direction (row/col)?

The cache-conscious iteration optimization only applies to the naive counting case where `nnz < minimumSize 
 = 1024`. Aside from this, I also refactored `MatrixSketch` and `KMVSketch` to replace the `getScalarValue() -> Integer` method with the `getValue() -> MatrixBlock` method to unify the handling of RowCol, Row, and Col cases (as per @Baunsgaard 's suggestion from an earlier PR). 

> I see that you are now iterating the dense and sparse inputs in a more cache-conscious manner (reducing CPU cache misses). Are there any other optimizations you are employing (e.g. reducing the number of intermediates)?

Really good point re intermediates. We are actually creating an intermediate `SmallestPriorityQueue` of max size `k = 64` per row/col. The way we are constructing the intermediates places an upper bound on additional memory:
1. RowCol: max 64 double values (64 bits each)
2. Row: `SmallestPriorityQueue spq` is reset between each iteration, so the intermediate is at most 64 double values
3. Col: identical to Row, but the iteration occurs in column-major order and `spq` is reset between iterations of consecutive columns

There are no further optimizations beyond cache-conscious iteration in CP case.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixCountDistinct.java
BACtaki: >
			// Result is a Mx1 matrix block.
			// TODO We are reusing memory allocated for input matrix block here - explore whether it would be better to
			//  allocate new
			MatrixBlock resultMatrix = in.slice(0, in.getNumRows() - 1, 0, 0);

I was trying to save on new memory allocations by reusing previously allocated, but you are right: it is probably unnecessary, not to mention that it is likely more expensive performance-wise to compute slice than to allocate new- fixed in next iteration.

BACtaki:
Not sure why [**.functions.builtin.part2.**](https://github.com/apache/systemds/runs/7579235369?check_suite_focus=true#logs) test is failing- maybe it's a transient issue?

```
script file: ./src/test/scripts/functions/builtin/multipleBuiltins.R
Error:    Run 2: MultipleBuiltinsTest.testMultipleBuiltinsDefaultSP:54->runMultipleBuiltinsTest:76->AutomatedTestBase.runRScript:1258 ERROR: R has ended irregularly
R Standard output :

R Standard Error  :
Error in library("DescTools") : there is no package called ‘DescTools’
Execution halted


script file: ./src/test/scripts/functions/builtin/multipleBuiltins.R
Error:    Run 3: MultipleBuiltinsTest.testMultipleBuiltinsDefaultSP:54->runMultipleBuiltinsTest:76->AutomatedTestBase.runRScript:1258 ERROR: R has ended irregularly
R Standard output :

R Standard Error  :
Error in library("DescTools") : there is no package called ‘DescTools’
Execution halted
```

phaniarnab:
Thanks, @BACtaki. The changes now look good to me. I will merge the PR in some time.
Don't worry about the test failure. That is due to some docker + R issues. @Shafaq-Siddiqi is working on it.

phaniarnab:
Thanks again @BACtaki. While merging I fixed a few minor warnings (redundant imports, missing cast, etc.).

BACtaki:
> Thanks again @BACtaki. While merging I fixed a few minor warnings (redundant imports, missing cast, etc.).

Thanks for your review @phaniarnab ! I will definitely be more careful with warnings going forward.

PR #1663, opened 2022-07-13
[MINOR] Fix FederatedLookupTable Eviction

ywcb00:
Hi,
This PR replaces the complete clear of the federated lookup table by removing only the entry of the corresponding coordinator which sent the respective _CLEAR_ request.
Additionally, this PR changes the _FederatedStatistics.coordinatorsTrafficBytes_ list to a CopyOnWriteArrayList in order to avoid the _ConcurrentModificationException_ in a multi-tenant scenario, and executes the federated requests of the _IndexingFEDInstruction_ in the correct order to broadcast the scalar object before calling the instruction.

Thanks for review :)


ywcb00:
I made a small change to the _Statistics_ class. We now obtain the size of the heavy hitters right after creating the separate array with entries in order to avoid nullpointer exceptions caused by wrong array lengths because another thread modifies the heavy hitters at the same time.

PR #1666, opened 2022-07-17
[SYSTEMDS-3018] Support function calls in `FedAll` and `Heuristic` planners

kev-inn:


sebwrede:
The federated planning tests for FOUT and Heuristic fail. I think some more work is needed before this PR can be merged. The other failing tests seem to be caused by something unrelated to this PR.

PR #1667, opened 2022-07-20
[MINOR] Fix federation for Python API

kev-inn:
Fixes a bug where instructions were not replaced by FED equivalent instructions, because the correct `CompilerConfig` option was not set.
In the future we want to check the statistics report in our federated testcases to check if execution was performed on federated workers.

### DIFF
--- a/src/main/java/org/apache/sysds/api/jmlc/Connection.java
+++ b/src/main/java/org/apache/sysds/api/jmlc/Connection.java
@@ -31,6 +31,7 @@ import java.util.Map;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.sysds.hops.OptimizerUtils;
 import org.apache.sysds.runtime.meta.MetaDataAll;
 import org.apache.sysds.api.DMLException;
 import org.apache.sysds.api.DMLScript;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/sysds/api/jmlc/Connection.java
```diff
@@ -145,23 +146,15 @@ public Connection(DMLConfig dmlconfig) {
...
-		_cconf = cconf;
-		
+		_cconf = OptimizerUtils.constructCompilerConfig(dmlconfig);
+		_cconf.set(ConfigType.IGNORE_UNSPECIFIED_ARGS, true);
+		_cconf.set(ConfigType.IGNORE_READ_WRITE_METADATA, true);
+		_cconf.set(ConfigType.IGNORE_TEMPORARY_FILENAMES, true);
+		_cconf.set(ConfigType.REJECT_READ_WRITE_UNKNOWNS, false);
+		_cconf.set(ConfigType.ALLOW_INDIVIDUAL_SB_SPECIFIC_OPS, false);
+		_cconf.set(ConfigType.ALLOW_CSE_PERSISTENT_READS, false);
+		_cconf.set(ConfigType.CODEGEN_ENABLED, false);
```
Baunsgaard, line 156: It looks good to me, but i think many of these settings should not be set custom here.

I we can i suggest removing all of them.

kev-inn: We need those, otherwise we can't execute the scripts as they read from files that do not exist, but are provided to the runtime by python.

Baunsgaard: but Codegen for instance?
I get the Reject_read_write unknowns should be enabled.

PR #1668, opened 2022-07-20
[SYSTEMDS-2835] Add a federated version of adult_neural and mnist tests

kev-inn:
The mnist version works. Adult neural still has problems, because `toOneHot()` is not executed federated and the federated Paramserver fails if features are federated, but labels are not.
Requires #1667

### DIFF
--- a/src/main/python/systemds/context/systemds_context.py
+++ b/src/main/python/systemds/context/systemds_context.py
@@ -31,6 +31,7 @@ from subprocess import PIPE, Popen
 from threading import Thread
 from time import sleep
 from typing import Dict, Iterable, Sequence, Tuple, Union
+from contextlib import contextmanager
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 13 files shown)


kev-inn:
The two helpers in the new testcases could be combined, but importing of helper classes is weird for testcases, so I don't plan on adding this for the commit.
Also please double check if the linked JIRA issue is fine

src/main/python/systemds/context/systemds_context.py
```diff
@@ -306,6 +311,71 @@ def __get_open_port(self):
...
+
+    @contextmanager
+    def capture_stats_context(self):
+        """
+        Context for capturing statistics. Should be used in a `with` statement.
+        Afterwards capturing will be reset to the state it was before.
+
+        Example:
+        ```Python
+        with sds.capture_stats_context():
```
Baunsgaard, line 339: i do not like the requirement of with here.

preferably i think something like:

sds.capture_stats()

...
x.compute().
...
sds.get_stats()

is perfectly fine.
this remove the need for this capture_stats_context.

kev-inn: This is completely optional, just offers another (arguably neater) way for the common pattern of
```Python
sds.capture_stats()
# some computation/computations
...
sds.capture_stats(False)
```

PR #1671, opened 2022-07-26
[MINOR] Fixes FrameBlock$FloatArray datatype.

paginabianca:
Fixes wrong datatype and adds missing switch cases in FrameBlock


Baunsgaard:
Hi @paginabianca 
Thanks for the PR. I will say that this is a minor change and it is not hooked up completely to the entire pipeline since we do no differentiate differentiate float previsions in the frame block yet, but thanks for the contribution.
I will merge shortly

mboehm7:
Thanks for merging - @paginabianca encountered the related bug with a frame having FP32 columns in its mtd file and consuming it in transformencode. It might be nice to create a respective test.

phaniarnab:
We didn't encounter this issue in the local transformencode simply because we don't call getColumnData there.
In the encoders we access cell-wise (getDouble) and the objectToDouble method rightly handles FP32 @mboehm7.
Thanks @paginabianca for even fixing the getInMemorySize method. Not supporting FP32 there leads to wrong size calculation and in turn array out-of-bound error during serializing.

PR #1676, opened 2022-08-03
[SYSTEMDS-3390] Improve performance of countDistinctApprox()

phaniarnab:
This patch improves the performance of countDistinctApprox() row/col
aggregation by replacing matrix slicing with direct ops on the input
matrix. This has the most impact in local CP execution mode, as
some simple experiments show:

(numbers represent average over 3 runs)
1. row aggregation
    (A) dense: 10000x1000 with sparsity=0.9
    1.198s with slicing, 0.874s without slicing - a 27% improvement

    (B) sparse: 10000x1000 with sparsity=0.1
    0.528s with slicing, 0.512s without slicing - a 3% improvement

As expected, the larger and the more dense the input matrix,
the larger the performance improvement.

2. col aggregation
    (A) dense: 1000x10000 with sparsity=0.9
    1.186s with slicing, 1.036s without slicing - a 13% improvement

    (B) sparse: 1000x10000 with sparsity=0.1
    1.272s with slicing, 0.647s without slicing - a 49% improvement

In this case, the sparser the input matrix, the larger the performance
improvement. This phenomenon is a result of employing a hash map M
in the implementation: as the RxC input matrix becomes denser, M's
keyset size approaches C, and the performance approaches the baseline,
which uses slicing.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/cp/AggregateUnaryCPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/cp/AggregateUnaryCPInstruction.java
@@ -82,8 +82,12 @@ public class AggregateUnaryCPInstruction extends UnaryCPInstruction {
 				in1, out, AUType.valueOf(opcode.toUpperCase()), opcode, str);
 		} 
 		else if(opcode.equalsIgnoreCase("uacd")){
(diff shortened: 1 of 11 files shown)


phaniarnab:
To run tests after rebasing PR #1650

PR #1677, opened 2022-08-03
[SYSTEMDS-3413] Add support for row/col aggregation to countDistinct builtin function

BACtaki:
This patch converts countDistinct() from a non-parameterized builtin to a parameterized builtin function to allow for 1 new parameter: dir for direction. The value of dir can be r and c, denoting row-wise and column-wise aggregation respectively. This patch only implements CP and the SP case will throw a NotImplementedException()- the latter case will be addressed in a subsequent patch.


BACtaki:
@phaniarnab @Baunsgaard @j143, would love your feedback!

phaniarnab:
I will look into this @BACtaki. 
Meanwhile, please check why the tests are failing.

BACtaki:
> I will look into this @BACtaki. Meanwhile, please check why the tests are failing.

Thanks @phaniarnab ! Ack, let me look into the test failures.

src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
```diff
@@ -420,7 +421,11 @@ else if ( opcode.equalsIgnoreCase("uacmin") ) {
 			AggregateOperator agg = new AggregateOperator(Double.POSITIVE_INFINITY, Builtin.getBuiltinFnObject("min"));
 			aggun = new AggregateUnaryOperator(agg, ReduceRow.getReduceRowFnObject(), numThreads);
 		}
-		
+		else if ( opcode.equalsIgnoreCase("uacd") || opcode.equalsIgnoreCase("uacdr")
+				|| opcode.equalsIgnoreCase("uacdc") ) {
```
Baunsgaard, line 425: Should we add the approximate methods as well?

src/main/java/org/apache/sysds/runtime/instructions/SPInstructionParser.java
```diff
@@ -126,7 +126,9 @@ public class SPInstructionParser extends InstructionParser
 		String2SPInstructionType.put( "uac*"    , SPType.AggregateUnary);
 		String2SPInstructionType.put( "uatrace" , SPType.AggregateUnary);
 		String2SPInstructionType.put( "uaktrace", SPType.AggregateUnary);
-		String2SPInstructionType.put( "uacdap"  , SPType.AggregateUnary);
+		String2SPInstructionType.put( "uacd"    , SPType.AggregateUnary);
+		String2SPInstructionType.put( "uacdr"   , SPType.AggregateUnary);
+		String2SPInstructionType.put( "uacdc"   , SPType.AggregateUnary);
```
Baunsgaard, line 131: same here, should we add the approximate methods as well?

src/test/scripts/functions/countDistinct/countDistinctRowCol.dml
```diff
@@ -0,0 +1,24 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+input = round(rand(rows = $2, cols = $3, min = 0, max = $1 -1, sparsity= $4,  seed = 7))
+res = countDistinct(input)  # default is dir="rc"
```
Baunsgaard, line 23: add another test for this, the call with "rc" should be equivalent to no "rc" argument.
While it is trivial and known, it is an easy test to add and it will enforce it later on.

Baunsgaard (commented):
Hi @BACtaki 

I like the PR, i do not see in the code why the tests would fail so i will force them to restart, 
I have left a few comments to the code in the PR.
My biggest concern is the arguments call, and i honestly do not know the conventions for the R like syntax we are using.

is the method supposed to be:

countDistinctCol(X) or countDistinct(X, "c")

Personally i would prefer if we had both, maybe @mboehm7 have an opinion?

BACtaki:
> I like the PR, i do not see in the code why the tests would fail so i will force them to restart,
The following careless edit was the culprit:
```
@@ -2361,10 +2375,6 @@ public class DMLTranslator
 		case SUM:
 		case PROD:
 		case VAR:
-		case COUNT_DISTINCT:
-			currBuiltinOp = new AggUnaryOp(target.getName(), DataType.SCALAR, target.getValueType(),
-					AggOp.valueOf(source.getOpCode().name()), Direction.RowCol, expr);
-			break;
```

> is the method supposed to be:
> countDistinctCol(X) or countDistinct(X, "c")
> Personally i would prefer if we had both, maybe @mboehm7 have an opinion?

That is a great idea! I have added `countDistinctRow` and `countDistinctCol` aliases for `countDistinct(X, dir="r")` and `countDistinct(X, dir="c")` respectively.

src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
BACtaki: I have added them in a separate "section" for sketch operators:
```
		String2SPInstructionType.put( "uacd"    , SPType.AggregateUnary);
		String2SPInstructionType.put( "uacdr"   , SPType.AggregateUnary);
		String2SPInstructionType.put( "uacdc"   , SPType.AggregateUnary);

		// Aggregate unary sketch operators
		String2SPInstructionType.put( "uacdap" , SPType.AggregateUnarySketch);
		String2SPInstructionType.put( "uacdapr", SPType.AggregateUnarySketch);
		String2SPInstructionType.put( "uacdapc", SPType.AggregateUnarySketch);
```

src/main/java/org/apache/sysds/runtime/instructions/SPInstructionParser.java
BACtaki: They are already there- they are in a separate "section" for sketch operators:
```
		String2SPInstructionType.put( "uacd"    , SPType.AggregateUnary);
		String2SPInstructionType.put( "uacdr"   , SPType.AggregateUnary);
		String2SPInstructionType.put( "uacdc"   , SPType.AggregateUnary);

		// Aggregate unary sketch operators
		String2SPInstructionType.put( "uacdap" , SPType.AggregateUnarySketch);
		String2SPInstructionType.put( "uacdapr", SPType.AggregateUnarySketch);
		String2SPInstructionType.put( "uacdapc", SPType.AggregateUnarySketch);
```

src/main/java/org/apache/sysds/lops/PartialAggregate.java
```diff
@@ -342,11 +342,20 @@ else if( dir == Direction.Col )
 			}
 
 			case COUNT_DISTINCT: {
-				if(dir == Direction.RowCol )
-					return "uacd";
+				switch (dir) {
+					case RowCol: return "uacd";
+					case Row: return "uacdr";
+					case Col: return "uacdc";
```
Baunsgaard, line 348: Maybe throw an exception here in the default case and not use the break.
This would make errors more informative.

src/main/java/org/apache/sysds/lops/PartialAggregate.java
```diff
@@ -355,6 +364,12 @@ else if( dir == Direction.Col )
 				}
 				break;
 			}
+
```
Baunsgaard, line 367: Same just above here , throw exception instead of break.

src/main/java/org/apache/sysds/runtime/instructions/InstructionUtils.java
```diff
@@ -420,7 +413,40 @@ else if ( opcode.equalsIgnoreCase("uacmin") ) {
...
+					.setDirection(Types.Direction.Row)
+					.setIndexFunction(ReduceCol.getReduceColFnObject());
+		}
+		else if ( opcode.equalsIgnoreCase("uacdapc") ) {
+			aggun = new CountDistinctOperator(AggregateUnaryCPInstruction.AUType.COUNT_DISTINCT_APPROX)
+					.setDirection(Types.Direction.Col)
+					.setIndexFunction(ReduceRow.getReduceRowFnObject());
+		}
+		else {
+			throw new IllegalArgumentException("Cannot parse given opcode");
```
Baunsgaard, line 447: I think this is what makes the tests fail. since some code rely on us returning the aggun as null i believe.

src/main/java/org/apache/sysds/runtime/matrix/operators/AggregateUnaryOperator.java
```diff
@@ -35,7 +35,7 @@ public class AggregateUnaryOperator extends MultiThreadedOperator {
 	private static final long serialVersionUID = 6690553323120787735L;
 
 	public final AggregateOperator aggOp;
-	public final IndexFunction indexFn;
+	public IndexFunction indexFn;
```
Baunsgaard, line 38: if we can please keep it final.

Baunsgaard (commented):
LGTM, I like it. Also i think i spotted the error and left a comment, please try it out.

src/main/java/org/apache/sysds/runtime/matrix/operators/AggregateUnaryOperator.java
BACtaki: 👍  I refactored the `CountDistinctOperator` ctor to make the API more intuitive

Baunsgaard:
Thanks for the PR, while merging i changed a wildcard to individual imports.
we should avoid wildcard imports.

PR #1689, opened 2022-08-26
[SYSTEMDS-3429] Multithreaded transformencode on fed workers.

paginabianca:
Use the local level of parallelism on worker nodes when `transfromencode` is used in federated mode.
@mboehm7


phaniarnab:
Can you please report what improvements you see with this change on a few use cases of FTBench? @paginabianca

PR #1693, opened 2022-09-06
[WIP] Python callback startup

Baunsgaard:


Baunsgaard:
This PR introduce updates to the Python API for faster startup and removes the need of looking at std out and err.
This overall leads to faster python API startup, and initial results show an improvement of all python tests from 
~350 sec to 277 sec.

Baunsgaard:
Single threaded tests:
Tests Before : 350 sec
Tests Now    : 186 sec

Startup SystemDS : 1.83
Python Now       : 1.878
Python old       : 2.0


SystemDS direct:
Screenshot from 2022-09-07 18-31-15

New Python API:
Screenshot from 2022-09-07 18-32-42

PR #1700, opened 2022-10-04
[MINOR] Update docker release script to publish

Baunsgaard:


Baunsgaard:
Previously the push of this script was set to false, this PR change it to true, so now we can publish our docker image with latest flag and version numbers.

PR #1709, opened 2022-10-21
[SYSTEMDS-3447] SDC AppendN

Baunsgaard:
This PR addresses the effeciency and implementation of AppendN in case of Offsets contained in all SDC Groups.
Initial performance numbers looks as follows with variable sparsity:
image

Here all increases in the read line for read performance is because of unsupported append N for SDC groups.


Baunsgaard:
Generic AppendN offsets:

image

Baunsgaard:
Specialized Append, and multi threading the append:

image

Unfortunately the parallelization have increased the base time, of the previous already good execution time, so further tuning is needed to make decisions for parallelization.

To break down the execution time better, i have chosen to change to only plot IO time, instead of end-to-end time:

image

Baunsgaard:
The compiler seems to be doing something questionable, since each execution have an overhead of 0.3 to 0.6 sec in compiler.
Is this something we can optimize @mboehm7 ?

PR #1710, opened 2022-10-24
[SYSTEMDS-3454] CLA Sheme primitive 

Baunsgaard:
This PR is for the code to enable extracting compression schemes and applying them to unseen data.
The application of this is to compress more data, and append to already compressed matrix blocks without the need of decompression.


Baunsgaard:
Something went wrong in the PR such that GitHub thinks it contains old commits.
Therefore closing PR. and opening a new one.

PR #1711, opened 2022-10-24
[SYSTEMDS-3454] CLA Sheme primitive

Baunsgaard:
This PR is for the code to enable extracting compression schemes and applying them to unseen data.
The application of this is to compress more data, and append to already compressed matrix blocks without the need of decompression.


Baunsgaard:
image

Always nice when everything is tested

Baunsgaard:
@phaniarnab , the async spark test fails here as well. Maybe we need to make them run in isolation for consistancy?

PR #1714, opened 2022-10-31
[SYSTEMDS-3254] Add new aliases for countDistinct built-in function

BACtaki:
This patch adds 2 new aliases for the `countDistinct` builtin function: `uniqueRow(X)` and `uniqueCol(X)` are aliases for `countDistinct(X, dir="r")` and `countDistinct(X, dir="c")` respectively.


src/main/java/org/apache/sysds/parser/DMLTranslator.java
```diff
@@ -2065,12 +2065,14 @@ private Hop processParameterizedBuiltinFunctionExpression(ParameterizedBuiltinFu
...
 
 			case COUNT_DISTINCT_ROW:
+			case UNIQUE_ROW:
 			case COUNT_DISTINCT_APPROX_ROW:
 				currBuiltinOp = new AggUnaryOp(target.getName(), DataType.MATRIX, target.getValueType(),
 						AggOp.valueOf(source.getOpCode().name()), Direction.Row, paramHops.get("data"));
 				break;
 
 			case COUNT_DISTINCT_COL:
+			case UNIQUE_COL:
```
Baunsgaard, line 2075: Are these valid? is Unique also supporting approximate?

Baunsgaard (commented):
Hi @BACtaki !

Thanks for the addition, 
When i made the task i was trying to make the methods the same as in R, since much of the syntax is similar.

https://www.geeksforgeeks.org/unique-function-in-r/

Unfortunately it seems i was a bit quick in my decision, since unique  != count distinct.
and what really is missing is the unique() function to return all unique elements in a matrix.

Sorry for the inconvenience.

Best regards
Sebastian.

BACtaki:
> Hi @BACtaki !
> 
> Thanks for the addition,
> When i made the task i was trying to make the methods the same as in R, since much of the syntax is similar.
> 
> https://www.geeksforgeeks.org/unique-function-in-r/
> 
> Unfortunately it seems i was a bit quick in my decision, since unique != count distinct.
> and what really is missing is the unique() function to return all unique elements in a matrix.
> 
> Sorry for the inconvenience.
> 
> Best regards
> Sebastian.

Thanks for clearing that up @Baunsgaard ! 

> [..] what really is missing is the unique() function to return all unique elements in a matrix.

I see a unique function in `builtins.java`:
```
	UNIQUE("unique", true),
```

In fact, there is an existing test for this function, which performs an equivalence check for the following:

- `unique.dml`
```
X = read($1);
R = unique(X = X);
write(R, $2);
```

- `unique.R`
```
args<-commandArgs(TRUE)
options(digits=22)
library("Matrix")

X = as.matrix(readMM(paste(args[1], "X.mtx", sep="")));
R = unique(X[order(X[,1]),]);
writeMM(as(R, "CsparseMatrix"), paste(args[2], "R", sep=""));
```

Is this the function you had in mind for the [JIRA](https://issues.apache.org/jira/browse/SYSTEMDS-3254) or was it something else?

Baunsgaard:
> > Hi @BACtaki !
> > Thanks for the addition,
> > When i made the task i was trying to make the methods the same as in R, since much of the syntax is similar.
> > https://www.geeksforgeeks.org/unique-function-in-r/
> > Unfortunately it seems i was a bit quick in my decision, since unique != count distinct.
> > and what really is missing is the unique() function to return all unique elements in a matrix.
> > Sorry for the inconvenience.
> > Best regards
> > Sebastian.
> 
> Thanks for clearing that up @Baunsgaard !
> 
> > [..] what really is missing is the unique() function to return all unique elements in a matrix.
> 
> I see a unique function in `builtins.java`:
> 
> ```
> 	UNIQUE("unique", true),
> ```
> 
> In fact, there is an existing test for this function - `BuiltinUniqueTest.java` - that performs an equivalence check for the following:
> 
>     * `unique.dml`
> 
> 
> ```
> X = read($1);
> R = unique(X = X);
> write(R, $2);
> ```
> 
>     * `unique.R`
> 
> 
> ```
> args<-commandArgs(TRUE)
> options(digits=22)
> library("Matrix")
> 
> X = as.matrix(readMM(paste(args[1], "X.mtx", sep="")));
> R = unique(X[order(X[,1]),]);
> writeMM(as(R, "CsparseMatrix"), paste(args[2], "R", sep=""));
> ```
> 
> Is this the function you had in mind for the [JIRA](https://issues.apache.org/jira/browse/SYSTEMDS-3254) or was it something else?

Yes, currently the tests case should parse, but for the wrong reasons.
I added it with the intention of comparing to R, but R does something different, that the test does not capture, since it returns the distinct elements from a unique call rather than the number of distinct elements.

BACtaki:
> Yes, currently the tests case should parse, but for the wrong reasons.
> I added it with the intention of comparing to R, but R does something different, that the test does not capture, since it returns the distinct elements from a unique call rather than the number of distinct elements.

Okay. If I understand you correctly, the `unique()` function has to be "reimplemented" in DML to mimic the same function in R- is that so? Is there a JIRA for it or should I create one?

Baunsgaard:
> > Yes, currently the tests case should parse, but for the wrong reasons.
> > I added it with the intention of comparing to R, but R does something different, that the test does not capture, since it returns the distinct elements from a unique call rather than the number of distinct elements.
> 
> Okay. If I understand you correctly, the `unique()` function has to be "reimplemented" in DML to mimic the same function in R- is that so? Is there a JIRA for it or should I create one?

I have not made such a JIRA yet, so it would be nice if you would !
I think the basic behavior is: unique returns all unique rows in a matrix input.

BACtaki:
👍 I have added [this JIRA](https://issues.apache.org/jira/browse/SYSTEMDS-3463) to track this feature.

Closing this PR.

PR #1721, opened 2022-11-03
[SYSTEMDS-3280] Docker Setup For HE

sebwrede:
This PR changes the dockerfile for tests to include SEAL for running the homomorphic encryption tests. Currently, the HE tests are ignored, but this PR removes the 'ignore' tags and adds the tests to the GitHub test workflows.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -23,6 +23,23 @@
 # A script to execute the tests inside the docker container.
 
 cd /github/workspace
+cd src/main/cpp
+# Download SEAL here?
+# SEAL
+#wget -qO- https://github.com/microsoft/SEAL/archive/refs/tags/v3.7.3.tar.gz | tar xzf -
+#cd SEAL-3.7.3
+#cmake -S . -B build -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=~/github/workspace
+#cmake --build build
+#cmake --install build
+#cd ..
+
+./build.sh
+cd ../../..
+#cd ~/github/workspace
+#cp -r src/main/cpp/lib/ lib/
+cp -r /usr/local/include/SEAL-3.7/seal /usr/local/include/seal
+cp /usr/local/lib/libseal.so.3.7 /usr/local/include/libseal.so.3.7
+cp /usr/local/lib/libseal.so.3.7 /usr/src/libseal.so.3.7
 
 export MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=128m"
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -22,10 +22,9 @@
 
 # A script to execute the tests inside the docker container.
 
-cd /github/workspace
-cd src/main/cpp
+cd /github/workspace/src/main/cpp
 ./build.sh
-cd ../../..
+cd /github/workspace
 
 export MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=128m"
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/sysds/test/functions/federated/paramserv/EncryptedFederatedParamservTest.java
```diff
@@ -73,7 +72,7 @@ public static Collection<Object[]> parameters() {
 				//{"TwoNN", 	5, 1000, 100, 1, 0.01, 	"BSP", "BATCH", "KEEP_DATA_ON_WORKER", 	"NONE",			"true",	"BALANCED",		200},
 				{"TwoNN",	2, 4, 1, 4, 0.01, 		"SBP", "BATCH", "KEEP_DATA_ON_WORKER", 	"BASELINE",		"false",	"IMBALANCED",	200},
 				{"TwoNN",	2, 4, 1, 4, 0.01, 		"SBP", "BATCH", "KEEP_DATA_ON_WORKER", 	"BASELINE",		"false",	"BALANCED",		200},
-				{"CNN",		2, 4, 1, 4, 0.01, 		"SBP", "EPOCH", "SHUFFLE",			 	"BASELINE",		"false",	"BALANCED",		200},
+				//{"CNN",		2, 4, 1, 4, 0.01, 		"SBP", "EPOCH", "SHUFFLE",			 	"BASELINE",		"false",	"BALANCED",		200},
```
Baunsgaard, line 75: i assume the CNN did not work?

docker/entrypoint.sh
```diff
@@ -22,6 +22,8 @@
 
 # A script to execute the tests inside the docker container.
 
+cd /github/workspace/src/main/cpp
+./build.sh
```
Baunsgaard, line 26: i do not like this, since it would mean we have to build our c++ part every test.

do you know how much the overhead is?

Baunsgaard (commented):
Hi Wrede!,

Looks fine to me, but i do not like we would build the c++ part every time, but on the other hand one could argue that we should have done that from the beginning every time.

Can we perhaps make a different docker file that contains the homomorphic encryption and not the R libs maybe, and then build inside this GitHub action?

src/test/java/org/apache/sysds/test/functions/federated/paramserv/EncryptedFederatedParamservTest.java
sebwrede: It cannot shuffle when it is encrypted. The CNN works.

sebwrede:
> Can we perhaps make a different docker file that contains the homomorphic encryption and not the R libs maybe, and then build inside this GitHub action?

How can we make sure that we don't need the R libs for the HE tests? Currently it may be true, but what if we add some other tests later on where both HE and R libs are needed?
Additionally, I think it would be a bit confusing if you need two different Docker images to run all tests instead of having a single Docker image to run all tests. 
Or did you mean that we should then include the HE dockerfile in the existing test dockerfile so that we still build a single image, but just have the configurations in two separate files?

Baunsgaard:
Lgtm. 

Do you know how to update the docker image ?

sebwrede:
> Do you know how to update the docker image ?

Not really. Do I even have permissions to update it?

Baunsgaard:
> > Do you know how to update the docker image ?
> 
> Not really. Do I even have permissions to update it?

as a committer you have the ability to build the images:
Go to actions
image

Then Docker release deployment action (for normal docker image)
image
Or For test image use Docker Test Image Update.


Then run workflow.
image

Note that currently you have to have the code on master for tit to work.

PR #1723, opened 2022-11-08
[MINOR] Reduce package time by excluding generated content

Baunsgaard:
This commit change our files to include in bin and src releases. Locally on my machine it started taking upwards of 5-8 minutes to do 'mvn clean package -P distribution', because we were packing libraries, datasets, and resources that was not needed for the release archives.

systemds.zip after 31.7 MB before: 664.6 MB
systemds-bin.zip after 75.2 MB before: 396.5 MB

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -95,7 +95,11 @@ jobs:
         architecture: 'x64'
   
     - name: Install pip Dependencies
-      run: pip install numpy py4j wheel scipy sklearn requests pandas unittest-parallel
+      run: |
+        # Install pip twice to update past the versions.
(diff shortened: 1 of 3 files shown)


Baunsgaard:
> > Locally on my machine it started taking upwards of 5-8 minutes to do 'mvn clean package -P distribution', because we were packing libraries, datasets, and resources that was not needed for the release archives.
> 
> Since when did it take that long?

It is because i build python and run those tests they generate files that are not ignored in our build.
Also the new federated monitoring tool, it was not ignored.

sebwrede:
> how long does it take for you ?

It takes 1 minute and 4 seconds (before your latest commit).

PR #1726, opened 2022-11-11
[MINOR] more sleep to added monitoring tests

Baunsgaard:
Debug PR to fix failing monitoring tests, 
The test fails because the monitoring tool does not startup with the sleep specified.

### DIFF
--- a/src/test/java/org/apache/sysds/test/AutomatedTestBase.java
+++ b/src/test/java/org/apache/sysds/test/AutomatedTestBase.java
@@ -123,7 +123,7 @@ public abstract class AutomatedTestBase {
 
 	// ms wait time 
 	public static final int FED_WORKER_WAIT = 1000; 
-	public static final int FED_MONITOR_WAIT = 5000; 
+	public static final int FED_MONITOR_WAIT = 10000; 
 	public static final int FED_WORKER_WAIT_S = 50;
 	
 
--- a/src/test/java/org/apache/sysds/test/functions/federated/monitoring/FederatedCoordinatorIntegrationCRUDTest.java
+++ b/src/test/java/org/apache/sysds/test/functions/federated/monitoring/FederatedCoordinatorIntegrationCRUDTest.java
@@ -19,6 +19,8 @@
 
 package org.apache.sysds.test.functions.federated.monitoring;
 
+import static org.junit.Assert.fail;
+
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.HttpStatus;
 import org.apache.sysds.runtime.controlprogram.federated.monitoring.models.CoordinatorModel;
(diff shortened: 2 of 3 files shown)


Baunsgaard:
In general a better design for many of the tests starting up a service would be to have callback methods.
But this require us to setup such a thing where on startup our services would send messages, and it would add code to the system that is only valid in test cases. 
Any opinions from anyone?

BACtaki:
> But this require us to setup such a thing where on startup our services would send messages, and it would add code to the system that is only valid in test cases.

I could see this being useful beyond tests in the future. For example, when we add a SystemDS control plane eventually, it would need to show that all services are up and running.

Baunsgaard:
> > But this require us to setup such a thing where on startup our services would send messages, and it would add code to the system that is only valid in test cases.
> 
> I could see this being useful beyond tests in the future. For example, when we add a SystemDS control plane eventually, it would need to show that all services are up and running.
> 
> Speaking of services, what other services do we have besides the federated monitoring service? Do we have plans to add any in the near future?

Currently we only have the federated monitoring tool, 
we could potentially add features to this to make it more general purpose, for the main system as well.
I think it would be best if we consolidate such monitoring / control interfaces in the same tool.
It could potentially be a very useful feature similar to the SparkUI.

BACtaki:
> Currently we only have the federated monitoring tool,
> we could potentially add features to this to make it more general purpose, for the main system as well.
> I think it would be best if we consolidate such monitoring / control interfaces in the same tool.
> It could potentially be a very useful feature similar to the SparkUI.

It makes sense to want to "unify" all monitoring in a single monitoring service. It would be amazing to have something similar to Spark/Flink UI eventually.

Re your original point, we have only 2 options:
1. Synchronous calls to a `/health` endpoint (polling)
2. Async call using something like `java.util.concurrent.Future` (callback)

In either case, we will need a timeout. After all, we don't want to wait indefinitely for the service to be up and running.

[1] is easier to implement and maintain IMO. [2] is equally valid though, and I think it really depends on whether somebody wants to implement it. In the latter case, it might make sense to create a JIRA story and tag it appropriately so people can discover it easily and work on it.

Baunsgaard:
> > Currently we only have the federated monitoring tool,
> > we could potentially add features to this to make it more general purpose, for the main system as well.
> > I think it would be best if we consolidate such monitoring / control interfaces in the same tool.
> > It could potentially be a very useful feature similar to the SparkUI.
> 
> It makes sense to want to "unify" all monitoring in a single monitoring service. It would be amazing to have something similar to Spark/Flink UI eventually.
> 
> Re your original point, we have only 2 options:
> 
>     1. Synchronous calls to a `/health` endpoint (polling)
> 
>     2. Async call using something like `java.util.concurrent.Future` (callback)
> 
> 
> In either case, we will need a timeout. After all, we don't want to wait indefinitely for the service to be up and running.
> 
> [1] is easier to implement and maintain IMO. [2] is equally valid though, and I think it really depends on whether somebody wants to implement it. In the latter case, it might make sense to create a JIRA story and tag it appropriately so people can discover it easily and work on it.

I am in general more a fan of the callback strategy, since this would allow the started processes to call, once ready rather than us repeatedly asking "are you ready" with polling :P.

I guess an easy design now that i think about it is to add a flag to make SystemDS send a HTTP message to an address once it is up. Then we would only need to make a REST API on the testing side that listens for ready workers, coordinators, monitoring services, etc.

All that said, i am about to close this PR, and i like the discussion so we should perhaps move it to the dev mailing list.

PR #1732, opened 2022-11-16
[SYSTEMDS-3272] applySchema.dml() builtin 

Shafaq-Siddiqi:
which will apply the given schema to a existing frame

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/cp/BinaryFrameFrameCPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/cp/BinaryFrameFrameCPInstruction.java
@@ -19,10 +19,13 @@
 
 package org.apache.sysds.runtime.instructions.cp;
 
+import org.apache.sysds.common.Types;
 import org.apache.sysds.runtime.controlprogram.context.ExecutionContext;
 import org.apache.sysds.runtime.frame.data.FrameBlock;
 import org.apache.sysds.runtime.matrix.operators.BinaryOperator;
 import org.apache.sysds.runtime.matrix.operators.Operator;
+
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/sysds/runtime/instructions/cp/BinaryFrameFrameCPInstruction.java
```diff
@@ -55,6 +58,16 @@ else if(getOpcode().equals("freplicate")) {
...
 			ec.setFrameOutput(output.getName(), retBlock);
 		}
+		else if(getOpcode().equals("applySchema")) {
+			// apply frame schema from DML
+			Types.ValueType[] schema = new Types.ValueType[inBlock2.getNumColumns()];
+			for(int i=0; i<inBlock2.getNumColumns(); i++)
+				schema[i] = Types.ValueType.fromExternalString(inBlock2.get(0, i).toString());
+			FrameBlock out = new FrameBlock(schema);
+			out.copy(inBlock1);
+			out.setSchema(schema);
```
Baunsgaard, line 68: I would have liked it to apply it to the data inside the columns, not first copying and then setting the schema.
this implementation just modify metadata you set, not reallocating the columns with the right data.

But on the other hand this is a great start, where we can easily add the functionality inside Frame Block and Arrays.

Do you want to do it ? or should i add that part?

Baunsgaard (commented):
Close but not quite there yet.

Baunsgaard:
Hi @Shafaq-Siddiqi ,
I am taking this and add the extension from the comments since i need it for further work.
Thanks for the help!

Baunsgaard:
Closing, see: #1737

PR #1733, opened 2022-11-16
[SYSTEMDS-3466] Asynchronous (Future-based) execution of Spark instructions

phaniarnab:
This patch introduces a future-based asynchronous execution of Spark actions. We wrap the matrix block with a future, create a matrix object handle, and maintain that in the symbol table. This extension allows triggering a chain of Spark instructions asynchronously and seeking the results only when needed.

TODO: Account the memory required for the future results, maintain the lineage of the broadcast variables to avoid premature removal.


BACtaki:
> This extension allows triggering a chain of Spark instructions asynchronously and seeking the results only when needed.

This is really interesting @phaniarnab . Curious: is there a concrete use case for this feature?

PR #1735, opened 2022-11-23
[SYSTEMDS-3468] Adding the outline for the survey of builtin functions

BACtaki:
This patch proposes a comparison of builtin functions based on a grouping by ML lifecycle phase and ML task. This patch presents only the basic outline- subsequent patches will do the actual comparison.


j143:
Thanks Badrul! Although this is a good idea, try to be familiar with the end to end pipelines or a single phase of the cycle with SystemDS. A plenty of working examples are here:
https://github.com/damslab/reproducibility

Keep posting any updates here though!

Baunsgaard:
what i think would be really useful in this direction.

- run experiments that evaluate performance over time of key algorithms and operations
- automatically plot these, and provide graphs in GitHub actions.

This can also compare to other systems  implementations automatically, to include in the graphs.
Optionally it can be an action we run weekly, or monthly?

PR #1738, opened 2022-11-25
[MINOR] Adding a factory method for MatrixSketch

BACtaki:
This patch introduces a factory method for sketches. This will centralize the creation of all sketches in one place and prevent duplication of operator switching and validation logic.


Baunsgaard:
LGTM

PR #1740, opened 2022-11-28
[SYSTEMDS-3463] Add unique() built-in function

BACtaki:
This patch converts the existing unique() function from a script to a built-in. The script-based approach is based on sorting, which is very expensive computationally, especially for large multiblock inputs. The new approach, on the other hand, is based on a new data sketch for the unique() function. This first patch creates the framework for the new unique sketch and implements the CP RowCol case; other cases - CP Row/Col and Spark RowCol/Row/Col - will be implemented in subsequent patches.

Tests:
    [X] Unit tests
    [ ] Integration tests
    [ ] N/A


BACtaki:
I will post a comparison of SystemDS' new built-in `unique()` and R's `unique()` functions here shortly

BACtaki:
I have added the following API design notes to the JIRA:

1. Row aggregation: remove duplicate rows

    * R: unique() removes duplicate rows, eg

    ```R
        > df <- matrix(rep(1:6,length.out=9),nrow = 3,ncol=3,byrow = T)
        > df
            [,1] [,2] [,3]
        [1,]    1    2    3
        [2,]    4    5    6
        [3,]    1    2    3
        > unique(df)
            [,1] [,2] [,3]
        [1,]    1    2    3
        [2,]    4    5    6
    ```

    * SystemDS: the same can be achieved using the unique() sketch like so:

        unique(X, dir="r")

2. Col aggregation: remove duplicate cols

    * R: unique() removes duplicate rows, so we can obtain the desired result using transpose, like so:

    ```R
        > df <- matrix(rep(1:6,length.out=9),nrow = 3,ncol=3)
        > df
            [,1] [,2] [,3]
        [1,]    1    4    1
        [2,]    2    5    2
        [3,]    3    6    3
        > df_t
            [,1] [,2] [,3]
        [1,]    1    2    3
        [2,]    4    5    6
        [3,]    1    2    3
        > df_t_ = unique(df_t)
        > df_t_
            [,1] [,2] [,3]
        [1,]    1    2    3
        [2,]    4    5    6
        > df_t_t = t(df_t_)
        > df_t_t
            [,1] [,2]
        [1,]    1    4
        [2,]    2    5
        [3,]    3    6
    ```

    * SystemDS: the same can be achieved using the unique() sketch like so:

        unique(X, dir="c")

3. RowCol aggregation: return only unique values in given matrix

    * SystemDS

        X = [[1, 1], [2, 2], [3, 3]]
        unique(X) will return X' = [[1], [2], [3]]
    
    * R

    This is similar to how unique() operates on vectors in R:

    ```R
        > df <- c(1, 1, 2, 2, 3, 3)
        > df
        [1] 1 1 2 2 3 3
        > unique(df)
        [1] 1 2 3
    ```

    The difference is that SystemDS' unique() will support the same for not only vectors, but also matrices.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixCountDistinct.java
```diff
@@ -19,25 +19,24 @@
...
-import org.apache.sysds.runtime.matrix.data.sketch.CountDistinctSketch;
 import org.apache.sysds.runtime.matrix.data.sketch.countdistinct.CountDistinctFunctionSketch;
 import org.apache.sysds.runtime.matrix.data.sketch.countdistinctapprox.KMVSketch;
 import org.apache.sysds.runtime.matrix.operators.CountDistinctOperator;
 import org.apache.sysds.runtime.matrix.operators.CountDistinctOperatorTypes;
 import org.apache.sysds.utils.Hash.HashType;
 
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
```
Baunsgaard, line 37: I am going to merge it, but please in the future, try not to change the import order.

Baunsgaard (commented):
Thanks for the upgrade and change to an internal unique function.
I did not run a comparison, but the code clearly indicate that it should be faster.

PR #1749, opened 2022-12-10
[MINOR] Create matrix sketches exclusively using MatrixSketchFactory

BACtaki:
This patch ensures that all matrix sketches are instantiated via MatrixSketchFactory


BACtaki:
Test failures are transient:
image
image

Need your help to re-trigger the tests please @Baunsgaard @phaniarnab

Baunsgaard:
Our Builtin1 tests, unfortunately sometimes run out of memory, resulting in the crash and failing test that you observed here.
We need to fix it by splitting Builtin1 tests, or reduce the resource requirement for the tests run in builtin 1.

PR #1750, opened 2022-12-12
[SYSTEMDS-3478] Bitset array for Frames

Baunsgaard:


Baunsgaard:
This pr improve performance and allocation of frames  in memory and on disk reduced size of 8x of boolean values.
 
on a 64kx2k frame with booleans:

- write time improved from 0.3-0.6 to 0.162
- read time improved from 0.4 to 0.119
- size on disk and memory from 128MB to 16MB

This commit is done to provide a lower bound case for dense boolean allocation.

PR #1756, opened 2022-12-21
[SYSTEMDS-3479] Mark Spark instructions to persist and locally cache

phaniarnab:
This patch adds the compiler flags and runtime support to checkpoint any Spark instruction which is marked for caching. During postprocessing of a marked instruction, we first inplace persist the RDD and then store the RDD in the local Lineage cache for reuse. This patch also fixes a bug in the last commit which was unpersisting the locally cached RDDs during rmvar. Future commits will add rewrites to mark the Spark instructions for caching in a cost-based manner.
Hyperparameter tuning of LmDS with 2.5k columns improves by 22x by caching the cpmm results in the executors.


Baunsgaard:
22x wau!

phaniarnab:
> 22x wau!

Yeah. Caching the shuffle-based Spark instruction results in the executors can generate huge benefits.

PR #1803, opened 2023-04-07
[SYSTEMDS-2102] Vectorize gradients for Factorization Machines function

j143:
image

### DIFF
--- a/scripts/nn/layers/fm.dml
+++ b/scripts/nn/layers/fm.dml
@@ -87,14 +87,19 @@ backward = function(matrix[double] dout, matrix[double] X, matrix[double] w0, ma
 
 
   # Second term -> V(i,f) * (X(i))^2
-  Xt = t( X^2 ) %*% dout  # shape (d,1)
+  # Xt = t( X^2 ) %*% dout  # shape (d,1)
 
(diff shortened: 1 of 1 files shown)


j143:
the code doesn't seem to be correct:

-- old g_V2 --
0.000 0.000
0.000 -0.001
0.000 -0.000
0.000 -0.002
0.000 0.000
0.000 -0.001
0.000 -0.014

-- g_V2 --
0.000 -0.017

Baunsgaard:
LGTM

but remove the commented code.

j143:
Hi @Baunsgaard - thanks for review.

I believe the vectorization I made is incorrect, especially output dimensions.
I will keep this in draft for a while.

Baunsgaard:
> Hi @Baunsgaard - thanks for review.
> 
> I believe the vectorization I made is incorrect, especially output dimensions. I will keep this in draft for a while.

aww sad, it looked good at a surface glance, maybe we do not have enough tests ? 
or is it that the NN is only in the Application tests again.

j143:
We verify the with gradients at the moment. but, we need to verify against R implementation once.

PR #1806, opened 2023-04-21
[SYSTEMDS-3428] Code coverage in GitHub Actions (Java)

kykrueger:
Add jacoco code coverage to java tests in github actions.

# Summary
As a user and developer in the SystemDS project, I want to have metrics to track the health of the project. Test coverage is one helpful indicator to track with the development of new features to ensure that coverage does not drop over time. The project uses github-actions for testing PRs and this is a helpful place to review the health metrics. Adding test-coverage for java components is non-trivial for the github actions, since the java tests have been parallelized over many executors in github-workflows, and the results need to be aggregated to generate a report or evaluation.

It was necessary to merge the existing java-test workflows into one in order to cleanly use the github-actions API within a workflow template for sharing files between jobs. It may have been possible to use the REST-API, but that would have increased the complexity of the implementation and been bad for maintainability. 

It may be possible, that the limits of jobs within one workflow are more restrictive, and before merging the documentation should be reviewed and cited.

This PR is not yet complete, and the issue will only be marked as closed in a later comment.


kykrueger:
@Baunsgaard I've added the directory of the jacoco report as an artifact at the end of the java test workflow. The Job cleans up all of the artifacts from the steps in between to avoid billing problems with the number of artifacts. 

I'd suggest we just keep the csv from the export to keep the artifact size down, but now we need to agree upon the metric for a test coverage badge. What would you think of just having one with the lines covered in percentage? 

It should also be possible to add some regression checks, by comparing the percentage covered in the branch being merged with the amount covered in main and failing the PR if the percentage drops. Perhaps that would be a good separate ticket though, I don't want to blow up the scope of this one.

kykrueger:
closes [SYSTEMDS-3428]

Still some work to be done on the badge, minimizing the size of the resulting code coverage report and maybe cleaning up the job names in the github-actions matrix.

Baunsgaard:
Sounds good. Will take a look later today.

Baunsgaard (commented):
Just looking at the code it seems to be all fine and i really like the contribution.

But we do have one issue

https://github.com/apache/systemds/actions/runs/4769263729

The test did not run, at least not in the last instance.
The previous instance did run some subset: https://github.com/apache/systemds/pull/1806/commits/c2a9bbe94bba882849fddded5c02583d6059c74c

Maybe there is a typo somewhere? typically if i had this error it is because of some indentation that was incorrect in the workflows file.

kykrueger:
@Baunsgaard thanks for the feedback. Its quite strange, the workflow runs in my fork, could you trigger a rerun with logging?

I've not been able to find any whitespace out of place.

kykrueger:
In my first attempt there was a failure that seemed to be due to something going wrong in githubs network (https://github.com/kykrueger/systemds/actions/runs/4769263418/jobs/8479892879), but the second ran through: https://github.com/kykrueger/systemds/actions/runs/4769263418/attempts/2

kykrueger:
A list of some things I'd like to adjust in either this ticket or a follow up:

1. Using ubuntu-latest could break reproducibility when new versions of ubuntu are added. I'd replace all calls to ubuntu-latest with ubuntu-22.04.
2. The applications-tests do not use setup-java and therefore use the default java installation of ubuntu-22.04, so if not replacing ubuntu-latest with a fixed version, I'd at least want to add a fixed version with setup java, but maybe even pin both a version for java and ubuntu.
3. Maybe also combine the functions, components and applications matricies into a shared matrix in order to avoid their matrix test groups from diverging, unless their matricies should remain different? I don't know enough about the project to say if this step would make sense.
4. Retention period for artifacts, the uploaded test-coverage reports may be saved for up to 400 days as artifacts. I don't think they have a ton of value to keep around that long since they can also be regenerated. I'd consider setting the retention long enough to make sure it is around at least a bit longer than it takes to get something merged, but I don't know how long this is for this project.

kykrueger:
It turns out that adding a custom status badge to the readme is a bit annoying since it needs to be saved somewhere, we have a few options:
1. Commit the status badge after the tests are successful. This creates some issues around the nature of open source projects, where you will have lots of branches coming from outside and the contributors will have to give the workflow permissions to commit to their branch. You will also need to give the workflows permission to commit inside of the github project. This also opens up a bit of a vulnerability for unintentional changes + a noisy commit history unless you switch to squash merging.
2. Host the status badge in some sort of cloud storage, and only let your workflows update it. Someone will need to provide an object storage account and add a secret in your github workflows, I can't do that myself from outside the project. This would also only work to display the most recent badge of main (could be misleading for forks). Someone had a neat solution of using gists to do this for free: https://discuss.circleci.com/t/how-to-show-badge-from-artifacts-in-github-readme/42109/2
3. I looked at storing the badges as artifacts, but there does not seem to be a static URL that we can use to fetch the latest badge from the main branch.

I'd suggest going with option 2 despite the misleading badge in forks

Baunsgaard:
> 1. Using ubuntu-latest could break reproducibility when new versions of ubuntu are added. I'd replace all calls to ubuntu-latest with ubuntu-22.04.

Sure we can make this change.
Personally i kind of like that we automatically keep up to date tho, but running both latest and 22.04 for all tests would be to much. If we can design it in such a way that all tests run on 22.04 and a subset (maybe applications only) run on latest then that would be cool, but only run the latest version if latest is different than 22.04.

In general that all seems like to much work compared to just having everything run on latest, also worth noting is that latest only update once every 2 years. so not a big concern in my opinion.

> 2. The applications-tests do not use setup-java and therefore use the default java installation of ubuntu-22.04, so if not replacing ubuntu-latest with a fixed version, I'd at least want to add a fixed version with setup java, but maybe even pin both a version for java and ubuntu.

This sounds like a bug on my side, feel free to make a minor PR for it.

> 3. Maybe also combine the functions, components and applications matricies into a shared matrix in order to avoid their matrix test groups from diverging, unless their matricies should remain different? I don't know enough about the project to say if this step would make sense.

The reason they were split was to give nice names on the UI when the tests were run to clearly indicate what tests were executed. the component.a** is not really quickly understandable.
Maybe if you want to combine with point 2 as a minor thing, and combine all in a single workflow to reduce the code duplication but we still want some way if clearly understanding what test is run in each workflow.


 >4. Retention period for artifacts, the uploaded test-coverage reports may be saved for up to 400 days as artifacts. I don't think they have a ton of value to keep around that long since they can also be regenerated. I'd consider setting the retention long enough to make sure it is around at least a bit longer than it takes to get something merged, but I don't know how long this is for this project.

I think you did a fine medium of 30 days, my only concern is that the artifact is 54.8 MB perhaps there is a way to compress this?
or/and set a dynamic retension of only last 30 days on main branch and 7 on PRs? 
This concerns me because it can grow very quickly with a few commits filling way to much.

> It turns out that adding a custom status badge to the readme is a bit annoying since it needs to be saved somewhere, we have a few options:

I think your option 3 is the one we want. there should be a way of doing this, since all the other batches are based on this as well. But this would be a followup project.

Finally a question.

How does it look, are we having a bad code coverage? --- i assume that it is fairly low since we did not keep it in mind the last few years.

kykrueger:
Looks like the workflow is failing because of a github action I was calling to delete all of the artifacts no longer needed at the end of the job. I'll write something myself in the workflow instead of calling that project. 

Anyhow, to your points:
> Sure we can make this change.
Personally i kind of like that we automatically keep up to date tho, but running both latest and 22.04 for all tests would be to much. If we can design it in such a way that all tests run on 22.04 and a subset (maybe applications only) run on latest then that would be cool, but only run the latest version if latest is different than 22.04.
> In general that all seems like to much work compared to just having everything run on latest, also worth noting is that latest only update once every 2 years. so not a big concern in my opinion.

Fair enough, it is just a preference, we can leave it on latest. It will just be important to keep that in mind as a possible source of test-failure if anything funny happens. 



 >> The applications-tests do not use setup-java and therefore use the default java installation of ubuntu-22.04, so if not replacing ubuntu-latest with a fixed version, I'd at least want to add a fixed version with setup java, but maybe even pin both a version for java and ubuntu.

> This sounds like a bug on my side, feel free to make a minor PR for it.

Will do

> The reason they were split was to give nice names on the UI when the tests were run to clearly indicate what tests were executed. the component.a** is not really quickly understandable.
Maybe if you want to combine with point 2 as a minor thing, and combine all in a single workflow to reduce the code duplication but we still want some way if clearly understanding what test is run in each workflow.

I'll see if I can come up with a way to clean up the names. If not I'll leave the test-groups separated as they were. 

> I think you did a fine medium of 30 days, my only concern is that the artifact is 54.8 MB perhaps there is a way to compress this?
or/and set a dynamic retension of only last 30 days on main branch and 7 on PRs?
This concerns me because it can grow very quickly with a few commits filling way to much.

I'm not sure about the billing for the artifacts and their size, but github actually already compresses the directory to a zip. This reduces the size to around 12.5 MB, and I think it would be cheeky for them to charge for the decompressed file size, but I can check their docs for more info. If there is no info about that in the docs and you are keen, we could pre-compress the directory to be sure. Otherwise, we we could reduce the size to under 1 MB (before compression) by just keeping the CSV with all the summarized results.

As for the test coverage, you can see it in my fork for now until I remove that action call. 
https://github.com/kykrueger/systemds/actions/runs/4769263418
The line coverage has 193,435 of 794,773 instructions missed (75% coverage)

Baunsgaard:
As far as I know they do not charge Apache. But i might be mistaken and honestly i do not know the deal made with Microsoft. 
I would not worry about any charging ATM. But being considerate is a good idea. 

If it is easy to switch to CSV we could make it CSV on PR and full on main ? 

Also it sounds good with code coverage and actually getting a number maybe @mboehm7 @phaniarnab are interested 😊.  (See last comment bottom)

phaniarnab:
Good work. For a system of this scale, I'd say 85% coverage would be a good aggregate (vary for the individual features. Require more coverage for the general code paths). For example, GPU code paths are not reachable by the unit tests. 75% is still quite okay. To get a better insight, we need to run code coverage for the individual features.

Baunsgaard:
It works!

thanks, will merge in shortly.

Baunsgaard:
Also as a small follow up we can merge the component application and function tests... since each of these no longer have much of a meaning because they all share the same base name now.

image

Baunsgaard:
and 

Unify Application Function and Component Test GitHub Action
https://issues.apache.org/jira/browse/SYSTEMDS-3531

kykrueger:
@Baunsgaard, sorry I wasn't ready to merge this PR because of the artifacts that hang around in addition to the final report. I'll wrap clean those up in the follow up issues you mentioned!

PR #1807, opened 2023-04-21
[MINOR] Cleanup and document PCA and Scale

Baunsgaard:
This commit adds documentation for the builtins PCA and Scale.

For Scale we add a NaN handling on internally calculated vectors, to allow the Scale builtin to work on inputs containing NaN, by essentially not scaling or shifting columns containing NaN.

PCA now include some code examples in the documentation and minor modifications such as a new argument onlyClusters that allow us to call PCA without having to apply the calculated eigenvectors to the input matrix. This modification is done because other instances i want to compare with only calculate PCA while not applying it. Also added is some minor comments inside PCA and removal of a matrix multiplication that is never used (and so far always compiled out as dead code when executing PCA)

### DIFF
--- a/scripts/builtin/pca.dml
+++ b/scripts/builtin/pca.dml
@@ -64,7 +64,9 @@
 # ScaleFactor  The Scaling of the values, to make each dimension same size.
 # -------------------------------------------------------------------------------------------------
 
-m_pca = function(Matrix[Double] X, Integer K=2, Boolean center=TRUE, Boolean scale=TRUE, Boolean onlyClusters= FALSE)
+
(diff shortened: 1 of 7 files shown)


Baunsgaard:
@phaniarnab 
I am down to 1 error. this is in the async things that you made.
Can you take a look?

phaniarnab:
> @phaniarnab I am down to 1 error. this is in the async things that you made. Can you take a look?

I reproduced and found an issue in my code. Happening due to an unhandled special case where the asynchronously collected matrix is larger than the buffer pool caching threshold.
For now, please add an `@Ignore` tag for the failing test `testPCAlm` in MaxParallelizeOrderTest. I will fix the issue in future commits.
Thanks for exposing the bug. 💯 
@Baunsgaard

Baunsgaard:
Great sounds good @phaniarnab

PR #1820, opened 2023-05-07
[SYSTEMDS-3528] Refine allocation of GitHub Artifacts

kykrueger:


kykrueger:
Open suggestions from the last PR:
https://github.com/apache/systemds/pull/1806

The two following suggestions are not a part SYSTEMDS-3528
https://github.com/apache/systemds/pull/1806#issuecomment-1519409025
https://github.com/apache/systemds/pull/1806#issuecomment-1521721841

@Baunsgaard, are you still keen on them?

Baunsgaard:
> Open suggestion from the last PR: #1806
> 
> The following suggestion is not a part [SYSTEMDS-3528](https://issues.apache.org/jira/browse/SYSTEMDS-3528) [#1806 (comment)](https://github.com/apache/systemds/pull/1806#issuecomment-1519409025)
> 
> @Baunsgaard, are you still keen on it?

Lets try to first clean up the artifacts as your PR suggests, then we can consider the CSV idea again afterwards.

kykrueger:
That answers my question, I'm fine with leaving the CSV change for another PR, just didn't want the request to get left behind.

kykrueger:
I need to debug my deletion script,  it should be quick to fix but I need to find the time to do it. Worst case I'll have some time this weekend.

kykrueger:
Not being able to delete the artifacts is related to this issue with the github rest API:
https://github.com/actions/upload-artifact/issues/53

Baunsgaard:
> Not being able to delete the artifacts is related to this issue with the github rest API: [actions/upload-artifact#53](https://github.com/actions/upload-artifact/issues/53)

I see 2 options.

1. We pause this PR untill some solution is found on the upload-artifacts action. I see this the same as abandoning this, since they did not resolve it in the last ... 2 years?
2. We set the retention period differently for the test artifacts such that each action only store its individual statistics for ... lets say 3 hours... and then we only keep around the full aggregated test statistics for longer.

kykrueger:
@Baunsgaard the retention can only be set through the actions in increments of days as far as I can tell. There is a third party github action I had been using initially that managed to delete the artifacts during the workflow run, and looking at the developers code, he used an undocumented javascript marked as a preview version of the github API to pull it off.

So, a few additional options. 

1. We could copy his approach if licensing allows, but it may still be a cause for security concerns. 

1. We can configure the repo to allow using his github action directly, so far the SystemDS repo is configured to disallow all third party actions, but it is possible to set up a whitelist.

1. Otherwise, I could also look at triggering an additional workflow after this one completes which deletes the artifacts, but I think that could increase the complexity of the task a bit more than I'd like in comparison to how simple it ought to be.

kykrueger:
This is the action I had used inititally, there are a few others, but I have not tested them to see if they also allow deleting artifacts within the same workflow. If there is a security review process that the team might have, I'd like to know if we could consider allowing this action.

https://github.com/marketplace/actions/delete-artifact

Baunsgaard:
i like the idea 3. of another workflow to run afterwards that delete the extra files.
1., is a no go, we do not copy from elsewhere.
 2. is questionable, i would prefer if we did not have to change the settings of the repository.

alternatively then we just set the retention period of 1 day and let that be it, i can live with this as well and only keep the full aggregated workflow around for a longer period..

kykrueger:
@Baunsgaard 
I just got back around to this today. Triggering a workflow upon completion of another works.
The deletion of the artifacts is successfull without changing any repo-permissions since the github-token of apache/systemds is already set to permissive. If the team later decides to tighten the security of the authentication token used in github-actions, the action will require a dedicated token in the repository secrets.

Testing the workflow in the main repository will only be possible once a merge is complete, this is because workflow_run triggers may only be started if the workflow is already a part of main.

I've tested it in my fork, and am cleaning up some some tests.

kykrueger:
On another note, I've only set up the deletion workflow to run if the triggering javaTests were a success. This is because some of the tests in require retries, and if you were to retry the tests after deleting the artifacts, the code-coverage report generated would be missing all of the prior successful tests. 
I'd like to suggest automating the retries of the common test failures in the java workflow. However, this comes with its own issues since tests real negative test results will also then be retried.

Baunsgaard:
> On another note, I've only set up the deletion workflow to run if the triggering javaTests were a success. This is because some of the tests in require retries, and if you were to retry the tests after deleting the artifacts, the code-coverage report generated would be missing all of the prior successful tests. I'd like to suggest automating the retries of the common test failures in the java workflow. However, this comes with its own issues since tests real negative test results will also then be retried.

(Unrelated to the PR)
Ideally we need to change our testing methodology for the failing tests, currently the ones that produce errors most commonly are because we do not do callback based testing, rather we wait a specified time frame, and that makes it sometimes fail when running in parallel with other tests.

Baunsgaard:
> @Baunsgaard I just got back around to this today. Triggering a workflow upon completion of another works. The deletion of the artifacts is successfull without changing any repo-permissions since the github-token of apache/systemds is already set to permissive. If the team later decides to tighten the security of the authentication token used in github-actions, the action will require a dedicated token in the repository secrets.
> 
> Testing the workflow in the main repository will only be possible once a merge is complete, this is because workflow_run triggers may only be started if the workflow is already a part of main.
> 
> I've tested it in my fork, and am cleaning up some some tests.

LGTM, will merge shortly.

.github/workflows/javaTests.yml
```diff
@@ -50,34 +50,10 @@ jobs:
...
-          "**.functions.dnn.**",
-          "**.functions.paramserv.**",
-          "**.functions.recompile.**,**.functions.misc.**,**.functions.mlcontext.**",
-          "**.functions.nary.**,**.functions.quaternary.**",
-          "**.functions.parfor.**",
-          "**.functions.pipelines.**,**.functions.privacy.**",
-          "**.functions.homomorphicEncryption.**",
-          "**.functions.unary.scalar.**,**.functions.updateinplace.**,**.functions.vect.**",
-          "**.functions.reorg.**,**.functions.rewrite.**,**.functions.ternary.**,**.functions.transform.**",
-          "**.functions.unary.matrix.**,**.functions.linearization.**,**.functions.jmlc.**"
```
Baunsgaard, line 80: we just need to reintroduce these in the PR.

kykrueger: Sure thing, I have a handful of changes not yet in this branch since I had to test this workflow in main. I'll update here once the changes are in.

kykrueger: I reverted the commit responsible for the minimal testset. 
Now all tests are enabled

Baunsgaard (approved):
LGTM, i like it retains different amounts of time depending on PR, Main or fork.

Baunsgaard:
Merged, Thanks for the contribution.

Baunsgaard:
image

Seems like it did not work :(

kykrueger:
@Baunsgaard sorry, havn't been on the project for a while. 

Looks like I broke it with a bad assignment of the name
https://github.com/apache/systemds/pull/1820/commits/edce86de35f17431625b1357a62ca841a76f23f8

Didn't catch it because changes to the workflow in a branch are not visible in its own PR. Looks like it kept running the version in main of my repo.

PR #1821, opened 2023-05-07
[SYSTEMDS-3531] merge application and component tests

kykrueger:
closes [SYSTEMDS-3531](https://issues.apache.org/jira/browse/SYSTEMDS-3531)

### DIFF
--- a/.github/workflows/javaTests.yml
+++ b/.github/workflows/javaTests.yml
@@ -19,7 +19,7 @@
 #
 #-------------------------------------------------------------
 
-name: Java Testing Workflow
+name: Java Test
 
 on:
   push:
@@ -81,7 +81,7 @@ jobs:
           "**.functions.reorg.**,**.functions.rewrite.**,**.functions.ternary.**,**.functions.transform.**",
           "**.functions.unary.matrix.**,**.functions.linearization.**,**.functions.jmlc.**"
         ]
-    name: ${{ matrix.tests }}
+    name: ${{ matrix.os }} Java ${{ matrix.java }} ${{ matrix.javadist }} ${{ matrix.tests }}
     steps:
     - name: Checkout Repository
       uses: actions/checkout@v3
--- a/.github/workflows/javaTests.yml
+++ b/.github/workflows/javaTests.yml
@@ -49,8 +49,6 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        java: ['11']
-        javadist: ['adopt']
         tests: [
           "org.apache.sysds.test.applications.**",
           "**.test.usertest.**",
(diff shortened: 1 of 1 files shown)


kykrueger:
@Baunsgaard any preferences for a new test-group name?
Also, do you want to merge the function tests?

Baunsgaard:
yup combine all of them, maybe the task description is not accurate, but we can just change it.

Baunsgaard:
name could be just `Java Test`

Baunsgaard:
This broke the tests. @kykrueger

kykrueger:
Forgot to remove the dependency on the old test group, that should be fixed now, and all merged into one matrix. I'm unsure about the use of adopt in the remaining test group. It used to be only present in one of the groups. Same goes for putting java 11 in the matrix.

Any feedback about that @Baunsgaard ?

.github/workflows/javaTests.yml
```diff
@@ -176,14 +102,14 @@ jobs:
           ${{ runner.os }}-maven-test-
 
     - name: ${{ matrix.tests }}
-      uses: ./.github/action/
-      id: test
-      with:
-        test-to-run: ${{ matrix.tests }}
```
kykrueger, line 182: Not sure why the jobs are failing, but this might have something to do with it, I'll check when I have some time this week.

Baunsgaard:
> Forgot to remove the dependency on the old test group, that should be fixed now, and all merged into one matrix. I'm unsure about the use of adopt in the remaining test group. It used to be only present in one of the groups. Same goes for putting java 11 in the matrix.

Just remove the other versions of java here, now unlike before the build tests also verify that it runs, and they include the adopt java tests.

also,

I would call it like this:

image


and not include the word Workflow
aka change it to Java Testing

Currently it is 

image

kykrueger:
@Baunsgaard Let me know if the new name of the workflow is what you had in mind

kykrueger:
Looks like combining all of the test types into one matrix makes a bit of a mess of a few things.

The components tests had been making use of the adopt java distribution, but the other tests had depended on openjdk and/or the R dependencies from the docker image for testing. 

I think it makes sense to keep things in the docker image since it allows devs to run the tests that depend on R with less worry about dependencies. Using the docker image comes with some extra complexity in the github-actions though. For example, caching the maven builds and testing against other java distributions or versions becomes more cumbersome, since these environments will require additional docker image builds.

I'll make the easiest change for now that meets the requirement of merging the test-groups together, which means running them all in the docker container. A caching strategy might be worth considering though to build the image once, and share it between the testing jobs.

Baunsgaard:
ya, i should have clarified this, 
It is supposed to run OpenJDK and the docker image because of the dependencies, and OpenJDK because that is what we actually code up against . It is working for the other JDKs but in actions it seems more consistant with OpenJDK.

Baunsgaard:
The image is shared between runs, and is in fact also build in another GitHub action if you look a bit around.

kykrueger:
Yeah, I saw the action that is being called to build the image, but I don't think the resulting image is cached. At least I didn't see any caching set up for it. 

This PR is from my perspective feature-complete though.

Do you have any change-requests?

Baunsgaard:
Nope, all good, Thanks for the PR it is now merged.

PR #1828, opened 2023-05-18
Transformapply WordEmbeddings

e-strauss:
- implemented transformapply for new encoder for word embeddings
- added new argument "embedding" in transform apply to pass the embedding matrix to the transform apply, the dictionary for the necessary recode is passed via "meta" argument
- added unit tests

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/compress/colgroup/AColGroup.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/colgroup/AColGroup.java
@@ -32,6 +32,8 @@ import org.apache.sysds.runtime.compress.colgroup.indexes.IColIndex.SliceResult;
 import org.apache.sysds.runtime.compress.colgroup.scheme.ICLAScheme;
 import org.apache.sysds.runtime.compress.cost.ComputationCostEstimator;
 import org.apache.sysds.runtime.compress.estim.CompressedSizeInfoColGroup;
+import org.apache.sysds.runtime.compress.estim.encoding.IEncode;
+import org.apache.sysds.runtime.compress.lib.CLALibCombineGroups;
 import org.apache.sysds.runtime.data.DenseBlock;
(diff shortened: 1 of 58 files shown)


src/main/java/org/apache/sysds/runtime/transform/encode/MultiColumnEncoder.java
```diff
@@ -314,11 +314,12 @@ public MatrixBlock apply(CacheBlock<?> in) {
 	public MatrixBlock apply(CacheBlock<?> in, int k) {
 		// domain sizes are not updated if called from transformapply
 		boolean hasUDF = _columnEncoders.stream().anyMatch(e -> e.hasEncoder(ColumnEncoderUDF.class));
+		boolean hasWE = _columnEncoders.stream().anyMatch(e -> e.hasEncoder(ColumnEncoderWordEmbedding.class));
 		for(ColumnEncoderComposite columnEncoder : _columnEncoders)
 			columnEncoder.updateAllDCEncoders();
 		int numCols = getNumOutCols();
 		long estNNz = (long) in.getNumRows() * (hasUDF ? numCols : (long) in.getNumColumns());
-		boolean sparse = MatrixBlock.evalSparseFormatInMemory(in.getNumRows(), numCols, estNNz) && !hasUDF;
+		boolean sparse = MatrixBlock.evalSparseFormatInMemory(in.getNumRows(), numCols, estNNz) && !hasUDF && !hasWE;
```
e-strauss, line 322: actually the "&& !hasWE" is not needed anymore, cuz I am using in the this version "quickSetValue" which should work for both sparse and dense, correct?

phaniarnab: Yes. For UDF encoders we forced dense as we cannot derive the sparsity of the outputs. But for embeddings, you already know the number of nonzeros.

PR #1833, opened 2023-05-25
[SYSTEMDS-3539] Delta Compression Encoding Interfaces

vlad-ponomarenko:
**Description**

-new classes were create for delta encoding task


src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupDeltaDDC.java
```diff
@@ -19,11 +19,271 @@
...
  * encoding (DeltaDDC).
  */
-public class ColGroupDeltaDDC  { // extends ColGroupDDC
+public class ColGroupDeltaDDC extends AColGroupCompressed implements AMapToDataGroup  {
+
+    private static final long serialVersionUID = -1045556313148564147L;
+    private AMapToData _data;
+    private DeltaDictionary _dict;
+    //TODO Confirm if casting from ADictionary to DeltaDictionary is reasonable
+    private ColGroupDeltaDDC(IColIndex colIndexes, ADictionary dict, AMapToData data) {
```
Baunsgaard, line 50: Change the constructor to take DeltaDictionary, this is a private constructor so it should be fine.
if you want to you can make another abstract class called ADeltaDictionary in the dictionary folder that then allows us to extend it. But at the current time it is optional.

This change allows you to not cast to DeltaDictionary, removing a potential source of error.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupDeltaDDC.java
```diff
@@ -19,11 +19,271 @@
...
+        super(colIndexes);
+        _data = data;
+        _dict = (DeltaDictionary) dict;
+    }
+
+    //TODO Confirm if other parameters for constructors are required, such as cachedCounts
+    public static AColGroup create(IColIndex colIndexes, ADictionary dict, AMapToData data) {
+        if(data.getUnique() == 1)
+            return ColGroupConst.create(colIndexes, dict);
```
Baunsgaard, line 59: Not valid. since const is not supporting delta encoding.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupDeltaDDC.java
```diff
@@ -19,11 +19,271 @@
...
+        else if(dict == null)
+            return new ColGroupEmpty(colIndexes);
```
Baunsgaard, line 61: not valid.
on the other hand if you observe that all instances point to a dictionary entry with only 0, then it is valid. 

consider the case that all deltas throughout the matrix is 0, then you know that it is equivalent to an empty matrix.

src/main/java/org/apache/sysds/runtime/compress/colgroup/dictionary/DeltaDictionary.java
```diff
@@ -20,26 +20,90 @@
 package org.apache.sysds.runtime.compress.colgroup.dictionary;
 
 import org.apache.commons.lang.NotImplementedException;
-import org.apache.sysds.runtime.functionobjects.Divide;
-import org.apache.sysds.runtime.functionobjects.Minus;
-import org.apache.sysds.runtime.functionobjects.Multiply;
-import org.apache.sysds.runtime.functionobjects.Plus;
+import org.apache.sysds.runtime.compress.colgroup.indexes.IColIndex;
+import org.apache.sysds.runtime.data.SparseBlock;
+import org.apache.sysds.runtime.functionobjects.*;
```
Baunsgaard, line 25: we do not allow wildcard imports, instead import only the function objects in use,
There should be a setting for your IDE to change this.

src/main/java/org/apache/sysds/runtime/compress/colgroup/dictionary/DeltaDictionary.java
```diff
@@ -20,26 +20,90 @@
...
+		throw new NotImplementedException();
+	}
+
+	@Override
+	public void aggregateColsWithReference(double[] c, Builtin fn, IColIndex colIndexes, double[] reference, boolean def) {
+		throw new NotImplementedException();
+	}
+
 	@Override
 	public DeltaDictionary applyScalarOp(ScalarOperator op) {
```
Baunsgaard, line 108: This content of applyScalarOp is not completely valid, if it is not correct, comment it out and put a NotImplementedException.

src/main/java/org/apache/sysds/runtime/compress/readers/ReaderColumnSelection.java
```diff
@@ -106,6 +106,30 @@ else if(rawBlock.getDenseBlock().numBlocks() > 1)
 			return new ReaderColumnSelectionDenseMultiBlock(rawBlock, colIndices, rl, ru);
 		return new ReaderColumnSelectionDenseSingleBlock(rawBlock, colIndices, rl, ru);
 	}
+	//TODO Add ReaderColumnSelectionDenseDeltaSingleBlock to the method
```
Baunsgaard, line 109: add a clean newline after all methods, after this you can add your TODO.

src/main/java/org/apache/sysds/runtime/compress/readers/ReaderColumnSelection.java
```diff
@@ -106,6 +106,30 @@ else if(rawBlock.getDenseBlock().numBlocks() > 1)
...
+	public static ReaderColumnSelection createDeltaReader(MatrixBlock rawBlock, IColIndex colIndices, boolean transposed,
+													 int rl, int ru) {
+		checkInput(rawBlock, colIndices, rl, ru);
```
Baunsgaard, line 112: The content of this method is not valid, therefore throw NotImplementedException.

src/main/java/org/apache/sysds/runtime/compress/readers/ReaderColumnSelectionDenseDeltaSingleBlock.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.sysds.runtime.compress.readers;
```
Baunsgaard, line 1: Missing license, you can take it from another file.

Baunsgaard (commented):
Thanks for the PR, this is a good start.

Next step is to add a test case of something simple.
For instance make a custom test case for the construction of a DDCDeltaEncoded group.

you can make this a new class in: 
`src/test/java/org/apache/sysds/test/component/compress/colgroup/ColGroupDeltaDDCTest.java`

There make a matrix just like we did on paper, and manually know how the DeltaDDC should be and call the constructor with that.

Then the next step is to add the decompression instruction to be able to decompress the DeltaDDC into a dense Matrix.

src/main/java/org/apache/sysds/runtime/compress/colgroup/dictionary/DeltaDictionary.java
vlad-ponomarenko: I will check it again before the next commit/push, the issue remains unresolved for now and not forgotten.

src/main/java/org/apache/sysds/runtime/compress/colgroup/ColGroupDeltaDDC.java
vlad-ponomarenko: Should I also considerer a case where every element in a column is the same? In a column (-3,-3,-3,-3), the first delta would  be -3 and all other deltas 0. In that case I would check if the first delta value is 0 and all other deltas are the same, then it would be possible to create a constant column group? Or have I misunderstunt the definition of a constant column group? 

 if (dict.getValues().length == 1 && dict.getValue(0)==0), that would represent not an empty Matrix, but a Matrix that only contains zeros, therefore, my recent code is invalid I suppose. 

The definition of an empty matrix in this context is a matrix with only zeroes or is it a matrix with at least one dimension that is equal to zero?

Baunsgaard: yes, if the column can be encoded as a constant or an empty column group the return should be an empty column group. 
The case you write is exactly the case we need to support.

But !

While it would be good that we return the specialized instance of const and empty column groups,  you do not have to add it if it is confusing, just make sure that the thing that is returned from the construct is correct, or throw an error if incorrect.

Baunsgaard:
You need to fix the failing tests such that i can merge this starting point into the system,
The fix in this case is a file missing an import.

image


Thanks.

PR #1836, opened 2023-06-04
[SYSTEMDS-3575] Column Group get compression Info 

Baunsgaard:


Baunsgaard:
Not to self, tests that are failing.
image

PR #1842, opened 2023-06-12
[SYSTEMDS-3579] New dense block for deduplication

e-strauss:
- added a new dense block, which is optimised for containing duplicate row for the word embedding transformation
- updated the word embedding encoder to use the new denseblock


phaniarnab:
Thanks @e-strauss, for this patch. I fixed a few minor things and merged the changes.

PR #1847, opened 2023-06-18
[SYSTEMDS-2834] python IO benchmarking

kykrueger:
This PR is for adding reusable benchmarking scripts initially discussed at the following link:
https://issues.apache.org/jira/browse/SYSTEMDS-2834

Implementation requirements have yet to be finalized, but the initial requirements may be summarized as benchmarks for:

- [x] loading csv/binary data with DML scripts
- [x] loading csv/binary data with the read command of the python api
- [x] loading csv/binary data to a frame from a pandas dataframe
- [x] loading csv/binary data to a matrix from a numpy array

The purpose is to evaluate the additional cost of loading data of the python api-calls.

### DIFF
--- a/scripts/perftest/README.md
+++ b/scripts/perftest/README.md
@@ -17,18 +17,35 @@ limitations under the License.
 {% end comment %}
 -->
 
-# Performance tests SystemDS
+# Performance Tests SystemDS
 
-To run all performance tests for SystemDS, simply download systemds, install the prerequisites and execute.
+To run all performance tests for SystemDS:
+ * install systemds,
+ * install the prerequisites,
+ * navigate to the perftest directory $`cd $SYSTEMDS_ROOT/scripts/perftest` 
+ * generate the data,
+ * and execute.
 
 There are a few prerequisites:
 
+## Install SystemDS
+
 - First follow the install guide: <http://apache.github.io/systemds/site/install> and build the project.
+- Install the python package for python api benchmarks: <https://apache.github.io/systemds/api/python/getting_started/install.html>
+- Prepare to run SystemDS: <https://apache.github.io/systemds/site/run>
+
+## Install Additional Prerequisites
 - Setup Intel MKL: <http://apache.github.io/systemds/site/run>
 - Setup OpenBlas: <https://github.com/xianyi/OpenBLAS/wiki/Precompiled-installation-packages>
 - Install Perf stat: <https://linoxide.com/linux-how-to/install-perf-tool-centos-ubuntu/>
 
-## NOTE THE SCRIPT HAS TO BE RUN FROM THE PERFTEST FOLDER
+## Generate Test Data
+
+Using the scripts found in `$SYSTEMDS_ROOT/scripts/perftest/datagen`, generate the data for the tests you want to run. Note the sometimes optional and other times required parameters/args. Dataset size is likely the most important of these.
+
+## Run the Benchmarks
+
+**Reminder: The scripts should be run from the perftest folder.**
 
 Examples:
 
@@ -36,7 +53,7 @@ Examples:
 ./runAll.sh
 ```
 
-Look inside the runAll script to see how to run individual tests.
+Or look inside the runAll script to see how to run individual tests.
 
-Time calculations in the bash scripts additionally subtract a number, e.g. ".4".
+Time calculations in the bash scripts may additionally subtract a number, e.g. ".4".
 This is done to accommodate for time lost by shell script and JVM startup overheads, to match the actual application runtime of SystemML.
--- a/scripts/perftest/datagen/genClusteringData.sh
+++ b/scripts/perftest/datagen/genClusteringData.sh
@@ -25,9 +25,9 @@ then
   exit 1;
 fi
 
-CMD=$1
-BASE=$2/clustering
-MAXMEM=$3
+CMD=${1:-systemds}
+BASE=${2:-"temp"}/clustering
+MAXMEM=${3:-80}
 
 FORMAT="binary" 
 DENSE_SP=0.9
--- a/scripts/perftest/datagen/genIOData.sh
+++ b/scripts/perftest/datagen/genIOData.sh
@@ -26,36 +26,47 @@ then
 fi
 
 CMD=${1:-systemds}
-BASE=${2:-"temp"}/dimensionreduction
-MAXMEM=${3:-80}
+DATADIR=${2:-"temp"}/io
+MAXMEM=${3:-1}
 
-FORMAT="binary"
+FORMAT="csv" # can be csv, mm, text, binary
 
-echo "-- Generating Dimension Reduction data." >> results/times.txt;
+echo "-- Generating IO data." >> results/times.txt;
+
+
+#generate XS scenarios (10MB)
+if [ $MAXMEM -ge 1 ]; then
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X500_250_dense R=500 C=250 Fmt=$FORMAT &
+fi
+
+#generate XS scenarios (10MB)
+if [ $MAXMEM -ge 10 ]; then
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X5k_250_dense R=5000 C=250 Fmt=$FORMAT &
+fi
 
 #generate XS scenarios (80MB)
 if [ $MAXMEM -ge 80 ]; then
-  ${CMD} -f ../datagen/genRandData4PCA.dml --nvargs R=5000 C=2000 OUT=$BASE/pcaData5k_2k_dense FMT=$FORMAT &
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X10k_1k_dense R=10000 C=1000 Fmt=$FORMAT &
 fi
 
 #generate S scenarios (800MB)
 if [ $MAXMEM -ge 800 ]; then
-  ${CMD} -f ../datagen/genRandData4PCA.dml --nvargs R=50000 C=2000 OUT=$BASE/pcaData50k_2k_dense FMT=$FORMAT &
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X100k_1k_dense R=100000 C=1000 Fmt=$FORMAT &
 fi
 
 #generate M scenarios (8GB)
 if [ $MAXMEM -ge 8000 ]; then
-  ${CMD} -f ../datagen/genRandData4PCA.dml --nvargs R=500000 C=2000 OUT=$BASE/pcaData500k_2k_dense FMT=$FORMAT &
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X1M_1k_dense R=1000000 C=1000 Fmt=$FORMAT &
 fi
 
 #generate L scenarios (80GB)
 if [ $MAXMEM -ge 80000 ]; then
-  ${CMD} -f ../datagen/genRandData4PCA.dml --nvargs R=5000000 C=2000 OUT=$BASE/pcaData5M_2k_dense FMT=$FORMAT
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X10M_1k_dense R=10000000 C=1000 Fmt=$FORMAT &
 fi
 
 #generate XL scenarios (800GB)
 if [ $MAXMEM -ge 800000 ]; then
-  ${CMD} -f ${EXTRADOT}./datagen/genRandData4PCA.dml --nvargs R=50000000 C=2000 OUT=$BASE/pcaData50M_2k_dense FMT=$FORMAT
+  ${CMD} -f ../utils/generateData.dml --nvargs Path=${DATADIR}/X100M_1k_dense R=100000000 C=1000 Fmt=$FORMAT &
 fi
 
 wait
--- a/scripts/perftest/python/io/load_native.py
+++ b/scripts/perftest/python/io/load_native.py
@@ -1,4 +1,3 @@
-#!/bin/bash
 #-------------------------------------------------------------
 #
 # Licensed to the Apache Software Foundation (ASF) under one
@@ -8,9 +7,9 @@
 # to you under the Apache License, Version 2.0 (the
 # "License"); you may not use this file except in compliance
 # with the License.  You may obtain a copy of the License at
-# 
+#
 #   http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing,
 # software distributed under the License is distributed on an
 # "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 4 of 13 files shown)


kykrueger:
@Baunsgaard 

I have a few questions about requirements now that I've gotten started with a proof-of-concept script.

1. I've selected the timeit module to help make the times more reproducible since it lets us elegantly ignore setup in the recorded time and disable some sources of additional processing time like the garbage-collector in python. However, I wasn't sure if this is something you want. Are you on-board with this approach, or did you want more of a big-picture benchmark with all overhead included? 
2.  I've noticed that the conversions from numpy and pandas data is transferred lazily to SystemDS in the JVM. So, calling `from_numpy` and `from_pandas`, alone isn't letting me evaluate the real overhead. If you choose a big-picture benchmark above, I figure here I'd be stuck adding some sort of simple operator and calling compute or get_lineage to force the data to load. Another option would be to trick systemds into loading the data by creating an empty `DMLScript` and calling  the hidden internal `__prepare_script()` method because it has a bit less overhead than `get_lineage()`. This second option is what I'd prefer since it is more isolated, does it meet your needs?

Baunsgaard:
> 

1. I am in favour of simplicity if it is clean and simple fine with me. Otherwise i am also fine with standard time calls in python. We have to remember what the use case is and in this case the difference in python garbage collector should not be a factor we need to consider. 
2. I suggest that we keep to the lazy interface and do not add some new shortcut. Since again we have to keep in mind what is the usecase. And in this case it is a user script. They care about the entire thing with compiling and sending over the python script to java. Not only the data transfer. I understand this make it more inconclusive about where time is spend. But if the data is of appropriate size then the overheads not associated with IO is amortized and will no longer be much visible in the observed times that we want to measure.

kykrueger:
@Baunsgaard thanks for the recommendations. The current approach should meet the requirements for measuring the loading time for all of the options in python. It uses an option I'd noticed after writing my last comment, and I believe it keeps all relevant compilation steps while avoiding any extra operations after loading.

It sticks with timeit instead of switching to the time functions, which I think continues to add a benefit without much added complexity. Looking at timeit benchmark scripts may first give the impression of an antipattern since the to be measured code is written in strings, but it seems to be the pythonic way.

There are a few things I see left for me to do:

- [ ] figure out why the pandas benchmark is hanging
- [ ] make the log/stdout match the style used in the other perftest scripts
- [ ] add an input option for the binary data format instead of csv
- [ ] wrap a DML script in bash which allows the user to choose a number of run times and pass a csv/binary file
- [ ] choose a variety of generated input data
- [ ] call all of the benchmark scripts in one bash script similar to the other perftest benchmarks.

That leaves me with a few more things to check with you before I complete the above tasks.
- Do the timeit benchmarks I've created so far cover all of the use-cases you had in mind?
- You asked for a variety of data-types when we started the ticket, I'd have expected that to mean floats vs ints vs strings vs doubles and so on. However, I've noticed that the systemds docs describe Frame, Matrix and Scalars as the primary data-types. So, I assume that you also meant only Frames and Matrices as the significant data-types to be tested?
- Initially I'd planned on writing some of my own data-generation scripts, but I noticed the existing generic one [scripts/utils/generateData.dml](https://github.com/apache/systemds/blob/main/scripts/utils/generateData.dml) and didn't see much value in creating a similar one. However, I am also considering adding a few extra runs using some of the generation scripts in scripts/perftest/datagen or scripts/datagen, but am not sure if that is in scope of the initial ticket. What do you think, is the use of the existing datagen what you were expecting?

kykrueger:
@Baunsgaard 
Initially I'd thought there was a problem with my benchmark for loading data with pandas, but it seems that the script runs correctly, and that the converter for pandas to frames cannot handle loading large datasets in a reasonable amount of time, am I mistaken, is this a known problem?

Im pretty sure I could come up with an easy fix for it by extending Py4JConverterUtils to take some sort of array format directly from pandas or dumping the pandas columns to numpy to reuse part of the existing ones from the Matrices. That defnitely would break the scope of this issue though.

kykrueger:
This is the code which isn't scaling:
https://github.com/apache/systemds/blob/5a907ca28ad1b3ffc7ad24c5c2227b91b0fd7795/src/main/python/systemds/utils/converters.py#L125-L129

Loading times blow up rapidly when comparing the numpy converter with growing data-sizes.

kykrueger:
I ended up just creating more Data-Generators for the benchmark. I reckon there is a way to improve run-times by using the same generated data for a few of the benchmarks, but I'd have to make changes to a lot of the other benchmarks which would be a bit tedious.

The pandas loading hangs for quite a long time with even the smallest 80 MB data set, but has eventually completed.

Baunsgaard:
> @Baunsgaard Initially I'd thought there was a problem with my benchmark for loading data with pandas, but it seems that the script runs correctly, and that the converter for pandas to frames cannot handle loading large datasets in a reasonable amount of time, am I mistaken, is this a known problem?

Yes this is a known problem for strings that we want to optimize. Hence the first step is to make quality measurements to start the process. I do not know if we have problems with other data types.

> Im pretty sure I could come up with an easy fix for it by extending Py4JConverterUtils to take some sort of array format directly from pandas or dumping the pandas columns to numpy to reuse part of the existing ones from the Matrices. That defnitely would break the scope of this issue though.

Yup it would break it and also go into another task already there to do ... exactly what you are proposing. 
We are aware that the performance is bad for Pandas currently. There are multiple reasons. But again. measurements first then we can fix the performance after.

Baunsgaard:
> This is the code which isn't scaling:
> 
> https://github.com/apache/systemds/blob/5a907ca28ad1b3ffc7ad24c5c2227b91b0fd7795/src/main/python/systemds/utils/converters.py#L125-L129
> 
> Loading times blow up rapidly when comparing the numpy converter with growing data-sizes.

Yup, you are identifying the same as me :) 

Since pandas use numpy arrays behind then we can definitely do something smarter. But lets wait and see for the other project is if gets going.

scripts/perftest/datagen/genDimensionReductionData.sh
```diff
@@ -25,9 +25,9 @@ then
   exit 1;
 fi
 
-CMD=$1
-BASE=$2/dimensionreduction
-MAXMEM=$3
+CMD=${1:-systemds}
```
Baunsgaard, line 28: i am not sure what this means. can you explain?

scripts/perftest/python/input/load_native.py
```diff
@@ -0,0 +1,50 @@
...
+"with SystemDSContext(logging_level=10, py4j_logging_level=50) as ctx:",
+"    node = ctx.read(src)",
+"    script = DMLScript(ctx)",
+"    script.build_code(node)",
+"    script.execute()"
+])
+
+def main(args):
+    gvars = {"src": args.src}
+    print(timeit.timeit(run, setup, globals=gvars, number=args.number))
```
Baunsgaard, line 41: this is pretty cool. did not know this library.

Baunsgaard (commented):
>    Do the timeit benchmarks I've created so far cover all of the use-cases you had in mind?

Well, not really. The data when read from CSV is always treated as worst case datatypes. 
this means in practice when we transfer the data over the scheme says string string string etc and therefore most likely is using the worst transfer.

I would like to see something where we maybe use detect schema, to infer the data types. and then use correctly allocated types to transfer.

>  You asked for a variety of data-types when we started the ticket, I'd have expected that to mean floats vs ints vs strings vs doubles and so on. However, I've noticed that the systemds docs describe Frame, Matrix and Scalars as the primary data-types. So, I assume that you also meant only Frames and Matrices as the significant data-types to be tested?

There is a big difference between a data structure such as Frame, Matrix and Scalar, and then the underlying value type inside. In the case of this task i expect us to transfer various allocated data values contained in structures over to data structures inside systemds. 

A Matrix (Data structure) always contains the value type of doubles inside the system, but the transfer allows use to use allocated numpy arrays of various numeric types and transfer into systemDS without materializing doubles on the python side. The best performing transfers at the moment should be  is using UINT8 in numpy. and i would like to see us measuring the transfer time for double, float, long, int, uint8, boolean and BitSets.

The frames map always from Pandas, and i have not done much testing with this. 
Ideally we want to test out mapping of all the data types, and have dedicated tests for individual types. In specific i would like to see cases where we transfer heterogenerous Pandas Dataframes, and homogeneous frames. Datatypes of interests are String, Double, Float, Long, Int, Byte, boolean and Bitsets.

>  Initially I'd planned on writing some of my own data-generation scripts, but I noticed the existing generic one [scripts/utils/generateData.dml](https://github.com/apache/systemds/blob/main/scripts/utils/generateData.dml) and didn't see much value in creating a similar one. However, I am also considering adding a few extra runs using some of the generation scripts in scripts/perftest/datagen or scripts/datagen, but am not sure if that is in scope of the initial ticket. What do you think, is the use of the existing datagen what you were expecting?

I would actually suggest that we generate data in memory to isolate the test.
This does not invalidate your work so far, but we need to make more files like the scripts/perftest/python/input/load_native.py 
That test via creating the numpy and pandas data in the python script (perhaps in the setup phase) and then evaluate the transfer speed into systemDS in memory, and similarly evaluate the extraction time all with the different datatypes.

Baunsgaard:
Also if you have any numbers of the things already, please just post them even if they are bad!

scripts/perftest/datagen/genDimensionReductionData.sh
kykrueger: This sets a default value if the Arg is null

kykrueger:
@Baunsgaard 
> Well, not really. The data when read from CSV is always treated as worst case datatypes.
this means in practice when we transfer the data over the scheme says string string string etc and therefore most likely is using the worst transfer.

Since I generated the data with systemds, it comes with a schema. When it comes to the interpolation, I had meant to say I didn't specify dtypes for numpy and pandas, but they had been smart enough to recognize the values as float64. 

In order to get benchmarks for the other data-types, I could pass in a dtype optionally to my scripts which forces numpy and pandas to read put the values into the type we want. I'll also check out overriding the types provided in the MTD files in the read, command for systemds. I think I remember seeing something in your docs which make that possible. This should make it easy to test for all of the data types you listed, without the creation of additional scripts, how about it?

- [ ] double, float, long, int, uint8, boolean and BitSets. 

I'm not sure about why you would like to generate the data in the scripts, that would leave us with writing the same generation code multiple times. I'd prefer to keep using the generation format similar to the existing perftest scripts. However, I don't think the large data-sizes are very useful. Since the 80MB dataset stalls pandas for much too long, I'm going to add a smaller generation option to the list and set it as the default. 

> Also if you have any numbers of the things already, please just post them even if they are bad!

As for the run-times, these values came out of a run with 1 repeat with a small dataset, of 100 columns and 10k rows if I remember correctly.

```
RUN IO Benchmarks:  Sun 25 Jun 2023 10:19:23 PM CEST
-- Running IO benchmarks on custom
read.dml:            2.932409849
load_native.py:      3.3357212970149703
load_numpy.py:       2.91045009897789
load_pandas.py:      33.540340821957216
```

Baunsgaard:
> Since I generated the data with systemds, it comes with a schema. When it comes to the interpolation, I had meant to say I didn't specify dtypes for numpy and pandas, but they had been smart enough to recognize the values as float64.



> In order to get benchmarks for the other data-types, I could pass in a dtype optionally to my scripts which forces numpy and pandas to read put the values into the type we want. I'll also check out overriding the types provided in the MTD files in the read, command for systemds. I think I remember seeing something in your docs which make that possible. This should make it easy to test for all of the data types you listed, without the creation of additional scripts, how about it?

The issue lies in matrices primarily, again currently the datatype is always double inside systemds. and transferring this out to numpy is always the same cost. What i would be interested to know is still the transfer in if that is efficient.

For pandas and frames we need the different types, and yes these can be specified in the mtd file. as long as the content of the CSV is correct it should work.

> I'm not sure about why you would like to generate the data in the scripts, that would leave us with writing the same generation code multiple times. I'd prefer to keep using the generation format similar to the existing perftest scripts. However, I don't think the large data-sizes are very useful. Since the 80MB dataset stalls pandas for much too long, I'm going to add a smaller generation option to the list and set it as the default.

Feel free to reduce the data size further. as said previously it is know that out pandas implementation is slow. we need to fix it after this task.

Baunsgaard:
```
RUN IO Benchmarks:  Sun 25 Jun 2023 10:19:23 PM CEST
-- Running IO benchmarks on custom
read.dml:            2.932409849
load_native.py:      3.3357212970149703
load_numpy.py:       2.91045009897789
load_pandas.py:      33.540340821957216

```

Cool so numpy to systemds is essentially same as reading. 
It should ideally be faster but i guess we still have performance on the table from not parallelizing the transfer maybe?

Baunsgaard:
> 

Deadline is 24 hours before your exam. Matthias have just extended it.

kykrueger:
Over the weekend I added the option to set dtypes in the io loading scripts.
I still want to add the additional benchmarks for materializing the data of different types in the other direction 
(systemds -> python).

For now, here are the results from a test with 20 repeats for the data-types configured in the test at them moment:
```
-- Running IO benchmarks on 500_250_dense
double:    read.dml;        40.781715454
double:    load_native.py;  39.19094614699134
int:       read.dml;        32.824596657
int:       load_native.py;  36.457156577002024
string:    read.dml;        34.440663763
string:    load_native.py;  38.71029913998791
boolean:   read.dml;        33.266684618
boolean:   load_native.py;  36.68671202700352
double:    load_numpy.py;   32.85507999898982
double:    load_pandas.py;  512.6433556610136
float:     load_numpy.py;   38.261559439997654
float:     load_pandas.py;  546.0650390849914
long:      load_numpy.py;   39.400702337006805
long:      load_pandas.py;  536.5950958920002
int64:     load_numpy.py;   32.98173662999761
int64:     load_pandas.py;  487.0634801320266
int32:     load_numpy.py;   32.48500068101566
int32:     load_pandas.py;  489.97116349000135
uint8:     load_numpy.py;   31.86706029099878
uint8:     load_pandas.py;  496.9151880980062
string:    load_numpy.py;
string:    load_pandas.py;  504.3096235789999
bool:      load_numpy.py;   33.19832509398111
bool:      load_pandas.py;  479.9256292580103
```

I'm undecided about the order to run them in to have logs which make the most sense for you to read. Currently they are just grouped in a way that made the proof-of-concept script easier to write.

kykrueger:
As far as I can tell, there is one subtask left:
- [ ] add materialization benchmarks from systemds to python

Otherwise, I don't have any tests for Bitsets, but I didn't see anything about them in the docs from systemds.

Baunsgaard:
cool, thanks!
I consider this solid work, and project done.

kykrueger:
@Baunsgaard thanks,

Perhaps two more notes from my side in addition to benchmarking the times to send the data from systemds to python. We could:

- [ ] add a few tips in the perftest readme about installing the python package for systemds since it is a new requirement
- [ ] add tips about generating the data as a prerequisite (I had noticed it was needed to run the existing tests, but didn't see any suggestions to do this, maybe I missed something)

Perhaps it would be a good follow-up task to check in the perftest shell scripts if the required data exists before running a test, and generating it if not yet present. That could also open some doors to sharing a folder and common naming for the test-data-sets in order to reuse the same cached data-set between tests.

kykrueger:
@Baunsgaard I took a look at isolating the time to materialize to python from systemds, and it doesn't look like there are any easy ways to do this at the moment.

There would be an ugly solution of creating  two new benchmarks which force a transformation and then the transformation + materialization, then calculate the difference between the operations as the time needed to pass from systemds to python. 

However, I'd take a step back from this one and suggest that first a materialization feature is added to systemds akin to what is available in spark. This might be useful in more contexts, and make more features easier to benchmark in the future.

Shall we close merge this PR without that feature after all?
Also, I've noticed that the PRs show as closed instead of merged, is that because of local merging, could you try merging from the github UI for better Github Achievements tracking?

Best,
-- Kyle

Baunsgaard:
> 

Hi Kyle, 

The easy way to see transfer speed from java into python is calling compute.
you can generate a matrix inside the SDS context with something like rand() and then call compute. The compute then transfer over the matrix, and if you enable statistics you can see the internal time use of the rand call, while extrapolating the time of transfer as the unreported time in the statistics.
Similarly you can read in a Frame in the SDS context and call compute to transfer it to pandas.

I think this would be to much for this PR, so it can be a subsequent PR.

> Also, I've noticed that the PRs show as closed instead of merged, is that because of local merging, could you try merging from the github UI for better Github Achievements tracking?

We are merging ourselves via rebase ans squash so we circumvent the GitHub UI. This is done for historic reasons, to reduce errors, and to avoid merge commits in the git log. If you want it to be shown as merged then the procedure is also simple. You rebase your commits on top of main, then squash all your commits into a single one. This then is registered as a merge in GitHub when we manually merge it.

Br 
Sebastian

kykrueger:
>  if you enable statistics you can see the internal time use of the rand call, while extrapolating the time of transfer as the unreported time in the statistics. 

I hadn't realized that the statistics from systemds would do soo much of the work for me of separating what is taking time in the calls once all of the steps materialize together. I'd been putting thought into how to standardize comparison benchmarks between the APIs for  tasks like this and set an example for later comparisons. However, I hadn't considered the statistics feature built into systemds for checking the time consumed at different steps. It would be worth a reconsideration for different python benchmarks in the future.

As for the rebase, I'll get right to it

kykrueger:
@Baunsgaard that should do it!

Baunsgaard:
Hi Kyle, thanks for the PR, it is now merged, 

Thanks for the contribution !

It clearly indicate that there are some things that can be improved.

best regards
Sebastian.

kykrueger:
Gladly!

PR #1857, opened 2023-07-03
[SYSTEMDS-3589] Frame single column ragged array

OlgaOvcharenko:
This PR adds a ragged array for a single column containing row values. It is just a wrapper around actual frame column Arrays.
 It can be used for the `transformencode` metadata frames.

Thanks for the review.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
+++ b/src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
@@ -24,6 +24,7 @@ import java.io.DataOutput;
 import java.io.IOException;
 import java.util.Arrays;
 
+import org.apache.commons.lang.NotImplementedException;
 import org.apache.sysds.common.Types.ValueType;
 import org.apache.sysds.runtime.DMLRuntimeException;
 import org.apache.sysds.runtime.frame.data.columns.ArrayFactory.FrameArrayType;
@@ -56,6 +57,15 @@ public class RaggedArray<T> extends Array<T> {
 	public RaggedArray(Array<T> a, int m) {
 		super(m);
 		this._a = a;
+	}
+
+	protected Array<T> getInnerArray(){
+		return _a;
+	}
+
+	@Override
+	public Object get() {
+		throw new NotImplementedException("Should not be called");
 	}
 
 	@Override
@@ -97,15 +107,6 @@ public class RaggedArray<T> extends Array<T> {
 
 	}
 
-	protected Array<T> getInnerArray(){
-		return _a;
-	}
-
-	@Override
-	public Object get() {
-		return _a.get();
-	}
-
 	@Override
 	public double getAsDouble(int i) {
 		return i < _a._size ? _a.getAsDouble(i) : 0;
@@ -120,18 +121,34 @@ public class RaggedArray<T> extends Array<T> {
 	public void set(int index, T value) {
 		if (index < _a._size)
 			_a.set(index, value);
+		else if(index < super.size()) {
+			_a.reset(index + 1);
+			_a.set(index, value);
+			LOG.warn("Reallocated ragged array");
+		}
+
 	}
 
 	@Override
 	public void set(int index, double value) {
 		if (index < _a._size)
 			_a.set(index, value);
+		else if(index < super.size()) {
+			_a.reset(index + 1);
+			_a.set(index, value);
+			LOG.warn("Reallocated ragged array");
+		}
 	}
 
 	@Override
 	public void set(int index, String value) {
 		if (index < _a._size)
 			_a.set(index, value);
+		else if(index < super.size()) {
+			_a.reset(index + 1);
+			_a.set(index, value);
+			LOG.warn("Reallocated ragged array");
+		}
 	}
 
 	@Override
@@ -149,6 +166,11 @@ public class RaggedArray<T> extends Array<T> {
 				_a.set(rl, ru, value);
 			else
 				throw new RuntimeException("RaggedArray set: value type should be same to RaggedArray type " + _a.getClass());
+		else if(rl >= 0 && rl < super.size() && ru < super.size()) {
+			_a.reset(rl + 1);
+			_a.set(rl, ru, value);
+			LOG.warn("Reallocated ragged array");
+		}
 	}
 
 
@@ -178,20 +200,37 @@ public class RaggedArray<T> extends Array<T> {
 
 	@Override
 	public void append(String value) {
-		_a.append(value);
+		Array<T> oldVals = _a.clone();
+		_a.reset(super.size() + 1);
+		_a.set(0, oldVals.size() - 1, oldVals);
+		_a.set(super.size(), value);
 		super._size += 1;
+
+		LOG.warn("Fully allocated ragged array");
 	}
 
 	@Override
 	public void append(T value) {
-		_a.append(value);
+		Array<T> oldVals = _a.clone();
+		_a.reset(super.size() + 1);
+		_a.set(0, oldVals.size() - 1, oldVals);
+		_a.set(super.size(), value);
 		super._size += 1;
+
+		LOG.warn("Fully allocated ragged array");
 	}
 
 	@Override
 	public Array<T> append(Array<T> other) {
-		super._size += other._size;
-		return _a.append(other);
+		Array<T> oldVals = _a.clone();
+		_a.reset(super.size() + other._size + 1);
+		_a.set(0, oldVals.size() - 1, oldVals);
+		_a.set(super.size(), super.size() + other.size() - 1, other);
+		super._size += other.size();
+
+		LOG.warn("Fully allocated ragged array");
+
+		return this;
 	}
 
 	@Override
@@ -211,7 +250,7 @@ public class RaggedArray<T> extends Array<T> {
 
 	@Override
 	public byte[] getAsByteArray() {
-		return _a.getAsByteArray();
+		throw new NotImplementedException("Unimplemented method 'getAsByteArray'");
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


phaniarnab:
I don't fully understand the use case of a ragged array-based frame. Is the metatadata from a `transformencode `the only case where this feature helps? Will splitting a metadata frame into a list of frames (one per column) not serve the purpose?

OlgaOvcharenko:
> I don't fully understand the use case of a ragged array-based frame. Is the metatadata from a `transformencode `the only case where this feature helps? Will splitting a metadata frame into a list of frames (one per column) not serve the purpose?

The ragged array is intended to save mem since the metadata frame is mostly nulls. We probably do not want to change the API of `transformencode`; therefore, I do not think a list of frames would be a suitable solution. This change makes no outside API changes and reduces memory allocation and the on-disk size of the metadata that we save after `transformencode`.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	public void setFromOtherTypeNz(int rl, int ru, Array<?> value) {
-		throw new NotImplementedException("Unimplemented method 'setFromOtherTypeNz'");
+		if(rl >= 0 && rl < _a._size && ru < _a._size)
+			_a.setFromOtherTypeNz(rl, ru, value);
 	}
 
 	@Override
 	public void append(String value) {
-		throw new NotImplementedException("Unimplemented method 'append'");
+		_a.append(value);
```
Baunsgaard, line 159: not valid. append appends to the end. this means in this case reallocate the entire thing and then append one element in the end.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	public void reset(int size) {
-		throw new NotImplementedException("Unimplemented method 'reset'");
+		_a.reset(size);
+		super._size = size;
 	}
 
 	@Override
 	public byte[] getAsByteArray() {
-		throw new NotImplementedException("Unimplemented method 'getAsByteArray'");
+		return _a.getAsByteArray();
```
Baunsgaard, line 192: throw some exception here. this will not work, since the caller expects the byte array to be certain m size * nBytes per value.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	@Override
 	public FrameArrayType getFrameArrayType() {
-		throw new NotImplementedException("Unimplemented method 'getFrameArrayType'");
+		return FrameArrayType.RAGGED;
 	}
 
 	@Override
 	public long getExactSerializedSize() {
-		throw new NotImplementedException("Unimplemented method 'getExactSerializedSize'");
+		return _a.getExactSerializedSize();
```
Baunsgaard, line 212: + int size.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	@Override
 	protected Array<Character> changeTypeCharacter() {
-		throw new NotImplementedException("Unimplemented method 'changeTypeCharacter'");
+		return _a.changeTypeCharacter();
 	}
 
 	@Override
 	public void fill(String val) {
-		throw new NotImplementedException("Unimplemented method 'fill'");
+		_a.fill(val);
```
Baunsgaard, line 257: not correct, also fill the ragged part therefore reallocate.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	@Override
 	public boolean isEmpty() {
-		throw new NotImplementedException("Unimplemented method 'isEmpty'");
+		return _a.isEmpty();
 	}
 
 	@Override
 	public Array<T> select(int[] indices) {
-		throw new NotImplementedException("Unimplemented method 'select'");
+		return _a.select(Arrays.stream(indices).filter(x -> x < _size && x >= 0).toArray());
```
Baunsgaard, line 277: no filter. also include indexes inside the ragged part, just as null.

src/test/java/org/apache/sysds/test/component/frame/array/FrameArrayTests.java
```diff
@@ -1779,6 +1782,22 @@ protected static Array<?> createDDC(FrameArrayType t, int size, int seed) {
 			case CHARACTER:
 				return DDCArray
 					.compressToDDC(ArrayFactory.create(generateRandomCharacterNUniqueLengthOpt(size, seed, nUnique)));
+			case RAGGED:
+				Random rand = new Random(seed);
+				switch(rand.nextInt(7)) {
+					case 0:
+						return ArrayFactory.create(generateRandomIntegerOpt(size, seed), size);
```
Baunsgaard, line 1789: we have to test ragged dedicated with smaller arrays inside the ragged part.

Baunsgaard (commented):
good start, some details missing.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	@Override
 	public void write(DataOutput out) throws IOException {
-		throw new NotImplementedException("Unimplemented method 'write'");
+		_a.write(out);
 	}
 
 	@Override
 	public void readFields(DataInput in) throws IOException {
-		throw new NotImplementedException("Unimplemented method 'readFields'");
+		_a.readFields(in);
```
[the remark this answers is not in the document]
OlgaOvcharenko: Done, fixed both read and write.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	@Override
 	public T get(int index) {
-		throw new NotImplementedException("Unimplemented method 'get'");
+		if (index > super._size || index < 0)
+			throw new ArrayIndexOutOfBoundsException("Index " + index + " out of bounds " + super._size);
+		return index < _a._size ? _a.get(index) : null;
+
+	}
+
+	protected Array<T> getInnerArray(){
```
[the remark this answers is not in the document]
OlgaOvcharenko: Sure, resolved.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
+		return _a;
 	}
 
 	@Override
 	public Object get() {
-		throw new NotImplementedException("Unimplemented method 'get'");
+		return _a.get();
```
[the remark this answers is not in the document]
OlgaOvcharenko: Fixed. However, OptionalArray.get returns an Object and it makes sense to me.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
 	}
 
 	@Override
 	public double getAsDouble(int i) {
-		throw new NotImplementedException("Unimplemented method 'getAsDouble'");
+		return i < _a._size ? _a.getAsDouble(i) : Double.NaN;
```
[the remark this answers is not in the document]
OlgaOvcharenko: You are right, fixed to 0.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
```diff
@@ -49,207 +49,261 @@ public class RaggedArray<T> extends Array<T> {
...
+	@Override
+	public double getAsNaNDouble(int i) {
+		return i < _a._size ? _a.getAsNaNDouble(i) : Double.NaN;
 	}
 
 	@Override
 	public void set(int index, T value) {
-		throw new NotImplementedException("Unimplemented method 'set'");
+		if (index < _a._size)
+			_a.set(index, value);
```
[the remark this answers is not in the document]
OlgaOvcharenko: Then it might be not ragged and a normal Array is cheaper then.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
OlgaOvcharenko: But then in this case it is not a ragged array anymore. But sure if we emulate normal behaviour.

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
OlgaOvcharenko: Why no filter? What if input is just wrong?

src/main/java/org/apache/sysds/runtime/frame/data/columns/RaggedArray.java
OlgaOvcharenko: Then, I believe that optional should also throw an exception to be consistent?
Since currently it is not the case:
```
public byte[] getAsByteArray() {
     // technically not correct.
    return _a.getAsByteArray();
}
```

Baunsgaard:
Thanks for the commit, i am merging and fixing remaining issues in other PR.

PR #1860, opened 2023-07-13
[MINOR] Remove usage of apache.commons.lang version 2

corepointer:
We use the latest version of apache.commons.lang but don't reflect that in the source code. This rather extensive commit replaces all occurrences of the old import naming scheme and fixes a few issues where functionality was removed or changed in the new version.


Baunsgaard:
Thanks @corepointer  ! it is now merged!

PR #1862, opened 2023-07-14
[MINOR] Support Non-Literals in Federated Reshape Instruction

ywcb00:
Hi,
This PR adds support for non-literal parameters _rows_ and _cols_ in _ReshapeFEDInstruction_ by simply constructing a literal operand for the respective scalar and replacing the corresponding operand of the instruction string. I also changed the DML script of the _FederatedReshapeTest_  such that we materialize one of these parameters as a non-literal.
This PR implicitly also includes the removal of unused variables inside the _FederatedReaderTest_ and the change of value range of the random matrices in _FederatedMisAlignedTest_.

Thanks for review :)


mboehm7:
LGTM - Thanks for the patch @ywcb00.

PR #1863, opened 2023-07-15
[SYSTEMDS-2944] Federated Sparsity Propagation

ywcb00:
Hi,
This PR adds the functionality of propagating the sparsity from the federated workers to the coordinator when executing an instruction without subsequent data consolidation at the worker. To achieve this, the workers respond to a EXEC_INST request with the number of non-zeros of their respective federated partition. Thereby, we can obtain and aggregate the number of non-zeros and subsequently set the number of non-zeros of the (still federated) matrix object.
In this PR, several instructions have been extended to support the before mentioned sparsity propagation. I also added a respective unit test which compares the propagated number of non-zeros to a reference. Unfortunately, I had to copy parts of the _DMLScript.java_ code into the test class because the test needs direct access to the execution context of the script execution.
Additionally, I left three types of FIXME comments in the test dml script. In two of these comments, the underlying local operator (_ifelse_ and _cumsum_) returns an incorrect number of non-zeros for some reason that I haven't discovered yet. The other FIXME comment relates to an incorrect computation of instructions like _sum(X!=0)_ which is introduced by rewriting the constructed HOP.
NOTE: This PR is based on #1862 . Hence, we should wait with the merge until the other PR has been resolved.

Thanks for review :)

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederationMap.java
@@ -658,6 +658,14 @@ public class FederationMap {
 			if(!overlap)
 				iter.remove();
 		}
+
+		boolean rowPartitioned = this.getType().isType(FType.ROW)
+			|| Arrays.stream(ret.getFederatedRanges()).allMatch(range -> range.getSize(1) == ret.getMaxIndexInRange(1));
+		boolean colPartitioned = this.getType().isType(FType.COL)
+			|| Arrays.stream(ret.getFederatedRanges()).allMatch(range -> range.getSize(0) == ret.getMaxIndexInRange(0));
+		if(rowPartitioned && colPartitioned)
+			ret.setType(FType.FULL);
+
 		return ret;
 	}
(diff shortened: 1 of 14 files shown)


Baunsgaard (approved):
LGTM

I especially like the tests. They test many operations in one go violating the single element testing, but it is clear what is going on.

ywcb00:
Thanks for the review @Baunsgaard :)
I also like that aspect of the tests. However, I don't like the copied code parts from _DMLScript_. Any idea how to prevent  this code duplication? Or do you think that it's fine like this?

Baunsgaard:
> Thanks for the review @Baunsgaard :) I also like that aspect of the tests. However, I don't like the copied code parts from _DMLScript_. Any idea how to prevent this code duplication? Or do you think that it's fine like this?

If you want to avoid the duplicate maybe extending the class can help. 
Unfortunately sometimes the easy way is to duplicate some code in the testing. And it did not seem excessive in this case.

mboehm7:
LGTM too - I'll merge it in and fix the remaining issues in the process.

PR #1867, opened 2023-07-17
[SYSTEMDS-3542] Compressed transfer to/from federated workers

Jaybit0:
This PR adds the option to add compression to the federated workers. Right now, only `none` and `zlib` are supported compression strategies that can be set in the config file at `sysds.federated.compression`. For benchmarking, some extra pipelines have been added (might not be the best way to get the statistics).


src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
```diff
@@ -22,14 +22,14 @@
 import java.io.Serializable;
 import java.net.ConnectException;
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
Baunsgaard, line 25: No wild card imports, you can configure your IDE to not do this

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
```diff
@@ -238,12 +233,18 @@ private static ChannelInitializer<SocketChannel> createChannel(InetSocketAddress
...
 				if(ssl)
 					cp.addLast(createSSLHandler(ch, address));
 				if(timeout > -1)
 					cp.addLast(new ReadTimeoutHandler(timeout));
-				cp.addLast(FederationUtils.decoder(), new FederatedRequestEncoder(), handler);
+				// cp.addLast(FederationUtils.decoder(), new FederatedRequestEncoder(), handler);
+				compressionStrategy.ifPresent(strategy -> cp.addLast(strategy.left));
+				cp.addLast(FederationUtils.decoder());
+				compressionStrategy.ifPresent(strategy -> cp.addLast(strategy.right));
+				cp.addLast(new FederatedRequestEncoder());
```
Baunsgaard, line 246: Is it correct when it always adds the decoder even if no compression is specified? 
Should it not only conditionally ad the decoder?
Or is this what the if Present does?

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorker.java
```diff
@@ -21,18 +21,24 @@
...
 import java.io.Serializable;
 import java.security.cert.CertificateException;
+import java.util.Optional;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLException;
 
+import io.netty.channel.*;
```
Baunsgaard, line 31: no wildcard imports

src/main/java/org/apache/sysds/runtime/controlprogram/federated/compression/CompressionEncoder.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.sysds.runtime.controlprogram.federated.compression;
```
Baunsgaard, line 1: License

src/test/scripts/functions/federated/io/FederatedCompressionTest.dml
```diff
@@ -0,0 +1,24 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+x = read($fedmatrix)
+x = x+1
+write(x, $out)
```
Baunsgaard, line 24: new line in end of file

src/test/scripts/functions/federated/io/FederatedCompressionTest1Reference.dml
```diff
@@ -0,0 +1,24 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#-------------------------------------------------------------
+
+x = read($fedmatrix)
+x = x+1
+write(x, $out)
```
Baunsgaard, line 24: newline in end of file

src/test/scripts/functions/federated/io/config/NoneCompressionConfig.xml
```diff
@@ -0,0 +1,29 @@
...
+<root>
+    <!-- The number of theads for the spark instance artificially selected-->
+    <sysds.local.spark.number.threads>2</sysds.local.spark.number.threads>
+    <!-- The timeout of the federated tests to initialize the federated matrixes -->
+    <sysds.federated.initialization.timeout>2</sysds.federated.initialization.timeout>
+    <!-- The timeout of each instruction sent to federated workers -->
+    <sysds.federated.timeout>128</sysds.federated.timeout>
+    <!-- sets the federated compression strategy (none, zlib) -->
+    <sysds.federated.compression>none</sysds.federated.compression>
+</root>
```
Baunsgaard, line 29: newline

src/test/java/org/apache/sysds/test/functions/federated/io/FederatedCompressionTest.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Test
+    public void testFederatedReadWriteCompressionStrategies() {
+        federatedReadWriteCompression();
+    }
+
+    public void federatedReadWriteCompression() {
+        System.out.println("CompressionStrategy: " + compressionStrategy);
+        System.out.println("Dim: " + dim);
```
Baunsgaard, line 92: try to avoid printing in the tests, to reduce the clutter when the test is run in github.

Baunsgaard (commented):
LGTM

The PR looks good, my comments are mainly cosmetic, but should be fixed before merge.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
Jaybit0: Exactly, the consumer to add the decoder / encoder will only be executed if the optional value is not empty (see [java docs](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-))

src/test/java/org/apache/sysds/test/functions/federated/io/FederatedCompressionTest.java
Jaybit0: Should I remove all printing (including the Federated Compression Statistics)? There, helpful information regarding the compression performance (duration and ratio) is printed to validate the compression strategy. Or should I use a Logger for that?

Jaybit0:
Apparently my test took a long time (assuming this is the reason for cancellation and 1h30m execution time). However, when running my test locally, I do not seem to have any problems (the test finishes in less than 10s). I'm not sure why this happens but it seems like LZ4 compression does not work properly here (according to the logs). Should I remove this compression strategy and try again?

src/test/java/org/apache/sysds/test/functions/federated/io/FederatedCompressionTest.java
Baunsgaard: Sure you can use the logger for this. 
Only requirement is that there is no prints if the test is behaving as it should, and informative prints if it fails.
Remember that there is over half a million tests and if each print... then that is to much printing.

Baunsgaard:
> Apparently my test took a long time (assuming this is the reason for cancellation and 1h30m execution time). However, when running my test locally, I do not seem to have any problems (the test finishes in less than 10s). I'm not sure why this happens but it seems like LZ4 compression does not work properly here (according to the logs). Should I remove this compression strategy and try again?

The test passed the second time. Sometimes, unfortunately, the testing instances encounter some difficulties, and waits for a thread that never finishes, making our tests timeout, In a second pass it did not so all is fine for your PR.

Baunsgaard:
Please mark the comments as resolved, if you are done with them. Thanks!

PR #1870, opened 2023-08-01
[SYSTEMDS-3579] optimisation for row sum, col sum, full sum, matrix multiplication + serialisation for deduplicated denseblock 

e-strauss:
- optimisation for row sum, col sum and full sum for deduplicated denseblock
- optimisation for matrix multiplication if left-hand side is a deduplicated denseblock 
- implemented specialised serialisation for deduplicated denseblock


phaniarnab:
Thanks @e-strauss for the support for the *sum instructions and serialization.
I removed the unused includes and moved the special handling code to separate util methods.

PR #1879, opened 2023-08-09

regaleo605:
This patch enables systemds to be able to impute missing value using KNN-algorithm. We calculate the similairy or distance between all pairs of records using the euclidean distances. The first method brute forces the imputation using the dist() method. However, this could leads to an expensive computation. Therefore, we proposed 2 other methods, the second method split the number of records (potentially large) and compute the distances with missing records(hopefully small). The third method is similar to the second method. However, we create a subset from the number of records to compute with missing records(large).


scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+#
+#-------------------------------------------------------------
+
+# Impute the data by using KNN-algorithm and finding the nearest neighbor using euclidean distance
+# Currently only for single column with multiple missing values and top 1 neighbor
+#
+# INPUT:
+# -------------------------------------------------------------------------------------
+# data    Data Matrix (numerical)
+# method  methods of calculating the KNN-algorith depending on the size of data and missing values
```
mboehm7, line 28: Include the valid method names in the documentation, where you pick meaningful names other than method2 and method3

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+# -------------------------------------------------------------------------------------
+#
+# OUTPUT:
+# -----------------------------------------------------------------------------------
```
mboehm7, line 32: please align the columns and surrounding lines appropriately.

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+# result     imputed dataset
+# -----------------------------------------------------------------------------------
+
+
+#method 1
```
mboehm7, line 37: move these documentation blocks into the documentation so they are automatically picked up when we generate the python API and documentation.

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+#impute By Mean, really small, coarse grained-operation
+#Extract Top 1 distances, impute the respective value
+#dist runtime * #features
```
mboehm7, line 40: this (dist runtime * #features) seems incorrect as the distance computation is already O(N2*#features).

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+m_imputeByKNN = function(Matrix[Double] data, String method = "default")
+return(Matrix[Double] result)
+{
+  print("KNN-IMPUTATION SCRIPT")
+
+  #Test data, initial data, will be replaced later with variable data
+  first = matrix ("1 3 4 8", rows = 4, cols = 1)
+  second = matrix ("2 4 6 7", rows = 4, cols = 1)
+  third = matrix ("4 NaN 5 NaN", rows = 4, cols = 1)
+  test_matrix = cbind(first,second,third)
```
mboehm7, line 59: This test data cannot be in the builtin function, which is supposed to be general purpose. Move this to the tests.

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+
+  #Add more data for more test cases
+  fifth = matrix("5 6 7", rows = 1, cols = 3)
+  sixth = matrix("3 4 5", rows = 1 , cols = 3)
+  tm2 = rbind(test_matrix, fifth,sixth)
+
+  #Create a mask for placeholder and to check for missing values
+  masked = is.nan(tm2)
+  print(toString(tm2))
```
mboehm7, line 68: remove such print statements, add a 'verbose' flag and if verbose output is requested print appropriate debug information.

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+    R = table(I2,1,imputedValue,odim1 = nrow(tm2), odim2=1)
+
+    #Replace the masked column with to be imputed Value
+    masked[,as.scalar(missing_col)] = masked[,as.scalar(missing_col)] * R
+
+    #Impute the value
+    result = replace(target = tm2, pattern = NaN, replacement = 0)
+    result = result + masked
+    print("Result method1")
+    print(toString(result))
```
mboehm7, line 119: move common code outside the method-specific branches (at the end of the builtin function).

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+    I = (rowSums(is.nan(tm2))!=0)
+    missing = removeEmpty(target=filled_matrix, margin="rows", select=I)
+
+    Y = (rowSums(is.nan(tm2))==0)
+    M3 = removeEmpty(target=filled_matrix, margin = "rows", select = Y)
+
+    #Create a random subset
+    random_matrix = ceiling(rand(rows = nrow(M3), cols = 1, min = 0, max = 1, sparsity = 0.5, seed = 33))
+    #ensure that random_matrix has at least 1 value
+    if(as.scalar(colSums(random_matrix)) < 1) { random_matrix = matrix(1, rows = nrow(M3), cols = 1)}
```
mboehm7, line 169: avoid such single line if then blocks

scripts/builtin/imputeByKNN.dml
```diff
@@ -0,0 +1,208 @@
...
+    print("Result method3")
+    print(toString(result))
+  } else {
+    print("Method is unknown or not yet implemented")
+  }
+
+#Default Results to be replaced with variable result
+result = data
+}
+
```
mboehm7, line 207: avoid multiple empty lines after end of builtin function

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImputeKNNTest.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String TEST_CLASS_DIR = TEST_DIR + BuiltinImputeKNNTest.class.getSimpleName() + "/";
+    @Override
+    public void setUp() {
+        TestUtils.clearAssertionInformation();
+        addTestConfiguration(TEST_NAME, new TestConfiguration(TEST_CLASS_DIR, TEST_NAME, new String[] {"C"}));
+    }
+
+    @Test
+    public void test()throws IOException{
+        runImputeKNN(Types.ExecType.CP);
```
mboehm7, line 44: add a second test with exectype Spark

mboehm7: Also replicate the CP/Spark test for each method.

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImputeKNNTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private void runImputeKNN(Types.ExecType instType) throws IOException {
+        Types.ExecMode platform_old = setExecMode(instType);
+        try {
+            loadTestConfiguration(getTestConfiguration(TEST_NAME));
+            String HOME = SCRIPT_DIR + TEST_DIR;
+            fullDMLScriptName = HOME + TEST_NAME + ".dml";
+            programArgs = new String[] {}; //
+            runTest(true, false, null, -1);
```
mboehm7, line 54: try to generate meaningful inputs and compare the results. You could for example check that the sum of the output matrix is roughly the same for all three methods. (computed with the brute force dist method)

mboehm7 (commented):
Thanks @regaleo605 for your contribution. This is a good start but needs a partially rework with regard to testing (no test data generation inside the builtin function, and proper result comparisons). Could you please address the following detailed comments and let me know (via a comment) once it's ready for a second review round. Thanks.

regaleo605:
Thank you @mboehm7 for the helpful feedbacks. However, I have a question regarding the test. So in the newest version I tried to write the output to 2 dml files and then compare the matrices using the test utils compare matrices with the tolerance. Do I understand correctly that it should not have any output or did I? Furthermore from what I have observed, the results from 3 different methods resulted in three different resutls. Perhaps either the test data is too small or because method3 contains randomly selected subset of rows which is why there are some random factors. What could be a good tolerance numbers?

scripts/builtin/imputeByKNN.dml
regaleo605: Is it fine to use "default", "small", and "large" or do you have other suggestions?

mboehm7: no, give meaningful names: such as dist, dist_missing, dist_sample - and they need to be explained in the header documentation.

mboehm7:
I would recommend to invoke a single test script with generated data (see other tests), pass the method as an argument, and check for example, that the sum of imputed missing values comes close to the expected value (which should be true for all methods and could can play around with the epsilon how far they can differ).

mboehm7:
Let me know once the PR is ready for another full review.

regaleo605:
@mboehm7  I have already addressed the previous feedbacks and did some test. As a result, the tests were unsatisfactory between dist and dist_missing/dist_sample, after looking through the code, I found that the problem is with the method of rowMins/RowIndexMin, Could you please help clarify why I am getting number that is not even from the calculation of euclidean distances? Interestingly these numbers are all roughly the same and as such the indeces were the same.

mboehm7:
Sure, but could you please clarify what you mean by "why I am getting number that is not even from the calculation of euclidean distances"?

Also for the tests, another good approach would be to generate random data in the java test, and replicate the a few times so you know what the top-1 nearest neighbor (and the sum of imputed values) is.

regaleo605:
After calculating the euclidean distance, I got a matrix of n rows, where n are the number of missing values. Each row contains the euclidean distances. For example the first row contains (-50,-3, 0 ,1 ,-20) but when I call the rowMins I got a number that is not in the example say -100 instead of -50.

mboehm7:
Hmm, that would be a bug and I would fix it immediately. For which intermediate (please indicate the line number) did you observe this behavior, right now there is not a single rowMins in there (while rowIndexMin returns the position of the minimum value per row).

regaleo605:
Line 107-108 and line 148-149. I noticed the problem comes from rowMins where I got an index that is greater than the dataset. Then I added print(toString(rowMins(t(D)))) after line 108/149 in my local version to check and the mentioned problem occured. I believe rowIndexMin leverage rowMin method.

mboehm7:
OK, so here is what happened: toString() prints by default only the first 100 rows and columns, leading to the rowMins not showing up in the output and rowIndexMax giving you an index larger than the truncated printed version. You can parameterize toString(t(D), rows=1000, cols=1000) and continue with your debugging. 

We do give a warning whenever we truncate such outputs, but unfortunately our tests run in log level ERROR which omits this crucial information. @Baunsgaard: since we recently discussed a second issue where this caused problems, you might want to change the log level back to WARN.

Baunsgaard:
> OK, so here is what happened: toString() prints by default only the first 100 rows and columns, leading to the rowMins not showing up in the output and rowIndexMax giving you an index larger than the truncated printed version. You can parameterize toString(t(D), rows=1000, cols=1000) and continue with your debugging.
> 
> We do give a warning whenever we truncate such outputs, but unfortunately our tests run in log level ERROR which omits this crucial information. @Baunsgaard: since we recently discussed a second issue where this caused problems, you might want to change the log level back to WARN.

well, this is up for debate, since some tests would write many many warning messages if we set the default to warning.
A middle ground is to set to warning as default, and we then in GitHub actions set it to a more restrictive version.

regaleo605:
Thank you for clearing up my confusion. It seems that I needed to test with other datasets to check whether this is only a coincidence or there is something wrong with the euclidean distance calculation. I will notify you if I figure something out or have some other problem.

regaleo605:
I fixed the euclidean distance, Could you please do a code review?

mboehm7:
LGTM - thanks @regaleo605. During the merge, I fixed remaining warnings, simplified the test script (which also fixes the spark test), update the docs, and added TODOs for future improvements regarding imputation for all columns with missing values, and proper randomization (where the passed seed is a root for all required seeds).

PR #1881, opened 2023-08-09
[MINOR] Adding decision tree to cleaning pipelines evaluation

Shafaq-Siddiqi:
Please do not merge as this PR require some cleanups.


scripts/builtin/vectorToCsv.dml
```diff
@@ -32,10 +32,10 @@
 # indexes  indexes
 # ----------------------------------------------------------------------------------------
 
-m_vectorToCsv = function(Matrix[Double] mask)
+m_vectorToCsv = function(Matrix[Double] mask = as.matrix(0), Integer n=1)
```
Baunsgaard, line 35: the error is here, or rather the parsers we have in Python does not like this syntax.

Baunsgaard (commented):
Hi @Shafaq-Siddiqi 

I think the error is in the Python parsers. that is not robust for all kinds of syntax of the function definitions.

Shafaq-Siddiqi:
> Hi @Shafaq-Siddiqi
> 
> I think the error is in the Python parsers. that is not robust for all kinds of syntax of the function definitions.

yes, I have added a parameter to an existing function and now the Python parser is failing. I am cleaning the PR to get rid of that parameter and then hopefully everything would work.

PR #1883, opened 2023-08-17
[SYSTEMDS-3545] img_crop_linearized

slnkahveci:
Crop method for row linearized images. This functionality is being developed as part of a student project.


slnkahveci:
@Baunsgaard Could there be a column - row mismatch in the existing img_crop script or test? It gives an error when the values in the test are not equal. This also seems like the problem I'm working on. So I'm not sure if I should look for the issue in my additions or the existing code.

Baunsgaard:
> @Baunsgaard Could there be a column - row mismatch in the existing img_crop script or test? It gives an error when the values in the test are not equal. This also seems like the problem I'm working on. So I'm not sure if I should look for the issue in my additions or the existing code.

Can you be more specific, and also write up the example that does not work, then I can also debug what is going on.
But there is a likely possibility that something does not work correctly when the number of rows and columns of an image is bad, since there are not that many tests of the image primitives.

slnkahveci:
> > @Baunsgaard Could there be a column - row mismatch in the existing img_crop script or test? It gives an error when the values in the test are not equal. This also seems like the problem I'm working on. So I'm not sure if I should look for the issue in my additions or the existing code.
> 
> Can you be more specific, and also write up the example that does not work, then I can also debug what is going on. But there is a likely possibility that something does not work correctly when the number of rows and columns of an image is bad, since there are not that many tests of the image primitives.


It doesn't work when the image is not square, for example for the values:
```
    private final static int rows = 512;
    private final static int cols = 256;
    private final static int x_offset = 12;
    private final static int y_offset = 24;
    private final static float size = 0.6f;
```

I’m suspecting this mismatch because the expected B has flipped indexes:
```
  %%MatrixMarket matrix coordinate real general
  153 153 21112
  1 1 241.32527313744777
  2 1 138.9662793588828
  4 1 233.29725808796292
  5 1 153.8881182614073
  ...
```

Baunsgaard:
> 153 153 21112

And this is with the current? 

and you would call it with something like:

```R
X = rand(rows = 10, cols = 12, min = 0, max = 9, seed=132)
X = ceil(X)
print(toString(X))
Xc = img_crop(X, 5, 4, 2, 3)
print(toString(Xc))
```

It correctly slice out a sub region:

```txt 
1.000 6.000 3.000 4.000 7.000 4.000 4.000 5.000 8.000 2.000 2.000 9.000
7.000 6.000 7.000 8.000 5.000 8.000 9.000 3.000 7.000 6.000 5.000 2.000
2.000 9.000 7.000 7.000 8.000 9.000 6.000 5.000 7.000 3.000 5.000 5.000
4.000 3.000 1.000 4.000 9.000 2.000 3.000 8.000 2.000 1.000 5.000 8.000
4.000 3.000 8.000 7.000 1.000 4.000 3.000 8.000 4.000 7.000 6.000 4.000
7.000 7.000 8.000 9.000 6.000 1.000 4.000 7.000 6.000 8.000 4.000 6.000
8.000 5.000 6.000 8.000 7.000 9.000 6.000 6.000 9.000 8.000 3.000 4.000
1.000 6.000 6.000 4.000 2.000 9.000 9.000 9.000 2.000 5.000 8.000 2.000
2.000 3.000 2.000 5.000 8.000 4.000 4.000 8.000 4.000 5.000 3.000 6.000
6.000 9.000 1.000 2.000 4.000 2.000 8.000 9.000 4.000 8.000 3.000 1.000

1.000 4.000 7.000 6.000 8.000
9.000 6.000 6.000 9.000 8.000
9.000 9.000 9.000 2.000 5.000
4.000 4.000 8.000 4.000 5.000
```

Baunsgaard:
> > 153 153 21112
> 
> And this is with the current?
> 
> and you would call it with something like:
> 
> ```r
> X = rand(rows = 10, cols = 12, min = 0, max = 9, seed=132)
> X = ceil(X)
> print(toString(X))
> Xc = img_crop(X, 5, 4, 2, 3)
> print(toString(Xc))
> ```
> 
> It correctly slice out a sub region:
> 
> ```
> 1.000 6.000 3.000 4.000 7.000 4.000 4.000 5.000 8.000 2.000 2.000 9.000
> 7.000 6.000 7.000 8.000 5.000 8.000 9.000 3.000 7.000 6.000 5.000 2.000
> 2.000 9.000 7.000 7.000 8.000 9.000 6.000 5.000 7.000 3.000 5.000 5.000
> 4.000 3.000 1.000 4.000 9.000 2.000 3.000 8.000 2.000 1.000 5.000 8.000
> 4.000 3.000 8.000 7.000 1.000 4.000 3.000 8.000 4.000 7.000 6.000 4.000
> 7.000 7.000 8.000 9.000 6.000 1.000 4.000 7.000 6.000 8.000 4.000 6.000
> 8.000 5.000 6.000 8.000 7.000 9.000 6.000 6.000 9.000 8.000 3.000 4.000
> 1.000 6.000 6.000 4.000 2.000 9.000 9.000 9.000 2.000 5.000 8.000 2.000
> 2.000 3.000 2.000 5.000 8.000 4.000 4.000 8.000 4.000 5.000 3.000 6.000
> 6.000 9.000 1.000 2.000 4.000 2.000 8.000 9.000 4.000 8.000 3.000 1.000
> 
> 1.000 4.000 7.000 6.000 8.000
> 9.000 6.000 6.000 9.000 8.000
> 9.000 9.000 9.000 2.000 5.000
> 4.000 4.000 8.000 4.000 5.000
> ```

That said, it does slice out from bottom right, not top left. So it is a bit confusing.

slnkahveci:
@Baunsgaard 

> > 153 153 21112

> And this is with the current?

It was the output from 256x256 input.

DML script functions as you’ve described. The issue is with the test. It doesn’t behave the same way as the R script. There is a mismatch in the rotation of the output matrix.

Input A.mtx: 
```
%%MatrixMarket matrix coordinate real general
512 256 131072
1 1 191.03824879907646
1 2 228.8056713992367
1 3 89.73826829980825
```
 

Output expected/B from R:
```
%%MatrixMarket matrix coordinate real general
153 307 42252
1 1 28.420986211683942
2 1 153.74324014564436
3 1 41.8703704404483
```

out/B.mtd from .dml:
```
…
    "rows": 307,
    "cols": 153,
…
```

Baunsgaard:
> ```
> %%MatrixMarket matrix coordinate real general
> 512 256 131072
> 1 1 191.03824879907646
> 1 2 228.8056713992367
> ```

It sounds like the output is transposed somehow in one code base and not the other since rows and columns are swapped.

scripts/builtin/img_crop_linearized.dml
```diff
@@ -0,0 +1,64 @@
...
+    img_out = img_in
+  }
+  else {
+    mask = matrix(0, rows=orig_h, cols=orig_w)
+    temp_mask = matrix(1, rows=h , cols=w )
+    mask[start_h:end_h, start_w:end_w] = temp_mask
+
+    linear_mask = matrix(mask, rows=1, cols=orig_w * orig_h)
+    
+    img_out = matrix(removeEmpty(target=(matrix(img_in+1, nrow(img_in), ncol(img_in))), margin="cols", select=linear_mask) - 1, nrows, w * h)
```
Baunsgaard, line 61: I'm not sure, but i think using remove empty is the wrong choice here. 
I suggest you use slicing instead to slice out specific columns of pixels instead.
But i could be wrong.

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImageCropLinTest.java
```diff
@@ -0,0 +1,120 @@
...
+	private final static String TEST_CLASS_DIR = TEST_DIR + BuiltinImageCropTest.class.getSimpleName() + "/";
+
+	private final static double eps = 1e-10;
+
+    private final static int s_cols = 27;
+    private final static int s_rows = 8;
+	private final static int rows = s_cols * s_rows;
+	private final static int cols = 13;
+	private final static double spSparse = 0.1;
+	private final static double spDense = 0.9;
```
Baunsgaard, line 44: here change the input cases to a parameterized list. line in #1903

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImageCropLinTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+
+    @Override 
+    public void setUp() {
+        addTestConfiguration(TEST_NAME,new TestConfiguration(TEST_CLASS_DIR, TEST_NAME,new String[]{"B"}));
+    }
+
+    @Test
+    public void testImageCropMatrixDenseCP() {runImageCropLinTest(false, ExecType.CP);
```
Baunsgaard, line 58: fix formatting, here the internals of the method should be on the next line.

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImageCropLinTest.java
```diff
@@ -0,0 +1,120 @@
...
+            programArgs = new String[]{"-nvargs",
+					"in_file=" + input("A"), "out_file=" + output("B"),
+					"x_offset=" + x_offset, "y_offset=" + y_offset, "cols=" + cols, "rows=" + rows, 
+                    "s_cols=" + s_cols, "s_rows=" + s_rows, "new_w=" + new_w, "new_h=" + new_h};
+
+            
+            fullRScriptName = HOME + TEST_NAME + ".R";
+            rCmd = "Rscript" + " " + fullRScriptName + " " + inputDir() + " " + expectedDir() 
+            + " " + size + " " + cols + " " + rows + " " + s_cols + " " + s_rows + " " + x_offset + " " + y_offset
+            + " " + new_w + " " + new_h;
```
Baunsgaard, line 98: I would prefer if we did not use R to verify instead we verify by knowing the output and calculating it right here. 

The crop should be simple to make with a few slice instructions in this code.

src/test/java/org/apache/sysds/test/functions/builtin/part1/BuiltinImageCropLinTest.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        finally {
+            rtplatform = platformOld;
+        }
+
+
+
+    }
+
+}
```
Baunsgaard, line 120: new line in the end.

Baunsgaard (commented):
Minor comments looks good in general. 
Is it still not working?

scripts/builtin/img_crop_linearized.dml
slnkahveci: Slicing out causes runtime errors and flakes for larger images/wider inputs (e.g. the last test parameters). I think restoring it to removeEmpty is a better option.

PR #1896, opened 2023-09-06
[MINOR] Frame Float detection refinement

Baunsgaard:
This commit refine the detection and selection of float values in the schema detection algorithm located in:

src/main/java/org/apache/sysds/runtime/frame/data/lib/FrameUtil.java

To improve the performance a custom matcher have been made to avoid using regexes if possible, and we try to avoid parsing the float value from the string representation if at all possible. The implementation is not completely fool proof and does not consider many adversarial inputs.


Baunsgaard:
depends on #1895

PR #1910, opened 2023-09-18
[SYSTEMDS-3545] img_translate_linearized

Barbar1432:
Hey Sebastian, 
I have closed the pr for this function and made a new one aiming my two PR's would be seperate but it seems they still share the same commit history. Other than that the translate operator in the first PR had some serious performance issues. I have tried to change it as much as possible. I did get rid of one of the loops. In the initial implementation we were assigning values to every index one by one which was inefficient for large matrices. This is not the case anymore. I also used parfor for the outer loop which helped with the performance. I have tried to compare it with the non linear translate operator with random matrices that are 512x512 size. It seems aproximately when we call it with more than 10 images . the linear function seems to catch up to the non linear one . I did this by comparing the system DS execution time that appears after the terminal after compiling the script. Is there a better way of comparing ? I think I could also compare with the time command but it gives 3 different measurements ( real, sys and .. ) I didnt know which one I should primarly focus on. I also added the dml script for test and the java test and looking forward to your feedback.


Barbar1432:
Hey Sebastian 
Also in this one should I put the ignore flag and recommit or are we going to run once more ?

PR #1911, opened 2023-09-19
[SYSTEMDS-3545] img_transform_linearized

slnkahveci:
Transform method for row linearized images. This functionality is being developed as part of a student project.


src/test/java/org/apache/sysds/test/functions/pipelines/ImageTransformLinTransformed.csv
Baunsgaard, line 1: if possible can we remove this file, or chose a smaller size for a test image.

scripts/builtin/img_transform_linearized.dml
```diff
@@ -0,0 +1,88 @@
...
+# The Linearized Image Transform function applies an affine transformation to linearized images.
+# Optionally resizes the image (without scaling).
+# Uses nearest neighbor sampling.
+#
+# INPUT:
+# -------------------------------------------------------------------------------------------
+# img_in       Linearized input images as 2D matrix with top left corner at [1, 1]
+# out_w        Width of the output matrix
+# out_h        Height of the output matrix
+# a,b,c,d,e,f  The first two rows of the affine matrix in row-major order
```
Baunsgaard, line 31: each variable have to have a new line in the  documentation

Or maybe these should be combined into the translation matrix by the caller?

src/test/resources/expected/ImageTransformLinInput.csv
Baunsgaard, line 1: similar here, if we can reduce this file or remove it it would be good.

Baunsgaard (commented):
looks okay.

scripts/builtin/img_transform_linearized.dml
slnkahveci: This was the same interface and the documentation format from the existing implementation.
  
Should I write the descriptions like so?:
a = The element located at [1,1] in the affine matrix

Baunsgaard: well, then the existing is wrong as well.

We have parsers that make the python interface code. it does support this syntax (surprised me now), but does not follow the rules of mapping variables to parameters, and therefore must be breaking something else.

image

https://apache.github.io/systemds/api/python/api/operator/algorithms.html#systemds.operator.algorithm.img_crop

PR #1922, opened 2023-10-16
[MINOR] MatrixBlock Sparse Optimizations

Baunsgaard:
This commit adds a specialized kernel for Sparse %*% Ultra Sparse Matrix multiplication, that previously
accessed all cells in rows where there could be a multiplication in the left matrix no matter if it was sparse or not. 

The optimization Improve default calls to Decision trees on kaggle playground-series-s3e23 from 100 sec to 30.

Also included is sparse handling of  Binary cell operations and a small optimization to Commons math eigen to skip a indirect call.


Baunsgaard:
Interestingly, the new list of matrix multiplication tests is have a speed up from  67 sec to 31 sec to execute all the multiplication tests. The main speedup comes from a fix to non zero counts, and the ultra sparse matrix multiplication speedups.

PR #1923, opened 2023-10-16
[SYSTEMDS-3631] Fix Integer casting bug in as.frame

Baunsgaard:
This commit fixes the bug by solidifying the isType detection to not allow casting to int on integers above Integer.MAX_VALUE and not cast to long on above Long.MAX_VALUE. Also added is a few tests to verify the behavior in some edge cases.


mboehm7:
Looks good to me and thanks for adding the test such that these issues don't happen again.

mboehm7:
Feel free to merge it in yourself.

PR #1925, opened 2023-10-18

christinadionysio:
This patch fixes a bug in the sampling method of the missing value imputation using KNN functionality. To identify the bug, the existing test was extended to cover the third method (sample_dist). Considering only records that do not contain any missing values for the nearest neighbor calculation solves the issue.

### DIFF
--- a/scripts/builtin/imputeByKNN.dml
+++ b/scripts/builtin/imputeByKNN.dml
@@ -85,7 +85,7 @@ m_imputeByKNN = function(Matrix[Double] X, String method="dist", Int seed=-1, Do
     imputedValue = compute_missing_values(X, filled_matrix, seed, sample_frac)
   }
   else {
-    print("Method is unknown or not yet implemented")
+    stop("Method is unknown or not yet implemented")
   }
 
   #Impute the value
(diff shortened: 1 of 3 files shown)


scripts/builtin/imputeByKNN.dml
```diff
@@ -66,103 +64,76 @@ m_imputeByKNN = function(Matrix[Double] X, String method="dist", Int seed=-1, Do
...
-    #Update the mask value
-    masked = t(imputedValue) * masked
+    imputedValue = compute_missing_values(X, filled_matrix, seed, 1.0)
   }
   else if(method == "dist_sample"){
     #assuming large missing values
+    imputedValue = compute_missing_values(X, filled_matrix, seed, sample_frac)
+  }
+  else {
+    print("Method is unknown or not yet implemented")
```
Baunsgaard, line 88: change this to stop instead of print

scripts/builtin/imputeByKNN.dml
```diff
@@ -66,103 +64,76 @@ m_imputeByKNN = function(Matrix[Double] X, String method="dist", Int seed=-1, Do
...
+  }
+
+  #Impute the value
+  result = replace(target=X, pattern=NaN, replacement=0)
+  result = result + (imputedValue * is.nan(X))
+}
+
+compute_missing_values = function (Matrix[Double] X, Matrix[Double] filled_matrix, Int seed, Double sample_frac)
+return (Matrix[Double] imputedValue)
```
Baunsgaard, line 97: indent with 4 spaces on the return or break lines in the function definitions.
(just to keep it consistent with other builtins.)

Baunsgaard (commented):
LGTM, i am not sure how detailed the tests are tho.

Baunsgaard:
how fast is it?

christinadionysio:
> how fast is it?

It works well for this small dataset, but I will add some experiments involving larger datasets in a different PR.

Baunsgaard:
Thanks for the commit!

I would suggest you update your git profile locally, to use your new email,
but i committed it with your TU graz one for now.

PR #1926, opened 2023-10-19
[MINOR] DML Hadoop Startup skip

Baunsgaard:
This PR contains a bit of a controversy.

At startup the first thing we do is to call Hadoop to parse the Hadoop specific arguments.
This takes ~ 200 ms at startup before we start our timing of SystemDS.

before:

```txt
1,6187 +- 0,0766 seconds time elapsed  ( +-  4,73% )
```

after:

```txt
  1,4366 +- 0,0496 seconds time elapsed  ( +-  3,46% )
```

The downside of this modification is that Hadoop arguments are not parsed to the Hadoop file system, that said, i am unaware if the arguments actually work at the moment because ---to the best of my knowledge---we have not tested if the system work when there are more than one HDFS system available.

The script used for the test is :

```R
print("Hello World!")
```


Baunsgaard:
@mboehm7 , for when you have time.

PR #1930, opened 2023-10-23
[WIP] Add hadoop native binaries for IO

Baunsgaard:
This PR adds native binaries to the python test.

It is WIP, since i need to add the files somewhere for download,
therefore i thought i could add it to the webpage.
But the webpage, is, slightly out of date, and needs more time to fix, since there are many packages that are out of date and have "critical vulnerabilities."


Baunsgaard:
Closing this pull request because it is going in the wrong direction. 

The correct way is to include the correct files inside the docker image.

PR #1940, opened 2023-11-03
ResNet Basic Block Implementation & Testing

MaximilianSchreff:
First pull request for the implementation of the ResNet architecture as a builtin network of SystemDS.
ResNets consists of residual blocks. There are two types of blocks - basic blocks and bottleneck blocks. Basic blocks are used in the smaller ResNets 18 and 34.
These commits add the forward pass of the basic blocks and provide testing for them as a NNComponentTest.
The testing is done by, first, testing for the correct handling of dimension throughout the residual bock. Secondly, we test for correct computation of the actual values of the output by using the torchvision.models.resnet.BasicBlock of PyTorch which is there implementation of the basic block for the ResNets 18 and 34. We let PyTorch randomly initialize all weights and then hard-code these weights into the component test with the expected values as well.


Baunsgaard:
What is a realistic input size? in ILSCRC2012 (Image net) an example size could be 500 x 375 pixels.

ILSVRC2012_val_00000003

Also  if you could write the '-stats' output on running this it would be great!

MaximilianSchreff:
Realisitc input size is HxWx3 with H and W. For the network to make a sensible prediction H and W should be bigger than 224. So, 500 x 375 pixels is fine.

MaximilianSchreff:
Here is the stat output when running a single forward pass of the basic residual block: 

SystemDS Statistics:
Total elapsed time:		14,218 sec.
Total compilation time:		1,144 sec.
Total execution time:		13,074 sec.
Cache hits (Mem/Li/WB/FS/HDFS):	97/0/0/0/0.
Cache writes (Li/WB/FS/HDFS):	3/33/0/0.
Cache times (ACQr/m, RLS, EXP):	0,000/0,001/0,006/0,000 sec.
HOP DAGs recompiled (PRED, SB):	0/7.
HOP DAGs recompile time:	0,037 sec.
Total JIT compile time:		3.086 sec.
Total JVM GC count:		2.
Total JVM GC time:		0.023 sec.
Heavy hitter instructions:
  #  Instruction          Time(s)  Count
  1  basic_block_forward   12,861      1
  2  conv2d_bias_add        6,923      3
  3  forward                5,463      3
  4  uacvar                 2,532      3
  5  bias_multiply          1,174      6
  6  bias_add               0,940      6
  7  uacmean                0,640      3
  8  max                    0,315      2
  9  rand                   0,208      9
 10  +                      0,142     16

The input has is of shape 32x64x120x100 (N = 32, C = 64, Hin = 120, Win = 100). The channel size is expanded to 128 within the residual block which is a realisitic residual block. There are 3 sequential conv2d forward passes happening which is the baseline in terms of runtime.

Baunsgaard:
> Here is the stat output when running a single forward pass of the basic residual block:
> 
> SystemDS Statistics: Total elapsed time: 14,218 sec. Total compilation time: 1,144 sec. Total execution time: 13,074 sec. Cache hits (Mem/Li/WB/FS/HDFS): 97/0/0/0/0. Cache writes (Li/WB/FS/HDFS): 3/33/0/0. Cache times (ACQr/m, RLS, EXP): 0,000/0,001/0,006/0,000 sec. HOP DAGs recompiled (PRED, SB): 0/7. HOP DAGs recompile time: 0,037 sec. Total JIT compile time: 3.086 sec. Total JVM GC count: 2. Total JVM GC time: 0.023 sec. Heavy hitter instructions:
> # Instruction Time(s) Count
> 
> 1 basic_block_forward 12,861 1 2 conv2d_bias_add 6,923 3 3 forward 5,463 3 4 uacvar 2,532 3 5 bias_multiply 1,174 6 6 bias_add 0,940 6 7 uacmean 0,640 3 8 max 0,315 2 9 rand 0,208 9 10 + 0,142 16
> 
> The input has is of shape 32x64x120x100 (N = 32, C = 64, Hin = 120, Win = 100). The channel size is expanded to 128 within the residual block which is a realisitic residual block. There are 3 sequential conv2d forward passes happening which is the baseline in terms of runtime.

This sounds painfully slow, have you compared to a single forward ResNet layer pyTorch?
I know that our conv2d_bias_add is slow, I do not recall if we have a native version of this, but if we have that should speed it up.
if you use the -conf flag when executing and add a sysds.native.blas parameter to your configuration maybe we can speed this up.

MaximilianSchreff:
I don't know how to do the configuration. This is SystemDS out of the box.

I didn't compare it to pyTorch but I suspect PyTorch is much faster with my current SystemDS configuration. But with my current configuration and the implementation of layers in SystemDS, there is not much to be optimized or runtime gained. I am already using the builtin conv2d layer and the convolutional layers need to happen the way they do in my implementation. 

Feel free to run this forward pass yourself:
[test_resnet_big.log](https://github.com/apache/systemds/files/13272287/test_resnet_big.log)

Baunsgaard:
> I don't know how to do the configuration. This is SystemDS out of the box.

okay, we will figure it out. https://apache.github.io/systemds/site/run contains some information about MKL in the bottom. 
But the docs for it is a bit... outdated.

> I didn't compare it to pyTorch but I suspect PyTorch is much faster with my current SystemDS configuration. But with my current configuration and the implementation of layers in SystemDS, there is not much to be optimized or runtime gained. I am already using the builtin conv2d layer and the convolutional layers need to happen the way they do in my implementation.

Last time i ran our convolution layers yes, pyTorch was (much) faster, we have not worked on that operation for a while.

> Feel free to run this forward pass yourself: [test_resnet_big.log](https://github.com/apache/systemds/files/13272287/test_resnet_big.log)

ahh, i see.
If you are profiling with this, then we cannot trust the results as much as we want.
I suggest adding a for loop around the call to the basic_block_forward, to allow the JIT compiler to help you.
it should improve the speed of each call. (Java is sometimes very bad if you do not call your things multiple times)

phaniarnab:
Just to clarify, this, and next few PRs are supposed to first bring the functionalities in terms of the popular CNNs. I will meanwhile profile the GPU side of these and compare with Keras/PyTorch.
Do we have an MKL implementation for conv2d? @Baunsgaard?

Baunsgaard:
> Just to clarify, this, and next few PRs are supposed to first bring the functionalities in terms of the popular CNNs. I will meanwhile profile the GPU side of these and compare with Keras/PyTorch. Do we have an MKL implementation for conv2d? @Baunsgaard?

I believe we have one, but is has been a while since we used it.

phaniarnab:
> > Just to clarify, this, and next few PRs are supposed to first bring the functionalities in terms of the popular CNNs. I will meanwhile profile the GPU side of these and compare with Keras/PyTorch. Do we have an MKL implementation for conv2d? @Baunsgaard?
> 
> I believe we have one, but is has been a while since we used it.

Cool. I will have a look.

phaniarnab:
Thanks for the changes @MaximilianSchreff. Now the PR is merged.
I noticed you used the old batch_norm2d instead of the builtin. While I changed the code to use the builtin, the test crashed, probably due to a wrong sparsity handling. Is that why you used the old implementation?
In any case, we need to fix the builtin in the future.

MaximilianSchreff:
Oh right. I completely forgot that.

I don't know why, but it always failed while the old implementation worked without problems. Since I didn't find an example code using the builtin function, I used the old implementation.

PR #1941, opened 2023-11-04
Graph Convolutional Layers - Builtin

MaximilianSchreff:
Adding a built in graph convolutional layer to our nn layer diretory. The graph convolutional layer follows the paper "Semi-Supervised Classification with Graph Convolutional Networks" from Kipf and Welling. This includes the forward and the backward pass of the layer and an example network on how to use the layers. 

We tested the implementation against PyTorch and our implementation computes the same exact values as PyTorch does and handles missing in-bound edges of a node the same way in the spectral convolution. This is also implemented as a component test as a NNComponentTest. There we hard-coded the initialized weights from PyTorch and the result into the component test.


MaximilianSchreff:
About the performance: My machine showed performance issues when testing against PyTorch for very very big inputs.

Stress test:     SystemDS: 340 seconds     -     PyTorch: 32 seconds

The stress test consisted of about 300 forward passes with about 10.000 x 10.000 matrices. This is likely a problem with my setup and not my implementation since the affine layer with the same inputs took 220 seconds. The GCL consists of a simple affine part and a convolutional part with the convolutional part being a lot more complex. So, the implementation is likely quite fast because the complex convolution part makes up less than a third of the runtime.

MaximilianSchreff:
About caching:

It would be a possibility to cache the normalized weights since they are quite complex to compute (counting of every degree and spectral convolution). In the stress testing, this only showed an improvement of 10 seconds from 340 seconds. Also, when testing the caching feature against smaller inputs than the huge inputs of the stress test, it was always slower.
With the added complexity and difficulty to handle cached weights and generally more outputs of the layer, I decided against caching.

Baunsgaard:
> About the performance: My machine showed performance issues when testing against PyTorch for very very big inputs.
> 
> Stress test: SystemDS: 340 seconds - PyTorch: 32 seconds
> 
> The stress test consisted of about 300 forward passes with about 10.000 x 10.000 matrices. This is likely a problem with my setup and not my implementation since the affine layer with the same inputs took 220 seconds. The GCL consists of a simple affine part and a convolutional part with the convolutional part being a lot more complex. So, the implementation is likely quite fast because the complex convolution part makes up less than a third of the runtime.

What is a realistic input size? in ILSCRC2012 (Image net) an example size could be 500 x 375 pixels.

ILSVRC2012_val_00000003

Also good progress, if you could write the '-stats' output it would be great!

Baunsgaard:
> About the performance: My machine showed performance issues when testing against PyTorch for very very big inputs.
> 
> Stress test: SystemDS: 340 seconds - PyTorch: 32 seconds
> 
> The stress test consisted of about 300 forward passes with about 10.000 x 10.000 matrices. This is likely a problem with my setup and not my implementation since the affine layer with the same inputs took 220 seconds. The GCL consists of a simple affine part and a convolutional part with the convolutional part being a lot more complex. So, the implementation is likely quite fast because the complex convolution part makes up less than a third of the runtime.

can you show the '-stats' output of calling it, to indicate where we are using time.
and maybe we want to profile it using a profiler: https://github.com/async-profiler/async-profiler

If you are in doubt how to use it, i can show you in office.

MaximilianSchreff:
I already profiled it. The convolutional layer consists of the linear layer and a convolutional layer. The linear layer takes up more than 70% of the runtime. This means there are probably some issues in my configs that restrict systemds from being faster since the linear part is the very base line of this layer.

Baunsgaard:
> I already profiled it. The convolutional layer consists of the linear layer and a convolutional layer. The linear layer takes up more than 70% of the runtime. This means there are probably some issues in my configs that restrict systemds from being faster since the linear part is the very base line of this layer.

with Linear layer, do you mean a simple fully connected affine layer (aka a matrix multiplication), or something else?

Can you maybe give me an example in code?

MaximilianSchreff:
Yes exactly. There is a matrix multiplication and adding a bias happening. This is what takes up 70% of the performance. 
You can see that in [scripts/nn/layers/graph_conv.dml](https://github.com/apache/systemds/pull/1941/files#diff-cf550d72402ce14bdc5b38206244cbb3152f6c02c5d42e2f066c5d5c0f91a661) in the forward pass function. That are only two lines that use most of the runtime. These two lines cannot be optimized by my layer.

As for the rest of the forward pass, I already took numerous steps to optimize it from, initially, 780 seconds to 340 seconds for the whole layer. This includes:
- Merging the different functions into a big one to combine for-loops, improving parallelization gain
- Optimizing the addition of self loops
- Caching normalization weights (removed it again, since it wasn't faster)

scripts/nn/layers/graph_conv.dml
```diff
@@ -0,0 +1,262 @@
...
+            D_in[as.integer(as.scalar(edge_index[j, 1])), 1]) != 0)
+        {
+            edge_weight[j, 1] = edge_weight[j, 1] * (1 / sqrt(D_in[as.integer(as.scalar(edge_index[j, 2])), 1]*
+                                                              D_in[as.integer(as.scalar(edge_index[j, 1])), 1]))
+        }
+        else
+        {
+            edge_weight[j, 1] = 0.0
+        }
+        X_out[as.integer(as.scalar(edge_index[j, 2]))] += as.scalar(edge_weight[j, 1]) * X_hat[as.integer(as.scalar(edge_index[j, 1]))]
```
Baunsgaard, line 120: all these as.integer(as.scalar(...)) should not be needed.

scripts/nn/layers/graph_conv.dml
```diff
@@ -0,0 +1,262 @@
...
+     *                   not yet have a self loop. Added self loops will have weight 1.
+     *
+     * Outputs:
+     * - X_out: convolved and transformed node features, matrix of shape (n, f_out).
+     */
+    n = nrow(X)
+    m = nrow(edge_index)
+
+    # transform
+    X_hat = X %*% W
```
Baunsgaard, line 60: This is the slow operation?

scripts/nn/layers/graph_conv.dml
```diff
@@ -0,0 +1,262 @@
...
+            edge_weight[j, 1] = 0.0
+        }
+        dOut_agg_rev[as.integer(as.scalar(edge_index[j, 1]))] += as.scalar(edge_weight[j, 1]) * dOut[as.integer(as.scalar(edge_index[j, 2]))]
+    }
+
+    # calculate gradient w.r.t. X (Formula: A_hat^T * dOut * W^T)
+    dX = dOut_agg_rev %*% t(W)
+
+    # calculate gradient w.r.t. W (Formula: X^T * A_hat^T * dOut)
+    dW = t(X) %*% dOut_agg_rev
```
Baunsgaard, line 231: or is it the gradient taking time?

Baunsgaard (commented):
I quickly looked through, and am unsure what operation you are referring to.
I assume the starting matrix multiplication is expensive, but that only depends on the input size, and it will always be expensive. Some of the other code contain many as.integer or as.scalar. most of them should not be needed but i do not think they have much impact on performance.

scripts/nn/layers/graph_conv.dml
MaximilianSchreff: I think I got some issues when not using them. I will take a look at them.

scripts/nn/layers/graph_conv.dml
MaximilianSchreff: Yes this is the slow thing. But adding the bias in the end takes just as much time.

scripts/nn/layers/graph_conv.dml
MaximilianSchreff: In the stress test, I only used the forward pass.

scripts/nn/layers/graph_conv.dml
MaximilianSchreff: This matrix multiplication takes around 35% of the runtime while adding the bias in the end also takes around 35%.

Baunsgaard: in practice adding the bias, should be very fast compared to the matrix multiplication. 
When looking at the code, it seems to me that you access indices, rather than adding vectors. 
This might be the issue.

MaximilianSchreff: All things related to the index is for the convolution part and not the linear part. But, it is actually possible to do the convolution part without any indices at all since the formula for a graph convolutional layer is OUT = D^-1 * A * D^-1 * X * W + b (A: Adjacency matrix n x n, D: degree matrix n x n, X: input n x features, W: weights f_in x f_out). As you can see, to do the convolution without any indices (normalization and message passing), you need to do 3 extra matrix multiplications instead. These matrix multiplications are (in normal use cases -> n >> features) even bigger than the linear part (X*W + b) because D and A are bigger matrices than X and W. 
So, since X*W + b takes 220 seconds (only the matrix multiplaction takes around 110 seconds), not using indices to do the normalization and convolution would take way longer than 340 seconds, likely around 600 seconds.

This is also the reason why famous other libraries (PyTorch, TF) also mainly use an edge list to do the convolution part through index accessing or use a sparse matrix datatype (which is basically also an edge list) in the GCL implementation.

Baunsgaard: You can try out the sequence of matrix multiplications the sparse representations in SystemDS are usually much faster than the other frameworks. I understand why you designed it the way you did because it does not intuitively make much sense to use the slower matrix multiplications. 

That said, the use case you currently have that runs for 220 seconds sounds a bit large because of the execution time. Can it be because the computer you use has few cores, or is the use-case you are running disproportionately large?

MaximilianSchreff: That is valid point, but I believe, the fact that my implementation was 10x slower than PyTorch is not down to my implementation but down to SystemDS' performance out of the box. It might be that this implementation is indeed slow but with the linear matrix multiplication alone being 3x slower than the whole layer in PyTorch it doesn't matter how fast the convolution part will get. It won't be acceptable anyway. So trying out the sparse matrices would be very effective with the current configuration.
I would suggest running the stress test on a correct and fast configuration to see whether even the matrix multiplication alone will be faster than PyTorch. From there, it is easier to know what actually needs to be improved.

About the machines that I used for testing - I tested it on 2 different machines. One was 96-core, 190GB RAM, x86 CPU and the other was a 14-core, 16GB RAM, ARM CPU. Since both of them don't have a GPU, SystemDS should be very competitive.

phaniarnab: I would be surprised if our matrix multiplication is 3x slower than PyTorch's in CPU, unless PyTorch uses MKL by default or the SystemDS execution is affected by other overheads such as disk eviction. I did not compare with PyToroch but with TensorFlow, and TensorFlow was using BLAS internally.

Baunsgaard: can it be that the memory configuration is off? what is your -Xms -Xmx arguments?

`-Xmx24g -Xms24g -Xmn2400m`

also, please when measuring did you call once? or multiple times. It is known in java that the first call is very slow.

MaximilianSchreff: My -Xms -Xmx arguments were 16g, I think.

Also, I called the tests multiple times. They were always very consistent, only changing by -3 to +3 seconds.

Baunsgaard: what i mean is not from the outside, but inside your script. 
or maybe that is what you do already?

MaximilianSchreff: Sorry, the stress test file itself would probably help. The stress test is a big forward pass over 3 layers, repeated 100 times.
So, from the outside and the inside.

Baunsgaard: ah okay, when will you be in office next time, then maybe we can talk about it?

MaximilianSchreff: Sure! I will be in the office tomorrow and on friday.

PR #1943, opened 2023-11-09

christinadionysio:
This patch adds a performance test for missing value imputation using KNN. I tested the individual methods (`dist`, `dist_missing`, `dist_sample`) with different dataset sizes where 10% of the rows contain missing values. We want to ensure that at least the `dist_sample` method can handle arbitrarily large datasets by reducing the sample size.

### DIFF
--- a/scripts/perftest/KnnMissingValueImputation.sh
+++ b/scripts/perftest/KnnMissingValueImputation.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


christinadionysio:
After running the perftest I created two figures that provide evidence that the first (`dist`) and second (`dist_missing `) methods do not perform well on larger datasets. However, the third method (`dist_sample`) is working for larger datasets by decreasing the sample size. The first figure shows the runtime for each method for different dataset sizes. 
As mentioned the first two methods do not perform well (java heap space exception) on larger datasets, which explains the missing values for `# rows 100000 1000000 10000000`  

[knn_perf_runtime.pdf](https://github.com/apache/systemds/files/13603755/knn_perf_runtime.pdf)

The second figure shows how the sampling size for the third method was decreased for larger datasets. 

[knn_perf_sampling_fac.pdf](https://github.com/apache/systemds/files/13603758/knn_perf_sampling_fac.pdf)

j143:
> As mentioned the first two methods do not perform well (java heap space exception) on larger datasets, which explains the missing values for # rows 100000 1000000 10000000

does this require tweaking with SystemDS-config.xml values or we need to improve the functionality?

@phaniarnab - could you suggest any config changes that could help?

thanks, Janardhan

PR #1946, opened 2023-11-19
[SYSTEMDS-3669] Computation of Shapley Values

louislepage:
This is a PR for a master thesis about computation of shapley values with systemds at scale.


louislepage:
I rewrote the generic shapley value computation with sampling and added an example script, as well as a jupyter notebook in which i compared the results of the official SHAP package and my systemds implementation.

The reults (at least in the case of scaled data) look good, however I found that the rbind() calls during preparation of the instances matrix are very slow and take the longest. Therefor I would revisit this to make further optimizations, but I think I need some advice on how make those appends/writes to large matrices faster.

Here is a quick plot of the computed values for the 107 features of the tranformencoded adult dataset.
Both implementations used the full ~32000 samples as background data and ran for 10000 iterations for each sample.
image

louislepage:
I was able to run it on 5.000 to 50.000 samples and, as expected, it did scale linearly.
However this also showed, that the python implementation just has a huge overhead for small sample sizes, but scales at virtually the same rate.

image


And rbind() is still the slowest single operation for very large sample-sizes, but i was unable to directly write to rows of a pre-allocated matrix, which could be benifical in this case.

```
Heavy hitter instructions:
  #  Instruction                    Time(s)  Count
  1  shapley_sampling                56.988      1
  2  shapley_sampling_prepare        29.756    107
  3  rbind                           22.870      2
  4  !                               10.097    107
  5  append                           6.878    219
  6  rand                             4.483    219
  7  leftIndex                        3.574    321
  ```

**Comment by Baunsgaard** [comment]
Hi @louislepage 

Once you want a review, mark it in this PR.

As an initial comment, try to avoid including the run times or other CSV files, and modify your Python notebook to be scripts that run instead. Removing the notebook gives us clearer ways of comparing the performance and removes the need to start a notebook.

Once you are happy with the behavior of your Shapley operator it would be great if you move it from staging to a builtin function. 
If it is unclear how to do this, comment on this PR, and we will help you out.

Thanks!

**Comment by christinadionysio** [comment]
Thank you @louislepage for this contribution.

As we talked about in the last meeting the code looks good so far.
Please add the license header to the `test_runtimes.sh` file. 
Additionally, I created a jira issue for your thesis, so you can replace `[WIP]` by `[SYSTEMDS-3669]`.

**Comment by louislepage** [comment]
After finishing the thesis, I started moving the explainer was to builtin as [scripts/builtin/shapExplainer.dml](https://github.com/louislepage/systemds/blob/abf1b8ae311353f414a169f13b12b60d7cd55233/scripts/builtin/shapExplainer.dml).

For the final merge, all files under [scripts/staging/shapley_values](https://github.com/louislepage/systemds/tree/shap-values/scripts/staging/shapley_values) will be removed. The experiments are available as part of the [reproducibility repository of the thesis](https://github.com/louislepage/scalable-shap-reproducibility).

Unit- and component tests were added in [test/functions/builtin/part2/BuiltinShapExplainerTest.java](https://github.com/louislepage/systemds/blob/abf1b8ae311353f414a169f13b12b60d7cd55233/src/test/java/org/apache/sysds/test/functions/builtin/part2/BuiltinShapExplainerTest.java) which use tests written in DML in [test/scripts/functions/builtin/shapExplainerUnit.dml](https://github.com/louislepage/systemds/blob/abf1b8ae311353f414a169f13b12b60d7cd55233/src/test/scripts/functions/builtin/shapExplainerUnit.dml) and [test/scripts/functions/builtin/shapExplainerComponent.dml](https://github.com/louislepage/systemds/blob/abf1b8ae311353f414a169f13b12b60d7cd55233/src/test/scripts/functions/builtin/shapExplainerComponent.dml). The DML file also stores the expected results for each test, results are compared within the javatest. This may be a bit unconventional, but it reduces the amount of files/code since otherwise all expected results would have to be written in java and each unit test would need to go to its own DML script. I hope this is fine.

File: scripts/builtin/shapExplainer.dml [hunk 1]
```diff
@@ -0,0 +1,730 @@
...
+# seed            A seed, in case the sampling has to be deterministic.
+# verbose         A boolean to enable logging of each step of the function.
+# ---------------------------------------------------------------------------------------
+#
+# OUTPUT:
+# -----------------------------------------------------------------------------
+# S              Matrix holding the shapley values along the cols, one row per instance.
+# expected       Double holding the average prediction of all instances.
+# -----------------------------------------------------------------------------
+s_shapExplainer = function(String model_function, list[unknown] model_args, Matrix[Double] x_instances,
```
christinadionysio, line 54: Please reformat the function declarations so that you use 4 spaces for every parameter that is introduced in a new line and 2 spaces for the return value and move the  opening bracket `{` to a new line. Please double check for all other functions in this file. 

```
example = function(String x, Double y, ...
    Matrix[Double] z)
  return Double r
{
}
```

scripts/builtin/shapExplainer.dml
```diff
@@ -0,0 +1,730 @@
...
+    mask_constructor=permutation
+  }
+
+  perm_cols = ncol(mask_constructor)
+
+  # we compute mask on reverse permutation wnd reverse it later to get desired shape
+
+  # create row indicator vector ctable
+  perm_mask_rows = seq(1,perm_cols)
+  #TODO: col-vector and matrix mult?
```
christinadionysio, line 235: Is this TODO still needed? If not please remove it and also double check for the other occurrences in this script.

src/test/java/org/apache/sysds/test/functions/builtin/part2/BuiltinShapExplainerTest.java
```diff
@@ -0,0 +1,155 @@
...
+    private static final boolean VERBOSE = true;
+
+    @Override
+    public void setUp() {
+        addTestConfiguration(TEST_NAME, new TestConfiguration(TEST_CLASS_DIR, TEST_NAME, new String[]{"R"}));
+    }
+
+    @Test
+    public void testPrepareMaskForPermutation() {
+        runShapExplainerUnitTest("prepare_mask_for_permutation");
```
christinadionysio, line 47: We use these tests to test the whole functionality of the scripts (more like your component test). The test cases should cover different input parameters and edge cases. It would be great if you could modify the tests in that regard to keep them consistent with our other tests. Please add additional component tests for the different modes (HYBRID, SPARK).

christinadionysio (commented):
Thank you for your contribution @louislepage! Overall it looks really good, I also like your reproducibility repository! I only left a few minor comments about formatting and tests.

mboehm7:
LGTM - thanks @louislepage for this great work. I will now merge it in, despite the remaining minor comments, in order to move this along and facilitate follow-up work. During the merge, I fixed some rebase conflicts, fixed the formatting of the java tests (tabs over spaces), eliminated remaining warnings, and added a FIXME for the unnecessary padding with zero in the test. Thanks.

PR #1948, opened 2023-11-24
[SYSTEMDS-3333] Improve matrix mult chain opt

gogokotsev00:
Now our chain of Hops gets rewritten when we have t(A %*% B) into t(B) %*% t(A), which contributes to a larger optimization space.


phaniarnab:
Thanks for the PR. It's a good start @gogokotsev00.
Here are a few general comments.
- Investigate why some of the tests failed.
- Add tests to verify your changes are working. Follow the examples of `RewriteMatrixMultChainOptTest` and `RewriteMatrixMultChainOptTest2`.
- Try to integrate the new rewrites in self-contained functions. This way we can easily disable the new rewrites if we encounter bugs or harmful impacts. You can also add static flags to enable/disable the rewrites (e.g. `PUSH_DOWN_TRANSPOSE = true`).
- Export developer code style (/dev/CodeStyle_eclipse.xml) to avoid unintended diffs such as space vs tabs.

phaniarnab:
@gogokotsev00, please do not merge branches. Instead, rebase your changes on the latest main branch.
Otherwise, it'd be hard to cherry-pick your changes while pushing your changes.

gogokotsev00:
We were not able to find rewrites which impede an optimal matrix multiplication chain. We have tested with the following examples: 

- -t(a) %\*% b %\*% -t(c) %\*% d
- -t(a) %\*% b
- -t(a) %\*% b %\*% c

j143:
Hi, 

this PR may not make it to 3.2.0 to avoid last minute internal behavior changes,
applying the `next-release` milestone

mboehm7:
LGTM - thanks for the patch @gogokotsev00 and team. During the merge, I moved the rewrite to a separate experimental rewrite which is not yet fully integrated (because I think it's not ready for that). The reason why you did not see your rewrite being applied as expected is that the transposes for folded into the generation of constant matrices (with flipped dimensions). By adding our special program block cut `while(FALSE){}` you can separate them and the tests fail with an unexpected number of transposes.

PR #1951, opened 2023-11-30
[SYSTEMDS-3653]  Ultra Sparse MM Optimization

Baunsgaard:
This PR optimize our ultra sparse matrix multiplication operation, to optimize it we do 3 things. 1. split the monolithic kernel into multiple smaller functions to allow JIT compilation without having to add any flags to java. 2. Remove indirection through MatrixBlock wherever possible.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/compress/CompressedMatrixBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/CompressedMatrixBlock.java
@@ -1152,12 +1152,17 @@ public class CompressedMatrixBlock extends MatrixBlock {
 	}
 
 	@Override
-	public void examSparsity(boolean allowCSR) {
+	public void examSparsity(boolean allowCSR, int k) {
 		// do nothing
 	}
 
 	@Override
-	public void sparseToDense() {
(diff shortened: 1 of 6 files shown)


Baunsgaard:
All execution times are 1000 calls.

Left side one non zero per row
100k by 1m  %*%  1m by 100 sp 0.1  ->  Before: 6.5  After : 4.5 sec
Left side two non zero per row
200k by 1m %*% 1m by 100 sp 0.1 -> Before 173.724 After : 26.3 sec

Baunsgaard:
With right side Dense

Left side one non zero per row
100k by 1m %% 1m by 100 sp 0.43 -> Before: 65.06  After : 29.039 sec

This speedup can be improved if our sparsity estimations are  better since the output first is allocated sparse and subsequently transformed into dense. Most of the speedup is in parallelizing our sparse to dense code.

Baunsgaard:
> Left side two non zero per row
> 200k by 1m %% 1m by 100 sp 0.1 -> Before 173.724 After : 26.3 sec 

Further speed up to 19.5 sec, via allocation and analysis of empty output earlier.

PR #1955, opened 2023-12-07
[SYSTEMDS-3636] Improved ultra-sparse tsmm left w/ sparse output

christinadionysio:
This patch provides the support for left transposed ultra-sparse tsmm. 
Similar to the the implementation of the right transpose ultra-sparse tsmm, 
binary search is used to populate the upper triangular part of a sparse output matrix.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixMult.java
+++ b/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixMult.java
@@ -2510,8 +2510,11 @@ public class LibMatrixMult
         SparseBlock c = ret.sparseBlock;
 		int m = m1.rlen;
 
-        if(leftTranspose) {
+		if(leftTranspose) {
 			// Operation t(X)%*%X, sparse input and output
+			for(int i=0; i<m; i++)
+				c.allocate(i, 8*SparseRowVector.initialCapacity);
+			SparseRow[] sr = ((SparseBlockMCSR) c).getRows();
 			for( int r=0; r<a.numRows(); r++ ) {
 				if( a.isEmpty(r) ) continue;
 				final int alen = a.size(r);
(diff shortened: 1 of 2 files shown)


christinadionysio:
> how much faster is it?

@Baunsgaard I only tested the `t(X) %*% X` operation on ultra sparse matrices locally for now. Using the germany_osm dataset we achieve a runtime of ~12s after the optimisation. (It is infeasible to run it locally without the optimisation, since a dense matrix is allocated and therefore a Java heap space exception occurs.)
When I run the operation on an ultra sparse matrix with 30000 rows and cols I get a speedup of 17x or 30x depending on which of the two methods is used.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixMult.java
```diff
@@ -2506,39 +2506,57 @@ private static void matrixMultTransposeSelfSparse( MatrixBlock m1, MatrixBlock r
...
+        if(leftTranspose) {
+			// Operation t(X)%*%X, sparse input and output
+			for( int r=0; r<a.numRows(); r++ ) {
+				if( a.isEmpty(r) ) continue;
+				final int alen = a.size(r);
+				final double[] avals = a.values(r);
+				final int apos = a.pos(r);
+				int[] aix = a.indexes(r);
+				int rlix = (rl==0) ? 0 : a.posFIndexGTE(r, rl);
+				rlix = (rlix>=0) ? apos+rlix : apos+alen;
```
Baunsgaard, line 2522: instead of this check make a if statement 

if (rlix >= 0) {
...
}

Baunsgaard (commented):
a few comments for improvements, hope you can decrypt it.

christinadionysio:
After the changes the runtime of the `matrixMultTransposeSelfUltraSparse` method improved by 2.5x.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixMult.java
```diff
@@ -2506,39 +2506,64 @@ private static void matrixMultTransposeSelfSparse( MatrixBlock m1, MatrixBlock r
...
+			for( int r=0; r<a.numRows(); r++ ) {
+				if( a.isEmpty(r) ) continue;
+				final int alen = a.size(r);
+				final double[] avals = a.values(r);
+				final int apos = a.pos(r);
+				int[] aix = a.indexes(r);
+				int rlix = (rl==0) ? 0 : a.posFIndexGTE(r, rl);
+
+				if(rlix>=0)
+					rlix = apos+rlix;
```
Baunsgaard, line 2527: what i ment with the if statement:

```
if(rlix >= 0){
				int len = apos + alen;
				for(int i = rlix; i < len && aix[i] < ru; i++) {
					for (int k = a.posFIndexGTE(r, aix[i]); k < len; k++) {
						sr[aix[i]].add(c.pos(k) + aix[k], avals[i] * avals[k]);
					}
				}
}
```
you do not need the else case because that terminate on first call to the for loop.

Baunsgaard:
LGTM

Baunsgaard:
Thanks for the PR, it is now merged!

PR #1956, opened 2023-12-07
Bug fix in cache management for nested lists

MaximilianSchreff:
SystemDS was previously not supporting nested lists correctly since the data of CacheableData objects within nested loops were always deleted after a function call. 
Normally, there are rmvar statements after function calls to remove alll variables used within the function. To protect CacheableData objects (e.g. matrices) from having their data removed by the rmvar statements we use a cleanup-enabled flag. This flag was not correctly set for variables that were within a nested list. These commits fix this problem by flagging all elements, also within nested lists.

Automated tests have been added to test the changes.

@phaniarnab

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/controlprogram/ParForProgramBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/controlprogram/ParForProgramBlock.java
@@ -31,6 +31,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+import java.util.Queue;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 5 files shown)


MaximilianSchreff:
The bug can be triggered through  a simple function call with a nested list as an argument.

Here is a dml script which does this. 
[list_bug.txt](https://github.com/apache/systemds/files/13599914/list_bug.txt)

Baunsgaard:
LGTM, 

I think the Queue make things slower instead of the boolean list, 
but if we need to support the dynamic lengths of the Lists or nested Lists, then this fix makes sense.
And the live variables management is a low overhead to begin with so, all good from my side.

We need to run the tests again once the main branch is clear of failing tests.

MaximilianSchreff:
@Baunsgaard 
About the Queue - previously, we iterated through all variables and counted the number of cacheable objects. This was done to calculate the size of the boolean array but in a shallow manner. With the changes, we would now need to iterate through every single variable, i.e. laso recursively for lists, in order to get the size. That is why I chose a dynamic data structure. I'm not sure whether this is faster though. Since the number of live variables per function call is usually pretty low, I wasn't able to see any performance differences.

Baunsgaard:
thanks for the PR, it is now merged.

PR #1957, opened 2023-12-07
ResNet Bottleneck Architecture

MaximilianSchreff:
This series of commits adds the forward pass of the residual network bottleneck architecture as introduces by "Deep Residual Learning for Image Recognition" by Kaiming He et. al. and refined in "ResNet v1.5 for PyTorch" by NVIDIA.

The individual components (bottleneck residual block, bottleneck residual layer) has been unit tested and the tests were added to the NN test suite and the whole bottleneck ResNet has been tested against the implementation of PyTorch which was not added to the test suite.

@phaniarnab


MaximilianSchreff:
Notice that this neural network will likely only work in the training loop with the fixed bug of nested lists from #1956

j143:
Hi @MaximilianSchreff , thanks a lot for working on this. Nice to see that the code is document in the functions.

can we create one small .md file with instructions on how to run, limitations, a quick example! If you have any images/diagrams (even hand drawn rough diagrams also fine). 

May be like this: https://pytorch.org/hub/nvidia_deeplearningexamples_resnet50/

Regards

MaximilianSchreff:
@j143 Thank you for the feedback!
I'm still developing the backward pass for the models. I will create a fully training loop example and instructions once I implemented the backward pass.

MaximilianSchreff:
Content merged by other PR

PR #1958, opened 2023-12-07
[SYSTEMDS-3658] Time-series Alignment Library (Python API)

christinadionysio:
This commit adds a Python Library for time-series alignment. The library supports alignment for video datasets and time-series (i.e. sensor data). It implements different alignment strategies from static to dynamic alignment and provides different similarity measures that can be used in the alignment process. The alignment strategies are tested on a synthetic dataset that can be found in the `data` directory.


mboehm7:
LGTM - thanks for contribution of this alignment library @christinadionysio .

PR #1982, opened 2024-01-15
[SYSTEMDS-3290] Multi-threaded Instruction Execution

sweetpellegrino:
This pull request introduces an implementation of a pipeline-aware linearization algorithm. 
It extends on the idea of [SYSTEMDS-3290] by marking the low-level operations for asynchronous execution. It achieves this by creating groups of Lops (pipelines) that are identified to be run in parallel.

This PR introduces only the marking of the operations; not the asynchronous execution itself.

Key changes:
- Pipeline-Aware Linearization
- Inclusion of new Linearization-Algorithm into existing test cases.
- Creation of new Test-Class for the new _ILinearize_ functions.

### DIFF
--- a/src/main/java/org/apache/sysds/lops/Lop.java
+++ b/src/main/java/org/apache/sysds/lops/Lop.java
@@ -158,6 +158,12 @@ public abstract class Lop
 	protected boolean _asynchronous = false;
 
 	/**
+	 * Refers to the pipeline to which this lop belongs to.
+	 * This is used for identifying parallel execution of lops.
+	 */
+	protected int _pipelineID = -1;
+
+	/**
 	 * Estimated size for the output produced by this Lop in bytes.
 	 */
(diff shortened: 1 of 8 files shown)


phaniarnab:
Thanks for the PR @sweetpellegrino. I will have a look.

phaniarnab:
Thanks again @sweetpellegrino, for the contribution. 
The changes look good to me. We can merge the changes (except the DAG printer class). However, before that, I will have to think a bit about how to integrate this instruction pipelining into the current operator ordering approaches.

sweetpellegrino:
Great to hear! I’ve changed the status to “Non-Draft PR.” If there’s anything that needs updating in the repo on my end due to the force-push, let me know.

mboehm7:
LGTM - thanks for the patch @sweetpellegrino. During the merge I removed the additional explain (as Arnab suggested), and fixed minor things in the added test (tabs vs spaces, and simplified types).

PR #1985, opened 2024-01-19
Cache Bug With Lists

MaximilianSchreff:
A description of the bug:

Usual Behavior of SystemDS:
- When: A function returns an object and this object is assigned the variable name "blub".
- And: A variable named "blub" already exists in the execution context.
- Then: Variable object should be replaced with new returned object and the old object should be cleaned up.

Bug:
- When: Old object and new object of variable named "blub" are lists and both contain the same element A.
- And: There is no other reference to A in the execution context.
- Then: The mutual list element will be cleaned up and its cached data deleted even though it is still in the new list. This happens since the new object isn't added to the execution context when cleaning up. Otherwise, the reference to A would be detected and its cached data not deleted.

This behavior can be seen when executing the newly added DML script. 

The solution for this is to only clean up the old objects after having removed all old objects from and added all new objects to the execution context. This way, the cached data of the mutual list elements will not be deleted since the clean up always checks whether there is another reference in the execution context to that element.

Testing has been added.


Baunsgaard (approved):
LGTM, Just change indentation of java files to tabs

phaniarnab:
Good catch. Looks good to me too. 
I will run a few tests and merge them once the comments are resolved.

PR #1986, opened 2024-01-20
Rewrite for Matrix Distributive Property

ReneEnjilian:
This pull request implements a simplification rewrite for the, input pattern `(A%*%B) + (A%*%C)`. The resulting rewrite is `A%*%(B+C)`.

However, for the rewrite to be applied a number of conditions need to be fulfilled. All three matrices `A`, `B`, and `C` have to be dense and certain properties like the number-of-non-zeros (`NNZ`) need to be available. Further, the rewrite is only applied if the number of floating point operations is actually reduced by the rewrite. Testing has been added.


src/main/java/org/apache/sysds/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
```diff
@@ -1137,46 +1138,88 @@ else if( right.getDim2()>1 ) //multi column vector
...
-	
-	private static Hop simplifySumDiagToTrace(Hop hi)
-	{
-		if( hi instanceof AggUnaryOp ) 
-		{
+
+	private static Hop simplifyMatrixFactorization(Hop parent, Hop hi, int pos) {
+		// Check for Hop being Binary
+		if(hi instanceof BinaryOp) {
+			Hop left = hi.getInput().get(0);
```
Baunsgaard, line 1171: i think we can avoid the nested get(0), via hi.getInput(0).

Baunsgaard (commented):
LGTM

src/main/java/org/apache/sysds/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
ReneEnjilian: Okay I changed all instances where this occurs. Further, I just realized that the name of the method simplifyMatrixFactorization() might be misleading because somebody might think we perform a matrix factorization (i.e., [decomposition](https://en.wikipedia.org/wiki/Matrix_decomposition)) instead of factoring out a term (i.e., here matrix A). Should I change the name to something more neutral like simplifyMatrixDistributiveFactorization() ?

Baunsgaard: That would be okay. But the best way ( in my opinion) would be to add a Java doc on the method to specify exactly what it is intended to do.

j143: > I just realized that the name of the method simplifyMatrixFactorization() might be misleading because somebody might think we perform a matrix factorization (i.e., [decomposition](https://en.wikipedia.org/wiki/Matrix_decomposition)) instead of factoring out a term (i.e., here matrix A). Should I change the name to something more neutral like simplifyMatrixDistributiveFactorization() ?


@ReneEnjilian -- if you prefer you can call this, may be `MatrixCombineMultiply`. It's up to you. 😺 

Factorization has a different meaning!

ReneEnjilian: I am using the name simplifyDistributiveMatrixMult() now.

mboehm7:
LGTM - thanks for the patch @ReneEnjilian. I'm merging it in with a few modifications.

mboehm7:
Thanks again - during the merge I change the rewrite condition to only apply to binary-plus (not binary in general) and tightened the cost conditions, so the rewrite does not apply for two vector-matrix multiplications (where a large matrix-matrix addition of B+C can be way more costly).

Furthermore, if you're interested contributing to improved rewrites, I created SYSTEMDS-3666 for adding various simplifications for not and comparison operations.

ReneEnjilian:
Thanks @mboehm7. I dont really understand why you chose m*n*l  (for A%*%B) as the formula for the FLOPs in costOriginal. Shouldnt it be 2*m*n*l ? I will start with SYSTEMDS-3666.

mboehm7:
It's related to fused-multiply-add (FMA) operations of modern processors for pairs of multiply-add as used in matrix multiplications. Also, given that matmult is generally more efficient than elementwise multiplications with comparatively large costs for allocation this gave a good (and still meaningful) break-even points for examples.

ReneEnjilian:
thats really good to know (regarding FMA), thanks Matthias.

PR #1988, opened 2024-01-24
[SYSTEMDS-3666] Simplify Not ('!') over comparisons

ReneEnjilian:
This pull request implements rewrites for the not-operator ('!') applied over comparisons. The implemented rewrites are:

- !(A>B) -> A<=B
- !(A<B) -> A>=B
- !(A==B) -> A!=B, including !(A==0) -> A!=0

I changed the last rewrite to a more general version that also covers !(A==0) -> A!=0. I would be grateful for feedback on this matter. Further, testing has been added.


mboehm7:
LGTM. Thanks for the patch @ReneEnjilian - during the merge I only moved the rewrite from dynamic to static rewrites (no size-based decision), tightened the source pattern (only if NOT is the only consumer), and simplified the test (single dml/R file).

PR #1989, opened 2024-01-25
[SYSTEMDS-3552] DCSR evaluation (PR2)

Jaybit0:
In this PR, I evaluated different sparse blocks (and a dense block as baseline) regarding their performance in execution time and memory consumption. I put my short documentation in the `docs/performance` folder. I commented my tests I added in `org.apache.sysds.performance.matrix` to avoid heavy tests being run on GitHub. @mboehm7


Baunsgaard:
Hi JayBit,

While Matthias will look into details i just want to clarify, 

1. None of the elements inside `org.apache.sysds.performance` are run on GitHub because they all are performance related, therefore, you do not need to comment your elements out.
2. The `org.apache.sysds.performance` package is intended to be run independently of the maven test framework, because the testing framework incurs overhead in the execution making the tests more inconsistent in performance numbers.

I suggest taking a look at the README how the performance tests were intended to be run:
https://github.com/apache/systemds/blob/main/src/test/java/org/apache/sysds/performance/README.md

If you want to make the changes to not run though the JUnit test framework, you can take a number for your specific runs in the Main file, and hook them up to your tests:

https://github.com/apache/systemds/blob/9f82e9f675c567c7f80360f3814e3a19138d7a62/src/test/java/org/apache/sysds/performance/Main.java#L104

I suggest taking something like 1000, 1001, 1002 etc. to not overlap with some other numbers.

Best regards
Sebastian

Jaybit0:
Thanks for the clarification. I will look into the README on how to create proper benchmarks.

Jaybit0:
I am not sure why these tests are failing. Does it have to do with my code?

j143:
Hi @Jaybit0 - thanks a lot for contribution to SystemDS. 😺 ✨ 

the test failure may not be related to this PR. Since, these tests won't run on github actions.

mboehm7:
LGTM - thanks for finalizing this patch @Jaybit0. During the merge I fixed the size estimation for MCSR (number of attributes, differentiation into SparseRowScalar and SparseRowVector, capacity instead of size), and removed the materialized images as they are already outdated.

PR #1990, opened 2024-01-26
[SYSTEMDS-3670] Added early-stopping mechanism to tSNE

ramesesz:
Added a tolerance parameter to tSNE and implemented an early-stopping mechanism based on the L1 norm of the difference. Furthermore I also added a test case, whose setup guarantees that the early-stopping mechanism is triggered.

Without early-stopping, tSNE always runs 1000 iterations (based on the default parameter value). More often than not, this leads to redundant iterations, where changes are minimal. This mechanism is a good addition as it adds a dimension of flexibility to the algorithm.


scripts/builtin/tSNE.dml
```diff
@@ -119,7 +161,7 @@ return(matrix[double] P)
   while (mean(abs(Hdiff)) > tol & itr < 50) {
     P = exp(-D * beta)
     P = P * ZERODIAG
-    sum_Pi = rowSums(P)
+    sum_Pi = rowSums(P) = 1e-12
```
Baunsgaard, line 164: this line confuses me, what is the intention?

scripts/builtin/tSNE.dml
```diff
@@ -141,4 +183,4 @@ return(matrix[double] P)
   P = P / sum(P)
   if(is_verbose)
     print("x2p finishing....")
-}
+}
```
Baunsgaard, line 186: add a new line in the end of the script

scripts/builtin/tSNE.dml
```diff
@@ -72,6 +106,13 @@ m_tSNE = function(Matrix[Double] X, Integer reduced_dims = 2, Integer perplexity
     sumW = rowSums(W)
     g = Y * sumW - W %*% Y
     dY = momentum*dY - lr*g
+
+    norm = sum(dY^2)
+    if(is_verbose & itr %%10 ==0){
```
Baunsgaard, line 111: maybe add a print iteration variable, to indicate how frequent the logging is and default it to 10.

src/test/java/org/apache/sysds/test/functions/builtin/part2/BuiltinTSNETest.java
```diff
@@ -403,4 +403,108 @@ private void runTSNETest(Integer reduced_dims, Integer perplexity, Double lr,
...
+                {3.665608219962555, 3.399666975542729},
+                {1.7618442622801385, 4.675570729512945},
+                {4.987236193552888, 0.41700477957766546},
+                {0.21496074278985922, 3.5781179414157616}
+        };
+
+
+			writeInputMatrixWithMTD("X", X, true);
+
+			runTest(true, false, null, -1);
```
Baunsgaard, line 503: we need to test for behavior. 
This test only verify that the script runs. not that it works.
you can maybe look at the print output and see that the iterations give lower and lower l1 norm?

Baunsgaard (commented):
This is a fine addition, there are some minor things to fix before we merge it.

The main thing missing is the testing verification should test for behavior.
I gave a concrete suggestion in the line comment.

best regards
Sebastian

scripts/builtin/tSNE.dml
ramesesz: My mistake, this should be `sum_Pi = rowSums(P) + 1e-12`. This is to prevent zero division errors that came up multiple times during the experiments.

scripts/builtin/tSNE.dml
ramesesz: Should this be a new input parameter or just a static variable in the script?

Baunsgaard: input variable.

j143:
Hi @ramesesz - thanks for contribution. 😄 ✨ 

is this good for merge?

ramesesz:
> Hi @ramesesz - thanks for contribution. 😄 ✨
> 
> is this good for merge?

Hi @j143, yes its ready for merge :)

Baunsgaard:
LGTM as well, thanks for the contribution.

PR #1991, opened 2024-01-26
[SYSTEMDS-3670] Added script of tsne with pca preprocessing to staging

ramesesz:
Added a script implementing tSNE with a pca-preprocessed input data.

This script implements the concept used in sklearn's tSNE, where they provide the option of pre-processing the input data with pca before giving them to the tSNE algorithm. For very large datasets, this could theoretically improve tSNE performance by removing a large portion of noise prior to the core algorithm. However, the current implementation does not speed up tSNE when applied on MNIST, so further research required to confirm this.


scripts/builtin/dist.dml
```diff
@@ -32,7 +32,8 @@
 # -----------------------------------------------------------------------------------------------
 
 m_dist = function(Matrix[Double] X) return (Matrix[Double] Y) {
-  G = X %*% t(X);
-  Y = sqrt(-2 * G + outer(diag(G), t(diag(G)), "+"));
+  n = nrow(X)
+  s = rowSums(X^2)
+  Y = sqrt(-2 * X %*% t(X) + s + t(s))
```
Baunsgaard, line 37: if you can limit it to not include this change in both PRs it would be great.
Alternatively you can move this specific dist change to a separate PR.

scripts/staging/pca-tsne.dml
Baunsgaard, line 1: This is a fine small script, 
I would suggest the following, add a comment in the documentation of TSNE that indicate clearly how to make a script with preprocessing PCA. and then you can reference this script.
Also we can move this specific script from staging to `scripts/tutorials`, where i think it would fit right in.
Just add a new folder inside there.

scripts/staging/pca-tsne.dml
```diff
@@ -0,0 +1,38 @@
...
+data = read($X)
+
+# Pre-process data with PCA
+[PCA, components, centering, scalefactor] = pca(X=data, K=$k)
+
+# Do tSNE with PCA output
+Y = tSNE(X=PCA)
+
+# Save reduced dimensions
+write(Y, $Y)
```
Baunsgaard, line 38: add newline in the end of the files

Baunsgaard (commented):
LGTM, minor changes needed, mainly moving files a bit around.

scripts/builtin/dist.dml
ramesesz: We already did a PR for the dist function, so is it better to revert the changes of dist() for the current PR?

Baunsgaard:
thaks for the PR, this is now part one merged.

PR #1993, opened 2024-01-30
[MINOR] Fix Frame tests for Github Actions

Baunsgaard:
> enable more Federated transform encode tests


Baunsgaard:
Invalid now.

PR #2001, opened 2024-02-19
Keyword Spotting Application

mufwan:
Added files to read wav files to cenvert wav files to a csv file.

Implemented an application for keyword Spotting.


Baunsgaard:
I have decided against merging this PR at the moment, since it is adding files some places not ideal.
If you want to clean it up @mufwan  then i will happily merge it.

Tasks:

1. Move the scripts into scripts/staging
2. Add documentation for how to run it inside the script, and what is expected.
3. Rebase the PR to not contain the changes for FFT

While merging the STFT is slightly changed and not hooked up to DML correctly,
The main issue was that one method got overloaded to take a long vs int as arguments to choose the version to call.

PR #2003, opened 2024-02-25
[SYSTEMDS-2926] AWS scripts update for EMR-7.0.0

lachezar-n:
The changes fix some general issues:
- creating and referencing the S3 buckets
- not assigning any sub-network for the cluster (bad practice + potential security vulnerability)

The changes also update the used EMR version to the currently most recent one: emr-7.0.0:
- configurations update
- exchanging Ganglia with AmazonCloudWatchAgent

While testing the script with the current repo version the following bug was observed: when launching SystemDS in execution mode "spark" an `IllegalCallerException` is thrown.
For running the command `spark-submit target/SystemDS.jar -f path/to/hello.dml -exec spark -stats -explain` the exact output in the console is:
```shell
...
--MAIN PROGRAM
----GENERIC (lines 1-1) [recompile=false]
------CP print Hello World.SCALAR.STRING.true _Var0.SCALAR.STRING 8
------CP rmvar _Var0


An Error Occurred : 
   IllegalCallerException -- java.lang.ref is not open to unnamed module @4eba373c

```

PR #2004, opened 2024-02-25
AWS script fix for EMR-7.0.0

lachezar-n:
Following the [pull request](https://github.com/apache/systemds/pull/2003#issue-2152722431) for updating the script to use emr-7.0.0, here is a fix to cover these issues:
- EMR uses by default Java 8 for Spark: Settings to enforce Java 11 were added
- Launching the cluster from outside the AWS client region was not configured originally: Here that is fixed for all shell scripts.

Successfully tested:
- Spin up the cluster, run DML script and terminate cluster from eu-centra-1 client into eu-central-1
- Spin up the cluster, run DML script and terminate cluster from eu-centra-1 client into us-central-1


mboehm7:
LGTM - thanks for the patch @lachezar-n. I resolved the merge conflicts (in the future please do not include the previous PR), and we should keep an eye open if the hard-coded `/usr/lib/jvm/jre-11` is always valid.

PR #2005, opened 2024-02-26
Sparse Block Iterator for non-zero Rows

ReneEnjilian:
**Description:**
This pull request introduces a new iterator designed to enhance our sparse block handling capabilities. Unlike the default IJV iterator, which iterates over all non-zero cells, this iterator focuses on the indexes of non-zero rows—skipping entirely zero rows. In a subsequent task, I will use this iterator to investigate which improvements the DCSR representation can bring over other representations like CSR in a number of key operations (e.g., matrix multiplication, unary aggregates). The new iterator supports the all four sparse formats: DCSR, MCSR, CSR, and COO. Testing has been added.

**Usage Example:**
```
SparseBlock block = ...; // Assume block is initialized
Iterator<Integer> iter = block.getIteratorNonZeroRows();
while (iter.hasNext()) {
    int rowIndex = iter.next();
    // Process non-zero row at rowIndex
}
```


mboehm7:
LGTM - thanks for the patch @ReneEnjilian. I'll only change minor things during the merge. Overall, I think it might be a better choice for performance and separation to move the non-empty row iterators into the individual sparse block implementations, but I'll explore that once it's in.

PR #2008, opened 2024-03-26
[SYSTEMDS-3687] Python API startup fixes

Baunsgaard:
This PR change the startup of the python interface to properly use the jar files, and fixes a release issue where if the SYSTEMDS_ROOT is not set the python API did not properly hook into the released jar artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,6 @@ buildNumber.properties
 _site/
 
 # Tutorial data mnist
-src/main/python/systemds/examples/tutorials/*/
 scripts/nn/examples/data/*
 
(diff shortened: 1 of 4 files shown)


Baunsgaard:
Still debugging a bit, therefore it is marked as a draft.

Baunsgaard:
@j143  Can you check if this fixes the issues for the release artifacts.

I have tested it with the released code, and it works for me.

If you set the SYSTEMDS_ROOT to the path of the unzipped bin release it works just like the other releases.
It also correctly detect the log4j files, in different configurations.

PR #2015, opened 2024-04-05
[SYSTEMDS-3572] Threadpool ParFor name threads 

Baunsgaard:
This PR contains code to improve an recent update in the common thread pool to properly use the threads pool cached allocations for PARFOR threads.


Baunsgaard:
```
X = rand(rows = 100, cols = 100, min = 0, max = 100)
Z = rand(rows = 1000, cols = 100, min = 0, max = 100)

parfor( i in 1: 4){
    for( i in 1 : 1000){
        Y = Z %*% X
    }

    print(sum(Y))
}
```

This script improve from 9.6 sec to 7.5 sec execution time (averaged over 10 runs) on my laptop from proper reuse of the threadpool.

PR #2017, opened 2024-04-08
[Not for merge] python logging issue

j143:
This is not fixing the log4j issue -- https://github.com/apache/systemds/commit/387b6c1c8e472dde65a36e8f3fd22091054ed347

```py

import logging

from systemds.context import SystemDSContext
from systemds.operator.algorithm import l2svm

with SystemDSContext() as sds:
    # Generate 10 by 10 matrix with values in range 0 to 100.
    features = sds.rand(10, 10, 0, 100)
    # Add value to all cells in features
    features += 1.1
    # Generate labels of all ones and zeros
    labels = sds.rand(10, 1, 1, 1, sparsity=0.5)

    model = l2svm(features, labels, verbose=False).compute()
    logging.info(model)
```

$ python ~/test.py 
24/04/08 03:26:16 WARN util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
$

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,6 @@ buildNumber.properties
 _site/
 
 # Tutorial data mnist
-src/main/python/systemds/examples/tutorials/*/
 scripts/nn/examples/data/*
 
 # User configuration files
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -39,5 +39,5 @@ exclude:
   - updateAPI.sh
 
 # These allow the documentation to be updated with newer releases
-SYSTEMDS_VERSION: 3.2.0-SNAPSHOT
+SYSTEMDS_VERSION: 3.2.0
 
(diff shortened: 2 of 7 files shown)


j143:
Hi @Baunsgaard -- I built the whl artifact and installed. And checked the logging doesn't seem to work as expected.

Baunsgaard:
what do you mean with not working as intended?
if it is the nativeCodeLoader warning that you see then it is correct.

j143:
It is supposed to print the script output, isn't it?

>     model = l2svm(features, labels, verbose=False).compute()
>     logging.info(model)

it providing no output!

Baunsgaard:
you should be able to make it if you set the logging level.

image

As you can see default behavior is nothing on info.

image

While only warnings are printed.

If you want more info. you need to set level.

image


Therefore everything is working as intended.

j143:
already published at https://pypi.org/project/systemds/

PR #2019, opened 2024-04-16
[MINOR] Double Buffering longer than buffer arrays

Baunsgaard:
This PR fixes a bug in the double buffering in cases of writing longer than buffer size buffers.
Also contained is minor cleanups, and clarifications through docs.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
+++ b/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
@@ -62,15 +62,23 @@ public class DoubleBufferingOutputStream extends FilterOutputStream {
 	{
 		try {
 			synchronized(_buff) {
-				//block until buffer is free to use
-				_locks[_pos].get();
 				
-				//copy for asynchronous write because b is reused higher up
-				System.arraycopy(b, off, _buff[_pos], 0, len);
+				byte [] b_pos = _buff[_pos];
 				
-				//submit write request 
-				_locks[_pos] = _pool.submit(new WriteTask(_buff[_pos], len));
-				_pos = (_pos+1) % _buff.length;
+				if(b_pos.length > len){
+					// block until buffer is free to use
+					_locks[_pos].get();
+					System.arraycopy(b, off, b_pos, 0, len);
+					// submit write request 
+					_locks[_pos] = _pool.submit(new WriteTask(b_pos, len));
+					// copy for asynchronous write because b is reused higher up
+					_pos = (_pos+1) % _buff.length;
+				}
+				else{
+					// we already have the byte array in hand, but we do not do it async
+					// since there would be no guarantee that the caller does not modify the array.
+					writeBuffer(b, off, len);
+				}
 			}
 		}
 		catch(Exception ex) {
--- a/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
+++ b/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
@@ -62,7 +62,7 @@ public class DoubleBufferingOutputStream extends FilterOutputStream {
 				final byte[] b_pos = _buff[_pos];
 				// block until buffer is free to use
 				_locks[_pos].get();
-				if(b_pos.length > len) {
+				if(b_pos.length >= len) {
 					// copy the block into the buffer.
 					System.arraycopy(b, off, b_pos, 0, len);
 					// submit write request guaranteed to be sequential since it is using a single thread.
--- a/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
+++ b/src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
@@ -75,6 +75,9 @@ public class DoubleBufferingOutputStream extends FilterOutputStream {
 					// To avoid this we simply write the given byte array without a buffer.
 					// This approach only works if the caller adhere to not modify the byte array given
 					_locks[_pos] = _pool.submit(() -> writeBuffer(b, off, len));
+					// get the task to reduce the risk ( and at least block the current thread) 
+					// to avoid race conditions from callers.
+					_locks[_pos].get(); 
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
```diff
@@ -43,42 +42,49 @@ public DoubleBufferingOutputStream(OutputStream out, int num, int size) {
...
 	public void write(int b) throws IOException {
-		throw new IOException("Not supported"); 
+		throw new IOException("Not supported");
 	}
 
 	@Override
-	public void write(byte[] b, int off, int len) 
-		throws IOException 
-	{
+	public void write(byte[] b, int off, int len) throws IOException {
```
Baunsgaard, line 59: The problem is this method, 

The rest of the PR is mainly just formatting and docs.

Baunsgaard:
In more detail:

```java

public static void writeWritableToLocal(String fname, Writable mb, boolean doubleBuffering)
		throws IOException
	{
		OutputStream fos = new FileOutputStream( fname );
		if( doubleBuffering )
			fos = new DoubleBufferingOutputStream(fos, 2, BUFFER_SIZE);
		FastBufferedDataOutputStream out = new FastBufferedDataOutputStream(fos, BUFFER_SIZE);
		
		try {
			mb.write(out);
		}
		finally {
			IOUtilFunctions.closeSilently(out); //incl double buffering
			IOUtilFunctions.closeSilently(fos);
		}
	}

```

When we enable the DoubleBuffering it writes from a FastBufferedDataOutputStream.

the FastBufferedDataOutputStream forwards byte[]'s if it is larger than it's buffer size of 8192.

```java
public void write(byte[] b, int off, int len) throws IOException {
	if(len > _bufflen) {
		// If we write a byte array that is larger than the buffer
		flushBuffer(); // flush the buffer first and
		// forward the array directly
		out.write(b, off, len);
	}
	else{
            ...
        }
}
```

Therefore the DoubleBufferingOutputStream can encounter inputs that are longer than it's allocated buffer.

to handle it i added a second case for it:

```java@Override
public void write(byte[] b, int off, int len) throws IOException {
	try {
		synchronized(_buff) {
			final byte[] b_pos = _buff[_pos];
			// block until buffer is free to use
			_locks[_pos].get();
			if(b_pos.length > len) {
				// copy the block into the buffer.
				System.arraycopy(b, off, b_pos, 0, len);
				// submit write request guaranteed to be sequential since it is using a single thread.
				_locks[_pos] = _pool.submit(() -> writeBuffer(b_pos, 0, len));
				// copy for asynchronous write because b is reused higher up
			}
			else {
				// The given byte array is longer than the buffer.
				// This means that the async buffer would overflow and therefore not work.
				// To avoid this we simply write the given byte array without a buffer.
				// This approach only works if the caller adhere to not modify the byte array given
				_locks[_pos] = _pool.submit(() -> writeBuffer(b, off, len));
			}
			_pos = (_pos + 1) % _buff.length;
		}
	}
	catch(Exception ex) {
		throw new IOException(ex);
	}
}
```

src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
mboehm7: Thanks, but could you also elaborate what the intended improvement is. From what I see you try to avoid a copy for tiny arrays, but it introduces brittle code that is prone to race conditions if the originating data stream flushes at certain points (danger that the array being currently written out to disk being overwritten by subsequent serialization).

Baunsgaard: The code should not modify the handling of tiny arrays, if it does, I made a mistake.
The change is primarily only addressing the handling of byte[]'s of lengths that are longer than 8192. (or what we set out BYTE_BUFFER sizes to).

Where is the race condition?
The idea of the pool in this code is to force it into a single async stream that has one thread allocated to processing (as you designed it) the write task in seqence outside the 'hot' path.

mboehm7:
OK thanks for the clarification but for a general-purpose implementation, we should write it sequentially (no pool submit) because somebody might write the same byte array to the stream from outside.

Baunsgaard:
> OK thanks for the clarification but for a general-purpose implementation, we should write it sequentially (no pool submit) because somebody might write the same byte array to the stream from outside.

We can clarify the code by not calling it a pool maybe?
Because it just makes the writing to this output stream asynchronous.

Baunsgaard:
> no, what I mean is that this code is incorrect - if somebody would allocate a large byte buffer, and repeatedly write this to the fastoutputstream, you get a race condition overwrite.

ahh true.

src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
```diff
@@ -75,6 +75,9 @@ public void write(byte[] b, int off, int len) throws IOException {
 					// To avoid this we simply write the given byte array without a buffer.
 					// This approach only works if the caller adhere to not modify the byte array given
 					_locks[_pos] = _pool.submit(() -> writeBuffer(b, off, len));
+					// get the task to reduce the risk ( and at least block the current thread)
```
Baunsgaard, line 78: This fixes the issue at least for one thread caller.

src/main/java/org/apache/sysds/runtime/util/DoubleBufferingOutputStream.java
```diff
@@ -75,6 +75,9 @@ public void write(byte[] b, int off, int len) throws IOException {
 					// To avoid this we simply write the given byte array without a buffer.
 					// This approach only works if the caller adhere to not modify the byte array given
 					_locks[_pos] = _pool.submit(() -> writeBuffer(b, off, len));
+					// get the task to reduce the risk ( and at least block the current thread) 
+					// to avoid race conditions from callers.
+					_locks[_pos].get();
```
Baunsgaard, line 80: This fixes the issue for a single thread caller.

PR #2024, opened 2024-05-07
Systemds 3696/add incsliceline builtin

fzoepffel:
In this PR a builtin for incremental sliceline was added and incorporated in SystemDS. The builtin currently does not work as intended, as it is only meant to provide the overall structure for further development. The function currently does mostly the same as the original non-incremental SliceFinder builtin. It was, however, extended to be able to receive a lattice of slices, which was output on a previous run of the algorithm, the corresponding statistics from the previous run, and the whole feature matrix as a combination of all prior input feature matrices. Further on, the function now outputs the lattice, which was computed within the algorithm, the corresponding statistics and the combined original and new feature matrices to use in the next runs. Also JUNIT tests were added that compare the results of incSliceLine with the ones of SliceFinder. 

Contributor:
Frederic Zoepffel
f.zoepffel@gmail.com

### DIFF
--- a/scripts/builtin/incSliceLine.dml
+++ b/scripts/builtin/incSliceLine.dml
@@ -28,7 +28,7 @@
 # INPUT:
 # ---------------------------------------------------------------------------------------
 # newX     Feature matrix in recoded/binned representation
-# oldX     Original feature matrix 
+# oldX     All-comprising feature matrix of previous runs in recoded/binned representation
 # e        Error vector of trained model
 # k        Number of subsets required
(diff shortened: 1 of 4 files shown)


mboehm7:
LGTM - thanks @fzoepffel for getting started on incremental slice line. During the merge, I left a TODO in the builtin to convert the new inputs and outputs to lists for easier collection and handling.

mboehm7:
In future, please also use tabs over spaces in Java tests.

PR #2026, opened 2024-05-30
[SYSTEMDS-3695] Fix frame builtin nary append for Spark backend

e-strauss:
This patch implements the nary append operation for frames (cbind and rbind). For the cbind, there are two physical implementation:
- map-size cbind (w/ broadcast)
- reduce-size cbind
I added test cases to check the functionality of the n-ary operators and I also enabled the Word-Error-Rate test case.

Additionally, this patch fixes the map-size cbind operation to handle misaligned frameblocks.
As discussed, I will make another PR for the misalignment fix of the other physical operator


mboehm7:
LGTM - thanks for the patch @e-strauss. During the merge I only fixes minor warnings and formatting issues.

PR #2027, opened 2024-05-31
raSelection function implementation

gghsu:
This raSelection-function takes a matrix data set as input from where it performs relational operations: selection


mboehm7:
LGTM - thanks for the patch @gghsu. This contribution already looked pretty good, I just made minor changes during the merge (formatting, unnecessary imports, and additional assertions to check that our rewrites simplify the function to a single comparison, which they do).

PR #2030, opened 2024-06-03
[SYSTEMDS-3213] Builtin for cluster-based quantization

canabdulla:
The quantizeByCluster function quantizes numerical vectors via product quantization. Vectors are split into sub-vectors and ran through K-Means clustering.


mboehm7:
LGTM - thanks for the patch @canabdulla. The new builtin and tests very already pretty good, during the merge I just fixed some syntactic issues (tab indentation in Java code, and unnecessary free lines at the end of dml scripts).

PR #2031, opened 2024-06-03
[SYSTEMDS-3695] Fix frame binary append for Spark for misaligned inputs (FrameAppendRSPInstruction)

e-strauss:
This patch fixes the binary cbind operation for the spark backend (FrameAppendRSPInstruction) for the edge case when the inputs have different block alignments.
The RHS input is aligned to the LHS by the following steps:
- collect on Index RDD of the LHS, sort the array of indices
- map each FrameBlock from RHS to the correct FrameBlock on the LHS by doing binary search on the index array
   - > if necessary slice block on the RHS and map it to multiple Blocks on the LHS
   
Additionally, this patch adds test cases for different edge cases regarding the alignment. E.g.
misaligned_cbind_example

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/spark/FrameAppendRSPInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/spark/FrameAppendRSPInstruction.java
@@ -21,6 +21,7 @@ package org.apache.sysds.runtime.instructions.spark;
 
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.function.Function;
+import org.apache.spark.api.java.function.PairFlatMapFunction;
 import org.apache.spark.api.java.function.PairFunction;
 import org.apache.sysds.hops.OptimizerUtils;
 import org.apache.sysds.runtime.controlprogram.context.ExecutionContext;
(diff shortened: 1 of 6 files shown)


mboehm7:
LGTM - thanks for the additional improvement @e-strauss. I liked the ASCII art in the tests. During the merge I left a TODO comment to keep the information if a frame RDD is already aligned (so we only conditionally align if needed), and fixed the alignment of ASCII art (which where a bit distorted due to a mix of spaces and tabs).

e-strauss:
Thanks - I guessed I will be simple way to understand the test case. When you say keep the information do mean store this information in the meta data field for each RDD or a function which checks for alignment, and than conditionally do the alignment.

mboehm7:
I was thinking it would be very good to add a primitive in `FrameRDDConverterUtils` to check if the distributed collection is already 1K-aligned (by scanning and aggregating all block meta data), another one to convert it to 1K blocks (this is expensive because of shuffling), and then keep this information in the `FrameObject`'s DataCharacteristics (or subclass like FameCharacteristics) as a simple boolean. Whenever we then come to binary operations like the frame cbind, we can check this meta data and potentially resort back to a normal block append. Feel free to have a look into it if you're interested, but it's nice to have right now.

PR #2032, opened 2024-06-16
Optimization of I/O path of python interface

cs-ally:


Baunsgaard:
Hi @cs-ally 

Thanks for the PR.

Before I go into details on the PR, I suggest you address the failing test cases in the Python API. The log indicates that they relate to your implementation.

Best regards
Sebastian

cs-ally:
processing_times_comparison
Processing time improvement visualisation

Baunsgaard:
> processing_times_comparison Processing time improvement visualisation

great, good results. The plot can be improved by marking relative speedup as in 5x faster etc.
It would be good if we can have the old implementation as a fall back solution in the cases with many columns, to not have any speed downs.

cs-ally:
Screenshot 2024-07-27 at 22 19 52

src/main/java/org/apache/sysds/runtime/util/Py4jConverterUtils.java
```diff
@@ -114,6 +120,82 @@ public static MatrixBlock convertPy4JArrayToMB(byte[] data, int rlen, int clen,
 		return mb;
 	}
 
+	public static Array<?> convert(byte[] data, int numElements, Types.ValueType valueType) {
```
Baunsgaard, line 123: we need to add some internal tests (java only) for this function.
however the implementation does look fine. There are some minor improvements possible if we do not cast the array types inside the inner loops but assign the cast arrays outside the loop and then use that allocation.

src/main/python/systemds/utils/converters.py
```diff
@@ -116,20 +121,46 @@ def pandas_to_frame_block(sds, pd_df: pd.DataFrame):
...
-        for j, col_name in enumerate(col_names):
-            col_data = pd_df[col_name].fillna("").to_numpy(dtype=str)
-            for i in range(col_data.shape[0]):
-                if col_data[i]:
-                    j_dataArray[i][j] = col_data[i]
-        fb = jc_FrameBlock(j_valueTypeArray, j_colNameArray, j_dataArray)
-
-        return fb
+        
+        if rows > 4:
```
Baunsgaard, line 125: add a comment for why we make this differentiation.

src/main/python/tests/iotests/test_io_pandas_systemds.py
```diff
@@ -0,0 +1,89 @@
...
+        print("result_df:")
+        result_df['C2'] = result_df['C2'].astype(int)
+        print(result_df)
+        print("self.df:")
+        print(self.df)   
+        self.assertTrue(isinstance(result_df, pd.DataFrame))
+        self.assertTrue(self.df.equals(result_df))
+
+if __name__ == "__main__":
+    unittest.main(exit=False)
```
Baunsgaard, line 89: newline in the end.

Baunsgaard (commented):
Overall LGTM, we just need to add an internal Java test, to make the code coverage happy.

src/main/java/org/apache/sysds/runtime/util/Py4jConverterUtils.java
cs-ally: I wasn’t able to find existing tests for the Py4jConverterUtils class. Could you please advise where I should add new tests?

Baunsgaard: I don't think we have any.
You will have to make some from scratch.
It should be fairly simple to allocate byte[] and then call the methods. 

Please put the new tests in 

src/test/java/org/apache/sysds/test/component/frame/array

Baunsgaard:
Merging in #2065  , 
Thanks for the contribution @cs-ally 
While merging I fix the remaining issues, with tabs vs spaces and such.

PR #2035, opened 2024-06-20
[MINOR] add SVD and Eigen to Python API

Baunsgaard:
This PR adds a SVD and Eigen to the Python API.

It also is intended as an example for adding more builtin functions to the python API.


mboehm7:
Thanks - LGTM and I'll let Tessa know that she can extend the TPCx-AI implementation accordingly.

PR #2037, opened 2024-06-26
raJoin and raGroupby function implementation

gghsu:
1. The raGroupby-function takes a matrix data set as input from where it performs relational operations: groupby
In the output matrix, the first column of each row is the group value, and then the remaining elements in the original row are put into the result matrix in order.

2. The raGroupby-function takes a matrix data set as input from where it performs relational operations: groupby
In the output matrix, joined elements of two relations are assigned to the same row in the result matrix.

3. I also have a mirror update in raSelection.dml, just an unmatched Quotation Mark.


mboehm7:
LGTM - thanks for the patch @gghsu. During the merge I fixed the builtin documentation (copy-paste error from selection), minor formatting issues (DML scripts use 2 space indentation), and added a method parameter to both new builtin functions. In following PRs it would be good to then add more methods (beyond nested loop) for efficient group-by and joins.

PR #2039, opened 2024-07-01
Systemds 3696/incSliceLine uses stats from previous runs

fzoepffel:
In this PR incremental SliceLine was further developed. Three main functionalities have been implemented:
1. IncSliceLine computes slice sizes for changed slices by adding the slice size from the previous run to the slice size in the current run. 
2. In the function createAndScoreBasicSlices unchanged slices are filtered out of the computation. 
3. Further more advanced tests were added that split the Salaries dataset into two parts to run incSliceLine in two iterations over the whole dataset. The final results are compared with the results retrieved from running the non-icremental version of SliceFinder. 

Currently a few lines are commented out, as the tests in which either tpEval or selFeat are true run into runtime errors. However when these inputs are deactivated all tests run through without errors.


mboehm7:
LGTM - thanks for the patch @fzoepffel and sorry for the delay, this is a great improvement.

PR #2044, opened 2024-07-05
Add a new sort-based as a second method for raJoin

gghsu:
Implement the sort-based function as a second method for raJoin. For this commit, I just compare the output matrix from sort-based to my known result to ensure the function's correctness.


mboehm7:
LGTM - thanks for the patch @gghsu. During the merge, I added the default join method, renamed the new method to 'sort-merge', and let a few TODOs for future improvements (in a follow up project). In the future, please start a new PR instead of including the old commits (which creates unnecessary work for conflict resolution). Thanks.

mboehm7:
closed manually because I forgot the sharp prefix in https://github.com/apache/systemds/commit/f81b76de742b1815f949c9d2aa9a13eb752499f2

PR #2046, opened 2024-07-08
Python API Compare Performance

R4nne:


Baunsgaard:
Hi @R4nne ,

Thanks for the PR.

The code unfortunately does not fit well, and the script placed in the root folder does not contribute to the system overall. Furthermore, the hard-coded paths make it impractical for reproducibility. 

If you would like to contribute I would suggest that you move your script to 'scripts/perftest/python/...' and improve on the comparison such that I would also understand what is going on. Additionally could you please translate your comments to English since the system and community use English for communication?

PR #2047, opened 2024-07-08
Add BuiltinImputeMARTest and ImputeMARTest.dml

11iy:


Baunsgaard:
hi @11iy ,

I can see this PR unfortunately breaks the system. 
I would suggest to remove the modifications of the files in '/bin/'

Furthermore, the intention of the task is to add a function that is able to detect the type. Therefore a test case is not sufficient.

mboehm7:
Thanks for the patch @11iy and sorry for the delay - as discussed offline, I will help to clean this up and move this to staging once its mostly operational.

PR #2048, opened 2024-07-10
[SYSTEMDS-3704] Add new tests to ILinearizeTest

aeup:
This pull request adds new tests to ILinearizeTest.
Each new test loads a different config file with a linearization order. The tests check that the linearization order is set correctly in the configuration of SystemDS. They also run a .dml file and compare the output with the output of an equivalent RScript file.

cc @mboehm7


mboehm7:
LGTM - thanks for the patch @aeup and sorry for the delay. During the merge I only replaced a remaining print check with an actual assertion. Thanks.

PR #2052, opened 2024-07-23
Add permutation-matrix as second method for raGroupby

gghsu:
In this pull request, I have done below changes
1. I add permutation-matrix as second method for raGroupby,  for a more efficient raGroupby which constructs a permutation matrix (0/1 matrix) in order to copy relevant tuples with a single matrix multiplication followed by a reshape to bring the matrix into the form of one row per group. I also test multiple methods at the the same time in the runRaGroupbyTest function.
2. I improve the sort-merge function in raJoin by avoid iterating over rows.

### DIFF
--- a/scripts/builtin/raGroupby.dml
+++ b/scripts/builtin/raGroupby.dml
@@ -26,7 +26,7 @@
 # ------------------------------------------------------------------------------
 # X         Matrix of input data [shape: N x M]
 # col       Integer indicating the column index to execute grupby command
-# method    Groupby implemention method (nested-loop)
+# method    Groupby implemention method (nested-loop, permutation-matrix)
 # ------------------------------------------------------------------------------
 #
 # OUTPUT:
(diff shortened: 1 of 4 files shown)


mboehm7:
Thanks for the further improvement @gghsu. May I ask you to resolve the following three issues? First, please test all group-by method like we do in the raJoin tests. Second, please remove the unnecessary print(toString) debug outputs which are overhead hazards with large data. Third, keep the 2-space indentation in dml scripts (which I fixed during previous merges). Thanks, and once these minor issues are resolved, we can merge it it.

gghsu:
Hi @mboehm7 

Thanks for pointing out, I have rewritten the regroupby testcases, removed unnecessary print, and align the line format in dml files. Please check, thank you!

mboehm7:
LGTM - thanks for the patch @gghsu, this is very good.

PR #2056, opened 2024-07-28
Add hash2 as fourth method for raJoin

gghsu:
In this pull request, I implemented hash2 as the fourth method for raJoin according to the original paper's function, which performs the join without sort or loop-based concatenation but currently only supports one-to-many, one-to-one, non matching joins. (The primary key of the left table needs to be unique)
Unfortunately, because I haven't resolved the issue that occurred in the hash collision at this time, so I made the hash table size larger than the largest number between two tables temporarily to avoid any hash collision.
m = max(max(left),max(right)) + 1;


mboehm7:
LGTM - thanks for the patch @gghsu and bringing in this additional method.

PR #2060, opened 2024-07-30
[SYSTEMDS-3701] Add new modalities and representations to Scuro

christinadionysio:
This PR adds new modalities and a revised structure for the combination of multiple modalities. 
Additionally, a number of different unimodal and multimodal representations are added to the library.


mboehm7:
LGTM - thanks for the improvements of the Scuro library @christinadionysio.

PR #2061, opened 2024-08-02
Adding TPCx-AI Benchmark

theidkamp:
Adding an initial implementation for the TPCx-AI benchmark on SystemDS. Includes files necessary for setup and a guide on how to set up the benchmark.


mboehm7:
Thanks for this very valuable contribution @theidkamp. I will think about how we handle these TPC/Intel licenses (ok to bring in?, ok to change to ASF?) which are currently breaking the build. Maybe we only bring in `UseCase01.py` through `UseCase10.py` which are utilizing the SystemDS Python API and thus should carry our ASF license?

PR #2045, opened 2024-07-05
Implementation of N-Gram statistics

Jaybit0:
This PR implements N-Gram statistics of instruction sequences when adding `-ngrams [<comma-separated tuple sizes> <topK>]` to the command line arguments. For instance, the CLI arguments `-ngrams 2,3 10` will record 2-grams and 3-grams of instruction sequences and will print the top 10 entries of all 2-grams and 3-grams. N-Grams are maintained for each thread individually and are merged together when printing the statistics. 

The table is currently sorted by the cumulative execution time of the instruction sequence (over all occurrences of that sequence). For example, a table could look like this:
```
Most common 2-grams (sorted by absolute time):
  #  N-Gram                 Time(s)  StdDev(t)/Mean(t)  Count
  1  (sp_rblk, createvar)     0,281    (0.857, 0.228)      2
  2  (createvar, sp_rblk)     0,281    (0.676, 0.857)      2
  3  (write, write)           0,134                 -      1
  4  (rmvar, write)           0,082                 -      1
  5  (write, rmvar)           0,053                 -      1
  6  (createvar, *)           0,014    (2.792, 1.971)     66
  7  (==, mvvar)              0,014    (3.299, 2.433)     11
  8  (round, rmvar)           0,012                 -      1
  9  (createvar, round)       0,011                 -      1
 10  (*, rmvar)               0,008    (2.303, 2.725)     57
```
@mboehm7

PR #2062, opened 2024-08-03
N-Gram statistics using lineage

Jaybit0:
This PR is an extension to PR #2045 and implements support for data-dependent n-grams using and extending the existing lineage functionality. As we are dealing with DAGs which are not linear sequences of instructions, I implemented the extension in such a way that it tracks every instruction path of the length `n`. If we had for example the DAG `(a*b + c/d)` and wanted to record bigrams, the two operation sequences `[(*, +), (/, +)]` would be added to the bigram store. I also keep track of the individual data-types of each instruction which is why I extended the existing lineage functionality as the `_data` string is sometimes empty and contains inconsistent information.

The n-gram table now looks like this, where the arguments within brackets show the input parameters of an instruction (separated by `°`) and the suffix `[i]` represents the parameter index for the following instruction (e.g. for the first entry the result of `rblk` is used as the second paremeter for `ba+*`):
```
Most common 2-grams (sorted by absolute time):
  #  N-Gram                          Time(s)  StdDev(t)/Mean(t)  Count
  1  (rblk·MATRIX·FP64(MATRIX·FP64)    1,144         (, 1.067)      4
     [1], ba+*·MATRIX·FP64(MATRIX·F                                  
     P64 ° MATRIX·FP64))                                             
  2  (rblk·MATRIX·FP64(MATRIX·FP64)    0,853         (, 0.469)      3
     [0], ba+*·MATRIX·FP64(MATRIX·F                                  
     P64 ° MATRIX·FP64))                                             
  3  (createvar·MATRIX·FP64()[0], r    0,343    (0.627, 0.929)      2
     blk·MATRIX·FP64(MATRIX·FP64))                                   
  4  (rblk·MATRIX·FP64(MATRIX·FP64)    0,285                 -      1
     [0], cpvar·MATRIX·FP64(MATRIX·                                  
     FP64))                                                          
  5  (+*·MATRIX·FP64(MATRIX·FP64 °     0,153                 -      1
     SCALAR·FP64 ° MATRIX·FP64)[0],                                  
      write·MATRIX·FP64(MATRIX·FP64                                  
      ° L_SCALAR·STRING ° L_SCALAR·                                  
     STRING ° L_SCALAR·INT64))                                       
```
@mboehm7

### DIFF
--- a/src/main/java/org/apache/sysds/api/DMLOptions.java
+++ b/src/main/java/org/apache/sysds/api/DMLOptions.java
@@ -57,6 +57,7 @@ public class DMLOptions {
 	public int                  statsCount    = 10;               // Default statistics count
 	public int[]                statsNGramSizes = { 3 };          // Default n-gram tuple sizes
 	public int                  statsTopKNGrams = 10;             // How many of the most heavy hitting n-grams are displayed
+	public boolean              statsNGramsUseLineage = true;     // If N-Grams use lineage for data-dependent tracking
 	public boolean              fedStats      = false;            // Whether to record and print the federated statistics
 	public int                  fedStatsCount = 10;               // Default federated statistics count
 	public boolean              memStats      = false;            // max memory statistics
@@ -219,7 +220,7 @@ public class DMLOptions {
 		dmlOptions.statsNGrams = line.hasOption("ngrams");
 		if (dmlOptions.statsNGrams){
 			String[] nGramArgs = line.getOptionValues("ngrams");
-			if (nGramArgs.length == 2) {
+			if (nGramArgs.length >= 2) {
 				try {
 					String[] nGramSizeSplit = nGramArgs[0].split(",");
 					dmlOptions.statsNGramSizes = new int[nGramSizeSplit.length];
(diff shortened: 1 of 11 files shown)


phaniarnab:
I am not fully aware of the goal of this project. But it seems to me that adding these new entries to the LineageItem class can significantly increase the size of each object, which in turn impacts the lineage tracing overhead, as well as the cache usage behavior. If materializing these entries for each object is absolutely necessary, I recommend writing a separate class (e.g., LineageItemExtended) connected via a single entry. However, as I am unsure about the full project, I will let @mboehm7 comment on this first.

Jaybit0:
Yes this makes sense. As this functionality is hardly used in production, it is probably better to avoid this kind of overhead. I think I can work my way around this by storing the additional data in a hashmap linking the additional data to the object. Then it would add no overhead if the n-gram statistics option is switched off.

Jaybit0:
I have now implemented the changes to not increase the object size of LineageItem. This will add some minor overhead to my code but execution speed should not matter too much when running those statistics.

mboehm7:
@Jaybit0 could you please rebase this PR? It seems to include the commits from the previous (already merged) PR. Thanks.

mboehm7:
LGTM - thanks for the patch @Jaybit0. During the merge I move the code from ProgramBlock into a dedicated method in statistics, and also removed all unnecessary imports.

PR #2063, opened 2024-08-07
Systemds 3696/add incsliceline builtin

fzoepffel:
In this PR incremental SliceLine was further developed.  The following functionalities were added:

UPDATE Aug. 9: also added handling of removed slices. Now a list of indices can be provided that indicate which slices of the previous dataset are removed from the dataset. This information is used for pruning, too.

- a params list parameter ensures consistent parameters over all runs
- further parameters were added that are necessary for pruning
- a function was added to determine the unchanged slices on each level
- several pruning strategies were added or adjusted. e.g. score pruning based on the lowest (updated) score of the previous top k was added (basic slices and normal slices are pruned), unchanged slices under the minSup are removed right away. 
- consistent one hot encoded matrices are ensured 
- encoding and decoding functions with domain size scaling were added. If the parameter encodeLat is set to true the lattice is stored in an encoded form for reduced size. 
- the prevStats only include the slice sizes as the other columns are not needed for pruning
- the code was cleaned up, comments were added, larger code snippets were replaced by functions
- first experiments have indicated a strong decrease in execution time, especially if the number of added tuples is significantly smaller than the total number of slices in an incremental run. This is the case because a larger number of added tuples usually tends to change a larger portion of slices. Only unchanged slices can lead to significant efficiency increases.


mboehm7:
@fzoepffel could you please rebase this PR in order to resolve the merge conflicts (as it seems to include the commits from the previous PR too)? Thanks

fzoepffel:
> @fzoepffel could you please rebase this PR in order to resolve the merge conflicts (as it seems to include the commits from the previous PR too)? Thanks

@mboehm7 Thank you for your reminder. Yes, the problem was that I had kept on working on the same branch I used for the previous PR. I tried to squash the duplicate commits, but that did not want to work for some reason. But, I just rebased the branch and now, there are no more conflicts with the base branch!

mboehm7:
LGTM - thanks for the patch and the rebasing help @fzoepffel. I'll merge it in and this week I will also start the work on our related join paper.

PR #2066, opened 2024-08-16
[SYSTEMDS-3719] Binary Cell Operations Specialization

Baunsgaard:
As part of BWARE, i optimized some of the uncompressed binary operations, 
for instance divide had some elements that could be specialized.


mboehm7:
Thanks for sharing @Baunsgaard - are these numbers after recompilation or on the first instance? Just asking, as we need to optimize for the former (i.e., post recompilation).

Baunsgaard:
> Thanks for sharing @Baunsgaard - are these numbers after JIT compilation or on the first instance? Just asking, as we need to optimize for the former (i.e., post JIT).

The numbers are average of 1000 repetitions and with 10 repetition warmup.

mboehm7:
very good - thanks.

PR #2067, opened 2024-08-16
Resource Elasticity: Enumeration Basics

lachezar-n:
Here are the bases for the Enumerators for Resource optimisation with the first 2 enumerators:
- GridBasedEnumerator
- InterestBasedEnumerator

In addition to that, some utility classes/methods were created for recompilation in the current case. It is still to be investigated how much better this recompilation approach is compared to the existing one.

To allow the desired functionality, small changes were made to the following classes of the existing SystemDS code base:
- src/main/java/org/apache/sysds/conf/CompilerConfig.java
   * creating a new configuration type to avoid further changes that would affect the basic functionality at script execution
- src/main/java/org/apache/sysds/hops/Hop.java
   * the only place the new configuration type comes in place to allow reassigning execution type
- src/main/java/org/apache/sysds/lops/compile/Dag.java
   * resetUniqueMembers() - to be used at (re)compilation 
- src/main/java/org/apache/sysds/runtime/controlprogram/context/SparkExecutionContext.java
   * initLocalSparkContext() to avoid real Spark context initialisation
- src/main/java/org/apache/sysds/runtime/controlprogram/parfor/stat/InfrastructureAnalyzer.java
   * allowing setting local parallelism from outside 

Besides these required changes I have made a tiny change that looked to me that could be useful for the SystemDS project overall:
- src/main/java/org/apache/sysds/hops/OptimizerUtils.java: replacing `getNumMappers` and `getNumReducers` to `getNumTasks` since Spark and not MapReduce is the new/current framework used for distributed execution

### DIFF
--- a/src/main/java/org/apache/sysds/conf/CompilerConfig.java
+++ b/src/main/java/org/apache/sysds/conf/CompilerConfig.java
@@ -78,7 +78,9 @@ public class CompilerConfig
 		CODEGEN_ENABLED,
 		
 		//federated runtime conversion
-		FEDERATED_RUNTIME;
+		FEDERATED_RUNTIME,
+		// resource optimization mode
+		RESOURCE_OPTIMIZATION;
 	}
 	
 	//default flags (exposed for testing purposes only)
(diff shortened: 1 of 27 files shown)


mboehm7:
LGTM - thanks for the patch and the tests @lachezar-n. During the merge I fixed the missing license headers, the incorrect formatting (tabs over spaces in Java code), and a few warnings. Initially, the tests `loadInstanceTableTest` and `loadCSVFileAWSTest` were failing for me (concurrent file access in Windows I/O) but after fixing the warnings (one of which was a file resource leak), these tests passed then as well. Thanks.

PR #2068, opened 2024-08-19
[SYSTEMDS-3701] Create Data Representation Search (DRSearch) Primitive

christinadionysio:
This PR introduces the new Data Representation Search (DRSearch) primitive for the Scuro library. 
The primitive includes a `.fit()` method that creates joint representations for all modality combinations using the provided representations and evaluates these representations against a given task. The included `.transform()` method takes the best parameters that were defined in the fit method to create the final data representation.


mboehm7:
LGTM - thanks for the patch @christinadionysio. That's a great starting point. I would recommend to also add basic tests (e.g., accuracy >= default accuracy of the ML application).

PR #2069, opened 2024-08-22
Decision Tree Predict GEMM strategy adapted to new layout

e-strauss:
This commit updates the GEMM strategy for evaluating decision tree to the new decision tree layout. Additionally, this patch adds three new test cases for decision tree predict.

I kept for now the "<=" [instead of "<"] because the DecisionTreeRealDataTest was failing after changing to "<", and added a comment in the code for future fix / investigation.


mboehm7:
LGTM - thanks for the patch @e-strauss. I also removed the TODO and kept the train/predict methods consistently at lte,

PR #2071, opened 2024-08-22
[WIP][SYSTEMDS-3259] Shampoo optimizer

OlgaOvcharenko:
This PR adds [Shampoo](https://arxiv.org/pdf/1802.09568) optimizer (matrix case, same to Algorithm 1).


mboehm7:
Thanks @OlgaOvcharenko for getting started on this new optimizer. I would recommend to return L and R as well from the update function (similar to we do it in Adam), add the init function (can be zero initialized), and maybe add the reference to the function documentation.

PR #2077, opened 2024-08-30
[SYSTEMDS-3735] Python API Builtin exp 

e-strauss:
This patch adds the built in operator for exponential to the python api.


Baunsgaard:
merged, I had to reformat the test while merging.

PR #2078, opened 2024-08-30
[SYSTEMDS-3734] Python API Builtin floor & ceil 

e-strauss:
This patch adds the built in operators for floor and ceiling to the python api.


Baunsgaard:
Merged, while merging I had to fix the formatting of the tests.

PR #2079, opened 2024-08-30
 [SYSTEMDS-3733] Python API Builtin log 

e-strauss:
This patch adds the built in operator for logarithm to the python api.


Baunsgaard:
Merged, similarly i had to reformat the test.

PR #2086, opened 2024-09-02
[SYSTEMDS-3701] Add LSTM representation to scuro

christinadionysio:
This commit adds an LSTM representation to scuro.


mboehm7:
LGTM - thanks for the patch @christinadionysio.

PR #2087, opened 2024-09-02
[SYSTEMDS-3740] Python API Builtin countDistinct

e-strauss:
[SYSTEMDS-3740] Python API Builtin countDistinct

This patch adds the builtin operator for countDistinct to the python api.

### DIFF
--- a/src/main/python/systemds/operator/nodes/matrix.py
+++ b/src/main/python/systemds/operator/nodes/matrix.py
@@ -244,6 +244,22 @@ class Matrix(OperationNode):
         raise ValueError(
             f"Axis has to be either 0, 1 or None, for column, row or complete {self.operation}")
 
+    def countDistinct(self, axis: int = None) -> 'OperationNode':
+        """Calculate the number of distinct values of matrix.
+
+        :param axis: can be 0 or 1 to do either row or column aggregation
+        :return: `Matrix` representing operation
+        """
+        if axis == 0:
+            return Matrix(self.sds_context, 'colCountDistinct', [self])
+        elif axis == 1:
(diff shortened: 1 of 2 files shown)


src/main/python/systemds/operator/nodes/matrix.py
```diff
@@ -244,6 +244,22 @@ def min(self, axis: int = None) -> 'OperationNode':
...
+        :return: `Matrix` representing operation
+        """
+        if axis == 0:
+            return Matrix(self.sds_context, 'colCountDistinct', [self])
+        elif axis == 1:
+            return Matrix(self.sds_context, 'rowCountDistinct', [self])
+        elif axis is None:
+            return Scalar(self.sds_context, 'countDistinct', [self])
+        raise ValueError(
+            f"Axis has to be either 0, 1 or None, for column, row or complete {self.operation}")
```
Baunsgaard, line 260: This PR is missing a test case for the error.

e-strauss: @Baunsgaard added, I also removed the print statements and squashed + rebased

e-strauss: @Baunsgaard the older test cases for the aggregation like min, max and co, it makes sense to add test cases there as well - or are there some in place which I havent found?

Baunsgaard: then we need to add it for them as well, 
unfortunately we do not see these things because we do not have code coverage for the Python API,
It would be cool to have that as well.

Baunsgaard:
LGTM, merged

PR #2090, opened 2024-09-03
[SYSTEMDS-3751, SYSTEMDS-3752, SYSTEMDS-3753] Python API missing buitin isNA, isNaN, isInf

e-strauss:
[SYSTEMDS-3751, SYSTEMDS-3752, SYSTEMDS-3753] Python API missing buitin isNA, isNaN, isInf

This patch adds the built in operator for isNA, isNaN, isInf to the python api.

### DIFF
--- a/src/main/python/systemds/operator/nodes/matrix.py
+++ b/src/main/python/systemds/operator/nodes/matrix.py
@@ -502,6 +502,29 @@ class Matrix(OperationNode):
         """
         return Scalar(self.sds_context, 'toString', [self], kwargs, output_type=OutputType.STRING)
 
+    def isNA(self) -> 'Matrix':
+        """ Computes a boolean indicator matrix of the same shape as the input, indicating where NA (not available)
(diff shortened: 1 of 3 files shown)


Baunsgaard:
is this operation only valid for matrices, if that is the case, we need to fix it in the system in general.
If not, we need to add it to the scalar node as well.

e-strauss:
> is this operation only valid for matrices, if that is the case, we need to fix it in the system in general. If not, we need to add it to the scalar node as well.

I will check that, for now I used the dml-language-reference as guideline and checked in java only when operator is not mentioned.

e-strauss:
updated for scalar as well

Baunsgaard:
LGTM, merged, while merging i formatted the test, it is only minor things.
I can suggest that you adhere to the 'black' formatted code, since this is the auto formatter that I 'unofficially' use for the code in the Python API.

PR #2093, opened 2024-09-03
[SYSTEMDS-3744, SYSTEMDS-3745, SYSTEMDS-3746, SYSTEMDS-3747, SYSTEMDS-3748] Python API Builtin cumsum, cumprod, summin, cummax, cumsumprod

e-strauss:
[SYSTEMDS-3744, SYSTEMDS-3745, SYSTEMDS-3746, SYSTEMDS-3747, SYSTEMDS-3748] Python API Builtin cumsum, cumprod, summin, cummax, cumsumprod

This patch adds the builtin operators cumsum, cumprod, summin, cummax, cumsumprod to the python api.

### DIFF
--- a/src/main/python/systemds/operator/nodes/matrix.py
+++ b/src/main/python/systemds/operator/nodes/matrix.py
@@ -671,5 +671,43 @@ class Matrix(OperationNode):
             },
         )
 
+    def cumsum(self) -> "Matrix":
+        """Column prefix-sum. (For row-prefix sum, use X.t().cumsum().t())
+
+        :return: The Matrix representing the result of this operation
+        """
+        return Matrix(self.sds_context, "cumsum", [self])
+
+    def cumprod(self) -> "Matrix":
+        """Column prefix-product. (For row-prefix prod, use X.t().cumprod().t())
+
+        :return: The Matrix representing the result of this operation
+        """
+        return Matrix(self.sds_context, "cumprod", [self])
+
+    def cumsumprod(self) -> "Matrix":
+        """Column prefix-sumprod of an 2-column matrix:
+        Y = X.comsumprod(),
+        where Y[i,1] = X[i,1] + X[i,2]*Y[i-1,1] for i in [1,2, .., nrow(X)]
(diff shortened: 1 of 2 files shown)


src/main/python/systemds/operator/nodes/matrix.py
```diff
@@ -522,5 +522,12 @@ def replace(self, pattern: VALID_INPUT_TYPES, replacement: VALID_INPUT_TYPES) ->
         """
         return Matrix(self.sds_context, "replace", named_input_nodes={"target": self, "pattern": pattern, "replacement": replacement})
 
+    def cumsum(self) -> 'Matrix':
+        """ Column prefix-sum. (For row-prefix sum, use X.t().cumsum().t())
```
Baunsgaard, line 526: interesting, do we not support cumulative sum on rows?

e-strauss: jep, afaik. It says so in the docu (even Arnab told me yesterday not trust it ^^) and java code suggests it too

e-strauss:
reformatted

src/main/python/systemds/operator/nodes/matrix.py
Baunsgaard: @mboehm7 

I think we should probably make a task for this, if it really is not supported.
It seems odd to to two transposes to do row cumulative operations.

mboehm7: well, its fine to add a task even through the focus was always R compliance and are defines it as column-wise cumulative aggregates (unary operation without additional parameters). The same also applies to rev(). Recently some of the operations like roll() are more inspired by numpy and there you have indeed usually axis parameters.

mboehm7: Also when committing this task here lets call it [SYSTEMDS-3744-48] ... and in the future let's create JIRA tasks for groups of operations that belong together. Thanks.

Baunsgaard: i have made a task for it and marked it as a student project, unrelated to this PR, such that we do not forget it.

PR #2094, opened 2024-09-04
[MINOR] Python API: generator option for manual added algorithm builtin files

e-strauss:
[MINOR] Python API: generator option for manual added algorithm builtin files

This patch adds an option for the Python API generator to include manually added algorithm buitlins in the init file.


e-strauss:
@Baunsgaard once, merged I will make another rebased PR for the first manually added operator

PR #2095, opened 2024-09-04
[SystemDS-3750] Python API Builtin prod

e-strauss:
[SystemDS-3750] Python API Builtin prod

This patch adds the builtin operator for prod to the python api.
Additionally I added test cases, which checks for errors when selecting a invalid axis for sum, mean, var, min, max


e-strauss:
reformatted with black

Baunsgaard:
LGTM and merged

PR #2096, opened 2024-09-04
[SystemDS-3755] Python API Builtin inv

e-strauss:
[SystemDS-3755] Python API Builtin inv

formatted with black


Baunsgaard:
LGTM, and merged

PR #2097, opened 2024-09-04
[SYSTEMDS-3743] Python API Builtin cov

e-strauss:
[SYSTEMDS-3743] Python API Builtin cov

formatted with black


Baunsgaard:
LGTM and merged

PR #2098, opened 2024-09-05
Systemds 3696/selector-for-pruning-strategies

fzoepffel:
In this PR a new input variable was added to incSLiceLine to allow the selection of the main incremental pruning strategies Either both (pruningStrat = 0), only size (pruningStrat = 1), only score (pruningStrat = 2), or no pruning (pruningStrat = 3) can be selected. This is mainly necessary for experiments and testing.


mboehm7:
LGTM - thanks for the patch @fzoepffel. During the merge I fixed unnecessary imports and wrong formatting in the Java tests.

PR #2100, opened 2024-09-05
Fixed null-pointer exception when LU Decomposition is done on a singular matrix

e-strauss:
Fixed null-pointer exception when LU Decomposition is done on a singular matrix

I added a test case for code coverage


Baunsgaard:
LGTM and merged

PR #2103, opened 2024-09-06
[SYSTEMDS-3729] Add roll reorg operations in CP, python script test

min-guk:
This PR implements a new roll reorg function in CP and provides test scripts using Python.

1. Implementation of roll function for dense and sparse matrices in CP, along with test code
- Similar to `rev`, dense column vectors are rolled column-wise, while the rest are rolled row-wise.

2. Support for Python test code and additional test cases
- Since R does not provide a `roll` function, baseline test scripts were written using Python's numpy `roll` function.
- Added support for executing the test code in Python. Although there is some overlap with the R script execution code, I avoided modifying or refactoring the existing code to minimize changes.
- However, due to slight differences in the `mtx` format between R and Python, I made the following modifications to the `mtx` file input/output code:
  - `TestUtils::writeTestMatrix()`: The last number in the mtx header should represent the number of non-zero elements in the matrix, but the previous input test matrix was showing the total number of elements, which caused an error in Python. (https://networkrepository.com/mtx-matrix-market-format.html)
  - `TestUtils::readRMatrixFromFS()`: While R writes the mtx header in 2 lines, Python's scipy writes it in 3 lines, including an extra line with just a %. I modified the code to ignore this blank line.
  
Please feel free to share any feedback or suggestions regarding the implementation direction.


src/main/java/org/apache/sysds/lops/Transform.java
```diff
@@ -111,7 +111,11 @@ private String getOpcode() {
 			case REV:
 				// Transpose a matrix
 				return "rev";
-			
+
+			case ROLL:
+				// Transpose a matrix
```
Baunsgaard, line 116: remove comment

src/main/java/org/apache/sysds/runtime/functionobjects/RollIndex.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public boolean computeDimension(DataCharacteristics in, DataCharacteristics out) {
+		out.set(in.getRows(), in.getCols(), in.getBlocksize(), in.getNonZeros());
+		return false;
+	}
+
+	@Override
+	public void execute(MatrixIndexes in, MatrixIndexes out) {throw new NotImplementedException();
```
Baunsgaard, line 58: missing, but i guess it is intentionally.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
```diff
@@ -83,6 +85,31 @@ private ReorgCPInstruction(Operator op, CPOperand in, CPOperand out, CPOperand c
...
+	 *            cp input operand
+	 * @param shift
+	 *            ?
+	 * @param out
+	 *            cp output operand
+	 * @param opcode
+	 *            the opcode
+	 * @param istr
+	 *            ?
+	 */
```
Baunsgaard, line 106: remove docs if they are empty like this.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
```diff
@@ -83,6 +85,31 @@ private ReorgCPInstruction(Operator op, CPOperand in, CPOperand out, CPOperand c
...
+	private ReorgCPInstruction(Operator op, CPOperand in, CPOperand out, CPOperand shift, String opcode, String istr) {
+		super(CPType.Reorg, op, in, out, opcode, istr);
+		_col = new CPOperand();
+		_desc = new CPOperand();
+		_ixret = new CPOperand();
+		_shift = shift;
```
Baunsgaard, line 112: we could consider reusing some of the other ones, 
however, If you want to keep the structure like this, then assign the other CPOperands to null if you do not use them instead of assigning new object that we will not use.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -2243,7 +2271,70 @@ private static void reverseSparse(MatrixBlock in, MatrixBlock out) {
...
+		out.allocateSparseRowsBlock(false);
+
+		//copy all rows into target positions
+		SparseBlock a = in.getSparseBlock();
+		SparseBlock c = out.getSparseBlock();
+
+		// roll matrix with axis=0
+		shift %= (m != 0 ? m : 1);
+
+		for( int i=0; i<m-shift; i++ ) {
```
Baunsgaard, line 2326: you can improve the sparse implementation by processing only non zero elements:
Use something like:

(assuming i is a given row.)
```java
if(a.isEmpty(i))
	continue;
final int apos = sb.pos(i);
final int alen = sb.size(i) + apos;
final int[] aix = sb.indexes(i);
final double[] avals = sb.values(i);
for(int k = apos; k < alen; k++)
    ...
```

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
```diff
@@ -1263,16 +1267,98 @@ protected void runRScript(boolean newWay) {
 		}
 	}
 
-	private static String buildOutputStringR(String standardOut, String standardError){
+	/**
+	 * Runs an Python script
+	 */
+	protected void runPythonScript() {
```
Baunsgaard, line 1273: this method looks a bit funky, it could probably be cleaned up a bit.

src/test/java/org/apache/sysds/test/functions/reorg/FullRollTest.java
```diff
@@ -0,0 +1,137 @@
...
+	public void testRollMatrixSparseCP() {
+		runRollTest(TEST_NAME2, true, true, ExecType.CP);
+	}
+
+	/**
+	 *
+	 * @param sparseM1
+	 * @param sparseM2
+	 * @param instType
+	 */
```
Baunsgaard, line 82: remove docs if empty

src/test/java/org/apache/sysds/test/functions/reorg/FullRollTest.java
```diff
@@ -0,0 +1,137 @@
...
+				Assert.assertTrue("Missing opcode: "+Instruction.SP_INST_PREFIX+"roll", Statistics.getCPHeavyHitterOpCodes().contains(Instruction.SP_INST_PREFIX+"roll"));
+		}
+		finally
+		{
+			//reset flags
+			rtplatform = platformOld;
+			DMLScript.USE_LOCAL_SPARK_CONFIG = sparkConfigOld;
+		}
+	}
+}
```
Baunsgaard, line 137: please format the file for consistantcy, I can see it is not formatted because it for instance is missing a newline in the end.

Baunsgaard (commented):
Looks very good, just missing a few things.

src/main/java/org/apache/sysds/lops/Transform.java
min-guk: I deleted it.

src/main/java/org/apache/sysds/runtime/functionobjects/RollIndex.java
min-guk: Yes, that function should not be called.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
min-guk: I filled in the blanks in the comment.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
min-guk: Modified the code to initialize the unused variable to null.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
min-guk: Modified the code to copy only non-zero elements.

src/test/java/org/apache/sysds/test/functions/reorg/FullRollTest.java
min-guk: Deleted the comment.

src/test/java/org/apache/sysds/test/functions/reorg/FullRollTest.java
min-guk: I added a newline.

src/test/java/org/apache/sysds/test/AutomatedTestBase.java
min-guk: I have written the `runPythonScript()` function by following the structure of the existing `runRScript()` function to run Python scripts. However, I would like to ask for clarification on how you would like me to proceed with further modifications.

Would you prefer that I refactor the code to combine the `runRScript()` and `runPythonScript()` functions into a single function (e.g., `runBaselineScript()`), and also consolidate class variables such as `rCmd`?

min-guk:
Hello Sebastian,
Thank you for the detailed feedback. I've addressed everything except for runPythonScript(), for which I left a comment above and would appreciate your input.

Additionally, the rollTest Python script works fine locally, but there are environment issues on GitHub Actions. I'll resolve this soon by updating the javaTests.yml.

min-guk:
The Java tests pass locally but continue to fail in GitHub Actions because the `numpy` package was not installed in the test Docker image.

I have updated the `testsysds.Dockerfile`, so I would appreciate it if you could execute the `Docker Test Image Update` in GitHub Actions after the Pull Request is merged.

.github/workflows/javaTests.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           "**.functions.reorg.**,**.functions.rewrite.**,**.functions.ternary.**",
           "**.functions.transform.**","**.functions.unique.**",
           "**.functions.unary.matrix.**,**.functions.linearization.**,**.functions.jmlc.**"
+          "**.functions.reorg.**,**.functions.rewrite.**,**.functions.ternary.**"
```
Baunsgaard, line 88: this looks like a duplicate of line 85, therefore only double the number of times we run the test.

docker/testsysds.Dockerfile
```diff
@@ -78,6 +78,13 @@ RUN apt-get install -y --no-install-recommends \
 	&& rm -rf installDependencies.R \
 	&& rm -rf /var/lib/apt/lists/*
 
+RUN apt-get install -y --no-install-recommends \
+	python3 python3-pip && \
+	apt-get clean && \
+	python3 -m pip install --upgrade pip && \
+	pip3 install numpy scipy && \
+    echo "Python3 installed"
```
Baunsgaard, line 86: This is fine to add, however, i would prefer not to.

We have an---unrelated to your PR---problem with the testsysds docker file that would require a larger fix to make it build properly. Therefore, it will take some time before we would run the update and therefore be able to enable the test to call python from inside the system.

Fundamentally this blocks your PR, because your work relies on Python being installed on the machine you want to test on.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
Baunsgaard: the formatting is off then, here is a correctly formatted one:

```java
	/**
	 * Apply binary row operation on the right side with one extra row evaluating with zeros.
	 * 
	 * @param op         The operation to this dictionary
	 * @param v          The values to use on the right hand side.
	 * @param colIndexes The column indexes to consider inside v.
	 * @return A new dictionary containing the updated values.
	 */
	public IDictionary binOpRightAndAppend(BinaryOperator op, double[] v, IColIndex colIndexes);
```

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
Baunsgaard: Great, it can be further optimized, by minor tweaks such as:

1. Move the inner parts of the for loop to a separate function to enable JIT (the two loops can call the same function with different shift arguments)
2. Move the shift addition out of the inner inner loop but to a temporary variable

Baunsgaard (commented):
It is a bit unfortunate that the python is not installed in our test image (see specific comment for more details). Therefore, it is hard for you to call python and test the way you want to.

Alternatively, I can suggest and would actually recommend:

1. Write the test up in java, where you know what the result should be without comparing to Numpy.
2. And then in our Python API, write the test again, to compare to Numpy. (we have many examples of doing this in recent commits)

.github/workflows/javaTests.yml
min-guk: Thanks for letting me know.

docker/testsysds.Dockerfile
min-guk: Since I’m not familiar with the SystemDS Python API, I’ll study it and proceed as you suggested.

src/main/java/org/apache/sysds/runtime/instructions/cp/ReorgCPInstruction.java
min-guk: I reformatted all the code written in this PR and the code in `ReorgCPInstruction.java` that didn’t match the comment format using the IDE.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
min-guk: I appreciate you pointing out the optimization points I hadn’t considered, and I’ve reflected those in the revisions.

docker/testsysds.Dockerfile
e-strauss: @min-guk Hi, I have worked recently on the python api, if I can help you lmk

min-guk: Thank you both for your kind words. I would like to ask a question to confirm one thing.

Should I implement a Python API to make the roll function, which is implemented in Java, callable from Python, and then write Python test code to compare it with NumPy?

And for such examples, should I refer to https://github.com/apache/systemds/pull/2085 and https://github.com/apache/systemds/pull/2097?

e-strauss: jep, in the Python test you can compare to numpy, you can take a look at the test in #2085 . (#2097 would be an example for an operator with two matrices as input, but since you got one matrix sa input #2085 fits better), 
so you can extend the https://github.com/apache/systemds/blob/main/src/main/python/systemds/operator/nodes/matrix.py with your operator.
You can also take a look at the quantile operator there on how to pass scalars to your function call (for your shift paramerter)

min-guk: Thank you for the clarification and the references. I'll take a look and proceed accordingly.

min-guk:
I have created a Python API and added Python test code for roll function.

I have rolled back the code related to the MTX file reader/writer, which was separated into another PR, as well as the previously modified Java test code and Docker file script.

docker/testsysds.Dockerfile
min-guk: @e-strauss Thanks to your help, I was able to successfully implement and complete the Python API.

If possible, could you share an example of a Python API that calls a Spark function, along with its corresponding test code? It would be really helpful to me.

Baunsgaard: We currently do not actively assign spark instructions through the python API, because we did not expose the flag for it yet. 

The CP test would be sufficient from python and then internally you can set the function test to test the operation with SP enabled for an input that you know the output for. 

Together this would be sufficient.

min-guk: Thank you for letting me know. I will verify CP by comparing it with numpy in the Python test, and for SP, I will verify it by comparing the output with CP using the Java test.

min-guk:
Previously, the roll function incorrectly handled sparse matrix indices, leading to a test failure, but I have resolved the issue.

Baunsgaard:
LGTM the implementation looks solid. 
However, we are missing code coverage in the test. The python test is great to ensure that we give the same results as numpy, however out code coverage only include internal java tests.

I would suggest to add a component test that just covers the call to LibMatrixReorg,
where you test sparse inputs and dense inputs.
It can be inspired by: src/test/java/org/apache/sysds/test/component/matrix/binary/BinaryOpTest.java

min-guk:
Thank you for continuously providing valuable feedback.

As I'm encountering code coverage for the first time, I have some questions. I've summarized the current situation and requests below as I understand them. 

Could you please confirm if this is correct?

- Situation Summary:
  - Java Test Rejection: 
    - Attempted to compare the Java test of the roll function with NumPy, but it was rejected because we cannot install Python packages in the Docker image.
  - Limitations of the Python Test: 
    - Verified correct implementation on dense and sparse matrices through Python tests comparing with NumPy.
    - However, since it uses the statically built Python API, it is not included in code coverage via Codecov.
  - Code Coverage Issue:
    - To improve code coverage, a Java-based test is necessary.

- Request:
  - Add Java Test:
    - Need to write a Java test for the roll function without using NumPy.
    - The test should generate sparse and dense matrices, apply the roll function to each, and verify that the results are identical.
    - The test code should be located at `src/test/java/org/apache/sysds/test/component/matrix/libMatrixReorg/RollTest.java` (or `src/test/java/org/apache/sysds/test/functions/reorg/FullReverseTest.java`).

Baunsgaard:
> Thank you for continuously providing valuable feedback.
> 
> As I'm encountering code coverage for the first time, I have some questions. I've summarized the current situation and requests below as I understand them.
> 
> Could you please confirm if this is correct?
> 
>     * Situation Summary:
>       
>       * Java Test Rejection:
>         
>         * Attempted to compare the Java test of the roll function with NumPy, but it was rejected because we cannot install Python packages in the Docker image.
>       * Limitations of the Python Test:
>         
>         * Verified correct implementation on dense and sparse matrices through Python tests comparing with NumPy.
>         * However, since it uses the statically built Python API, it is not included in code coverage via Codecov.
>       * Code Coverage Issue:
>         
>         * To improve code coverage, a Java-based test is necessary.
> 
>     * Request:
>       
>       * Add Java Test:
>         
>         * Need to write a Java test for the roll function without using NumPy.
>         * The test should generate sparse and dense matrices, apply the roll function to each, and verify that the results are identical.
>         * The test code should be located at `src/test/java/org/apache/sysds/test/component/matrix/libMatrixReorg/RollTest.java` (or `src/test/java/org/apache/sysds/test/functions/reorg/FullRollTest.java`).

Yes this sounds correct.

Therefore:

1. keep the python tests that you made, because they ensure compatibility with numpy,
2. make internal tests (component tests) that, as you say, verify the roll function without using numpy and verify the results correctly.


Thanks.

min-guk:
I have added the JavaTest as mentioned earlier. I would appreciate it if you could review it. Thank you!

min-guk:
I fixed an issue where the roll function was not properly invoked due to an incorrect output file setting and the lack of calling recomputeNonZeros in the Java test.

mboehm7:
LGTM - @min-guk could you please rebase this PR (there are quite a number of merge conflicts over these many commits) and then I would merge it in.

min-guk:
Thank you @mboehm7. I have proceeded with the rebase as requested. 

Once the PR is merged, I will immediately create a PR for the Spark version as well.

Additionally, I have been studying how to implement it using FED, but I’m finding it a bit difficult to grasp. Could you perhaps provide a brief explanation or suggest some reference materials I could look into?

min-guk:
@mboehm7 I apologize for the inconvenience.

It seems I made a mistake due to my inexperience with Git and GitHub. I will make sure to study more to prevent this from happening again in the future.

Also, thank you for recommending the papers. I will take the time to read and study them.

PR #2106, opened 2024-09-11
minor debugging fix

fzoepffel:
The adjusted one hot encoding and maxScore computation led to a few errors. These were fixed. 
- One hot encoding ran into an error if cix and rix are all 0 values (see line 567. Table could not be built with 0 values in cix or rix. The removeEmpty before would sometimes return vectors with exactly one 0.0 value). 
- maxscore computation ran into an error in case the previous lattice was pruned after it was stored (in that case, "S * sc" operation in line 721 was performed on differing sizes of matrices.)


Baunsgaard:
LGTM

mboehm7:
LGTM - thanks for the patch @fzoepffel. While none of the tests caught this, I also ran into the same issue yesterday on larger datasets and specific configurations of new datapoints. Thanks.

PR #2107, opened 2024-09-14
[MINOR] Adaptation of new IncSliceLine Pruning strategy

fzoepffel:
In this PR only a minor bug was fixed that led to errors in some edge cases. Also, the disable tag was added to the newly added max score pruning to allow disabling it for experimentation.


mboehm7:
LGTM - during the merge I added the missing documentation for `pruningStrat` and extended the flags for the new approximate max-score feature pruning.

PR #2112, opened 2024-09-23
[SYSTEMDS-3729] Add roll reorg operations in SP

min-guk:
This PR implements a new roll reorg function in SP.
- In Javatest, the results of CP and SP are compared to validate the SP implementation.
- To avoid code duplication between CP and SP, a function for copying matrices was created, and both functions now use it.

Please feel free to share any feedback or suggestions regarding the implementation direction.


mboehm7:
LGTM - thanks for the patch @min-guk. During the merge I fixed some remaining warnings (unnecessary imports) and the check for opcodes which made the tests fail.

PR #2113, opened 2024-09-23
[SYSTEMDS-3758] Python API Builtin triu, tril, argmin, argmax and casting Scalar <-> Matrix <-> Frame

e-strauss:
This patch adds the builtin functions tril, triu, argmin (rowIndexMin) and argmax (rowIndexMax) to the python API. Additionally, I added functionalities to the python API to 
 - cast matrices, frames and scalars 
 - reshape matrices

I also extended the Scalar Output Type with boolean and integer

### DIFF
--- a/src/main/python/systemds/operator/nodes/frame.py
+++ b/src/main/python/systemds/operator/nodes/frame.py
@@ -44,8 +44,14 @@ if TYPE_CHECKING:
     from systemds.context import SystemDSContext
 
 
-class Frame(OperationNode):
+def to_frame(self):
+    return Frame(self.sds_context, "as.frame", [self])
 
+
+OperationNode.to_frame = to_frame
+
+
+class Frame(OperationNode):
     _pd_dataframe: pd.DataFrame
 
     def __init__(
--- a/src/main/python/systemds/operator/nodes/matrix.py
+++ b/src/main/python/systemds/operator/nodes/matrix.py
@@ -41,6 +41,13 @@ from systemds.utils.helpers import (
 )
 
 
+def to_matrix(self):
+    return Matrix(self.sds_context, "as.matrix", [self])
+
+
+OperationNode.to_matrix = to_matrix
+
+
 class Matrix(OperationNode):
     _np_array: np.array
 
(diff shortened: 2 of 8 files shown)


src/main/python/systemds/operator/nodes/matrix.py
```diff
@@ -634,6 +634,14 @@ def to_string(self, **kwargs: Dict[str, VALID_INPUT_TYPES]) -> "Scalar":
             self.sds_context, "toString", [self], kwargs, output_type=OutputType.STRING
         )
 
+    def to_scalar(self):
+        return Scalar(self.sds_context, "as.scalar", [self])
+
+    def to_frame(self):
+        from systemds.operator import Frame
```
Baunsgaard, line 641: cyclic dependency, i understand it is how i thought about it as well, however, maybe we need to add it afterwards dynamically instead.

See:

https://stackoverflow.com/questions/972/adding-a-method-to-an-existing-object-instance-in-python

This would allow us to add the to_scalar, to_frame and to_matrix on the operation_node directly inside frame, matrix and scalar themselves to overwrite the operation node.

This would also allow us to have documentation on the method, since you add an empty method with docs that would throw an exception on the operation_node, and then on the import of Matrix etc it overwrite the Class method such that new instances use the overwritten method in Matrix.

src/main/python/systemds/operator/nodes/matrix.py
```diff
@@ -810,5 +818,89 @@ def quantile(self, p, weights: "Matrix" = None) -> "OperationNode":
...
+                f"Axis has to be either 0, 1 or None, for column, row or complete {self.operation}"
+            )
+
+    def reshape(self, rows, cols=1):
+        """Gives a new shape to a matrix without changing its data.
+
+        :param rows: number of rows
+        :param cols: number of columns, defaults to 1
+        :return: `Matrix` representing operation"""
+        return Matrix(self.sds_context, "matrix", [self, rows, cols])
```
Baunsgaard, line 903: the rest of the functions looks good.

src/main/python/systemds/operator/nodes/scalar.py
Baunsgaard, line 1: please move the formatting to another PR

src/main/python/systemds/script_building/dag.py
Baunsgaard, line 1: same here, move the formatting to another PR.

src/main/python/systemds/script_building/dag.py
```diff
@@ -43,6 +43,8 @@ class OutputType(Enum):
     NONE = auto()
     SCALAR = auto()
     STRING = auto()
+    INT = auto()
+    BOOLEAN = auto()
```
Baunsgaard, line 47: These are not valid, since both are instances of SCALAR,
where are they used?

Baunsgaard (commented):
Most of the content is good, however i really hope the casting can be modified according to my comment, i am not 100% sure it is possible but it would be great.

Furthermore, the PR contains formatting of frame and scalar, that should be moved to another PR.

src/main/python/systemds/operator/nodes/frame.py
```diff
@@ -112,68 +136,94 @@ def transform_apply(self, spec: "Scalar", meta: "Frame"):
...
+
+    def to_scalar(self):
+        return Scalar(self.sds_context, "as.scalar", [self])
 
     def __str__(self):
         return "FrameNode"
 
-    def nRow(self) -> 'Scalar':
-        return Scalar(self.sds_context, 'nrow', [self])
+    def nRow(self) -> "Scalar":
```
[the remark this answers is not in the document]
e-strauss: All the formatting comes from black, as you recommended

src/main/python/systemds/script_building/dag.py
e-strauss: But isnt STRING also a value type of Scalar?

I added boolean and int there since in the scalar class, the value type of the scalar is determined by this. But you're right it would make more sense to put the value type in a different enum and create a new attribute, so that there is one field for the data type (matrix, frame, scalar) and one for the value types (double, integer, string, boolean)

src/main/python/systemds/operator/nodes/frame.py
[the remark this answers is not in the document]
Baunsgaard: okay, maybe we should add a verification in Github actions that analyze if the code base is all formated according to black, i have seen this in other repositories such as Modyn

The check would be simple to add in a new action using: 
https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#check

src/main/python/systemds/script_building/dag.py
Baunsgaard: you are right, that would make more sense.

I think the reason why it is as it is now, is because someone (probably me) was lazy when making the initial version.

src/main/python/systemds/operator/nodes/matrix.py
e-strauss: I see, currently the cyclic dependency error is avoided since the import happens inside the function, which won't be executed while importing afaik.

src/main/python/systemds/script_building/dag.py
e-strauss: I will continue on this PR once the other Datatype / Valuetype PR is gone through

e-strauss:
@Baunsgaard I adapted the casting, please take a look

Baunsgaard:
LGTM

PR #2115, opened 2024-09-24
[MINOR] added github workflow for black formatter check

e-strauss:
black formatter check for src/main/python/systemds and src/main/python/tests
for now I excluded the auto-generated builtins in operator/algorithm


Baunsgaard:
LGTM as well, if you verify that there are no other python API Pr's then now would be a good time to format everything with black, and then merge this. The order would be first this PR in a single commit, followed by the reformatting commit.

PR #2116, opened 2024-09-24
[SYSTEMDS-3696] updated state management does not lose data throughout updates, updated pruning disabling

fzoepffel:
In this PR only the individual disabling of the newly introduced max score pruning is introduced, which is needed for experimental analysis.

UPDATE:
The current state management was not working correctly for incremental updates that are based on stats and lattice from a run of incSliceLine that already depend on stats and lattice from an even earlier run. This is because the slices and stats that are pruned through incremental pruning methods are not stored for subsequent increments even though we mostly have all the necessary information. I updated the code to track the lost slices and stats and add them to the output lattice and statistics (L and Stats) lists as well. Also, currently, some tests run into errors when enabling the new approx pruning method, which is why for now I set the enabling boolean to FALSE.

### DIFF
--- a/scripts/builtin/incSliceLine.dml
+++ b/scripts/builtin/incSliceLine.dml
@@ -52,7 +52,8 @@
 # prevTK           previous top-k slices (for incremental updates)
 # prevTKC          previous top-k scores (for incremental updates)
 # encodeLat        flag for encoding output lattice for less memory consumption
-# pruningStrat     pruning strategy: 0 all pruning, 1 only score pruning, 2 only size pruning, 3 only max score pruning, 4 no pruning
+# pruningStrat     pruning strategy: 0 all pruning, 1 only score pruning, 2 only size pruning, 
+#                                     3 only max score pruning, 4 only approx pruning, 5 no pruning 
 # ---------------------------------------------------------------------------------------
 #
 # OUTPUT:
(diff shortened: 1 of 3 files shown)


mboehm7:
could you please rebase this PR and resolve the merge conflicts yourself @fzoepffel - the PR actually starts from the wrong code base and overwrites the new disableIncApproxPruning flag (introduced 10 days ago) which makes all the tests fail.

fzoepffel:
> could you please rebase this PR and resolve the merge conflicts yourself @fzoepffel - the PR actually starts from the wrong code base and overwrites the new disableIncApproxPruning flag (introduced 10 days ago) which makes all the tests fail.

Thank you for the reminder; it slipped my mind. I just rebased the PR, and to save some lines of code, I adjusted the disabling a bit to **enable** rather than **disable** the strategies, as all strategies are only enabled in two cases (1. all enabled, 2. only the respective strategy enabled), while they'd have to be disabled in more cases.

fzoepffel:
I just added another bigger change to this PR. As you, @mboehm7, commented in my thesis, the current state management is not working correctly for incremental updates that are based on stats and lattice from a run of incSliceLine that already depends on stats and lattice from an even earlier run. This is because the slices and stats that are pruned through incremental pruning methods are not stored for subsequent increments even though we mostly have all the necessary information. I updated the code to track the lost slices and stats and add them to the output lattice and statistics (L and Stats) lists as well. Also, currently, some tests run into errors when enabling the new approx pruning method, which is why for now I set the enabling boolean to FALSE.

mboehm7:
LGTM - thanks for the patch @fzoepffel. The merged branch still created conflicts, so I assembled the patch from scratch. The improved state handling is indeed very important. Thanks.

PR #2117, opened 2024-09-24
[MINOR] add correct python dependencies

christinadionysio:
This PR removes the dependency to keras and adds additional dependencies needed in scuro.


Baunsgaard:
LGTM, and merged.

PR #2119, opened 2024-09-24
[MINOR] Python API: removal of enum output_type

e-strauss:
this PR removes the redundant information about the datatype from the python DAGNodes, the datatype can be instead inferred from the instance class and the value type directly from the Java object

### DIFF
--- a/src/main/python/systemds/context/systemds_context.py
+++ b/src/main/python/systemds/context/systemds_context.py
@@ -39,9 +39,9 @@ import pandas as pd
 from py4j.java_gateway import GatewayParameters, JavaGateway, Py4JNetworkError
 from systemds.operator import (Frame, List, Matrix, OperationNode, Scalar,
                                Source, Combine,  MultiReturn)
-from systemds.script_building import DMLScript, OutputType
+from systemds.script_building import DMLScript
 from systemds.utils.consts import VALID_INPUT_TYPES
-from systemds.utils.helpers import get_module_dir
+from systemds.utils.helpers import get_module_dir, valuetype_from_str
 
 
 class SystemDSContext(object):
@@ -425,41 +425,6 @@ class SystemDSContext(object):
         named_input_nodes = {'rows': shape[0], 'cols': shape[1]}
         return Matrix(self, 'matrix', unnamed_input_nodes, named_input_nodes)
 
-
-    def fft(self, real_input: 'Matrix') -> 'MultiReturn':
-        """
-        Performs the Fast Fourier Transform (FFT) on the matrix.
-        :param real_input: The real part of the input matrix.
-        :return: A MultiReturn object representing the real and imaginary parts of the FFT output.
-        """
-
-        real_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-        imag_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-
-        fft_node = MultiReturn(self, 'fft', [real_output, imag_output], [real_input])
-
-        return fft_node
-
-
-    def ifft(self, real_input: 'Matrix', imag_input: 'Matrix' = None) -> 'MultiReturn':
-        """
-        Performs the Inverse Fast Fourier Transform (IFFT) on a complex matrix.
-        
-        :param real_input: The real part of the input matrix.
-        :param imag_input: The imaginary part of the input matrix (optional).
-        :return: A MultiReturn object representing the real and imaginary parts of the IFFT output.
-        """
-
-        real_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-        imag_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-
-        if imag_input is not None:
-            ifft_node = MultiReturn(self, 'ifft', [real_output, imag_output], [real_input, imag_input])
-        else:
-            ifft_node = MultiReturn(self, 'ifft', [real_output, imag_output], [real_input])
-
-        return ifft_node
-
     def seq(self, start: Union[float, int], stop: Union[float, int] = None,
             step: Union[float, int] = 1) -> 'Matrix':
         """Create a single column vector with values from `start` to `stop` and an increment of `step`.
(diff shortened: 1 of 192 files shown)


src/main/python/systemds/operator/operation_node.py
```diff
@@ -45,27 +45,25 @@ class OperationNode(DAGNode):
...
+    _datatype_is_unknown: bool
 
     def __init__(self, sds_context: 'SystemDSContext', operation: str,
                  unnamed_input_nodes: Union[str,
                                             Iterable[VALID_INPUT_TYPES]] = None,
                  named_input_nodes: Dict[str, VALID_INPUT_TYPES] = None,
-                 output_type: OutputType = OutputType.MATRIX,
                  is_python_local_data: bool = False,
-                 brackets: bool = False):
+                 brackets: bool = False,
```
e-strauss, line 55: I added two new boolean attributes to operation_node: datatype_is_none and datatype_is_unknown to handle operations with no return value (write, print,..) or unkown datatype (list access), which where previously indicated by the OutputType.NONE

src/main/python/systemds/operator/nodes/list.py
```diff
@@ -75,11 +74,6 @@ def pass_python_data_to_prepared_script(self, sds, var_name: str, prepared_scrip
             prepared_script.setMatrix(var_name, numpy_to_matrix_block(
                 sds, self._np_array), True)  # True for reuse
 
-    def code_line(self, var_name: str, unnamed_input_vars: Sequence[str],
-                  named_input_vars: Dict[str, str]) -> str:
-        code_line = super().code_line(var_name, unnamed_input_vars, named_input_vars)
-        return code_line
-
```
e-strauss, line 82: removed since we call anyways the super class method, so no need to overwrite imo

src/main/python/systemds/operator/nodes/matrix.py
```diff
@@ -810,5 +809,36 @@ def quantile(self, p, weights: "Matrix" = None) -> "OperationNode":
...
+        real_output = Matrix(self.sds_context, '')
+        imag_output = Matrix(self.sds_context, '')
+
+        if imag_input is None:
+            ifft_node = MultiReturn(self.sds_context, 'ifft', [real_output, imag_output], [self])
+        else:
+            ifft_node = MultiReturn(self.sds_context, 'ifft', [real_output, imag_output], [self, imag_input])
+
+        return ifft_node
+
```
e-strauss, line 842: moved from systemds context because the there the outputs were set as OperationNodes with OutputType Matrix to avoid a cyclic dependancy.

src/main/python/systemds/operator/nodes/multi_return.py
```diff
@@ -68,18 +67,18 @@ def _parse_output_result_variables(self, result_variables):
         result_var = []
         jvmV = self.sds_context.java_gateway.jvm
         for idx, v in enumerate(self._script.out_var_name):
-            out_type = self._outputs[idx].output_type
-            if out_type == OutputType.MATRIX:
+            output = self._outputs[idx]
```
e-strauss, line 70: to avoid cyclic dependencies I used instead of isinstance the to_string value

src/main/python/systemds/operator/nodes/source.py
```diff
@@ -54,10 +53,7 @@ def get_func(self, sds_context, source_name) -> MethodType:
         output_object = self.parse_outputs()
 
         definition = f'def {self._name}(self{argument_string}):'
-        if self._outputs is None:
-            output = f'out = {output_object}(self.sds_context, {operation}, named_input_nodes=named_arguments, output_type=OutputType.NONE)'
-        else:
-            output = f'out = {output_object}(self.sds_context, {operation}, named_input_nodes=named_arguments)'
```
e-strauss, line 60: the ifelse is not needed anymore since we can infer the output type directly from the output_object

src/main/python/systemds/operator/nodes/source.py
```diff
@@ -54,10 +53,7 @@ def get_func(self, sds_context, source_name) -> MethodType:
...
+        output = f'out = {output_object}(self.sds_context, {operation}, named_input_nodes=named_arguments)'
```
e-strauss, line 56: the ifelse is not needed anymore since we can infer the output type directly from the output_object

src/main/python/systemds/operator/nodes/source.py
```diff
@@ -135,7 +131,7 @@ class Source(OperationNode):
 
     def __init__(self, sds_context, path: str, name: str):
         super().__init__(sds_context,
-                         f'"{path}"', output_type=OutputType.IMPORT)
+                         f'"{path}"')
```
e-strauss, line 134: I have not created attribute for the IMPORT type since I have not seen code where the flag is relevant

src/main/python/systemds/operator/operation_node.py
```diff
@@ -174,11 +172,15 @@ def write(self, destination: str, format: str = "binary", **kwargs: Dict[str, VA
...
         """ Prints the given Operation Node.
         There is no return on calling.
         To get the returned string look at the stdout of SystemDSContext.
         """
-        return OperationNode(self.sds_context, 'print', [self], kwargs, output_type=OutputType.NONE)
+        return OperationNode(self.sds_context, 'print', [self], kwargs)
+
+    def data_type(self):
+        return "operation_node"
+
```
e-strauss, line 186: can be removed, artefact from an older approach

src/main/python/systemds/context/systemds_context.py
```diff
@@ -425,41 +425,6 @@ def full(self, shape: Tuple[int, int], value: Union[float, int]) -> 'Matrix':
...
-        real_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-        imag_output = OperationNode(self, '', output_type=OutputType.MATRIX, is_python_local_data=False)
-
-        if imag_input is not None:
-            ifft_node = MultiReturn(self, 'ifft', [real_output, imag_output], [real_input, imag_input])
-        else:
-            ifft_node = MultiReturn(self, 'ifft', [real_output, imag_output], [real_input])
-
-        return ifft_node
-
```
e-strauss, line 462: moved to matrix.py, cuz of the indirect creation of the Matrix-OperationNode (which was done to avoid cyclic dependency)

src/main/python/systemds/operator/operation_node.py
Baunsgaard: if we can remove it, lets.

PR #2122, opened 2024-09-25
[SYSTEMDS-3701] Add tests for Scuro

christinadionysio:
This patch adds a data generator that creates a simple multimodal dataset and python tests for scuro.


mboehm7:
thanks for the patch @christinadionysio. Could you please git rebase this PR as it looks to include commits from prior PRs also. Thanks.

PR #2123, opened 2024-09-25
[MINOR] investigating not terminating python test

e-strauss:


e-strauss:
adding the --verbose option to the unittest-parallel python test fixes the non-terminating test

PR #2131, opened 2024-10-21
[SYSTEMDS-3729] Fix Federated Response in Roll Function

min-guk:
I previously wrote the federated `roll function` to return the result matrix block through a `federated response` for debugging purposes. Although this does not affect the computation results, it unnecessarily degrades performance, so I will remove it. 

Moving forward, I will carefully review the code before committing. I apologize for the inconvenience caused.


mboehm7:
Thanks for the patch @min-guk and sorry for missing this debugging relict during the merge of the previous PR.

min-guk:
@mboehm7 Not at all, I should have been more mindful. 

I really appreciate your always kind and insightful review!

PR #2132, opened 2024-10-21
[SYSTEMDS-3172] Sparse CSC Block

ReneEnjilian:
This PR is just a supplement to a previous task and implements the Sparse CSC Block. MCSC and CSC will most likely only be used for some experiments (@Baunsgaard ) and there will not be a deeper integration for now. Other systems such as PyTorch support both row and column oriented matrix compression formats and now SystemDS does too. I implemented both the row and the column APIs. Further, testing has been added.


mboehm7:
LGTM - thanks for the additional sparse block @ReneEnjilian. During the merge I fixed remaining warnings, and removed the wildcard imports. Furthermore, some operations are really slow (the sparse component tests now run 300+ seconds) - accordingly, I will do a few runtime improvements in order to reduce this test runtime.

PR #2133, opened 2024-10-29
[WIP][SYSTEMDS-3777] Builtin function for ADASYN

ReneEnjilian:
This PR adds a new builtin function for ADASYN (Adaptive Synthetic Sampling) for generating synthetic data in case of class imbalances in ML datasets (binary classification). The method itself is implemented but I still need to add the test class in java. I manually tested the method on a real dataset called [Pima Indians Diabetes](https://www.kaggle.com/datasets/uciml/pima-indians-diabetes-database), which the authors also used in the original paper. The generated synthetic data looked very reasonable when compared to the original data. I will make a more detailed description here once I added the test cases and conducted more experiments on the other datasets mentioned in the paper.


mboehm7:
OK, thanks for getting started on this @ReneEnjilian. I'll now take it over, add the missing tests, and then debug it accordingly.

PR #2134, opened 2024-11-07
Boolean Simplification Rewrites

aarnaty:
Boolean Rewrites:
# boolean variable a 
a & !a --> FALSE
a | !a --> TRUE


mboehm7:
LGTM - thanks for the patch @aarnaty. During the merge I fixed the missing license headers, missing comparison of results (the results were never written out), formatting (tabs vs spaces in Java code), and generalized the rewrites such that the negation can occur either on left or right hand sides.

mboehm7:
and a minor thing for the future - please avoid reformatting large existing files.

PR #2137, opened 2024-11-13
[SYSTEMDS-3789] Fix the CovarianceFEDInstruction for weighted covariance

gaturchenko:
Previous implementation of the CovarianceFEDInstruction worked correctly only for regular federated covariance. The following issues were discovered when tested for the weighted covariance case:
1. In the processCov method the value to the wtBlock variable was assigned incorrectly, as input2 which is the second input matrix was used instead of the weights matrix stored as input3. Moreover, the slice method should be called for wtBlock as well when passed as an argument to the executeFederatedOperation method.
2. The processMean method was implemented incorrectly for the case of weighted covariance. In its previous implementation it modified the original instruction string so as to execute the uamean instruction. However, in order to calculate weighted covariance weighted mean is required. Therefore, the method was substantially modified to make sure the proper mean values are calculated.
3. The aggregation of partial results from the federated workers was not implemented properly. To begin with, in the getResponses method mean values were collected only from the second and third federated worker. Then, in the aggCov method the implementation did not support the aggregation of weighted means, since in this case the sum of weights should be employed instead of vector lengths per worker. Both methods were modified to ensure the correct aggregation of covariance.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/fed/CovarianceFEDInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/fed/CovarianceFEDInstruction.java
@@ -380,13 +380,11 @@ public class CovarianceFEDInstruction extends BinaryFEDInstruction {
 			// divide sum(X*W) by sum(W)
 			String[] partsSum2 = sumInstr2.split("°");
 			String divInstrOutput = incrementVar(sumInstr2Output, 1);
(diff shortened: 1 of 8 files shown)


mboehm7:
LGTM - thanks for catching and fixing this issue @gaturchenko. During the merge I fixed only minor formatting issues (tabs over spaces, and too verbose parameter lists)

PR #2139, opened 2024-11-15
Restoration of Pre-FED Privacy Test Code

min-guk:
The test code for FED Privacy, which was previously removed in commit d9822ce9c7eb26412c00cbcaf0decc11e02ae533, has been restored. Moving forward, we plan to implement FED Privacy features incrementally in a bottom-up approach and aim to pass each of these test cases one by one as the functionality is developed.

- Currently, all FED Privacy implementations have been removed, and as a result, the restored test code cannot be compiled or executed at this stage. 
- Instead of using the `@Ignore` annotation for the tests, all restored test code has been commented out to avoid build or runtime issues.

This pull request focuses solely on restoring the test code for future development and testing purposes.


mboehm7:
LGTM - thanks for the patch @min-guk. During the merge, I picked the fedplanner tests, adapted their implementation so they succeed, enabled them in the github actions.

min-guk:
Thank you. I will take a look at the revisions.

PR #2143, opened 2024-11-20
[SYSTEMDS-3701] Add test suite for Scuro

christinadionysio:
This PR adds an initial test suite for the Scuro library. It includes a data generator to create modalities (text, audio, video for now). Additionally, two tests were implemented to test the data loaders and the first version of Data Representation Search.


mboehm7:
LGTM - thanks for adding these tests @christinadionysio.

PR #2145, opened 2024-11-21
[SYSTEMDS-3782] Bag-of-words Encoder for SP

e-strauss:
This patch extends the bag-of-words operation to support transform apply and the distributed spark backend.

To support the bow encoder for sparse outputs I had to adapt the OutputMatrixPreProcessing step in the transform apply framework, since we need to know the number of non-zero values per row when allocating the output sparse matrix. This is easy for the other encoders, since they resulted in just one non-zero value per encoder.
In transform encode we gained this information about the number of non-zeroes in the Build phase. Since we don't have a build phase in transform apply, we don't dont have this information. To solve this issue, I added a simplified build-like phase in the  OutputMatrixPreProcessing, which computes the #nnz for each bow encoder, if the output is sparse and the transformation involves a bow encoder. This phase is parallelized across the bow encoder columns.
The decision if the output is sparse is based on an estimation of the total #nnz of bow encoder, which is calculated based on a sample of the input. 

In one of the spark test cases, where I cbind the input frame with itself before executing the transform encode, I encountered an alignment issue while the map-based frame append, which I could bypass for now by adding a breaker (while(FALSE). I am investigating the bug in a separate PR.

### DIFF
--- a/src/test/java/org/apache/sysds/test/functions/transform/ColumnEncoderMixedFunctionalityTests.java
+++ b/src/test/java/org/apache/sysds/test/functions/transform/ColumnEncoderMixedFunctionalityTests.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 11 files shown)


src/test/scripts/functions/transform/TransformFrameEncodeApplyBagOfWords.dml
```diff
@@ -0,0 +1,84 @@
...
+}
+
+[Data_enc, Meta] = transformencode(target=Data, spec=jspec);
+while(FALSE){}
+
+i = 0
+total = 0
+j = 0
+# set to 20 for benchmarking
+while(i < 30){
```
e-strauss, line 65: can be set to "i < 1"

src/test/java/org/apache/sysds/test/functions/transform/ColumnEncoderMixedFunctionalityTests.java
e-strauss, line 1: test cases to improve code coverage

mboehm7:
LGTM - thanks for the patch @e-strauss. During the merge I only fixed minor formatting things (use of this, empty lines) as well as removed assertions and unnecessary imports.

PR #2146, opened 2024-11-26
[SYSTEMDS-3701] Add additional text representations to Scuro

christinadionysio:
This PR adds additional text representations like Bag-of-Words, Word2Vec, TF-IDF, and Glove to Scuro. They will be used in DRSearch to find the best unimodal representations for text data.


mboehm7:
LGTM - thanks for the patch @christinadionysio

PR #2148, opened 2024-12-02
[SYSTEMDS-3800] Add tests to increase code coverage of FED instructions

gaturchenko:
The tests for the following federated instructions have been added:

Central moment (weighted case)
Ternary aggregation (cumulative product & sumproduct)
Ternary instruction (account for non-literal scalar inputs)
The `TernaryFEDInstruction` was also slightly modified, since in its previous implementation the `processMatrixScalarInput` method returned an "unknown variable" error when processing an `ifelse` instruction with one matrix input and non-literal scalar inputs.


mboehm7:
LGTM - thanks for the additional tests and improved code coverage @gaturchenko. A minor thing for the future: please do not include the commits from previous PRs which creates lots of merge conflicts during the rebase/squash process. Instead I just picked the actual 3 new commits. Thanks.

PR #2150, opened 2024-12-04
[MINOR] Exclude Scuro if dependencies not installed

christinadionysio:
This patch checks if the dependent packages for Scuro including their correct versions are installed and otherwise skips importing Scuro.

### DIFF
--- a/src/main/python/systemds/__init__.py
+++ b/src/main/python/systemds/__init__.py
@@ -18,10 +18,44 @@
 # under the License.
 #
 # -------------------------------------------------------------
-
+from importlib.metadata import version, PackageNotFoundError
 from systemds import context
 from systemds import operator
 from systemds import examples
-from systemds import scuro
 
-__all__ = ["context", "operator", "examples", "scuro"]
+__all__ = ["context", "operator", "examples"]
+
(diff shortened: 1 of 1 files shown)


Baunsgaard:
I like it, to make it even better can we make the requirements open ended, such that if someone has newer versions installed it is also valid?

christinadionysio:
> I like it, to make it even better can we make the requirements open ended, such that if someone has newer versions installed it is also valid?

Sure, that's why I check if the installed version is `>=` the specified version.

src/main/python/systemds/__init__.py
```diff
@@ -18,10 +18,46 @@
...
+    ("h5py", "3.11.0"),
+    ("transformers", "4.46.3"),
+    ("nltk", "3.9.1"),
+    ("gensim", "4.3.3"),
+]
+
+
+def check_package_version(package_name, required_version):
+    try:
+        return version(package_name) >= required_version
```
Baunsgaard, line 43: i see, missed it was greater than, 

LGTM

Baunsgaard:
Thanks for the commit, It is now merged.

PR #2153, opened 2024-12-13
implement new "-explain" command to print the generated code to the console

Frxms:


mboehm7:
Thanks for the contribution @Frxms. Please do the minor cleanups we talked about yesterday, and then we can merge it in.

PR #2154, opened 2024-12-17
[SYSTEMDS-3548] Optimize IO path Python interface for SystemDS

Nakroma:
Draft for the student project `SYSTEMDS-3548`.

Current contributions:
- Fixes some minor bugs related to the performance tests
- Parallelizes pandas_to_frame_block column processing (see image below for speed up, tested on my machine)

load_pandas


christinadionysio:
LGTM! Thank you for your contribution @Nakroma!

Nakroma:
@Baunsgaard I used the IO benchmark scripts for the figure provided above:

https://github.com/apache/systemds/blob/main/scripts/perftest/runAllIO.sh
https://github.com/apache/systemds/blob/main/scripts/perftest/python/io/load_pandas.py

Baunsgaard:
> @Baunsgaard I used the IO benchmark scripts for the figure provided above:
> 
> https://github.com/apache/systemds/blob/main/scripts/perftest/runAllIO.sh https://github.com/apache/systemds/blob/main/scripts/perftest/python/io/load_pandas.py

Great! 

Then you can get better numbers:

https://github.com/apache/systemds/blob/d3fcfb13ccb69f4cfc4338cd0323b44dc1be4c55/scripts/perftest/python/io/load_pandas.py#L37

Modify the script to start the context not in the 'run' part, instead move it to the 'setup' part, and remember to shut down the system with ctx.close() after you are done measuring.

Baunsgaard:
> @Baunsgaard Okey yeah that makes sense - pushed a commit for that 👍 I didnt move it to the setup but rather inside the global context, so .close() time is not included in the timing and also to support the `args.number` parameter.

what are the times then?

Nakroma:
> what are the times then?

seems to be about a difference of 1-2s

load_pandas

Baunsgaard:
> > what are the times then?
> 
> seems to be about a difference of 1-2s, at least on my local machine
> 
> load_pandas

60% speedup on int32 and 100% on int64 is great!
However, it does seem to me like there is something else taking time from your results. I would expect speedup closer to the number of cores in your system.

Baunsgaard:
> So there is some more constant time, the building of the frameblock a few lines before the `.convert` calls for example is around 400ms.

We could put the allocation into the parallel transfer call for each column?

> I looked at the profiling a bit more and it seems like most time is spent on socket communication between Java and Python. My assumption would be that this adds quite a bit of overhead and doesn't parallelize well.

I was afraid this was the case, Maybe there is a way to set the thread pool to use on the java side to make it better at receiving requests in parallel. However, that is probably just a dream.

Baunsgaard:
LGTM, I have now merged it.

While merging I played around with parallelizing the Python API and found out that it spawns a thread per connection. There is indeed an overhead in this connection, but it is not the main problem.

To make the transfer more efficient we could: 

1. Reduce the number of calls by fusing many operations into single calls to java.
2. Reduce the current serialization bottleneck by slicing up the array into many smaller byte arrays when sending over.

I see around 20% utilization of my CPU when transferring 10k by 10k integer matrices, so there is room for improvement.

PR #2155, opened 2024-12-17
[SYSTEMDS-3701] Rework Modality Data Structure

christinadionysio:
This patch disentangles the loading mechanisms of raw data from the representations and contains rewrites of the Modality data structure. It adds a base loader that can be extended for every modality type and needs to be adapted according to the file structure of the raw data. Additionally, the modality  data structure was reworked to enable stacking of operations. This includes slight API changes and the introduction of `.apply_representation()` and `.combine()` methods. Individual modality types are replaced by a universal `UnimodalModality` that takes a `ModalityType` to indicate the original modality of the data.


mboehm7:
@christinadionysio thanks for the patch - could you please have a look into the failing python test? Recently, ubuntu-latest switched to Ubuntu 24 which also updated a number of dependencies and hence we moved to Python 3.8 (which might have affected some of these issues).

mboehm7:
LGTM - thanks for the modifications @christinadionysio.

PR #2167, opened 2024-12-29
[MINOR] reformat releaseIntern in cacheableData

Baunsgaard:


Baunsgaard:
Perhaps this change is not needed. I remember a problem in some edge case that prompted me to make this change, but I do not remember what. Therefore, this change is separated from BWARE and perhaps not to be merged.

PR #2174, opened 2025-01-04
Feat/sqrt in dml

engineerswiftie97:


mboehm7:
Thanks for the contribution @engineerswiftie97 and @trp-ex - this is a good start. As the next step, I would recommend to do the following improvements:

- Restore the deleted files, unrelated to the new builtin function
- Remove the hello.dml, .Rapp.history files as well as the testing area in the dml builtin function
- Add the missing license headers to all files
- Add the missing input/output documentation to the dml builtin function (we automatically generate the Python API call and respective documentation from it)
- Differentiate the names of the two registered builtin functions (if the names are equal the parser cannot differentiate them)
- Add the dispatching logic of two different strategies (dml-builtin vs java-builtin) in the dml-builtin script.
- Add tests for both functions with results comparison and check that the right function was called

PR #2176, opened 2025-01-08
made changes to pushDownSumOnBinaryAdditive and removeUnnecessaryReor…

aarnaty:
…gOperation. Added row and col test cases for the former.


mboehm7:
LGTM - thanks for the improved rewrites @aarnaty. Overall this looked pretty good already. During the merge I fixed the old negative tests and reduced the epsilon for test comparisons.

PR #2180, opened 2025-01-13
Printing matrix without toString()

ReneEnjilian:
This patch enables the DML to print matrices and other datatypes (frame, list)  without using the toString() method.
Previously, printing a matrix A required: `print(toString(A))`.
Now, printing can be done via: `print(A)`.
The same applies to frames and lists. However, there are some remaining issues when datastructures are used with indexing. For example, while `A[1,]` is also a matrix (first row of A), `print(A[1,])` will not print the correct output (first row) but rather only the first element of the first row. Internally, the compiler misinterprets the matrix as a scalar and does not propagate the correct size information. The propagated size information on the rewrite-level are (-1 x -1), meaning unknown. A work-around would be either to still use the `toString` method or, as shown in the test cases, to materialize the slicing before printing: `B=A[1,]` , `print(B)`.
Similar problems exist with lists and frames but can be addressed by materializing the slicing/indexing outside the print as shown in the previous example. A future task should address this issue. 
I promoted the rewrite from RewriteAlgebraicSimplificationStatic to having a separate class in the rewrite directory as discussed. Testing has been added.


ReneEnjilian:
All checks fail. The error message for all of the 44 failing checks is something like: 
 Error: fatal: unable to access 'https://github.com/apache/systemds/': The requested URL returned error: 502
Can I assume that this issue is unrelated to my pushed code ?

mboehm7:
LGTM - Thanks for the patch @ReneEnjilian. During the merge, I moved the rewrite to static rewrites (in program rewriter), fixed the failing test, and removed commented code.

PR #2181, opened 2025-01-15
[SYSTEMDS-3815] Fused table sequence 

Baunsgaard:
This commit contains a new fused operator for:

table(seq(1, nrow(A)), A, w)

This operation is useful for efficient word embedding of a token matrix A.


mboehm7:
Is that an extension (for weighted variants?) of the existing rexpand instruction?

Baunsgaard:
> Is that an extension (for weighted variants?) of the existing rexpand instruction?

assuming I understand the current implementation.
I made it work for both. It is constructed to avoid allocating the seq(1, nrow(A)), since this was the main bottleneck.

Baunsgaard:
just to answer the question, yes it is an extension of rexpand.

mboehm7:
What irritates me a bit is that the rexpand kernels are implemented in `LibMatrixReorg`, which you not even touch. If this code is for the weighted versions of rexpand, please move it there which would also prevent creating too many Lib* classes which were meant for types of operations with many kernels, not individual operations.

Baunsgaard:
> What irritates me a bit is that the rexpand kernels are implemented in `LibMatrixReorg`, which you not even touch. If this code is for the weighted versions of rexpand, please move it there which would also prevent creating too many Lib* classes which were meant for types of operations with many kernels, not individual operations.

okay will do, I just had a tendency lately to make Lib files since it made it 'cleaner' in my mind for containing parallel and non parallel calls.

Baunsgaard:
Performance improvements:

```txt
Before:
Total elapsed time:             16.876 sec.
  1  ctableexpand   12.974    100
  2  seq             2.885    100]

After Uncompressed:
Total elapsed time:             10.496 sec.
  1  ctableexpand    9.236    100

After Compressed:
Total elapsed time:             4.973 sec.
  1  ctableexpand    4.089    100
```

The change unfortunately makes the current fed_tableexpand instruction invalid and therefore have to be reimplemented.

PR #2186, opened 2025-01-23
[SYSTEMDS-3556] Counter based random number generator

ichbinstudent:
# Summary
Adding counter based RNG which improves random quality and speed

# Details
The PRNG uses Philox4x64_10 to generate batches of random double values with a 64 bit randomness. The algorithm was tested for correctness by comparing its output for certain counters and keys with an existing implementation in c.
This reference implementation called [openRAND](https://github.com/msu-sparta/OpenRAND) was tested by the authors using various statistical methods described [here](https://www.sciencedirect.com/science/article/pii/S2352711024001444).

## Advantages
- Quality: Instead of 32 bits in java.util.Random, Philox4_64 produces 64 bits of randomness. While java.util.Random has a period of only (2^48), while the period of Philox4_64 is 2^256 - 1. 
- Speed: While the Java version of Philox4x64_10 is only about half as fast as java.util.Random, there is a Cuda kernel version available producing the exact same sequence of random numbers. This means that the Cuda and Java versions can be used interchangeably. If a system has support for Cuda, the kernel ist used, if not, the Java version can be used as a fallback. The kernel version is around 200 times faster than java.util.Random, and even faster, if the results are not copied to the CPU but kept in the GPU's memory.
- Parallelisation: When using state based PRNGs, it is impossible to generate the same random matrix when changing the block size. With counter based PRNGs it is possible to change the block size but still compute the same random matrix by using the global index (row * row_size + col) as the counter.


src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixDatagen.java
```diff
@@ -36,12 +36,7 @@
 import org.apache.sysds.runtime.controlprogram.parfor.util.IDSequence;
 import org.apache.sysds.runtime.data.DenseBlock;
 import org.apache.sysds.runtime.data.SparseBlock;
-import org.apache.sysds.runtime.util.CommonThreadPool;
-import org.apache.sysds.runtime.util.NormalPRNGenerator;
-import org.apache.sysds.runtime.util.PRNGenerator;
-import org.apache.sysds.runtime.util.PoissonPRNGenerator;
-import org.apache.sysds.runtime.util.UniformPRNGenerator;
-import org.apache.sysds.runtime.util.UtilFunctions;
+import org.apache.sysds.runtime.util.*;
```
Baunsgaard, line 39: please do not use wildcard imports. it is probably a setting in your IDE.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixDatagen.java
```diff
@@ -590,13 +602,26 @@ private static void genDense(MatrixBlock out, int clen, int blockrows, int block
...
-		double min, double range, PRNGenerator valuePRNG)
+		double min, double range, IPRNGenerator valuePRNG, long[] ctr)
 	{
-		for(int i = rowoffset; i < rowoffset+blockrows; i++) {
-			double[] cvals = c.values(i);
-			int cix = c.pos(i, coloffset);
-			for(int j = 0; j < blockcols; j++)
-				cvals[cix+j] = min + (range * valuePRNG.nextDouble());
+		if (valuePRNG instanceof PRNGenerator) {
+			for(int i = rowoffset; i < rowoffset+blockrows; i++) {
```
Baunsgaard, line 608: please split it into two different functions for each branch of this if statement

src/main/java/org/apache/sysds/runtime/matrix/data/RandomMatrixGenerator.java
```diff
@@ -20,25 +20,22 @@
 package org.apache.sysds.runtime.matrix.data;
 
 import org.apache.sysds.runtime.DMLRuntimeException;
-import org.apache.sysds.runtime.util.NormalPRNGenerator;
-import org.apache.sysds.runtime.util.PRNGenerator;
-import org.apache.sysds.runtime.util.PoissonPRNGenerator;
-import org.apache.sysds.runtime.util.UniformPRNGenerator;
+import org.apache.sysds.runtime.util.*;
```
Baunsgaard, line 23: no wildcard

src/test/java/org/apache/sysds/test/component/matrix/LibMatrixDatagenTest.java
```diff
@@ -0,0 +1,94 @@
+package org.apache.sysds.test.component.matrix;
```
Baunsgaard, line 1: license

src/test/java/org/apache/sysds/test/component/matrix/LibMatrixDatagenTest.java
```diff
@@ -0,0 +1,94 @@
...
+		LibMatrixDatagen.generateRandomMatrix(mb, rgen, null, 123123123123L);
+		double mean = mb.mean();
+		double[] bv = mb.getDenseBlockValues();
+		double variance = Arrays.stream(bv).map(x -> Math.pow(x - mean, 2)).sum() / bv.length;
+		assertEquals("Mean should be 0", 0, mean, 0.01);
+		assertEquals("Variance should be 1", 1, variance, 0.001);
+	}
+
+	@Test
+	@Ignore
```
Baunsgaard, line 45: specific reasons to ignore your added tests?

Baunsgaard (commented):
Many good things here, please address the minor comments, and describe the performance differences in the PR.

Next steps, is to integrate it into the compiler as options, i would suggest to do it via arguments to random "CB_uniform" could be the argument for instance.

Baunsgaard:
If you resolve some of the comments, please feel free to mark them as such.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixDatagen.java
ichbinstudent: I changed it to a solution using iterators. This removes the duplicate code. What do you think?

src/test/java/org/apache/sysds/test/component/matrix/LibMatrixDatagenTest.java
ichbinstudent: They were just for testing the performance. Shouldn't have been included in the PR.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixDatagen.java
```diff
@@ -590,13 +615,22 @@ private static void genDense(MatrixBlock out, int clen, int blockrows, int block
...
 	private static void genFullyDense(DenseBlock c, int blockrows, int blockcols, int rowoffset, int coloffset, 
-		double min, double range, PRNGenerator valuePRNG)
+		double min, double range, IPRNGenerator valuePRNG, long[] ctr)
 	{
-		for(int i = rowoffset; i < rowoffset+blockrows; i++) {
+		Iterator<Double> rngStream;
+		if (valuePRNG instanceof PRNGenerator) {
+			rngStream = Stream.generate(() -> min + (range * ((PRNGenerator) valuePRNG).nextDouble())).iterator();
+		} else if (valuePRNG instanceof CounterBasedPRNGenerator) {
+			rngStream = Arrays.stream(((CounterBasedPRNGenerator)valuePRNG).getDoubles(ctr, blockrows * blockcols)).map(i -> min + (range * i)).iterator();
```
Baunsgaard, line 624: yup the stream looks nice, however, since it modifies old code, we need to be careful that the new code returns the same values as the previous.

scripts/staging/cuda-counter-based-prng/readme.md
```diff
@@ -0,0 +1,410 @@
...
+)
+{
+    ... cuda kernel code ...
+}
+
+```
+To use this ptx file in the SystemDS project, you can use the following code:
+
+```java
+import jcuda.*;
```
Baunsgaard, line 83: lets remove the large blocks of code in the readme, instead refer to the individual files.

scripts/staging/cuda-counter-based-prng/PhiloxJNvrtcExample.java
```diff
@@ -0,0 +1,90 @@
+import jcuda.*;
```
Baunsgaard, line 1: missing licenses.

scripts/staging/cuda-counter-based-prng/philox_kernel.ptx
```diff
@@ -0,0 +1,772 @@
+//
```
Baunsgaard, line 1: if this is a compiled file, based on any of your files, then lets remove it, or git-ignore it.

chris-1187: This is based on the kernel.cu file, but when removed, the examples wont work without additional compilation. Still remove?

scripts/staging/cuda-counter-based-prng/readme.md
chris-1187: resolved.

scripts/staging/cuda-counter-based-prng/PhiloxJNvrtcExample.java
chris-1187: resolved.

scripts/staging/cuda-counter-based-prng/philox_kernel.ptx
Baunsgaard: assuming the readme contains how to compile/make this file, then yes gitignore it.

ichbinstudent:
Ok, deleted the compiled kernel file.

Baunsgaard:
Hi @ichbinstudent  and @chris-1187 

I will take this PR over and get it merged. See #2260

Thanks for the contribution.

PR #2187, opened 2025-01-23
Opcode enum

aarnaty:
Put all the opcodes in an enum and replaced them in other files.


mboehm7:
Thanks for getting started on this opcode enum @aarnaty , but the submitted code only contains the already merged rewrite from your last PR.

PR #2194, opened 2025-01-29
[SYSTEMDS-3823] Compression test case for bultin kmeans

e-strauss:
This fix adds a new test case for the builtin kmeans with compression enabled. It also adds a missing decompression handling for indexed MatrixBlock copy, which previously broke the compressed kmeans test case.


Baunsgaard (commented):
LGTM, just two lines to change.

Baunsgaard:
Thanks!

PR #2197, opened 2025-01-30
[SYSTEMDS-3704] Resource-aware Linearizers

aeup:
This pull request adds two new linearizers for the LOP DAG linearization.
The linearizers focus on reducing the minimal memory requirement by scheduling the LOPs in a way that affects the memory requirement for live intermediates.

**RESOURCE_AWARE_OPTIMAL** uses a brute-force approach that generates the optimal sequence but requires a lot of time to compile which makes it impractical for larger algorithms and most real world use cases.
**RESOURCE_AWARE_FAST** uses a heuristic approach that does not always generate the optimal sequence but is a lot faster than the other approach. The linearizer can be used for tasks with large data sets, but there are also some cases where the compile time is larger than expected or the memory requirement is larger in comparison to other linearizer.

cc @mboehm7


mboehm7:
LGTM - thanks for the patch @aeup. During the merge I expanded wildcard imports, and added a test for the optimal scheduler. Unfortunately, this optimal scheduler seems to run into an endless loop. I tested it with a toy example of 5 operators. Accordingly, I commented the code and we'll come back to this once we do a deeper comparison.

PR #2199, opened 2025-01-30
In-Place Dense Matrix Transposition 

jessicapriebe:
Added a new kernel for **In-Place Dense Matrix Transposition**, based on **Algorithm 467** by Brenner ([DOI: 10.1145/355611.362542](https://dl.acm.org/doi/pdf/10.1145/355611.362542)).

**Performance:**
Compared to the existing kernel, the added method provides significant performance benefits in a single-threaded context:
- Faster in **82.7% of cases**.
- Average speedup of **25.8%**.

**Note:** 
Performance measurements were restricted to cases where the existing kernel yields correct results. Similar or even better performance could be observed across all cases.

**Future Work:**
The divisors operate on different indices of the array, allowing for parallelization and offering additional performance improvements in multi-threaded scenarios.

@mboehm7

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
+++ b/src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
@@ -4332,7 +4332,6 @@ public class LibMatrixReorg {
 				iExponents[ip] = 0;
 				div /= powers[ip];
 			}
-			break;
 		}
 
 		denseBlock.setDims(new int[] {cols, rows});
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -4227,4 +4234,231 @@ public Object call() {
...
+			// decrease the remaining shift count by orig and comp
+			count -= 2;
+
+			if(prevOrig == start) {
+				// cycle and comp are distinct
+				matrix[orig] = val;
+				matrix[comp] = cval;
+				break;
+			}
+			if(prevComp == start) {
```
Baunsgaard, line 4383: else if?

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -4227,4 +4234,231 @@ public Object call() {
...
+		int workSize, int start, int comp) {
+
+		int orig = start;
+		double val = matrix[orig];
+		double cval = matrix[comp];
+
+		int prevOrig = prevIndexCycle(orig, rows, (maxIndex + 1) / rows);
+		int prevComp = maxIndex - prevOrig;
+
+		while(true) {
```
Baunsgaard, line 4369: you can probably help the compiler and improve performance a bit, by moving the content of this loop to another function.

furthermore, is this loop possible to do in parallel?

src/test/java/org/apache/sysds/test/component/matrix/libMatrixReorg/TransposeInPlaceBrennerTest.java
```diff
@@ -0,0 +1,98 @@
...
+		testTransposeInPlaceDense(347, 27953, 0.86);
+	}
+
+	private void testTransposeInPlaceDense(int rows, int cols, double sparsity) {
+		MatrixBlock X = MatrixBlock.randOperations(rows, cols, sparsity);
+		MatrixBlock tX = LibMatrixReorg.transpose(X);
+
+		LibMatrixReorg.transposeInPlaceDenseBrenner(X, 1);
+
+		TestUtils.compareMatrices(X, tX, 1e-8);
```
Baunsgaard, line 95: set epsilon to 0.

Baunsgaard (commented):
Very cool, thanks for the PR, and finding bugs in my transpose in place.

I have only a few comments.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
jessicapriebe: thank you for your feedback! I've addressed your comments in the latest commit :)

regarding parallelization, unfortunately, it's not possible at this level. it would need to be implemented at a higher level, specifically for the div loop.

mboehm7:
LGTM - thanks for the new kernel @jessicapriebe. I'll merge it in.

PR #2205, opened 2025-01-31
Fix recompute non zeros for column aggregations

Frxms:


mboehm7:
LGTM - thanks for catching this inconsistency and the patch @Frxms.

PR #2206, opened 2025-01-31
[SYSTEMDS-3553] Additional DNN/factorization optimizers and preconditioners (AdamW, ScaledGD)

ReneEnjilian:
This patch implements two new optimization algorithms: 

- Adam with weight decay (AdamW): https://arxiv.org/abs/1711.05101v3
- Scaled Gradient Descent (ScaledGD): https://arxiv.org/pdf/2005.08898

I did not add the shampoo optimizer because another student is working on that.

### DIFF
--- a/scripts/nn/optim/adamw.dml
+++ b/scripts/nn/optim/adamw.dml
@@ -1,0 +1,95 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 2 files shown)


mboehm7:
LGTM - thanks for the patch @ReneEnjilian. During the merge, I fixed the formatting and it would be nice to add tests in a future PR and report on the convergence speed of these optimizers.

PR #2220, opened 2025-02-05
[SYSTEMDS-3830] Add join operator to Scuro

christinadionysio:
This patch adds a new join operator to scuro. The join operation takes two modalities as well as a join condition as input and joins the two modalities on their common dimension (temporal for now). This includes two new modalities and the ability to apply new representations on top of a joined modality. In the future the join operator will also serve as a simple alignment operator by joining two modalities by a given offset.


Baunsgaard (commented):
LGTM, 

From a systems perspective we could separate the changes to `src/main/python/systemds/__init__.py` into another commit.

christinadionysio:
> LGTM,
> 
> From a systems perspective we could separate the changes to `src/main/python/systemds/__init__.py` into another commit.

@Baunsgaard, sounds good. I removed the changes to the __init__.py and will create another commit with the version changes.

Baunsgaard:
> > LGTM,
> > From a systems perspective we could separate the changes to `src/main/python/systemds/__init__.py` into another commit.
> 
> @Baunsgaard, sounds good. I removed the changes to the **init**.py and will create another commit with the version changes.

Great, I have now merged it.

PR #2224, opened 2025-02-07
[MINOR] Update Scuro dependency versions

christinadionysio:
In this patch the dependency versions for scuro were updated.


Baunsgaard:
Thanks, and now merged.

PR #2227, opened 2025-02-13
SP and FED Opcodes

aarnaty:
Shifted SP and FED opcodes to enum. Consolidated instruction types.


mboehm7:
LGTM - thanks for the patch @aarnaty. During the merge I just removed the unnecessary imports and fixed an instance in the compression lop where the use of opcodes was replaced back with strings.

PR #2229, opened 2025-02-16
Extended I/O Framework: Readers/Writers for Parquet

sayedkeika:
### Overview 
This pull request adds support for the Parquet file format. The implementation includes both readers and writers that can handle Parquet files sequentially and concurrently.

### Details
- Single-threaded reader and writer for smaller datasets
- Multi-threaded reader and writer for larger datasets
- Component tests for both sequential and parallel operations that cover different schemas
- Detailed documentation


mboehm7:
LGTM - thanks for the patch @sayedkeika. During the merge I resolved the merge conflicts, fixed the warnings and formatting (tabs over spaces), added additional tests (sparse data), and left a FIXME (for removing the ExampleParquetWriter). Additionally, I fixed the parallel write task to call the sequential write instead of the parallel writer again (which worked because the number of part files was 1 due to the small size).

PR #2230, opened 2025-02-17
[SYSTEMDS-3541] Exploratory workload-aware compression on intermediates

e-strauss:
This PR explores the aggressive compression on intermediates, explicitly for the kmeans builtin algorithm. This commit adds new compressed operations to avoid the decompression and minimize the compression time of intermediates.

The runtime of the kmeans algorithm on the census dataset was reduced from initially 50s with intermediate compression down to 17.5s with all the optimizations. Which is an overall improvement of 33% in comparison to the baseline runtime of workload-aware, non-aggressive compression of 27s.

A summary of the changes:

- added config option for aggressive compression
-  removed scalars from possible compressible intermediate candidates
- extended the compression workload analyzer to pick up aggregation operations if the input is compressed as a single column group
- extended the compression workload analyzer to pick up binary matrix-vec op, if of both inputs have the same col indices
- updated cost estimation for compression on already compressed inputs with single column group
- added support for double compressed binary matrix-matrix op
- single-threaded compressed binary matrix-vector operation with single column group encoding
- multi threaded compressed binary matrix-vector operation with single column group encoding
- added support for left binary matrix-vector operation for sparse outputs

### DIFF
--- a/src/main/java/org/apache/sysds/hops/rewrite/RewriteCompressedReblock.java
+++ b/src/main/java/org/apache/sysds/hops/rewrite/RewriteCompressedReblock.java
@@ -171,11 +171,12 @@ public class RewriteCompressedReblock extends StatementBlockRewriteRule {
 			satisfies |= HopRewriteUtils.isTernary(hop, OpOp3.CTABLE) 
 				&& hop.getInput(0).getDataType().isMatrix() 
 				&& hop.getInput(1).getDataType().isMatrix();
-			satisfies |= HopRewriteUtils.isData(hop, OpOpData.PERSISTENTREAD) && !hop.isScalar();
+			satisfies |= HopRewriteUtils.isData(hop, OpOpData.PERSISTENTREAD);
 			satisfies |= HopRewriteUtils.isUnary(hop, OpOp1.ROUND, OpOp1.FLOOR, OpOp1.NOT, OpOp1.CEIL);
 			satisfies |= HopRewriteUtils.isBinary(hop, OpOp2.EQUAL, OpOp2.NOTEQUAL, OpOp2.LESS,
 				OpOp2.LESSEQUAL, OpOp2.GREATER, OpOp2.GREATEREQUAL, OpOp2.AND, OpOp2.OR, OpOp2.MODULUS);
 			satisfies |= HopRewriteUtils.isTernary(hop, OpOp3.CTABLE);
+			satisfies &= !hop.isScalar();
 		}
 		if(LOG.isDebugEnabled() && satisfies)
 			LOG.debug("Operation Satisfies: " + hop);
--- a/src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
+++ b/src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
@@ -20,6 +20,7 @@
 package org.apache.sysds.runtime.compress.lib;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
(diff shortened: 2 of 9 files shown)


src/main/java/org/apache/sysds/hops/rewrite/RewriteCompressedReblock.java
```diff
@@ -171,11 +171,12 @@ public static boolean satisfiesAggressiveCompressionCondition(Hop hop) {
 			satisfies |= HopRewriteUtils.isTernary(hop, OpOp3.CTABLE) 
 				&& hop.getInput(0).getDataType().isMatrix() 
 				&& hop.getInput(1).getDataType().isMatrix();
-			satisfies |= HopRewriteUtils.isData(hop, OpOpData.PERSISTENTREAD) && !hop.isScalar();
+			satisfies |= HopRewriteUtils.isData(hop, OpOpData.PERSISTENTREAD);
 			satisfies |= HopRewriteUtils.isUnary(hop, OpOp1.ROUND, OpOp1.FLOOR, OpOp1.NOT, OpOp1.CEIL);
 			satisfies |= HopRewriteUtils.isBinary(hop, OpOp2.EQUAL, OpOp2.NOTEQUAL, OpOp2.LESS,
 				OpOp2.LESSEQUAL, OpOp2.GREATER, OpOp2.GREATEREQUAL, OpOp2.AND, OpOp2.OR, OpOp2.MODULUS);
 			satisfies |= HopRewriteUtils.isTernary(hop, OpOp3.CTABLE);
+			satisfies &= !hop.isScalar();
```
Baunsgaard, line 179: great catch

src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
```diff
@@ -52,13 +66,15 @@
...
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
+import org.apache.sysds.runtime.matrix.data.Pair;
 import org.apache.sysds.runtime.matrix.operators.BinaryOperator;
 import org.apache.sysds.runtime.matrix.operators.LeftScalarOperator;
 import org.apache.sysds.runtime.matrix.operators.RightScalarOperator;
 import org.apache.sysds.runtime.matrix.operators.ScalarOperator;
 import org.apache.sysds.runtime.util.CommonThreadPool;
 import org.apache.sysds.utils.DMLCompressionStatistics;
 import org.apache.sysds.utils.stats.Timing;
+import org.jetbrains.annotations.NotNull;
```
Baunsgaard, line 77: replace this import with code for != null

src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
```diff
@@ -104,11 +120,35 @@ public static MatrixBlock binaryOperationsLeft(BinaryOperator op, CompressedMatr
 	private static MatrixBlock binaryOperationsRightFiltered(BinaryOperator op, CompressedMatrixBlock m1,
 		MatrixBlock that) throws Exception {
 		BinaryAccessType atype = LibMatrixBincell.getBinaryAccessTypeExtended(m1, that);
-		if(that instanceof CompressedMatrixBlock && that.getInMemorySize() < m1.getInMemorySize()) {
+		if(that instanceof CompressedMatrixBlock){
```
Baunsgaard, line 123: take the nested if statements, and refactor into a utility method named something appropriate.

Then make the nested statements into one, followed by another function call to a new method, that returns.


```java
if(newfunction(...)){
    return secondFunction(...);
}
```

src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
```diff
@@ -403,25 +447,133 @@ else if(nnz == 0) // all was 0 -> return empty.
...
+				nnz += f.get();
+		}
+		finally {
+			pool.shutdown();
+		}
+
+		// map each unique indicator to an index
+		HashMapToInt<Integer> hm = new HashMapToInt<>(nCols*2);
+		int[] colMap = new int[nRows];
+		for(int j = 0; j < tasks.size(); j++) {
```
Baunsgaard, line 526: you can get speedups (because of JIT compilation) by moving these two for loops to a different function.

src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
```diff
@@ -403,25 +447,133 @@ else if(nnz == 0) // all was 0 -> return empty.
...
+
+		// get indicators (one-hot-encoded comparison results)
+		long nnz = 0;
+		final ArrayList<BinaryMVColTaskCompressed> tasks = new ArrayList<>();
+		final ExecutorService pool = CommonThreadPool.get(op.getNumThreads());
+		try {
+			for(int i = 0; i < nRows; i += blkz) {
+				tasks.add(new BinaryMVColTaskCompressed(m1, m2, i, Math.min(nRows, i + blkz), op, left));
+			}
+			for(Future<Long> f : pool.invokeAll(tasks))
```
Baunsgaard, line 516: consider if we can avoid calling the tasks here.

I do think we can get away with calling later, and extend the try catch block around subsequent code.
This would parallelize the allocation of the subsequent colmap.

src/test/java/org/apache/sysds/test/component/compress/lib/CLALibBinaryCellOpTest.java
```diff
@@ -62,7 +62,7 @@ public class CLALibBinaryCellOpTest {
 		// (LessThanEquals.getLessThanEqualsFnObject()), //
 		// (GreaterThan.getGreaterThanFnObject()), //
 		// (GreaterThanEquals.getGreaterThanEqualsFnObject()), //
-		// (Multiply.getMultiplyFnObject()), //
+		(Multiply.getMultiplyFnObject()), //
```
Baunsgaard, line 65: i think i missed this in some commit i made because i was debugging, we should enable all of these again.

Baunsgaard (commented):
very good!

However, we can get a bit more performance out of it, via small optimizations.

src/main/java/org/apache/sysds/runtime/compress/lib/CLALibBinaryCellOp.java
e-strauss: but we have to to merge the results in a single thread anyways or do you want to use a concurrent hashmap?

Baunsgaard:
Thanks for the contribution @Elias-Strauss  . It is now merged.

PR #2234, opened 2025-02-20
implement new colProd and rowProd tests

Frxms:


mboehm7:
LGTM - thanks for the extended and new tests @Frxms.

PR #2237, opened 2025-02-24
NN Built-In for Embedding Layers

MaximilianSchreff:
This PR adds the embedding layer as a built-in operator in our nn/layers library. The functionality is similar to pytorch.nn.Embedding (https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html)

The layer receives indices as input which refer to indices of an embedding dictionary and returns an embedding matrix where row i refers to embedding vector indices[i] of the embedding dictionary.

This layer is used in every transformer architecture. Here the indices usually come from a tokenizer and the embedding matrix is the input to the actual transformer model.

### Testing
- Testing forward pass and backward pass for correctness
- Implemented as a component test in NNComponentTest.java
- Manually calculated test cases for the forward pass
- For backward pass, comparison against pytorches autograd module

### DIFF
--- a/scripts/nn/layers/embedding.dml
+++ b/scripts/nn/layers/embedding.dml
@@ -19,8 +19,33 @@
 #
 #-------------------------------------------------------------
 
-forward = function()
-    return () {}
+forward = function(matrix[double] indices, matrix[double] embedding_dict)
+    return (matrix[double] embeddings) {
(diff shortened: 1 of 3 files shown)


phaniarnab:
Thanks @MaximilianSchreff. I will merge it in.

PR #2240, opened 2025-03-01
Improve test coverage for rewrite package

ReneEnjilian:
This pull request is the first in a series aimed at enhancing the test coverage of the rewrite package. Specifically, it adds additional tests for the algebraic simplifications defined in `RewriteAlgebraicSimplificationStatic`. The goal is also to proactively identify potential bugs, similar to our previous efforts. No bugs were found in this patch.

### DIFF
--- a/src/main/java/org/apache/sysds/hops/OptimizerUtils.java
+++ b/src/main/java/org/apache/sysds/hops/OptimizerUtils.java
@@ -199,7 +199,7 @@ public class OptimizerUtils
 	public static boolean ALLOW_SUM_PRODUCT_REWRITES2 = true;
 
 	/**
-	 * Enables additional mmchain optimizations. in the future, this might be merged with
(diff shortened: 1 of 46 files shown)


mboehm7:
Thanks @ReneEnjilian - could you please address the failing rewrite tests (I just kicked them off again, but there seem to be systematic failures).

ReneEnjilian:
I implemented a fix and hope that this resolves the issue now. The error does not occur on my machine but from what I can see is that R struggled with the class type of the result matrix. Changing `outer(v, 1:m, "==")` to `outer(as.vector(v), 1:m, "==")` should resolve the error because R is now recognized as type matrix and not only as type array anymore.

mboehm7:
LGTM - thanks @ReneEnjilian for the effort improving our test coverage.

PR #2243, opened 2025-03-14
[MINOR] Fix in Python API GatewayServerListener

e-strauss:
two things:
- the current inner class DMLGateWayListener implements function from the GatewayServerListener interface, which are never invoked by the GatewayServer (since the GatewayServer, which also implements GatewayServerListener, does not implement these methods
- DMLGateWayListener previously called, Sys.exit(), which I think is not correct, since it breaks the proper shutdown of the GatewayServer

I added a new unit case, which checks the functionality of the DMLGateWayListener.

### DIFF
--- a/src/main/java/org/apache/sysds/api/PythonDMLScript.java
+++ b/src/main/java/org/apache/sysds/api/PythonDMLScript.java
@@ -21,17 +21,20 @@ package org.apache.sysds.api;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.log4j.Level;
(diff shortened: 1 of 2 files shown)


Baunsgaard:
Hi @e-strauss , 

We added the shutdown calls to enforce that the Java process would terminate in case the Python API terminated. It does violate the default shutdown process since any live connections would be forcefully terminated. Because only one Python process is guaranteed to be using this listener, the design was valid.

But I do see your points, and I like the PR. I just want to verify the Python API startup and shutdown time of the change before merging.

Baunsgaard:
Merged, while merging i verified that there was no regressions.

PR #2246, opened 2025-03-31
Added Vector API and FFM API prototypes to staging area

moatabay:


mboehm7:
LGTM - thanks for this contribution @moatabay and sorry for the delay. Since we now switched to JDK 17, we can slowly start bringing some of your kernels into the main codeline too. Thanks.

PR #2247, opened 2025-04-01
[MINOR] Remove outputdir .asf.yaml

Baunsgaard:
Currently the format of the .asf.yaml sends emails to everyone on all commits. According to https://issues.apache.org/jira/browse/INFRA-26700 the error is in our project. This commit, therefore, removes the outputdir, as specified in: https://github.com/apache/infrastructure-asfyaml?tab=readme-ov-file#jekyll-cms

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -58,4 +58,3 @@ github:
 staging:
   profile: beta
   whoami: main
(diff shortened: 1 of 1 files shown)


Baunsgaard:
Merged

PR #2250, opened 2025-04-02
[SYSTEMDS-3182] Builtin ampute() for introducing missing values in data

mmoesm:
This patch adds ampute(), as known from R's MICE package, to SystemDS as a DML-bodied built-in function, including unit tests on real datasets.

The method accepts the following parameters: the dataset itself, the proportion of missing values to introduce (a double value), the patterns, the frequency of each pattern, mechanism of missing values (MAR, MCAR, MNAR), pattern weights, and an RNG seed.

Everything is "approximately" done, meaning I have a few notes, all marked by `FIX ME`s in ampute.dml:

*  I use a couple of `while (TRUE) {}` blocks to circumvent an issue with (possibly) conflicting dynamic vs. static matrix behaviors on [Lines 76](https://github.com/apache/systemds/pull/2250/files#diff-782ee837628fcd8bae928f8e67eb396ea25a3d8cae464ed999c708367fa4e04fR76) and [244](https://github.com/apache/systemds/pull/2250/files#diff-782ee837628fcd8bae928f8e67eb396ea25a3d8cae464ed999c708367fa4e04fR244) in `ampute.dml`
*  There is currently an issue with capturing the return values of `freq`, `patterns`, and `weights` on [Line 49](https://github.com/apache/systemds/pull/2250/files#diff-782ee837628fcd8bae928f8e67eb396ea25a3d8cae464ed999c708367fa4e04fR49) in ampute.dml. This can be circumvented by capturing the output in throwaway variables and reassigning to `freq`, `patterns`, and `weights`, as shown in comments on lines 50-52.
* As a minor cosmetic issue, returned values of `u_isEmpty` in ampute.dml are captured in local variables before being used inside `if` headers, since such function calls cannot be used in this context yet.

### DIFF
--- a/scripts/builtin/ampute.dml
+++ b/scripts/builtin/ampute.dml
@@ -1,0 +1,311 @@
+#-------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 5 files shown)


mboehm7:
Thanks for the patch and your contribution @mmoesm. LGTM and during the merge I only fixed the formatting of the Java test (tabs over spaces in our Java code, I would recommend to enable whitespace characters to see them in your IDE). Also thanks for catching these issues, which I will follow-up and fix.

PR #2256, opened 2025-04-30
[SYSTEMDS-3870] Update to Java 17

Baunsgaard:
This PR contains the code to update the base system to Java 17, 
It includes updated docker images for testing and POM updates for Maven. 

However, it will probably take more edits for a full change to Java 17.


Baunsgaard:
Please be advised, I have already pushed the Java 17 test image, therefore all other PRs will run with Java 17. 
I did test it locally if it works with the old POM and the answer is yes. 
However, I do not know if it works in the cloud and therefore we might have some issues.

PR #2268, opened 2025-05-26
[SYSTEMDS-3330] Documentation of builtin functions

anuunchin:
### Description 

This PR resolves [issue 3303](https://issues.apache.org/jira/browse/SYSTEMDS-3330) and is a part of the AMLS course. 

### Overview (as of now)

To resolve the issue, the following tasks are to be carried out:
- [x] Fix the warnings and errors currently thrown by the `src/main/python/generator/generator.py`.
- [x] Starting with the following builtin functions, add a python and dml example in a format of a code block, but under a different tag:
   - [x] dist.dml
   - [x] lm.dml
   - [x] normalize.dml
   - [x] toOneHot.dml
   - [x] randomForest.dml
   - [ ] img_*.dml
- [x] Adjust `src/main/python/generator/generator.py` and `src/main/python/generator/dml_parser.py` accordingly to handle the python examples and transform them into test cases, possibly using a new script under `src/main/python/generator/test_generator.py`.


scripts/builtin/glove.dml
```diff
@@ -119,7 +141,7 @@ gloveWithCoocMatrix = function(matrix[double] cooc_matrix, frame[Unknown] cooc_i
     G = cbind(cooc_index[,2], as.frame(G));
 }
 
-glove = function(
+f_glove = function(
```
Baunsgaard, line 147: we use the f_ to indicate that this is the public function, therefore we should not modify this line.
the rest of the functions in this file are private, and therefore not available from outside.

scripts/builtin/shapExplainer.dml
```diff
@@ -51,9 +51,8 @@
 # S              Matrix holding the shapley values along the cols, one row per instance.
 # expected       Double holding the average prediction of all instances.
 # -----------------------------------------------------------------------------
-s_shapExplainer = function(String model_function, list[unknown] model_args, Matrix[Double] x_instances,
-    Matrix[Double] X_bg, Integer n_permutations = 10, Integer n_samples = 100, Integer remove_non_var=0,
-    Matrix[Double] partitions=as.matrix(-1), Integer seed = -1, Integer verbose = 0)
+
+s_shapExplainer = function(String model_func, list[unknown] model_args, Matrix[Double] x_instances, Matrix[Double] X_bg, Integer n_permutations = 10, Integer n_samples = 100, Integer remove_non_var=0, Matrix[Double] partitions=as.matrix(-1), Integer seed = -1, Integer verbose = 0)
```
Baunsgaard: avoid changing the names of the arguments, such as model_function -> model_func.

This would break backwards compatibility.

Baunsgaard (commented):
I have a few early comments.

anuunchin:
This PR is separated into a prep PR #2291 that resolves existing issues thrown by generator.py and.

PR #2276, opened 2025-06-23
Sparse row templates

Frxms:
I have implemented SparseRowPrimitives into the Binary and Unary templates. I have also added an "S" to every result variable that is sparse to keep track of sparse intermediates.


mboehm7:
There seem to be still some codegen errors - do you need any help @Frxms ?

Frxms:
As far as i understood it, the Unary class was missing its ROW_VARS case. Could you run the tests again please?

PR #2278, opened 2025-06-26
Systemds 3863 batliner

KilianBat:


mboehm7:
LGTM - thanks for the patch @KilianBat. During the merge, I only cleaned up the minor formatting, unnecessary imports, and tests (Python dependency).

PR #2286, opened 2025-07-02
[SYSTEMDS-3253] Add combined rewrite and lop instruction for union

chihsinh:
This patch refines the current union operation to an internal LOP operation. Currently, two subsequent operations -- rbind() and unique(), are used to perform the union operation. We rewrite the operation with an internal LOP that uses a HashSet to compute the unique entries and returns them in a matrix. This improves the efficiency of the operation, as it avoids unique(). The order of the input entries is preserved in the output.


mboehm7:
Btw, the rewrite test seems to fail because `rewriteSimplifyUnionDistinct.dml` cannot be found - this likely due to the local tests running in Windows, which is case-insensitive, while the github action tests run in Linux which is case-sensitive. Please fix the capitalization but also rename to something else; otherwise git on windows will have problems because it cannot differentiate old and new files, again because of the case insensitivity.

PR #2290, opened 2025-07-13
[SYSTEMDS-3730] Multi-threaded rev reorg operation

j143:
Basic multi-thread rev implementation.

## **the rev() Operation**

Given a matrix $$\( A \)$$ of size $$\( m \times n \)$$, `rev()` produces a matrix $$\( B \)$$ of the same dimensions such that:

$$B_{i,j} = A_{m-1-i, j}$$

for all $$\( 0 \leq i < m \)$$ and $$\( 0 \leq j < n \)$$.

**In words:**  
- Each row of the output is the corresponding row from the bottom of the input matrix.
- The last row of the input becomes the first row of the output, the second last becomes the second, ... the first row becomes the last.

---

## **Example**

Suppose:

$$A = \begin{bmatrix}
1 & 2 & 3 \\
4 & 5 & 6 \\
7 & 8 & 9 \\
\end{bmatrix}$$

Then after `rev(A)`, you get:

$$B = \begin{bmatrix}
7 & 8 & 9 \\
4 & 5 & 6 \\
1 & 2 & 3 \\
\end{bmatrix}$$

---

## **Multi-threaded Implementation**

- **Split the matrix** into chunks of rows.
- **Each thread** copied its assigned chunk of output rows, each time taking from the appropriate (reversed) input row:
    - If a thread was responsible for output row $$\( i \)$$, it took input row $$\( m-1-i \)$$.
- All threads worked in parallel.

---

## **Summary Table**

| Input Row $$\( i \)$$ | Output Row $$\( m-1-i \)$$ |
|---------------------|--------------------------|
| 0                   | last row                 |
| 1                   | second last row          |
| ...                 | ...                      |
| m-1                 | first row                |

---

## **test**

for every cell $$\( (i, j) \)$$, `output[i][j] == input[rows - 1 - i][j]`


src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -128,7 +128,11 @@ public static MatrixBlock reorg( MatrixBlock in, MatrixBlock out, ReorgOperator
 				else
 					return transpose(in, out);
 			case REV:
-				return rev(in, out);
+//				System.out.println("Reorg: rev() called with numThreads: " + op.getNumThreads());
```
mboehm7: remove these comment sysout lines.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
 		return out;
 	}
 
+	public static MatrixBlock rev(MatrixBlock in, MatrixBlock out, int k) {
+		if (k <= 1 || in.isEmptyBlock(false)) {
+			System.out.println("choosing single thread");
```
mboehm7: as above

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
 		return out;
 	}
 
+	public static MatrixBlock rev(MatrixBlock in, MatrixBlock out, int k) {
+		if (k <= 1 || in.isEmptyBlock(false)) {
```
mboehm7: add a check for a minimum size for parallelization, otherwise fall back to single-threaded as we do it for example in aggregations (LibMatrixAgg).

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
...
+
+		// Before starting threads, ensure the output sparse block is allocated!
+		if (sparse) {
+			out.allocateSparseRowsBlock(false);
+		}
+
+		// Set up thread pool
+		ExecutorService pool = CommonThreadPool.get(k);
+		try {
+			int blklen = (int) Math.ceil((double) numRows / k);
```
mboehm7, line 415: I would recommend to create smaller tasks (e.g., numRows/k/4) which yields better load balancing.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
...
+			List<Future<?>> tasks = new ArrayList<>();
+
+			for (int i = 0; i < k; i++) {
+				final int startRow = i * blklen;
+				final int endRow = Math.min((i + 1) * blklen, numRows);
+
+				tasks.add(pool.submit(() -> {
+					if (!sparse) {
```
mboehm7, line 423: create a static method for this kernel which is called from both the single-threaded implementation as well as the multi-threaded implementation

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
...
+						// Dense case
+						System.out.println("dense case");
```
mboehm7: remove

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
...
+						double[] inVals = in.getDenseBlockValues();
+						double[] outVals = out.getDenseBlockValues();
+						for (int r = startRow; r < endRow; r++) {
+							int revRow = numRows - r - 1;
+							System.arraycopy(inVals, revRow * numCols, outVals, r * numCols,
+									numCols);
```
mboehm7: avoid this line break

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -389,10 +393,78 @@ public static MatrixBlock rev( MatrixBlock in, MatrixBlock out ) {
...
+			// Wait for all threads
+			for (Future<?> task : tasks) {
+				task.get();
+			}
+		} catch (Exception ex) {
+			throw new DMLRuntimeException(ex);
+		} finally {
+			pool.shutdown();
+		}
+		out.recomputeNonZeros();
```
mboehm7: avoid this unnecessary recomputation of the the number of non-zeros. We should just set the nnz of the output to the nnz of the input, as this reorganization does not change values but only placed them differently

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -2461,6 +2533,7 @@ public static int[] mergeNnzCounts(int[] cnt, int[] cnt2) {
 	private static void reverseDense(MatrixBlock in, MatrixBlock out) {
 		final int m = in.rlen;
 		final int n = in.clen;
+		System.out.println("inside reverseDense");
```
mboehm7: remove

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -2486,6 +2559,7 @@ private static void reverseDense(MatrixBlock in, MatrixBlock out) {
 
 	private static void reverseSparse(MatrixBlock in, MatrixBlock out) {
 		final int m = in.rlen;
+		System.out.println("inside reverseSparse");
```
mboehm7: remove

src/main/java/org/apache/sysds/runtime/matrix/data/TestMultiThreadedRev.java
```diff
@@ -0,0 +1,62 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sysds.runtime.matrix.data;
+
+
+public class TestMultiThreadedRev {
```
mboehm7: remove this class, and rather ensure that some of the existing tests run with large enough inputs that it triggers the multi-threaded code path too.

src/test/java/org/apache/sysds/test/functions/reorg/FullReverseTest.java
```diff
@@ -165,6 +188,76 @@ else if ( instType == ExecType.SPARK )
...
+		}
+		boolean sparkConfigOld = DMLScript.USE_LOCAL_SPARK_CONFIG;
+		if( rtplatform == ExecMode.SPARK )
+			DMLScript.USE_LOCAL_SPARK_CONFIG = true;
+
+		String TEST_NAME = testname;
+
+		try
+		{
+			System.setProperty("sysds.parallel.threads", String.valueOf(numThreads));
```
mboehm7, line 220: remove (I don't think we use this property internally)

mboehm7 (commented):
Thanks for the patch @j143 - overall the changes are a very good start. Please address the minor comments and then I would merge it in.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixReorg.java
j143: I have added something like 3000_000 in the hop for now, for the tests to trigger the multi-thread.

j143:
At this point, the multi-threading is working. Can we checkpoint this PR till now and refine further in next PR?

Next:
1. CP -> we decide threading, SP -> let the spark decide
2. increase the test coverage
3. use a common static block to use for both the single thread, multi-thread case
4. decide on the blocksize or the threshold at which to trigger to threading

mboehm7:
Thanks for the patch @j143. During the merge, I fixed the remaining stdout printing, moved the parallelization check from compilation to runtime, used the existing common kernels, fixed test errors (dense operations where running into null-pointer exceptions), and consolidated the tests.

PR #2295, opened 2025-07-18
[MINOR] PythonAPI: Fix for correct logging setup for debug

e-strauss:
I think there is a bug in the current setup of the logging in the Python SystemDS context, because the debug logging was not working on machine.

The first commit contains test if setup worked correctly, which fails.
The second commit fixes the setup.


Baunsgaard (commented):
LGTM

PR #2300, opened 2025-07-30
[SYSTEMDS-3901] Updated Python CI Test setup

e-strauss:
Changes:
- split down of python test runs: w/ env, w/o env, federated
- separate python formatting test

These changes reduce the runtime of the single tests, and also make the Python test without environment to actually run without the SYSTEMDS_HOME installation.


Baunsgaard (commented):
LGTM

PR #2302, opened 2025-08-03
[SYSTEMDS-3900] Stream aware write operation for OOC (option 2)

j143:
..

I've chosen two pass method, write data separately and hold metadata in memory and concat at the end.

### DIFF
--- a/src/test/java/org/apache/sysds/test/functions/ooc/UnaryTest.java
+++ b/src/test/java/org/apache/sysds/test/functions/ooc/UnaryTest.java
@@ -31,6 +31,7 @@ import org.apache.sysds.runtime.io.MatrixWriterFactory;
 import org.apache.sysds.runtime.matrix.data.MatrixBlock;
 import org.apache.sysds.runtime.matrix.data.MatrixValue;
 import org.apache.sysds.runtime.meta.MatrixCharacteristics;
+import org.apache.sysds.runtime.util.DataConverter;
 import org.apache.sysds.runtime.util.HDFSTool;
 import org.apache.sysds.test.AutomatedTestBase;
 import org.apache.sysds.test.TestConfiguration;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/sysds/runtime/instructions/cp/VariableCPInstruction.java
```diff
@@ -1060,6 +1054,36 @@ private void processWriteInstruction(ExecutionContext ec) {
 			HDFSTool.writeScalarToHDFS(ec.getScalarInput(getInput1()), fname);
 		}
 		else if( getInput1().getDataType() == DataType.MATRIX ) {
+			MatrixObject mo = ec.getMatrixObject(getInput1().getName());
+			int blen = Integer.parseInt(getInput4().getName());
+			LocalTaskQueue<IndexedMatrixValue> stream = mo.getStreamHandle();
+
+			if (stream != null) {
```
j143, line 1071: Due to recompilation, this is visited the second time. Since the stream is null the second time, it fails.

So, I tried setting the values explicitly. But, recompilation still happening.

```java
mo.updateDataCharacteristics(mc);
HDFSTool.copyFileOnHDFS(fname, mo.getFileName());
mo.setDirty(false);
```

mboehm7: I would recommend to simple probe if a stream exists and not consume the matrix if it does.

mboehm7:
You don't have to invent any new things here: we take the stream of blocks and directly write them into a sequence file (like our existing binary writer, which actually takes the overall matrix and chunks them off into 1k-1k blocks). The resulting sequence files should be identical, and there is no file-system merge needed.

src/main/java/org/apache/sysds/runtime/instructions/cp/VariableCPInstruction.java
j143: Hi, I've not understood this part! :)

PR #2305, opened 2025-08-09
[SYSTEMDS-3904] OOC Matrix Vector operations

j143:
..


mboehm7:
LGTM - thanks for the nice addition to the OOC backend. During the merge, I fixed the formatting (tabs over spaces in Java code), extended the tests slightly (with what you already planned), used the slice function for consistency, removed the task.get() which would block until the entire input is processed, and optimized the case of a single column block (which does not need any aggregation and allows streaming).

PR #2310, opened 2025-08-20
Add early block outputs for OOC matrix-vector multiplication

jessicapriebe:
This extends the OOC matrix-vector multiplication by supporting early block outputs


mboehm7:
LGTM - thanks @jessicapriebe for this patch and improvement. During the merge, I only slightly moved the count maintenance around such that we only replace these counts (which is a hashmap update) if we do not anyway output the block and then remove the count. As a future step, we should think about some util methods / classes for encapsulating the count maintenance and early emission of blocks.

PR #2311, opened 2025-08-20
[SYSTEMDS-??] cujava: Custom java-cuda bindings for SystemDS

ReneEnjilian:
This PR starts the implementation of custom Java–CUDA bindings for SystemDS, called **cujava**. The long-term goal is to replace **JCuda** altogether. The project is split into a Java layer and a C++/JNI layer. 
### Project structure
`src/main/java/org.apache.sysds/cujava:`

**cujava/
├─ cublas/
├─ cudnn/
├─ cusolver/
├─ cusparse/
├─ driver/
├─ interop/
└─ runtime/
CudaDataType
CudaException
CuJavaLibLoader
NativePointerObject
Pointer
Sizeof**

Each directory contains the corresponding Java-side implementation. 

`src/main/cpp/jni:`

**jni/
├─ common/
├─ cublas/
├─ cudnn/
├─ cusolver/
├─ cusparse/
├─ driver/
└─ runtime/
build_cujava_libs.sh
CMakeLists.txt**

These directories hold the C++/JNI implementations that mirror the Java side. Each directory has its own CMakeLists.txt to produce a dedicated shared library. The libraries are emitted under `src/main/cpp/lib`. 
I completed the runtime, driver, cusparse, and cublas packages (java + cpp). The cusolver and cudnn remain for future work.

### Example usage
Writing code with cuJava is reminiscent of JCuda—only the imports change:

`import org.apache.sysds.cujava.runtime.CuJava;`
`CuJava.cudaMalloc(...);`

### Notes

- **Parallel to JCuda:** cuJava currently exists alongside JCuda and is not integrated into the SystemDS GPU backend yet. The existing GPU backend remains unchanged and continues to use JCuda. The newly added cuJava runtime methods have been tested successfully.
- **API scope:** cuJava intentionally implements only the CUDA/NVIDIA library calls that SystemDS actually uses. This downsized approach—contrary to JCuda’s full coverage—enables faster upgrades to newer CUDA versions: we only need to update deprecated APIs we rely on. In contrast, JCuda must update every deprecated method, which is cumbersome and wasteful.
- **Linux vs. Windows:** At the moment, shared libraries are generated only for Linux (producing `.so` files). I don’t have a Windows machine, so Windows builds are omitted for now. It should be straightforward to adapt the `CMakeLists` and `CuJavaLibLoader.java` to support Windows and emit `.dll` files as well.

@mboehm7


ReneEnjilian:
The errors resulted from a mistake in the javadoc. Apparently, one cant use `<`. I fixed that now.

ReneEnjilian:
The two failing tests are unrelated to my changes. The first failing check stems from the federated backend and the second is in the python layer of SystemDS. cujava touches neither of them and is in fact isolated from the rest of the codebase. I guess this is likely a CI/CD issue ?

ReneEnjilian:
okay, all tests pass now. After this PR is merged, I will make a follow-up PR, where I complete the cujava runtime package. I will (hopefully) complete that PR either tomorrow or on Saturday.

PR #2317, opened 2025-08-28
[WIP][SYSTEMDS-3907] tee ooc operator

j143:
implement tee ooc operator

TODO:
1. address review comment - moving TeeOp to DataOp
2. presently only, meant for t(X) %*% X, need to see if this works in general.

### DIFF
--- a/src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
+++ b/src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
@@ -48,194 +48,194 @@ import java.util.Set;
  */
 public class RewriteInjectOOCTee extends HopRewriteRule {
 
-  private static final Set<Long> rewrittenHops = new HashSet<>();
-  private static final Map<Long, Hop> handledHop = new HashMap<>();
+	private static final Set<Long> rewrittenHops = new HashSet<>();
+	private static final Map<Long, Hop> handledHop = new HashMap<>();
 
-  // Maintain a list of candidates to rewrite in the second pass
-  private final List<Hop> rewriteCandidates = new ArrayList<>();
+	// Maintain a list of candidates to rewrite in the second pass
+	private final List<Hop> rewriteCandidates = new ArrayList<>();
 
-  /**
-   * Handle a generic (last-level) hop DAG with multiple roots.
-   *
-   * @param roots high-level operator roots
-   * @param state program rewrite status
-   * @return list of high-level operators
-   */
-  @Override
-  public ArrayList<Hop> rewriteHopDAGs(ArrayList<Hop> roots, ProgramRewriteStatus state) {
-    if (roots == null) {
-      return null;
-    }
+	/**
+	 * Handle a generic (last-level) hop DAG with multiple roots.
+	 *
+	 * @param roots high-level operator roots
+	 * @param state program rewrite status
+	 * @return list of high-level operators
+	 */
+	@Override
+	public ArrayList<Hop> rewriteHopDAGs(ArrayList<Hop> roots, ProgramRewriteStatus state) {
+		if (roots == null) {
+			return null;
+		}
 
-    // Clear candidates for this pass
-    rewriteCandidates.clear();
+		// Clear candidates for this pass
+		rewriteCandidates.clear();
 
-    // PASS 1: Identify candidates without modifying the graph
-    for (Hop root : roots) {
-      root.resetVisitStatus();
-      findRewriteCandidates(root);
-    }
+		// PASS 1: Identify candidates without modifying the graph
+		for (Hop root : roots) {
+			root.resetVisitStatus();
+			findRewriteCandidates(root);
+		}
 
-    // PASS 2: Apply rewrites to identified candidates
-    for (Hop candidate : rewriteCandidates) {
-      applyTopDownTeeRewrite(candidate);
-    }
+		// PASS 2: Apply rewrites to identified candidates
+		for (Hop candidate : rewriteCandidates) {
+			applyTopDownTeeRewrite(candidate);
+		}
 
-    return roots;
-  }
+		return roots;
+	}
 
-  /**
-   * Handle a predicate hop DAG with exactly one root.
-   *
-   * @param root  high-level operator root
-   * @param state program rewrite status
-   * @return high-level operator
-   */
-  @Override
-  public Hop rewriteHopDAG(Hop root, ProgramRewriteStatus state) {
-    if (root == null) {
-      return null;
-    }
+	/**
+	 * Handle a predicate hop DAG with exactly one root.
+	 *
+	 * @param root  high-level operator root
+	 * @param state program rewrite status
+	 * @return high-level operator
+	 */
+	@Override
+	public Hop rewriteHopDAG(Hop root, ProgramRewriteStatus state) {
+		if (root == null) {
+			return null;
+		}
 
-    // Clear candidates for this pass
-    rewriteCandidates.clear();
+		// Clear candidates for this pass
+		rewriteCandidates.clear();
 
-    // PASS 1: Identify candidates without modifying the graph
-    root.resetVisitStatus();
-    findRewriteCandidates(root);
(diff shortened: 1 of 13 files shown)


src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -212,7 +228,7 @@ private void applyTopDownTeeRewrite(Hop sharedInput) {
 	private boolean isNotAlreadyTee(Hop hop) {
 		if (hop.getParent().size() > 1) {
 			for (Hop consumer : hop.getParent()) {
-				if (consumer instanceof TeeOp) {
+				if (consumer instanceof DataOp) {
```
j143, line 231: this check only checking for DataOp.

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -175,7 +178,20 @@ private void applyTopDownTeeRewrite(Hop sharedInput) {
 		ArrayList<Hop> consumers = new ArrayList<>(sharedInput.getParent());
 
 		// Create the new TeeOp with the original hop as input
-		TeeOp teeOp = new TeeOp(sharedInput);
+//		TeeOp teeOp = new TeeOp(sharedInput);
+		DataOp teeOp	= new DataOp("tee_out_" + sharedInput.getName(),
```
j143, line 182: presently we are creating different variable names than the ones created automatically _mvar1 ... --> is this the correct approach?

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -175,7 +178,20 @@ private void applyTopDownTeeRewrite(Hop sharedInput) {
...
+						sharedInput.getDataType(),
+						sharedInput.getValueType(),
+						Types.OpOpData.TRANSIENTWRITE,
```
j143, line 185: Is it supposed to be TRANSIENTWRITE or TRANSIENTREAD?

src/main/java/org/apache/sysds/hops/DataOp.java
```diff
@@ -32,12 +32,8 @@
 import org.apache.sysds.conf.CompilerConfig.ConfigType;
 import org.apache.sysds.conf.ConfigurationManager;
 import org.apache.sysds.hops.rewrite.HopRewriteUtils;
-import org.apache.sysds.lops.Data;
-import org.apache.sysds.lops.Federated;
-import org.apache.sysds.lops.Lop;
+import org.apache.sysds.lops.*;
 import org.apache.sysds.common.Types.ExecType;
-import org.apache.sysds.lops.LopsException;
-import org.apache.sysds.lops.Sql;
```
j143, line 40: I'll remove *

mboehm7: yes exactly - let's try to avoid the wildcard imports (to avoid ambiguity).

src/main/java/org/apache/sysds/hops/DataOp.java
```diff
@@ -60,6 +56,8 @@ public class DataOp extends Hop {
 	
 	private boolean _recompileRead = true;
 
+	private boolean _isTeeOp = false;
```
mboehm7, line 59: I would rather make it a op-type of DataOp like TWRITE.

src/main/java/org/apache/sysds/hops/DataOp.java
```diff
@@ -251,56 +253,66 @@ public Lop constructLops()
...
+					throw new LopsException("Invalid operation type for Data LOP: " + _op);
+			}
+			setLineNumbers(l);
+			setLops(l);
 		}
-		
-		setLineNumbers(l);
-		setLops(l);
+//		setLineNumbers(l);
+//		setLops(l);
```
mboehm7, line 315: the line numbers and lops should also be set for the TeeOp

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
mboehm7: no worries, these hop names are just for debugging - while generating instructions, we generate the _mvar, _fvar, _var accordingly.

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -0,0 +1,257 @@
...
+			return false;
+		}
+
+		boolean multipleConsumers = hop.getParent().size() > 1;
+		boolean isNotAlreadyTee = isNotAlreadyTee(hop);
+		boolean isOOCEnabled = DMLScript.USE_OOC;
+		boolean isTransposeMM = isTranposePattern(hop);
+		boolean isMatrix = hop.getDataType() == Types.DataType.MATRIX;
+
+		return isOOCEnabled && multipleConsumers && isNotAlreadyTee && isTransposeMM && isMatrix;
```
mboehm7, line 162: it's fine to currently only do it for transpose - it will not affect other things, and later we generalize the approach.

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	private boolean isTranposePattern (Hop hop) {
+		boolean hasTransposeConsumer = false; // t(X)
+		boolean hasMatrixMultiplyConsumer = false; // %*%
+
+		for (Hop parent: hop.getParent()) {
+			String opString = parent.getOpString();
+			if (parent instanceof ReorgOp) {
+				if (opString.contains("r'") || opString.contains("transpose")) {
```
mboehm7, line 246: please always use the following utils: `HopRewriteUtils.isReorgOp(parent, ReOrgOp.TRANS)`

src/main/java/org/apache/sysds/hops/rewrite/RewriteInjectOOCTee.java
```diff
@@ -0,0 +1,257 @@
...
+					hasTransposeConsumer = true;
+				}
+			}
+			else if (parent instanceof AggBinaryOp)
+				if (opString.contains("*") || opString.contains("ba+*")) {
```
mboehm7, line 251: same here as above.

src/main/java/org/apache/sysds/lops/Tee.java
```diff
@@ -0,0 +1,65 @@
...
+		return "Operation = Tee";
+	}
+
+	@Override
+	public String getInstructions(String input1, String outputs) {
+
+		String[] out = outputs.split(Lop.OPERAND_DELIMITOR);
+		String output2 = outputs + "_copy";
+
+		// This method generates the instruction string: OOC°tee°input°output1°output2...
```
mboehm7, line 55: I would recommend not making tee an operation with two outputs but just one. (copying the input). After thinking a bit more about it, maybe the entire notion of a tee was not a good idea. Potentially, we could make the stream resettable, and the tee operator takes a normal stream, and outputs a resettable stream which can then be consumed by arbitrary many other operations (which also solves the problem of complex control flow interactions where the multiple parents are not immediately visible)

j143: thanks @mboehm7 . Are you referring to the way, where we have lightweight marker for the stream so the multiple consumers could utilize it. Please let me raise a different PR with the option.

mboehm7: yes exactly - I'm thinking about something like `setResettable()` on the taskqueue, which means the blocks are never removed (and potentially subject to eviction) but different operations can get new iterables that always iterate the all blocks. The first operation would simply pull all blocks into memory, and the rmvar then drops all of them. If there is a single consumer we would not need to make it resettable, but can just stream them in.

PR #2327, opened 2025-09-15
Improve website contrast by making description text dark black

VirenS13117:
- Changed main content text color from #363f3f to #000000
- Improves readability and contrast for the SystemDS description text
- Addresses accessibility concerns with low contrast text


VirenS13117:
@j143 got it, thanks

PR #2329, opened 2025-09-23
Fix L2-SVM invalid label validation

VirenS13117:
Convert warning to error when Y contains invalid number of binary labels. This prevents the algorithm from continuing with malformed data which could lead to unpredictable training results.


Baunsgaard:
Unfortunately it is not so simple to resolve the TODO, there are other functions that currently are calling L2SVM and sometimes with incorrect arguments. To resolve the support I think we need to introduce error handling in SystemDS to have something like try catch semantics in the runtime.

PR #2330, opened 2025-09-23
Fix autoencoder bias gradient updates

VirenS13117:
Use bias gradients (b1_grad, b2_grad, etc.) instead of bias values (b1, b2, etc.) in momentum updates. This critical fix ensures proper backpropagation and training convergence in the 2-layer autoencoder.


Baunsgaard:
It is definitely a bug, and a nice catch. Will merge once I verify the GitHub actions.

VirenS13117:
@j143 well it's a logical bug as this was taking bias values instead of gradients.

PR #2336, opened 2025-10-08
[MINOR] Move Package Dependencies for Python Testing 

christinadionysio:
This commit divides the installation of packages into a part that installs all generally needed packages and one part that is only installed for specific test runs like Scuro. Additionally, torch hub get's cached which reduces setup time.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -31,7 +31,7 @@ on:
       - 'src/assembly/**'
       - 'dev/**'
     branches:
-      - main
+      - pkg_dependencies
   pull_request:
     paths-ignore:
       - 'docs/**'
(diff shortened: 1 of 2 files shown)


Baunsgaard:
it would be good to additionally cache the datasets you use, see: 

```
    - name: Cache Datasets
      uses: actions/cache@v4
      with:
        path: |
          src/main/python/systemds/examples/tutorials/mnist
          src/main/python/systemds/examples/tutorials/adult/data.zip
        key: ${{ runner.os }}-mnist-${{ hashFiles('src/main/python/systemds/examples/tutorials/mnist.py') }}-${{ hashFiles('src/main/python/systemds/examples/tutorials/adult.py') }}

```

PR #2399, opened 2026-01-14
[SYSTEMDS-3867] Safe Fast Frame Copy

yj-40-56:
Student project for DIA at TU Berlin: improving copy()

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/frame/data/FrameBlock.java
+++ b/src/main/java/org/apache/sysds/runtime/frame/data/FrameBlock.java
@@ -154,7 +154,7 @@ public class FrameBlock implements CacheBlock<FrameBlock>, Externalizable {
 
 	/**
 	 * FrameBlock constructor with constant
-	 * 
+	 *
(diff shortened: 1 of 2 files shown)


Baunsgaard:
Just to pitch in, the implementations are indeed not thread safe for some specific variations of calling. However, if the caller does not have overlapping ranges that it sets, it is perfectly safe for OptionalArrays. 

The BitSetArray was intentionally fixed by introducing the lock (as the @yj-40-56  is pointing out), however ideally we should (for good performance) avoid using the lock by aligning updates to the nearest 64 values, to allow full parallelism.

The RaggedArray and the DDCArray/ACompressedArray are both not designed to be frequently updated, so should not be considered a valid copy target.

A better design than the proposed would be to specialize each of the fundamental arrays to take from the Source arrays. We have primitives like:

```
	public abstract void setFromOtherType(int rl, int ru, Array<?> value);
```

That allows very efficient cloning from another array into the current one. This should just be called in parallel across different input columns.

PR #2435, opened 2026-02-18
[MINOR] Clean up of isDebugEnabled() usages

e-strauss:
This PR is a followup to #2432 .
We have 131 Usages of LOG.isDebugEnabled() 
```bash
grep -rnF "LOG.isDebugEnabled" src/main/java/org/apache/sysds | wc -l
131
```
This PR adds a new utils for Parameterized Logging
```java
LOG.debug("Begin Function {%15d}", fkey);
// instead of
if(LOG.isDebugEnabled){
    LOG.debug(String.format("Begin Function %15d", fkey))
}
```

This helps with readability and simplifies code coverage

### DIFF
--- a/src/main/java/org/apache/sysds/hops/DataOp.java
+++ b/src/main/java/org/apache/sysds/hops/DataOp.java
@@ -19,13 +19,14 @@
 
 package org.apache.sysds.hops;
 
+import static org.apache.sysds.parser.DataExpression.FED_RANGES;
+
 import java.util.HashMap;
 import java.util.Map.Entry;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.apache.sysds.api.DMLScript;
 import org.apache.sysds.common.Types.DataType;
+import org.apache.sysds.common.Types.ExecType;
 import org.apache.sysds.common.Types.FileFormat;
 import org.apache.sysds.common.Types.OpOpData;
 import org.apache.sysds.common.Types.ValueType;
@@ -35,22 +36,21 @@ import org.apache.sysds.hops.rewrite.HopRewriteUtils;
 import org.apache.sysds.lops.Data;
 import org.apache.sysds.lops.Federated;
 import org.apache.sysds.lops.Lop;
-import org.apache.sysds.common.Types.ExecType;
 import org.apache.sysds.lops.LopsException;
 import org.apache.sysds.lops.Sql;
 import org.apache.sysds.lops.Tee;
 import org.apache.sysds.parser.DataExpression;
-import static org.apache.sysds.parser.DataExpression.FED_RANGES;
 import org.apache.sysds.runtime.controlprogram.caching.MatrixObject.UpdateType;
 import org.apache.sysds.runtime.meta.DataCharacteristics;
 import org.apache.sysds.runtime.util.LocalFileUtils;
+import org.apache.sysds.utils.ParameterizedLogger;
 
 /**
  *  A DataOp can be either a persistent read/write or transient read/write - writes will always have at least one input,
  *  but all types can have parameters (e.g., for csv literals of delimiter, header, etc).
  */
 public class DataOp extends Hop {
-	private static final Log LOG =  LogFactory.getLog(DataOp.class.getName());
+	private static final ParameterizedLogger LOG = ParameterizedLogger.getLogger(DataOp.class);
 	private OpOpData _op;
 	private String _fileName = null;
 	
@@ -130,9 +130,7 @@ public class DataOp extends Hop {
 			String s = e.getKey();
 			Hop input = e.getValue();
 			getInput().add(input);
-			if (LOG.isDebugEnabled()){
-				LOG.debug(String.format("%15s - %s",s,input));
-			}
+			LOG.debug("{%15s} - {%s}", s, input);
 			input.getParent().add(this);
 
 			_paramIndexMap.put(s, index);
--- a/src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
+++ b/src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
@@ -19,11 +19,23 @@
 
 package org.apache.sysds.hops.codegen;
 
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map.Entry;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.SystemUtils;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.apache.sysds.api.DMLScript;
 import org.apache.sysds.common.Types.AggOp;
 import org.apache.sysds.common.Types.DataType;
@@ -97,24 +109,11 @@ import org.apache.sysds.runtime.lineage.LineageItemUtils;
 import org.apache.sysds.runtime.matrix.data.Pair;
 import org.apache.sysds.utils.Explain;
 import org.apache.sysds.utils.NativeHelper;
+import org.apache.sysds.utils.ParameterizedLogger;
 import org.apache.sysds.utils.stats.CodegenStatistics;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.Map.Entry;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
-
 public class SpoofCompiler {
-	private static final Log LOG = LogFactory.getLog(SpoofCompiler.class.getName());
+	private static final ParameterizedLogger LOG = ParameterizedLogger.getLogger(SpoofCompiler.class);
 
 	//internal configuration flags
 	public static CompilerType JAVA_COMPILER           = CompilerType.JANINO;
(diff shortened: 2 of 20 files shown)


src/main/java/org/apache/sysds/hops/cost/CostEstimator.java
```diff
@@ -155,18 +155,16 @@ else if( pb instanceof BasicProgramBlock )
 						//awareness of recursive functions, missing program
 						if( !memoFunc.contains(fkey) && pb.getProgram()!=null ) 
 						{
-							if(LOG.isDebugEnabled())
-								LOG.debug("Begin Function "+fkey);
-							
+							LOG.debug("Begin Function " + fkey);
```
Baunsgaard: The main difference here, is that the LOG.isDebugEnabled() is very cheap, while the string concatenation can be very expensive. Unfortunately to call the LOG.debug java has to resolve the string variable input, and that is the main time used, especially if Logging is disabled. 

I do not know if there is a way to avoid the string concatenation, or somehow getting it delayed to after the logging is analysed to be off.

e-strauss: I see your point, but isnt it in practice in micro-level noise for these simple concats? i think we might over-engineer here

Why are we sticking to apache commons logging? with the newer SLF4J we would could do stuff like this:
```java
LOG.debug("Begin Function {}", fkey);
```
also SLF4J might already be used inside apachae commons logging, since it's already in the pom

Baunsgaard: This would be cool, but we should not change the logging framework, so we just need to verify.

src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
```diff
@@ -19,11 +19,23 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map.Entry;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
```
Baunsgaard, line 34: can we avoid moving these imports around in this PR?

src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
```diff
@@ -538,22 +537,21 @@ public static ArrayList<Hop> optimize(ArrayList<Hop> roots, boolean recompile)
...
 					}
 
-					//explain debug output cplans or generated source code
-					if( LOG.isInfoEnabled() || DMLScript.EXPLAIN.isHopsType(recompile) ) {
-						LOG.info("Codegen EXPLAIN (generated cplan for HopID: " + cplan.getKey() + 
-							", line "+tmp.getValue().getBeginLine() + ", hash="+tmp.getValue().hashCode()+"):");
-						LOG.info(tmp.getValue().getClassname()
-							+ Explain.explainCPlan(cplan.getValue().getValue()));
+					// explain debug output cplans or generated source code
+					// (explicit guard: Explain.explainCPlan is expensive and evaluated eagerly)
```
Baunsgaard, line 541: remove this filling comment. 
It makes sense that we do not do the check for this logging.

src/main/java/org/apache/sysds/hops/codegen/SpoofCompiler.java
```diff
@@ -538,22 +537,21 @@ public static ArrayList<Hop> optimize(ArrayList<Hop> roots, boolean recompile)
...
 						LOG.info(CodegenUtils.printWithLineNumber(src));
-						
-						if(API == GeneratorAPI.CUDA) {
-							LOG.info("CUDA Codegen EXPLAIN (generated code for HopID: " + cplan.getKey() +
-									", line " + tmp.getValue().getBeginLine() + ", hash=" + tmp.getValue().hashCode() + "):");
 
+						if(API == GeneratorAPI.CUDA) {
+							LOG.info("CUDA Codegen EXPLAIN (generated code for HopID: {}, line {}, hash={}):",
+								cplan.getKey(), tmp.getValue().getBeginLine(), tmp.getValue().hashCode());
 							LOG.info(CodegenUtils.printWithLineNumber(src_cuda));
```
Baunsgaard, line 555: I think a small cleanup is to move the tmp.getValue outside in a variable. 

Can we do that as part of the cleanup ?

Maybe even the .hashCode().

src/main/java/org/apache/sysds/hops/DataOp.java
```diff
@@ -19,13 +19,14 @@
 
 package org.apache.sysds.hops;
 
+import static org.apache.sysds.parser.DataExpression.FED_RANGES;
+
 import java.util.HashMap;
```
Baunsgaard, line 24: Keep the import reordering of this file.
It clearly needed cleanup.

src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
```diff
@@ -905,15 +904,13 @@ public synchronized void exportData(String fName, String outputFormat, FileForma
...
-			LOG.trace("Export data "+hashCode()+" "+fName);
+		LOG.trace("Export data {} {}", hashCode(), fName);
 		long t0 = DMLScript.STATISTICS ? System.nanoTime() : 0;
 		//prevent concurrent modifications
 		if ( !isAvailableToRead() )
 			throw new DMLRuntimeException("MatrixObject not available to read.");
 
-		if( LOG.isTraceEnabled() )
-			LOG.trace("Exporting " + this.getDebugName() + " to " + fName + " in format " + outputFormat);
+		LOG.trace("Exporting {} to {} in format {}", this.getDebugName(), fName, outputFormat);
```
Baunsgaard, line 913: we need to doublecheck that this.getDebugName is not constructing a string, otherwise, we might want to instead introduce a toString() method for CacheableData instead. This would avoid the eager string construction of the caller.

src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
```diff
@@ -1043,8 +1040,7 @@ else if( getRDDHandle()!=null && getRDDHandle().isPending()
 		else 
 		{
 			//CASE 4: data already in hdfs (do nothing, no need for export)
-			if( LOG.isTraceEnabled() )
-				LOG.trace(this.getDebugName() + ": Skip export to hdfs since data already exists.");
+			LOG.trace("{}: Skip export to hdfs since data already exists.", this.getDebugName());
```
Baunsgaard, line 1043: same doublecheck the getDebugName

src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
```diff
@@ -1076,28 +1072,27 @@ protected boolean isBlobPresent() {
...
 		String cacheFilePathAndName = getCacheFilePathAndName();
-		long begin = LOG.isTraceEnabled() ? System.currentTimeMillis() : 0;
-		
-		if( LOG.isTraceEnabled() )
-			LOG.trace ("CACHE: Restoring matrix...  " + hashCode() + "  HDFS path: " + 
-						(_hdfsFileName == null ? "null" : _hdfsFileName) + ", Restore from path: " + cacheFilePathAndName);
-				
-		if (_data != null)
-			throw new DMLRuntimeException(cacheFilePathAndName + " : Cannot restore on top of existing in-memory data.");
+		long begin = LOG.currentTimeMillisIfTraceEnabled();
```
Baunsgaard, line 1075: I feel like adding this currentTImeMillis is a bit overkill. but fine we can keep it.

src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
```diff
@@ -1076,28 +1072,27 @@ protected boolean isBlobPresent() {
...
-			throw new DMLRuntimeException (cacheFilePathAndName + " : Restore failed.");
-		
-		if( LOG.isTraceEnabled() )
-			LOG.trace("Restoring matrix - COMPLETED ... " + (System.currentTimeMillis()-begin) + " msec.");
+
+		// check for success
+		if(_data == null)
+			throw new DMLRuntimeException(cacheFilePathAndName + " : Restore failed.");
+
+		LOG.trace("Restoring matrix - COMPLETED ... {} msec.", LOG.currentTimeMillisIfTraceEnabled() - begin);
```
Baunsgaard, line 1095: Here to be a bit pedantic, we could let the currentTimeMillisIfTraceEnabled method take an argument called offset. 


then the 

```
currentTimeMillisIfTraceEnabled(long off){
  if(trace)
    return System.currentTimeMillis - off;
}
```

src/main/java/org/apache/sysds/runtime/controlprogram/caching/CacheableData.java
```diff
@@ -1295,8 +1287,7 @@ protected void acquire (boolean isModify, boolean restore) {
 				throw new DMLRuntimeException("MODIFY-MODIFY not allowed.");
 		}
 
-		if( LOG.isTraceEnabled() )
-			LOG.trace("Acquired lock on " + getDebugName() + ", status: " + _cacheStatus.name() );
+		LOG.trace("Acquired lock on {} status: {}", getDebugName(), _cacheStatus.name());
```
Baunsgaard, line 1290: doublecheck .name is constant time and with no allocation.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -290,11 +289,8 @@ private static void printStatistics() {
 	}
 
 	private static void logRequests(FederatedRequest request, int nrRequest, int totalRequests) {
-		if(LOG.isDebugEnabled()) {
-			LOG.debug("Executing command " + (nrRequest + 1) + "/" + totalRequests + ": " + request.getType().name());
-			if(LOG.isTraceEnabled()) 
-				LOG.trace("full command: " + request.toString());
-		}
+		LOG.debug("Executing command {}/{}: {}", nrRequest + 1, totalRequests, request.getType().name());
```
Baunsgaard, line 292: Doublecheck method is constant time and do not allocate

src/main/java/org/apache/sysds/runtime/controlprogram/paramserv/FederatedPSControlThread.java
```diff
@@ -147,11 +146,9 @@ public void setup(double weightingFactor) {
 		if(_runtimeBalancing == PSRuntimeBalancing.BASELINE)
 			_cycleStartAt0 = true;
 
-		if( LOG.isInfoEnabled() ) {
-			LOG.info("Setup config for worker " + this.getWorkerName());
-			LOG.info("Batch size: " + _batchSize + " possible batches: " + _possibleBatchesPerLocalEpoch
-					+ " batches to run: " + _numBatchesPerEpoch + " weighting factor: " + _weightingFactor);
-		}
+		LOG.info("Setup config for worker {}", this.getWorkerName());
```
Baunsgaard, line 149: doublecheck it does not allocate, otherwise make a case class, and have a overloaded toString()

src/main/java/org/apache/sysds/runtime/instructions/cp/CompressionCPInstruction.java
```diff
@@ -197,8 +196,7 @@ private void processMatrixBlockCompression(ExecutionContext ec, MatrixBlock in,
 		if(LOG.isTraceEnabled())
 			LOG.trace(compResult.getRight());
 		MatrixBlock out = compResult.getLeft();
-		if(LOG.isInfoEnabled())
-			LOG.info("Compression output class: " + out.getClass().getSimpleName());
+		LOG.info("Compression output class: {}", out.getClass().getSimpleName());
```
Baunsgaard, line 199: this allocates a string, case class again ?

src/main/java/org/apache/sysds/runtime/instructions/cp/ParamservBuiltinCPInstruction.java
```diff
@@ -611,12 +608,11 @@ private void partitionLocally(PSScheme scheme, ExecutionContext ec, List<LocalPS
...
-		if (pfs.size() < workers.size()) {
-			if (LOG.isWarnEnabled()) {
-				LOG.warn(String.format("There is only %d batches of data but has %d workers. "
-					+ "Hence, reset the number of workers with %d.", pfs.size(), workers.size(), pfs.size()));
-			}
-			if (getUpdateType().isSBP() && pfs.size() <= getNumBackupWorkers()) {
+		if(pfs.size() < workers.size()) {
+			LOG.warn(
+				"There is only {} batches of data but has {} workers. " + "Hence, reset the number of workers with {}.",
+				pfs.size(), workers.size(), pfs.size());
```
Baunsgaard, line 614: Since some of them are reused from the if statement, it could maybe be cleaner to make variables?

src/main/java/org/apache/sysds/utils/ParameterizedLogger.java
```diff
@@ -0,0 +1,170 @@
...
+		_log.error(message, t);
+	}
+
+	/**
+	 * Substitutes {@code {}} and {@code {%fmt}} placeholders in {@code pattern} with {@code args}. Non-empty braces
+	 * that aren't a {@code String.format} spec (e.g. {@code {n}}) are left intact, and any unconsumed args are appended
+	 * as {@code  [arg]} — keep call sites strict if accidental drift matters. Invalid {@code {%fmt}} specs fall back to
+	 * {@code String.valueOf(arg)} so a log call never throws.
+	 */
+	public static String format(String pattern, Object... args) {
```
Baunsgaard, line 120: Is there not a Java String formatter ? and is it not as fast as this ?

Baunsgaard (commented):
Looks pretty good, I am not accepting yet, because there are some of the call instances that does function calls that counteract the benefit of this PR, we should clean this up somehow, perhaps Case classes ? or Overloaded toString methods?

I also think there is some followup to reduce logging statements that are right next to each other 

e.g.
 
```
LOG.debug('a');
LOG.debug('b');
LOG.debug('c');
```

should probably be changed to :

```
LOG.debug('a\nb\nc');
```

But we should do that in another followup PR.

PR #2460, opened 2026-04-22
[SYSTEMDS-3941] Add new rewrites

smelihportakal:
This PR implements the rewrites requested in [SYSTEMDS-3941](https://issues.apache.org/jira/browse/SYSTEMDS-3941).

Changes:
- Add a static rewrite for `rowSums(a*A) => a*rowSums(A)`
- Add a static rewrite for `colSums(a*A) => a*colSums(A)`
- Add a dynamic rewrite for `sum(matrix(a, rows=b, cols=c)) => a*b*c`
- Add tests for all new rewrites under `src/test/java/org/apache/sysds/test/functions/rewrite`
- Adjust `RewriteFusedRandTest` to avoid interference from the new `rowSums` rewrite 

Validation:
- Added rowSums rewrite tests with rewrites enabled and disabled
- Added colSums rewrite tests with rewrites enabled and disabled
- Added constant matrix sum rewrite tests with rewrites enabled and disabled
- Updated `RewriteFusedRandTest` so it continues to test the intended fused random rewrite behavior

### DIFF
--- a/src/main/java/org/apache/sysds/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
+++ b/src/main/java/org/apache/sysds/hops/rewrite/RewriteAlgebraicSimplificationDynamic.java
@@ -184,6 +184,7 @@ public class RewriteAlgebraicSimplificationDynamic extends HopRewriteRule
 			hi = simplifySumDiagToTrace(hi);                  //e.g., sum(diag(X)) -> trace(X); if col vector
 			hi = simplifyLowerTriExtraction(hop, hi, i);      //e.g., X * cumsum(diag(matrix(1,nrow(X),1))) -> lower.tri
 			hi = simplifyConstantCumsum(hop, hi, i);          //e.g., cumsum(matrix(1/n,n,1)) -> seq(1/n, 1, 1/n)
+			hi = simplifySumConstantMatrix(hop, hi, i);       //e.g., sum(matrix(a,rows=b,cols=c)) -> a*b*c
 			hi = pushdownBinaryOperationOnDiag(hop, hi, i);   //e.g., diag(X)*7 -> diag(X*7); if col vector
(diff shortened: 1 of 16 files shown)


janniklinde:
Thanks for the additional rewrite implementations and corresponding tests @smelihportakal. In your tests, could you validate that your rewrites are actually applied as an additional assert? You could use the heavy hitters similar to `RewriteSimplifyNegatedSubtractionTest` (e.g., `Assert.assertEquals(1, Statistics.getCPHeavyHitterCount("-"));`).

For colsum/rowsum pushdown, you could test an alternative expression such as `2*colSums(3*X)`, where after your rewrite there should only be a single multiply present as a heavy hitter.

smelihportakal:
> Thanks for the additional rewrite implementations and corresponding tests @smelihportakal. In your tests, could you validate that your rewrites are actually applied as an additional assert? You could use the heavy hitters similar to `RewriteSimplifyNegatedSubtractionTest` (e.g., `Assert.assertEquals(1, Statistics.getCPHeavyHitterCount("-"));`).
> 
> For colsum/rowsum pushdown, you could test an alternative expression such as `2*colSums(3*X)`, where after your rewrite there should only be a single multiply present as a heavy hitter.

Implemented the requested rewrite-validation asserts in the tests.
For SimplifySumConstantMatrix I added a heavy-hitter check to ensure the constant matrix is not materialized when the rewrite is enabled:

```
Assert.assertFalse(heavyHittersContainsString("rand"));
```
For the row/col sum pushdown tests, I changed the left-scalar cases to use expressions of the form:

```
2 * rowSums(3 * X)
2 * colSums(3 * X)
```
and added heavy-hitter assertions to validate the rewrite effect. In the rewrite-enabled case, there is a single scalar multiply heavy hitter:
```
Assert.assertEquals(1, Statistics.getCPHeavyHitterCount("n*"));
```
and in the no-rewrite case, there are two elementwise multiply heavy hitters:
```
Assert.assertEquals(2, Statistics.getCPHeavyHitterCount("*"));
```

PR #2466, opened 2026-05-06
[SYSTEMDS-3948] Row-wise Sparsity Estimator

ywcb00:
Hi,
this PR adds the row-wise sparsity estimator from:
Lin, Chunxu, Wensheng Luo, Yixiang Fang, Chenhao Ma, Xilin Liu and Yuchi Ma;
On Efficient Large Sparse Matrix Chain Multiplication;
Proceedings of the ACM on Management of Data 2 (2024): 1 - 27.

Note that the row sparsity propagation, as described in the publication, applies to MM chains only. Other operations use fallback methods for sparsity estimation w/ row sparsity vectors, which create a cut in the sparsity estimation DAG.

### DIFF
--- a/src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
+++ b/src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
@@ -72,11 +72,11 @@ public class EstimatorRowWise extends SparsityEstimator {
 		return DoubleStream.of(rsOut).average().orElse(0);
 	}
 
-	private void estimInternChain(MMNode node) {
-		estimInternChain(node, null, null);
+	private double[] estimInternChain(MMNode node) {
+		return estimInternChain(node, null, null);
 	}
 
-	private void estimInternChain(MMNode node, double[] rsRightNeighbor, OpCode opRightNeighbor) {
+	private double[] estimInternChain(MMNode node, double[] rsRightNeighbor, OpCode opRightNeighbor) {
 		double[] rsOut;
 		if(node.isLeaf()) {
 			MatrixBlock mb = node.getData();
@@ -86,75 +86,80 @@ public class EstimatorRowWise extends SparsityEstimator {
 				rsOut = getRowWiseSparsityVector(mb);
 		}
 		else {
+			MMNode nodeLeft = node.getLeft();
+			MMNode nodeRight = node.getRight();
 			switch(node.getOp()) {
 				case MM:
-					estimInternChain(node.getRight(), rsRightNeighbor, opRightNeighbor);
-					estimInternChain(node.getLeft(), (double[])(node.getRight().getSynopsis()), node.getOp());
-					rsOut = (double[])node.getLeft().getSynopsis();
+					double[] rsRightMM = estimInternChain(nodeRight, rsRightNeighbor, opRightNeighbor);
+					rsOut = estimInternChain(nodeLeft, rsRightMM, node.getOp());
 					break;
 				case CBIND:
-					/** NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
+					/**
+					 * NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
 					 * the right neighbor cannot be aggregated into a cbind operation when having only row sparsity vectors
 					 */
-					estimInternChain(node.getLeft());
-					estimInternChain(node.getRight());
-					double[] rsCBind = estimInternCBind((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					double[] rsLeftCBind = estimInternChain(nodeLeft);
+					double[] rsRightCBind = estimInternChain(nodeRight);
+					double[] rsCBind = estimInternCBind(rsLeftCBind, rsRightCBind);
 					if(rsRightNeighbor != null) {
-						rsOut = (double[])estimInternMMFallback(rsCBind, rsRightNeighbor);
+						rsOut = estimInternMMFallback(rsCBind, rsRightNeighbor);
 						if(opRightNeighbor != OpCode.MM)
 							throw new NotImplementedException("Fallback sparsity estimation has only been " +
 								"considered for MM operation w/ right neighbor yet.");
 					}
 					else
-						rsOut = (double[])rsCBind;
+						rsOut = rsCBind;
 					break;
 				case RBIND:
-					/** NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
+					/**
+					 * NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
 					 * the right neighbor cannot be aggregated into an rbind operation when having only row sparsity vectors
 					 */
-					estimInternChain(node.getLeft());
-					estimInternChain(node.getRight());
-					double[] rsRBind = estimInternRBind((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					double[] rsLeftRBind = estimInternChain(nodeLeft);
+					double[] rsRightRBind = estimInternChain(nodeRight);
+					double[] rsRBind = estimInternRBind(rsLeftRBind, rsRightRBind);
 					if(rsRightNeighbor != null) {
-						rsOut = (double[])estimInternMMFallback(rsRBind, rsRightNeighbor);
+						rsOut = estimInternMMFallback(rsRBind, rsRightNeighbor);
 						if(opRightNeighbor != OpCode.MM)
 							throw new NotImplementedException("Fallback sparsity estimation has only been " +
 								"considered for MM operation w/ right neighbor yet.");
 					}
 					else
-						rsOut = (double[])rsRBind;
+						rsOut = rsRBind;
 					break;
 				case PLUS:
-					/** NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
+					/**
+					 * NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
 					 * the right neighbor cannot be aggregated into an element-wise operation when having only row sparsity vectors
 					 */
-					estimInternChain(node.getLeft());
-					estimInternChain(node.getRight());
-					double[] rsPlus = estimInternPlus((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					double[] rsLeftPlus = estimInternChain(nodeLeft);
+					double[] rsRightPlus = estimInternChain(nodeRight);
+					double[] rsPlus = estimInternPlus(rsLeftPlus, rsRightPlus);
 					if(rsRightNeighbor != null) {
-						rsOut = (double[])estimInternMMFallback(rsPlus, rsRightNeighbor);
+						rsOut = estimInternMMFallback(rsPlus, rsRightNeighbor);
 						if(opRightNeighbor != OpCode.MM)
 							throw new NotImplementedException("Fallback sparsity estimation has only been " +
 								"considered for MM operation w/ right neighbor yet.");
 					}
 					else
-						rsOut = (double[])rsPlus;
+						rsOut = rsPlus;
 					break;
 				case MULT:
-					/** NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
+					/**
+					 * NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
 					 * the right neighbor cannot be aggregated into an element-wise operation when having only row sparsity vectors
 					 */
-					estimInternChain(node.getLeft());
-					estimInternChain(node.getRight());
-					double[] rsMult = estimInternMult((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					double[] rsLeftMult = estimInternChain(nodeLeft);
+					double[] rsRightMult = estimInternChain(nodeRight);
+					double[] rsMult = estimInternMult(rsLeftMult, rsRightMult);
 					if(rsRightNeighbor != null) {
-						rsOut = (double[])estimInternMMFallback(rsMult, rsRightNeighbor);
+						rsOut = estimInternMMFallback(rsMult, rsRightNeighbor);
 						if(opRightNeighbor != OpCode.MM)
 							throw new NotImplementedException("Fallback sparsity estimation has only been " +
 								"considered for MM operation w/ right neighbor yet.");
 					}
 					else
-						rsOut = (double[])rsMult;
+						rsOut = rsMult;
 					break;
 				default:
 					throw new NotImplementedException("Chain estimation for operator " + node.getOp().toString() +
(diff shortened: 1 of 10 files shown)


src/test/java/org/apache/sysds/test/component/estim/OpElemWTest.java
```diff
@@ -128,7 +129,18 @@ public void testSampleMult() {
 	public void testSamplePlus() {
 		runSparsityEstimateTest(new EstimatorSample(), m, n, sparsity, plus);
 	}
-	
+
+	// Row Wise Sparsity Estimator
```
Baunsgaard, line 152: You can optionally have all of these classes extend another class where you add these tests. Instead of copying the code on all of them.

src/test/java/org/apache/sysds/test/component/estim/OpSingleTest.java
```diff
@@ -40,7 +41,7 @@ public class OpSingleTest extends AutomatedTestBase
 	private final static int m = 600;
 	private final static int k = 300;
 	private final static double sparsity = 0.2;
-//	private final static OpCode eqzero = OpCode.EQZERO;
+	// private final static OpCode eqzero = OpCode.EQZERO;
```
Baunsgaard: Suggestion remove this edit, or do a followup PR that cleans it up.

src/test/java/org/apache/sysds/test/component/estim/OpSingleTest.java
```diff
@@ -237,7 +238,33 @@ public void testLGCasetrans() {
...
+	// Row Wise Sparsity Estimator
+	// @Test
+	// public void testRowWiseEqzero() {
+	// 	runSparsityEstimateTest(new EstimatorRowWise(), m, k, sparsity, eqzero);
+	// }
+
+	// @Test
+	// public void testRowWiseDiag() {
+	// 	runSparsityEstimateTest(new EstimatorRowWise(), m, m, sparsity, diag);
+	// }
```
Baunsgaard: why are these removed?

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,350 @@
...
+		MMNode nodeRight = node.getRight();
+		switch(node.getOp()) {
+			case MM:
+				return new MatrixCharacteristics(nodeLeft.getRows(), nodeRight.getCols(),
+					OptimizerUtils.getNnz(nodeLeft.getRows(), nodeRight.getCols(), spOut));
+			case MULT:
+			case PLUS:
+			case NEQZERO:
+			case EQZERO:
+				return new MatrixCharacteristics(nodeLeft.getRows(), nodeLeft.getCols(),
```
Baunsgaard: consider calling the getters above the switch case, to avoid all these repeated calls.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,350 @@
...
+			case TRANS:
+			case RESHAPE:
+				throw new NotImplementedException("Characteristics derivation for trans and reshape has not been " +
+					"implemented yet, but could be implemented similar to EstimatorMatrixHistogram.java");
+			default:
+				throw new NotImplementedException();
+		}
+	}
+
+	public static class RSVector {
```
Baunsgaard: Wrapping the double[] primitive is not really good practice. This adds an indication that the JIT compiler have to resolve, and slows you down. I would suggest not adding this to the code. and instead on the individual functions code up the solution without this functional indirection.

Baunsgaard (commented):
Some comments. hope it helps.

src/test/java/org/apache/sysds/test/component/estim/OpSingleTest.java
ywcb00: Done. Thanks for spotting this.

src/test/java/org/apache/sysds/test/component/estim/OpElemWTest.java
ywcb00: Thank you for the comment. I do not fully understand how you would consolidate them. Could you please clarify?

Baunsgaard: What I ment was something like :

```
OpElemWTest extends OpElemBaseTest {

}
```

and then you define a new OpElemBaseTest class that contains the variables all the different OpElem  test cases share, and define in that base test the fundamental tests that all of the OpElem should run. 

I hope this would work nicely, since it makes it easy to make the same test flavour run in all of the variations.

ywcb00: Thank you again for the explanation, I gave it another thought. While I also like such a hierarchical structure for tests in general, I cannot see the clear advantage here--given that I understood your suggestion correctly.
There are five common variables among the test cases `m, n, sparsity, mult, plus`, which would then be defined inside OPElemBaseTest class instead of OPElemWTest class. I would not consolidate tests for multiplication and addition (neither for different estimators) in one single function, as this would break the typical structure in which tests are defined (right?). Hence, I think such a base class would just add another class file, or have I misunderstood something?

Baunsgaard: I still think the restructure would be advantageous. Since many of these tests have the same fundamental structure, and are not very well tested with only one matrix size. Alternatively you could change the entire class to be a parameterized test instead. That would also help you make more tests easily. 

e.g.

```

@RunWith(value = Parameterized.class)
public class OpElm {
  public OpElm(Estimator e, int m, int n, double s, BinaryOp BOp){
   ...
  }


  @Parameters
  public static Collection<Object[]> data() {
  ...


  }


}

```

See for instance `src/test/java/org/apache/sysds/test/component/compress/CompressedTestBase.java`.


To make the tests faster as well you can set as input the actual matrix, and reuse the same matrix across many tests. The current structure create the matrix for each test, and that is by far the most expensive part of these tests.

ywcb00: Thank you for your patience. I really like the suggestion of parameterized tests, and changed it accordingly. :)

I also tried to reuse the input matrices instead of generating them from scratch every time. However, as it did not result in a speedup of the tests, I removed it again.

ywcb00:
Thank you for the comments @Baunsgaard . I have updated the PR accordingly. :)

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+					estimInternChain(node.getRight());
+					double[] rsCBind = estimInternCBind((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					if(rsRightNeighbor != null) {
+						rsOut = (double[])estimInternMMFallback(rsCBind, rsRightNeighbor);
+						if(opRightNeighbor != OpCode.MM)
+							throw new NotImplementedException("Fallback sparsity estimation has only been " +
+								"considered for MM operation w/ right neighbor yet.");
+					}
+					else
+						rsOut = (double[])rsCBind;
```
Baunsgaard: no need to cast to double[]

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+				case MM:
+					estimInternChain(node.getRight(), rsRightNeighbor, opRightNeighbor);
+					estimInternChain(node.getLeft(), (double[])(node.getRight().getSynopsis()), node.getOp());
+					rsOut = (double[])node.getLeft().getSynopsis();
+					break;
+				case CBIND:
+					/** NOTE: considering the current node as new DAG for estimation (cut), since the row sparsity of
+					 * the right neighbor cannot be aggregated into a cbind operation when having only row sparsity vectors
+					 */
+					estimInternChain(node.getLeft());
```
Baunsgaard: Move these .getLeft and .getRight to variables above the switch statement.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+					estimInternChain(node.getRight());
+					double[] rsCBind = estimInternCBind((double[])(node.getLeft().getSynopsis()), (double[])(node.getRight().getSynopsis()));
+					if(rsRightNeighbor != null) {
+						rsOut = (double[])estimInternMMFallback(rsCBind, rsRightNeighbor);
```
Baunsgaard: remove cast

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+			.toArray();
+		return rsOut;
+	}
+
+	// NOTE: this is the best estimation possible when we only have the two row sparsity vectors
+	private double[] estimInternMMFallback(double[] rsM1, double[] rsM2) {
+		// NOTE: Considering the average would probably not be far off while saving computing time
+		// double avgRsM2 = DoubleStream.of(rsM2).average().orElse(0);
+		// double[] rsOut = DoubleStream.of(rsM1).map(
+		// 	rsM1I -> (double) 1 - Math.pow((double) 1 - (rsM1I * avgRsM2), rsM2.length)).toArray();
```
Baunsgaard: Consider, formatting comments as docstrings above the method, or removing them.

Do not leave commented code.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+		double[] rsOut = DoubleStream.of(rsM1).map(
```
Baunsgaard: Im pretty sure the double stream here is slower than a normal explicit for loop.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,325 @@
...
+			rsM1I -> (double) 1 - DoubleStream.of(rsM2).reduce((double) 1,
+				(currentVal, rsM2J) -> currentVal * ((double) 1 - (rsM1I * rsM2J)))).toArray();
+		return rsOut;
+	}
+
+	private double[] estimInternCBind(double[] rsM1, double[] rsM2) {
+		// FIXME: this assumes that the number of columns is equivalent for both inputs
+		return IntStream.range(0, rsM1.length).mapToDouble(
+			idx -> (rsM1[idx] + rsM2[idx]) / (double) 2).toArray();
```
Baunsgaard: try to time this function and compare it to:

```
double[] ret = new double[rsM1.length];

for(int i = 0; i < rsM1.length; i++){
    ret[i] = (rsM1[i] + rsM2[i]) / 2.0;
}

return ret;
```

Baunsgaard (commented):
A few comments

ywcb00:
Thank you for the valuable comments. :)

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,349 @@
...
+				currentVal *= (double) 1 - rsM2[cIdx];
+			}
+			rsOut[rIdx] = 1 - currentVal;
+		}
+		return rsOut;
+	}
+
+	/**
+	 * NOTE: this is the best estimation possible when we only have the two row sparsity vectors
+	 * NOTE: Considering the average of the second matrix would probably not be far off while saving computing time
```
Baunsgaard: this is a wrong statement. 

The estimator here is the uniform estimator ( also called average-case or in fancy terms Naive Bayes estimator )

It is not the 'best', if that even exists, and it is an interesting point to consider if you are going in this direction for your research. 

a clear way to see it is not best or worst case is:

```
rsM1 = [1,0,0] and rsM2 = [1/3,1/3,1/3]
```

then:

```
rsOut[0] = 1 - (1 - 1·1/3)^3
         = 1 - (2/3)^3
         = 1 - 8/27
         = 19/27 ≈ 0.704
```

which is impossible to get in practice. 

for these values the matrices could be :


```
M1 = [1 1 1]      M2 = [1 0 0]
     [0 0 0]           [0 1 0]
     [0 0 0]           [0 0 1]

C = M1·M2 = [1 1 1]   →  row-0 actual density = 1.0
            [0 0 0]
            [0 0 0]
```
formular says 0.704 while the matrix actually is 1.0 


another fun one is :

```
M1 = [1 1 1]      M2 = [1 0 0]
     [0 0 0]           [1 0 0]
     [0 0 0]           [1 0 0]

C = M1·M2 = [3 0 0]   →  row-0 actual density = 1/3 ≈ 0.333
            [0 0 0]
            [0 0 0]
```

it ends up as 0.333 density, but the estimator says 0.7 again.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,349 @@
...
+	 */
+	private double[] estimInternMMFallback(double[] rsM1, double[] rsM2) {
+		double[] rsOut = new double[rsM1.length];
+		for(int i = 0; i < rsM1.length; i++) {
+			double currentVal = 1;
+			for(int j = 0; j < rsM2.length; j++) {
+				currentVal *= (double) 1 - (rsM1[i] * rsM2[j]);
```
Baunsgaard: consider on code like this to do :

```
currentVal *= 1.0 - (rsM1[i] * rsM2[j]);
```

you do not need to write the cast.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
```diff
@@ -0,0 +1,349 @@
...
+			rsOut[rIdx] = 1 - currentVal;
+		}
+		return rsOut;
+	}
+
+	/**
+	 * NOTE: this is the best estimation possible when we only have the two row sparsity vectors
+	 * NOTE: Considering the average of the second matrix would probably not be far off while saving computing time
+	 */
+	private double[] estimInternMMFallback(double[] rsM1, double[] rsM2) {
```
Baunsgaard, line 225: similar to the cbind this depends on number of rows/cols in the input, right ?

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
Baunsgaard: also consider early out, if you encounter zero, do not enter the inner loop 

```
rsM1[i] == 0;
```

Baunsgaard (approved):
PR LGTM, there are only micro optimizations to do now.

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
ywcb00: Thank you very much for this detailed explanation. :)

src/main/java/org/apache/sysds/hops/estim/EstimatorRowWise.java
ywcb00: The size of the output depends on the size of the first input.
However, it is independent of the size of the second input and does not take the number of columns into account, as it is uses only row sparsity vectors.

Baunsgaard: okay makes sense.

PR #2481, opened 2026-06-08
[BWARE] Track more compressed-friendly ops in FederatedWorkloadAnalyzer

Baunsgaard:
Extends the federated workload counter so that compression decisions account for additional instruction shapes beyond AggregateBinary.

- Pass the right-hand column count to incOverlappingDecompressions so the cost model reflects the actual decompression size rather than counting a single column
- Count MMChainCPInstruction as one LMM and one RMM contribution per invocation
- Count AggregateUnaryCPInstruction: when reducing columns with a sum/mean operator, treat it as a dict-op (compression-friendly); otherwise count it as a decompression
- Minor formatting cleanup in compressRun


Baunsgaard:
Hey @christinadionysio  , 

I have seen this error before for the CI of Scuro, where the machine runs out of disk space, maybe you want to take a look at the logs?

PR #2488, opened 2026-06-11
[MINOR][CI] Fix leaked threads hanging Java test forks

Baunsgaard:
## Prevent leaked threads from hanging Java test forks in CI

Several Java test suites (most visibly `**.component.c**` and `data.misc/lineage`) intermittently ran until the GitHub Actions job timeout even though the tests themselves had completed. The cause was **leaked non-daemon threads keeping the surefire fork JVM alive**, so the fork never exited and the job stalled until cancelled. There were two sources: in-JVM federated workers (`FederatedWorker`'s Netty event loops and the test-side worker wrapper threads were non-daemon), and `CommonThreadPool`'s fallback pools — when called off the main thread, it returned `Executors.newFixedThreadPool`/`newCachedThreadPool`, which default to non-daemon threads, while only the ForkJoinPool-backed variants were already daemon.

This PR makes those threads daemon at the source: `FederatedWorker` now creates its Netty event-loop groups with a daemon thread factory, and `CommonThreadPool` routes its fixed/cached fallbacks through one too, so daemon behavior is uniform across all pool variants. On the test side, `AutomatedTestBase` marks spawned worker threads as daemon, `TestUtils.shutdownThread` bounds its `join` (30s, warns on stragglers, restores the interrupt flag), and the lineage tests (`LineageFedReuseAlg`, `FedFullReuseTest`, `FedUDFReuseTest`) now shut workers down in a `finally` block so failures no longer leak workers (the large line counts there are just reindentation from the `try/finally` wrap). The `javaTests.yml` job cap stays at 30 minutes, with a comment documenting why it sits above the 600s per-fork surefire timeout, which remains the backstop for genuine hangs.

### DIFF
--- a/.github/workflows/javaTests.yml
+++ b/.github/workflows/javaTests.yml
@@ -50,7 +50,10 @@ concurrency:
 jobs:
   java_tests:
     runs-on: ubuntu-24.04
-    timeout-minutes: 30
+    # Job cap kept above the per-fork surefire timeout (test-forkedProcessTimeout,
+    # 600s) so surefire can kill hung forks and finish the remaining tests in a
+    # suite before GitHub Actions cancels the whole job.
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/javaTests.yml
+++ b/.github/workflows/javaTests.yml
@@ -51,9 +51,8 @@ jobs:
   java_tests:
     runs-on: ubuntu-24.04
     # Job cap kept above the per-fork surefire timeout (test-forkedProcessTimeout,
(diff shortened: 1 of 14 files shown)


Baunsgaard:
@janniklinde thanks to your points in the previous commit, I found the source of the issue for timeouts in many tests. It basically is because our parallel threads were not deamon threads. 

Would you like to review?

janniklinde (approved):
Looks good to me. Nice to finally fix those hanging actions.

One minor note: `shutdownThread(process)` still prints the stack trace on interrupt, while `shutdownThread(thread)` restores the interrupt status with `Thread.currentThread().interrupt()`.

PR #2489, opened 2026-06-15
[MINOR] OOC Fix Blocking Threads

janniklinde:
Blocking threads now use a dynamic thread pool.

### DIFF
--- a/src/main/java/org/apache/sysds/runtime/instructions/ooc/OOCInstruction.java
+++ b/src/main/java/org/apache/sysds/runtime/instructions/ooc/OOCInstruction.java
@@ -1196,7 +1196,7 @@ public abstract class OOCInstruction extends Instruction {
 	}
 
 	protected CompletableFuture<Void> submitOOCTask(Runnable r, StreamContext ctx) {
-		ExecutorService pool = CommonThreadPool.get();
+		ExecutorService pool = CommonThreadPool.getDynamicPool();
 		final CompletableFuture<Void> future = new CompletableFuture<>();
 		try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/sysds/runtime/instructions/ooc/OOCInstruction.java
```diff
@@ -1196,7 +1196,7 @@ protected <T> CompletableFuture<Void> submitOOCTasks(OOCStream<T> queue, Consume
 	}
 
 	protected CompletableFuture<Void> submitOOCTask(Runnable r, StreamContext ctx) {
-		ExecutorService pool = CommonThreadPool.get();
+		ExecutorService pool = CommonThreadPool.getDynamicPool();
```
Baunsgaard, line 1199: This should not make a difference, The CommonThreadPool.get() should be the right way to get our shared threadpool that does get killed with the Java process.

janniklinde: This should make a difference because `submitOOCTask` may be used for blocking work (e.g. `dequeue()`). If we now use a fixed sized thread pool (which especially causes problems in environments with fewer enabled CPUs thus smaller number of workers on the default thread pool), this may block scheduled tasks from making progress that are needed for those workers to unblock.

Baunsgaard: okay did not know you had more tasks than threads in the pool.

PR #2490, opened 2026-06-16
[MINOR] test use deterministic alphabetical test run order in surefire

Baunsgaard:
Set surefire runOrder to alphabetical so test classes execute in the same sequence on every machine.

This makes local reproduction of tests easier.


Baunsgaard:
Not needed, since it is not reproducing the error.

PR #2492, opened 2026-06-17
[MINOR][CI] Add hard guard that dumps stacks on stalled test forks

Baunsgaard:
The **.component.c**.** Java test job still has intermittently runs until the 30-minute GitHub Actions cap with no further output: a surefire fork stalls in a way that surefire's own timeouts never catch (a fork wedged around the booter handshake, or a starved maven parent), so neither forkedProcessTimeoutInSeconds nor forkedProcessExitTimeoutInSeconds fires and the job is cancelled with nothing to diagnose. The stall does not reproduce locally, so the only place to capture evidence is CI.

Add an outer guard in the docker test entrypoint that watches the test log for a stall (no new line for a window kept just above the 600s per-fork surefire timeout) and an absolute runtime ceiling below the job cap. On either trigger it force-dumps thread stacks from every JVM in the test process tree via SIGQUIT (relayed into the job log) plus a jstack file backup, then force-kills the tree so the job fails fast WITH stacks instead of being cancelled empty-handed. Limits are overridable via SYSDS_TEST_STALL_LIMIT and SYSDS_TEST_MAX_RUNTIME.

Also set surefire runOrder to alphabetical so the hang reproduces at a stable class boundary across runs, making the responsible class identifiable from the captured dumps.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -51,15 +51,94 @@ if [ "$compile_transient_failure" = true ]; then
 else
 	echo "No transient Maven repository error detected; no retry needed."
 fi
-mvn -ntp -B test -D maven.test.skip=false -D automatedtestbase.outputbuffering=true -D test=$1 2>&1 \
+# Outer guard: catch test-fork hangs that surefire's own timeouts miss, dump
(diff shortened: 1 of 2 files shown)


Baunsgaard:
Root course found, we close this PR.

PR #2500, opened 2026-06-22
[MINOR][CI] Extend hard guard that dumps stacks on stalled test forks

janniklinde:


Baunsgaard:
Great!, this gave us the error in my code. 

it is because of a deadlock in AOffset.java

I will make a PR to fix it!

Baunsgaard:
I think we can close this PR again? WDYT @janniklinde ?

PR #2511, opened 2026-06-25
[SYSTEMDS-3949] Add native Delta Lake matrix read/write via Delta Kernel

Baunsgaard:
Introduce a DELTA file format that reads and writes Delta Lake tables natively through the Spark-free Delta Kernel library, for matrices on the single-node CP path. DML read/write with format="delta" now operates directly on Delta tables without a Spark DataFrame round-trip.

- Add FileFormat.DELTA and exclude it from the text formats
- Accept format="delta" with unknown dimensions in the parser (like CSV) and set blocksize -1 for the columnar format
- Wire DELTA into the matrix reader and writer factories
- Add DeltaKernelUtils plus ReaderDelta and WriterDelta with column-at-a-time, boxing-free data transfer
- Refresh cached matrix metadata after a Delta read (discovered dimensions)
- Pin parquet 1.13.1 and jackson core/annotations to 2.15.2 to align with Delta Kernel's transitive requirements

Append/overwrite table semantics, distributed execution, frames, and time travel are out of scope.

### DIFF
--- a/.github/workflows/javaTests.yml
+++ b/.github/workflows/javaTests.yml
@@ -60,7 +60,7 @@ jobs:
           "org.apache.sysds.test.applications.**",
           "**.test.usertest.**",
           "**.component.c**.**",
-          "**.component.e**.**,**.component.f**.**,**.component.m**.**,**.component.o**.**",
+          "**.component.e**.**,**.component.f**.**,**.component.i**.**,**.component.m**.**,**.component.o**.**",
           "**.component.p**.**,**.component.r**.**,**.component.s**.**,**.component.t**.**,**.component.u**.**",
           "**.functions.a**.**,**.functions.binary.matrix.**,**.functions.binary.scalar.**,**.functions.binary.tensor.**",
           "**.functions.blocks.**,**.functions.data.rand.**,",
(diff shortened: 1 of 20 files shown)


Baunsgaard:
### Native Delta read performance vs binary reader

Benchmark (`org.apache.sysds.performance` `DeltaIO`), 32 vCPU, JDK 17, `-Xmx16g`,
20 timed repetitions after warmup, isolated (no competing load).

**Workload:** 5,000,000 × 16 dense FP64 matrix (~640 MB in memory, sparsity 1.0).

| Read path            | Time (ms)      | Disk throughput |
|----------------------|----------------|-----------------|
| Delta (serial)       | 5411 ± 79      | 79 MB/s         |
| **Delta (parallel)** | **875 ± 19**   | **490 MB/s**    |
| Binary (parallel)    | ~100–170       | 3.7–6.0 GB/s    |

**Takeaways**
- The parallel reader decodes data files concurrently and is **~6.2× faster than the
  serial Delta reader** (5411 ms → 875 ms).
- The binary reader is still **~5–9× faster** than parallel Delta. This is expected:
  binary is a near-raw `double[]` dump, while Delta Kernel 3.3.0 decodes parquet
  per-value (row-by-row converters, not vectorized). File-level parallelism is the
  main lever available, which is what this reader exploits.
- **On disk Delta is smaller:** 450 MB vs 645 MB for binary (0.70×), thanks to parquet
  encoding — so the read-time cost buys a ~30% smaller footprint.

For reference, the write path is encode-bound: Delta write ~7.0 s vs binary ~0.15 s
(parquet encoding dominates); not the focus of this PR but noted for completeness.

PR #2527, opened 2026-07-01
[CI] Add Java Format CI check for pull requests

Baunsgaard:
Add a workflow that enforces dev/CodeStyle_eclipse.xml on the Java files changed by a pull request. It applies the Eclipse formatter (via formatter-maven-plugin) to only the changed src/main/java and src/test/java files and fails if that produces any diff, printing the required changes.

Also add dev/format-changed.sh, a helper that runs the same scoped formatting locally on the files you changed relative to the base branch (plus staged/unstaged/untracked edits). This avoids a bare `mvn formatter:format`, which would reformat the entire tree.

The check is scoped to the PR diff rather than the whole tree because the existing sources are not yet fully formatter-clean; this lets the codebase converge file-by-file without failing unrelated PRs.


janniklinde:
I like this approach, but I'm wondering if it is too strict. It would imply that as a contributor, any slightly modified file must be completely reformatted by the PR. Second, different IDEs may apply slightly different formatting under the same xml so the only reliable way to reformat is using the exact same formatter.

Baunsgaard:
> I like this approach, but I'm wondering if it is too strict. It would imply that as a contributor, any slightly modified file must be completely reformatted by the PR. Second, different IDEs may apply slightly different formatting under the same xml so the only reliable way to reformat is using the exact same formatter.

Ya, the current solution is a bit too strict, i wanted to do something where we only fail if the there is a diff in conflict with the changes proposed, and then the CI job here would give a diff back that only edits those lines, not anything else in the file. 

I will try some more to make it streamlined.


About different IDEs, I think it should just be consistent and one source of truth, this seems Reliable with maven and this proposed solution.

Baunsgaard:
@janniklinde  want to take a look? now it is targeting the exact diff you have as subparts of the files you edited.

janniklinde:
LGTM @Baunsgaard (except of course the probe code) :)

We may treat other codestyle policies similarly in future (e.g., private variable naming etc.). The main blocker was this exact failing for unchanged lines of code

PR #2528, opened 2026-07-02
[SYSTEMDS-3929] Speed up Parquet frame reader/writer

Jakob-al28:
[SYSTEMDS-3929] Speed up Parquet frame reader/writer

The old reader created a `Group` object per row, boxing every decoded value into a wrapper object inside it, even though SystemDS frames are always flat. The parallel reader also converted every boxed value to a string and re-parsed it back to its target type in `FrameBlock.set`. This PR reads columns via parquet's internal column API (`ColumnReadStoreImpl`/`ColumnReader`) instead, avoiding the `Group` allocation.
On TPC-H lineitem (sf=5, ~30M rows, median of 7 runs on a cloud VM, n2-highmem-8, 40GB heap), with both readers on files written at the same codec/dictionary settings, read time went from 102.1s to 53.4s (1.91x faster). (An earlier version reported 2.6x, but that compared the readers on files written with different settings.)

matched_encoding_parquet

Also rewrites the writer to use a custom `WriteSupport` with tuned defaults, adds INT96 timestamp read support, and fixes row offsets for the parallel reader.

The old writer had a tunable batch of 1000 rows. The batch size was benchmarked looking for a better default, but `ParquetWriter` already batches internally by buffered byte size, making the manual row-count batch on top of it redundant, so the new writer drops it and instead exposes that internal buffer size as the tunable row-group size. Compression codec and dictionary-encoding were benchmarked as well: `ZSTD` and per-column dictionary encoding (`STRING_ONLY`) are used as defaults. Row group size was benchmarked too and kept at parquet's default of 128MB. Note the batch-size comparison runs uncompressed to match the legacy writer's codec, while the row-group benchmark uses the ZSTD default, so absolute times differ between the two plots.

tpch_batch_sizes
tpch_row_group_sizes
tpch_compression
tpch_encoding

TPC-H lineitem's comment column is ~84% unique, so dictionary encoding on it doesn't pay off, which is why `ALL_OFF` wins on this benchmark. Typical frame workloads are assumed to have low-cardinality string columns where dictionaries pay off, so `STRING_ONLY` is kept as the default.

Also adds public Parquet test files under `src/test/resources/datasets/parquet/` (duckdb, apache/parquet-testing, Titanic from HuggingFace) as a check against real-world files, and tests covering INT96 decoding, null handling and parallel/sequential equivalence.

Possible next step: row-group-level parallel reads within a single file.

Limitations: INT96 decoding truncates nanosecond precision to milliseconds, and there's no logical-type mapping for dates/timestamps since FrameBlock has no such types.


Baunsgaard:
Im curious how this compares to the Delta IO i have in https://github.com/apache/systemds/pull/2515 .

There the main bottleneck is decoding the Parquet files.

src/test/resources/datasets/parquet/all.parquet
Baunsgaard, line 1: If possible avoid adding parquet files to the system, 

instead rely on some known good Parquet writer to generate your initial inputs. e.g. Spark has a writer you can use for making inputs.

Jakob-al28: Parquet files have been deleted and instead, Spark is used to write the test files. Will take a look at how this compares to the Delta IO solution in #2515.

Jakob-al28: Looked into the decode bottleneck in #2515. Delta Kernel's default engine decodes through parquet-mr's row-record API (ParquetReader + ReadSupport/RecordMaterializer), which is built to reassemble nested records, each cell goes through a PrimitiveConverter callback. Frames are flat in SystemDS, so it doesn't apply here.

I built a thin Engine decorator on top of #2515 that overrides readParquetFiles: flat schemas decode through the same column API this PR uses (ColumnReadStoreImpl/ColumnReader) instead, everything else (predicates, nested/decimal/timestamp columns, unknown metadata columns) falls through to the wrapped default engine. 

On the same 30M row TPC-H lineitem (median of 7 runs, cloud VM): 103.9s to 55.2s (1.88x). On the DeltaFrameRead benchmark from #2515: 1.68x serial, 1.50x parallel (k=8).

delta_tpch_30m
deltaframeread

Two cases needed explicit handling: tables with deletion vectors enabled ask every read for a synthetic _metadata.row_index column that isn't actually in the file, and older files in a table that's had a column added later don't contain that column at all. Both would have failed otherwise; now the first is computed and the second comes back as null, covered by tests. There's also a config kill switch (sysds.io.delta.reader.columnapi, default on).

If you're interested, I can open this as a new PR now that #2515 is in, happy to do that whenever works.

Baunsgaard: If you want to make a PR that makes the Delta read and write faster as well that is great!

About the results, these are just the delta read and write. What would be interesting is comparing your raw parquet read and write functions (the PR you are making) and comparing it to the delta read and write, and as an extra bonus compare it to our internal binary format.

I know you have the results above, I just need you to combine them in a single table.

Jakob-al28: Opened #2535 with the combined table, all runs at the same codec/dictionary settings and heap size. Also corrected the read numbers in this PR's description (1.91x with matched settings).

src/main/java/org/apache/sysds/runtime/io/FrameReaderParquet.java
```diff
@@ -89,65 +97,147 @@ public FrameBlock readFrameFromHDFS(String fname, ValueType[] schema, String[] n
...
+						// See https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp
+						Binary binary = creader.getBinary();
+						ByteBuffer buf = ByteBuffer.wrap(binary.getBytes()).order(ByteOrder.LITTLE_ENDIAN);
+						long nanosOfDay = buf.getLong();
+						int julianDay = buf.getInt();
+						long millis = (julianDay - JULIAN_EPOCH_OFFSET_DAYS) * MILLIS_IN_DAY
+							+ nanosOfDay / NANOS_PER_MILLISECOND;
+						dest.set(row, col, millis);
+						break;
+					}
```
Baunsgaard: I think, we probably should not support this Int96, seems like a feature going out of the format.

src/main/java/org/apache/sysds/runtime/io/FrameReaderParquet.java
```diff
@@ -89,65 +97,147 @@ public FrameBlock readFrameFromHDFS(String fname, ValueType[] schema, String[] n
...
+	 */
+	private void readColumn(FrameBlock dest, ColumnReader creader, int col, int rowStart,
+		int rowsInGroup, int maxDef, PrimitiveType.PrimitiveTypeName ptype) throws IOException
+	{
+		for (int i = 0; i < rowsInGroup; i++) {
+			int row = rowStart + i;
+			if (creader.getCurrentDefinitionLevel() == maxDef) {
+				switch (ptype) {
+					case INT32:
+						dest.set(row, col, creader.getInteger());
```
Baunsgaard: Instead of going though the FrameBlock interface, we should directly allocate and use underlying Arrays. Just like the Delta Pr you just made.

src/main/java/org/apache/sysds/runtime/io/FrameReaderParquet.java
```diff
@@ -89,65 +97,147 @@ public FrameBlock readFrameFromHDFS(String fname, ValueType[] schema, String[] n
...
-			if (row != rlen) {
-				throw new IOException("Mismatch in row count: expected " + rlen + ", but got " + row);
+	/**
+	 * Reads one column of a row group, writing each value (or null) into the destination FrameBlock.
+	 */
+	private void readColumn(FrameBlock dest, ColumnReader creader, int col, int rowStart,
+		int rowsInGroup, int maxDef, PrimitiveType.PrimitiveTypeName ptype) throws IOException
+	{
+		for (int i = 0; i < rowsInGroup; i++) {
+			int row = rowStart + i;
```
Baunsgaard: Here similarly to the other Pr, you can move the row calculation out of the for loop and iterate 

int row = rowStart ; i < rowsInGroup + rowstarte; i++.

Please use the same variable names as your other PR.

src/test/java/org/apache/sysds/test/functions/io/parquet/FrameParquetSchemaTest.java
```diff
@@ -76,11 +77,11 @@ public void testParquetWriteReadAllSchemaTypes() {
...
-			{ 1.0,	1.1f, 10,  100L,  true,  "A" },
-			{ 2.0,	2.1f, 20,  200L,  false, "B" },
-			{ 3.0,	3.1f, 30,  300L,  true,  "C" },
-			{ 4.0,	4.1f, 40,  400L,  false, "D" },
-			{ 5.0,	5.1f, 50,  500L,  true,  "E" }
+			{ 1.0,    1.1f, 10,  100L,  true,  "A" },
+			{ 2.0,    2.1f, 20,  200L,  false, "B" },
+			{ 3.0,    3.1f, 30,  300L,  true,  "C" },
+			{ 4.0,    4.1f, 40,  400L,  false, "D" },
+			{ 5.0,    5.1f, 50,  500L,  true,  "E" }
```
Baunsgaard: formatting only change, please refrain from things like this.

src/test/java/org/apache/sysds/test/functions/io/parquet/FrameReaderParquetLegacy.java
```diff
@@ -0,0 +1,186 @@
...
+			throw new IOException("Mismatch in row count: expected " + rlen + ", but got " + row);
+		}
+	}
+
+	// Constants for decoding legacy INT96 timestamps
+	private static final int JULIAN_EPOCH_OFFSET_DAYS = 2_440_588;
+	private static final long MILLIS_IN_DAY = TimeUnit.DAYS.toMillis(1);
+	private static final long NANOS_PER_MILLISECOND = TimeUnit.MILLISECONDS.toNanos(1);
+
+	protected int readSingleParquetFile(Path path, Configuration conf, FrameBlock dest,
```
Baunsgaard: This seems odd, why do we have a custom reader in the test ? 

I suggest, instead of this, we compare to the Spark Parquet reader (that is builtin to the spark dependency)

src/test/java/org/apache/sysds/test/functions/io/parquet/FrameReaderParquetLegacy.java
Baunsgaard, line 1: Remove these legacy files.

src/test/java/org/apache/sysds/test/functions/io/parquet/FrameWriterParquetLegacy.java
Baunsgaard, line 1: remove

src/test/java/org/apache/sysds/performance/frame/ParquetReaderBenchmark.java
Baunsgaard, line 1: Do not add a benchmark file here, Instead put it in test/java/org/aparche/sysds/performance/

src/test/java/org/apache/sysds/test/functions/io/parquet/ParquetWriterBenchmark.java
Baunsgaard, line 1: remove the writer benchmark, move it similarly to the reader.

Baunsgaard (commented):
Overall, it is also getting close. However, there is some cleanup to do, Critically, add tests to verify compatibility with external readers and writers!

Jakob-al28:
Thanks, addressed all comments:

The reader decodes directly into pre-allocated column backing arrays, with the same structure and naming as the Delta column decode, the row offset is moved out of the decode loops. Columns whose parquet physical type does not match the requested frame type fall back to a per-cell conversion with the previous FrameBlock.set semantics (e.g. numeric files read with a string schema).
INT96 support is removed. Requesting an INT96 column fails with a error while the remaining columns of the same file stay readable via projection.
Legacy files are removed. Tests now compare against Spark as the reference implementation in both directions: Spark-written inputs are compared against Spark's own read, and SystemDS-written single- and multi-part outputs are read back with Spark. The sequential reader also expands part-file directories now, and both readers validate the file row counts against the expected dimensions.
Benchmarks are moved to org.apache.sysds.performance.frame, the batch-size comparison was dropped together with the legacy writer it measured.
Tests use TestUtils.compareFrames/generateRandomFrameBlock and redundant tests were removed.

Baunsgaard (approved):
Overall LGTM. Will merge.

PR #2535, opened 2026-07-06
[SYSTEMDS-3949] Column API parquet decode for Delta frame reads

Jakob-al28:
Follow-up to #2515, as discussed in #2528.

Delta Kernel's default engine decodes parquet through parquet-mr's row-record API (ParquetReader + ReadSupport/RecordMaterializer), which is built to reassemble nested records; each cell goes through a PrimitiveConverter callback. Frames are flat in SystemDS, so this adds a thin Engine decorator that overrides ParquetHandler.readParquetFiles: flat schemas decode through the column API (ColumnReadStoreImpl/ColumnReader). Everything else (predicates, nested/decimal/timestamp columns, unknown metadata columns) falls through to the wrapped default engine, and deletion vectors, partition values and column mapping are still applied by the kernel above the handler.

On TPC-H lineitem, 30M rows (median of 7 runs, cloud VM,  n2-highmem-8, 40GB heap): 101.0s to 48.5s (2.08x). On the DeltaFrameRead benchmark from #2515: 1.73x serial, 1.51x parallel (k=8).

| Format / Reader | Median (ms) |
|-----------------|------------:|
| Parquet: Legacy (Group) | 102,099 |
| Parquet: Column API | 53,381 |
| Delta: Row-record | 101,028 |
| Delta: Column API | 48,513 |
| Binary Block | 21,576 |

All runs at the same settings: ZSTD, no dictionary for Parquet and Delta; Binary Block uncompressed.

matched_encoding_all
delta_deltaframeread

Two cases needed explicit handling: tables with deletion vectors enabled ask every read for a synthetic _metadata.row_index column that isn't actually in the file, and older files in a table that's had a column added later don't contain that column at all. Both would have failed otherwise; now the first is computed and the second comes back as null, covered by tests (DeltaFrameSparkContractTest).

Notes: the ParquetHandler contract has grown across kernel versions (the row_index requirement came with deletion vector support), so delta-kernel version bumps should re-run the contract tests. The decode relies on parquet-mr's column API (org.apache.parquet.column.impl); its signatures are unchanged from parquet 1.13 through 1.17.

The write path stays on the default engine. Two checks suggest there is little to gain there: the parquet writer rewrite in #2528, which removed the per-row Group materialization, only moved write time from 92s to 87s, and with matched codecs the Delta write is within about 8% of that plain parquet writer (2.7s Delta vs 2.5s parquet on 1M lineitem rows, both set to snappy for this check). Neither points to a large win available on the write side.


Baunsgaard:
Overall this looks good, but i would change some things. 

1. can we instead of adding a new way of reading the delta format, simply overwrite the old. We do not want to maintain may ways, if your way is better lets default to that, and only support that. 
2. I am wondering if we can avoid using the Engine abstraction with the columnAPI. This would allow us to avoid allocating the result FrameBlock's arrays times, and instead directly put the values into preallocated columns. 


Lets me know what you think?

Jakob-al28:
I agree, the changes have been applied. The engine decorator and the config flag are gone, the column decode is now the path for a table's data files whenever it applies, it writes directly into the preallocated frame columns. The kernel is still used for the log/snapshot and for the cases where the data files alone aren't enough: deletion vectors and partitioned tables fall back to the kernel engine read. This simplification improved the speedup from 1.89x to 2.08x. Updated numbers are in the PR description.

src/main/java/org/apache/sysds/runtime/io/DeltaKernelUtils.java
```diff
@@ -162,6 +174,199 @@ public static int countSelected(int size, boolean[] selected) {
...
+	 * @throws IOException on read failure
+	 */
+	public static int decodeDataFileInto(String filePath, StructType physicalSchema, int[] readCodes, Object[] dest,
+		int destOff, int limit, String tablePath) throws IOException {
+		final Configuration conf = ConfigurationManager.getCachedJobConf();
+		final int ncol = physicalSchema.length();
+		int off = destOff;
+		try(ParquetFileReader reader = ParquetFileReader.open(HadoopInputFile.fromPath(new Path(filePath), conf))) {
+			MessageType parquetSchema = reader.getFooter().getFileMetaData().getSchema();
+			String createdBy = reader.getFooter().getFileMetaData().getCreatedBy();
```
Baunsgaard: Avoid calling reader.getFooter().getFileMetadata() twice, just put another variable.

src/main/java/org/apache/sysds/runtime/io/DeltaKernelUtils.java
```diff
@@ -162,6 +174,199 @@ public static int countSelected(int size, boolean[] selected) {
...
+			String[] colNames = resolveParquetColumns(physicalSchema, parquetSchema);
+			GroupConverter root = dummyConverter(parquetSchema.getFieldCount());
+			PageReadStore pages;
+			while((pages = reader.readNextRowGroup()) != null) {
+				int nrow = (int) pages.getRowCount();
+				// guard before decoding: writing past the limit would overflow into the
+				// next file's slice (or off the array) in the pre-allocated output
```
Baunsgaard: This comment is redundant, the code bellow is sufficient

src/main/java/org/apache/sysds/runtime/io/DeltaKernelUtils.java
```diff
@@ -162,6 +174,199 @@ public static int countSelected(int size, boolean[] selected) {
...
+	 * cells (definition level below max) keep the array default (0 for numerics, null for strings).
+	 */
+	private static void decodeColumnInto(ColumnReader creader, int maxDef, int nrow, int readCode, Object dest,
+		int off) {
+		switch(readCode) {
+			case T_DOUBLE: {
+				double[] a = (double[]) dest;
+				for(int r = 0; r < nrow; r++) {
+					if(creader.getCurrentDefinitionLevel() == maxDef)
+						a[off + r] = creader.getDouble();
```
Baunsgaard: we can move the off + r out to the forloop, this can be done for all the cases.

```
int end = nrow + off;
for(int r = off; r < end; r++) {
}
```

src/main/java/org/apache/sysds/runtime/io/DeltaKernelUtils.java
```diff
@@ -162,6 +174,199 @@ public static int countSelected(int size, boolean[] selected) {
...
+	 * Decode one parquet column of the current row group into a pre-allocated typed array at the given offset. Null
+	 * cells (definition level below max) keep the array default (0 for numerics, null for strings).
+	 */
+	private static void decodeColumnInto(ColumnReader creader, int maxDef, int nrow, int readCode, Object dest,
+		int off) {
+		switch(readCode) {
+			case T_DOUBLE: {
+				double[] a = (double[]) dest;
+				for(int r = 0; r < nrow; r++) {
+					if(creader.getCurrentDefinitionLevel() == maxDef)
```
Baunsgaard, line 373: What is this getCurrentDefinitionLevel actually doing ? 

Is it nessesarry for all the rows to call this function ?

Baunsgaard (commented):
A few things to consider changing, but overall a fine PR. I do not think my performance suggestions will make much of a difference since the main overhead is in parquet parsing, but it does not hurt to do.

Can you doublecheck that this API is working with various shape and size inputs, and it is faster in those cases. 

e.g.  100 rows, -> 1mil rows and 1 column -> 1000 columns ?

src/main/java/org/apache/sysds/runtime/io/DeltaKernelUtils.java
Jakob-al28: The Definition Level is the parquet null encoding. The pages only store data for not-null values, and the definition level says whether a value is present. A level below the maximum means the cell is null and the typed getter must not be called. 

For variable-length columns skipping the check desyncs the value stream, I tried it and the read fails with an EOFException. The check itself is basically free: the reader decodes the level stream anyway while advancing, getCurrentDefinitionLevel is just a field read. It would only be skippable for required columns where the max level is 0, but Delta writes all data columns as optional, so that case does not occur in practice.

Jakob-al28:
The inline comments are applied. For the shapes and sizes I added DeltaFrameShapeCoverageTest, which round-trips 27 shapes from 1x1 up to 1M rows and 1 to 1000 columns, including the 4096 writer batch boundaries, all-null string columns, NaN/infinity/min/max values and unicode strings.

For the speed I benchmarked this branch against the kernel read path on main, same on-disk tables, same JVM flags, alternating runs, best of three each. On a 10Mx3 table with the default file layout (208MB, string/long/double with nulls) the direct decode is 1.7x faster serial and 1.5x parallel. Smaller and wider tables gain more, between 1.8x and 5.2x across the grid, no shape was slower.

While validating this I found and fixed one gap: type widening can leave INT32 data files under a column that is now bigint. The kernel converts those on read, the typed decode cannot. decodeDataFileInto now checks each file's physical types up front and falls back to the kernel read for that file only; readTypeWidenedIntFilesAsLongColumn builds such a table and fails on the previous code.

src/test/java/org/apache/sysds/test/component/io/DeltaFrameShapeCoverageTest.java
```diff
@@ -0,0 +1,279 @@
...
+		int ncol = expected.getNumColumns();
+		for(int c = 0; c < ncol; c++) {
+			assertEquals(label + ": schema col " + c, expected.getSchema()[c], actual.getSchema()[c]);
+			assertEquals(label + ": name col " + c, expected.getColumnNames()[c], actual.getColumnNames()[c]);
+		}
+		int nrow = expected.getNumRows();
+		for(int r = 0; r < nrow; r++)
+			for(int c = 0; c < ncol; c++)
+				assertEquals(label + ": cell (" + r + "," + c + ")", expected.get(r, c), actual.get(r, c));
+	}
```
Baunsgaard: I think in TestUtils we have a frame comparison. Can we use this?

https://github.com/apache/systemds/blob/e5578364bfe4167e6e9b289d439bfd0ae034173b/src/test/java/org/apache/sysds/test/TestUtils.java#L882

src/test/java/org/apache/sysds/test/component/io/DeltaFrameShapeCoverageTest.java
```diff
@@ -0,0 +1,279 @@
...
+		return schema;
+	}
+
+	/**
+	 * Deterministic test frame for a covered shape: random data over the cycled schema plus adversarial patterns
+	 * injected into every string column (nulls every 7th row, empty strings every 11th, multi-byte unicode every 13th)
+	 * so the null definition-level handling and UTF-8 decode are covered at every size, including across file and batch
+	 * boundaries of the larger shapes.
+	 */
+	private static FrameBlock genFrame(int nrow, int ncol, long seed) {
```
Baunsgaard: Similarly, we have frame generators.

one of many examples:

https://github.com/apache/systemds/blob/e5578364bfe4167e6e9b289d439bfd0ae034173b/src/test/java/org/apache/sysds/test/TestUtils.java#L2357

src/test/java/org/apache/sysds/test/component/io/DeltaFrameShapeCoverageTest.java
```diff
@@ -0,0 +1,279 @@
...
+			if(multiFile)
+				ConfigurationManager.clearLocalConfigs();
+			FileUtils.deleteQuietly(dir.toFile());
+		}
+	}
+
+	private static void assertRoundTrip(int nrow, int ncol, long seed, boolean multiFile, boolean checkBuffered)
+		throws Exception {
+		roundTripAllReaders(nrow + "x" + ncol, genFrame(nrow, ncol, seed), multiFile, checkBuffered);
+	}
```
Baunsgaard: While I appreciate the amount of testing, It is a bit excessive ATM. 

You have many calls now to assert the round trip is correct. 
please reduce it to a reasonable amount (remember we will run this test many many times)

src/test/java/org/apache/sysds/test/component/io/DeltaFrameShapeCoverageTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+	@Test
+	public void millionRowsRoundTrip() throws Exception {
+		// the 1M-row end of the row scaling, as multi-file tables so the parallel
+		// reader really splits across files; the kernel-path parity at this scale
+		// is covered via the forced-buffered read of the 1Mx3 table
+		assertRoundTrip(1_000_000, 1, 501, true, false);
+		assertRoundTrip(1_000_000, 3, 502, true, true);
+	}
```
Baunsgaard, line 89: These ones are especially excessively large. Do they really test anything differently than the smaller tests?

Baunsgaard (commented):
Great, Really like how this PR is turning out.

But, we need to reduce the testing to a reasonable amount for the round trips. 

Another flavor of tests we could instead refine is the spark round trip testing, to ensure our reader is understanding the external engines.  However this would be a followup PR.

Jakob-al28:
Done, thanks, reduced the tests to a smaller set (8 round trips total). Also swapped the comparison and data generation to use TestUtils.compareFrames and TestUtils.generateRandomFrameBlock. Agreed on the Spark round-trip refinement being a good followup.

Baunsgaard:
@Jakob-al28  Thanks for the PR, and addressing the comments, It is now merged!

Jakob-al28:
Thanks for the review!

PR #2542, opened 2026-07-11
[SYSTEMDS-3524] Add multi-threaded unique in LibMatrixSketch

shieru1214:
### Summary

This PR adds a multi-threaded path for `unique` in `LibMatrixSketch`.

The existing single-threaded implementation is kept and is still used as the baseline/fallback. I added a new overload: `getUniqueValues(MatrixBlock blkIn, Types.Direction dir, int k)`, where `k` controls the requested number of threads.

The implementation follows the existing SystemDS `unique` direction semantics:

- `RowCol`: unique scalar values over the input matrix, returned as one column
- `Row`: row-wise unique scalar values, preserving the number of rows
- `Col`: column-wise unique scalar values, preserving the number of columns

### Implementation

For `k > 1` and sufficiently large inputs, the code uses balanced row or column ranges and `CommonThreadPool`.

The parallel logic is direction-specific:

- `RowCol`: each worker builds a local `HashSet<Double>` over its row range, and the local sets are merged afterwards.
- `Row`: rows are partitioned across workers. The code first computes the maximum number of unique values per row, allocates the output, and then fills the row-wise unique values in parallel.
- `Col`: columns are partitioned across workers. The code first computes the maximum number of unique values per column, allocates the output, and then fills the column-wise unique values in parallel.

There are two parallel paths:

- regular parallel deduplication, used when the estimated local memory usage is safe
- batched parallel deduplication, used when full thread-local temporary data may be too large

The batched path processes a limited number of ranges at a time, clears temporary local data earlier, and then continues with the next batch. If the input is small, `k <= 1`, or batching is not useful, the code falls back to the original single-threaded path.

### Testing

I added a focused JUnit test:

- `src/test/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketchUniqueParallelTest.java`

It checks `RowCol`, `Row`, and `Col`, and compares the multi-threaded result with the single-threaded baseline.

I ran the following tests:

- `mvn -Dtest=LibMatrixSketchUniqueParallelTest test`
- `mvn -Dtest=UniqueRow,UniqueRowCol test`

The focused test completed with 4 tests passing, and the existing unique tests completed with 9 tests passing. I also ran the focused test with `-Xmx64m` to exercise the memory-aware batched path.

### Runtime experiments

I ran local runtime experiments with fixed `k=4`, comparing against the `k=1` baseline.

The experiments were run locally on an Apple Silicon Mac using Java 17. The JVM heap was fixed at 3 GB, and all inputs used dense `MatrixBlock` storage.

For each case, I used three warm-up iterations followed by five measured iterations. The order of `k=1` and `k=4` was alternated between iterations to reduce ordering effects, and the median runtime is reported. Input generation is not included in the measurements.

The experiments cover:

- directions: `RowCol`, `Row`, and `Col`
- input sizes: small and large
- unique shares: `1%`, `10%`, `50%`, and `100%`
- thread counts: `k=1` and `k=4`

In the tables, `Unique Values` has a different meaning for each direction:

- for `RowCol`, it is the total number of distinct scalar values in the input
- for `Row`, it is the number of distinct scalar values within each row
- for `Col`, it is the number of distinct scalar values within each column

The exact integer count used for each case is shown in the tables. Speedup is calculated as `k=1 time / k=4 time`, so a value above 1 means that the multi-threaded path was faster.

The 3 GB heap limit was chosen to keep the large dense inputs within memory while allowing the memory guard to select the batched path for the large `Row` and `Col` cases. With this setting, the smaller cases use the regular parallel path, while the large `Row` and `Col` cases use batched parallel processing.

These are local measurements and are intended as an initial comparison rather than a full benchmark.

#### RowCol

| Size | Input | Unique Share | Unique Values | k=1 (ms) | k=4 (ms) | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| small | 100000x1 | 1% | 1000 | 1.598 | 2.362 | 0.68 |
| small | 100000x1 | 10% | 10000 | 2.041 | 4.897 | 0.42 |
| small | 100000x1 | 50% | 50000 | 2.145 | 5.307 | 0.40 |
| small | 100000x1 | 100% | 100000 | 3.092 | 7.047 | 0.44 |
| large | 2000000x1 | 1% | 20000 | 23.294 | 20.875 | 1.12 |
| large | 2000000x1 | 10% | 200000 | 32.340 | 112.501 | 0.29 |
| large | 2000000x1 | 50% | 1000000 | 79.177 | 115.331 | 0.69 |
| large | 2000000x1 | 100% | 2000000 | 60.601 | 110.218 | 0.55 |

#### Row

| Size | Input | Unique Share | Unique Values per Row | k=1 (ms) | k=4 (ms) | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| small | 10000x64 | 1% | 1 | 10.561 | 3.722 | 2.84 |
| small | 10000x64 | 10% | 6 | 12.057 | 4.279 | 2.82 |
| small | 10000x64 | 50% | 32 | 14.747 | 5.811 | 2.54 |
| small | 10000x64 | 100% | 64 | 18.601 | 9.271 | 2.01 |
| large | 200000x64 | 1% | 1 | 209.592 | 66.916 | 3.13 |
| large | 200000x64 | 10% | 6 | 237.736 | 85.111 | 2.79 |
| large | 200000x64 | 50% | 32 | 291.955 | 112.381 | 2.60 |
| large | 200000x64 | 100% | 64 | 362.697 | 165.816 | 2.19 |

#### Col

| Size | Input | Unique Share | Unique Values per Column | k=1 (ms) | k=4 (ms) | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| small | 64x10000 | 1% | 1 | 14.640 | 4.453 | 3.29 |
| small | 64x10000 | 10% | 6 | 15.888 | 5.031 | 3.16 |
| small | 64x10000 | 50% | 32 | 20.366 | 6.899 | 2.95 |
| small | 64x10000 | 100% | 64 | 22.102 | 9.930 | 2.23 |
| large | 64x200000 | 1% | 1 | 296.105 | 87.217 | 3.40 |
| large | 64x200000 | 10% | 6 | 323.869 | 99.900 | 3.24 |
| large | 64x200000 | 50% | 32 | 418.123 | 135.996 | 3.07 |
| large | 64x200000 | 100% | 64 | 447.553 | 220.738 | 2.03 |

The row-wise and column-wise cases show consistent speedups because rows and columns can be processed independently. Their speedup generally decreases as the number of unique values grows, since larger sets and outputs require more work.

`RowCol` behaves differently because its thread-local sets must still be merged into one global set. The large input with a 1% unique share shows a small improvement, but in the other tested cases the allocation and merge overhead is greater than the saved parallel work.


src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+	 *
+	 * @param blkIn input matrix block
+	 * @param dir unique direction
+	 * @return matrix block containing unique values
+	 */
 	public static MatrixBlock getUniqueValues(MatrixBlock blkIn, Types.Direction dir) {
-		//similar to R's unique, this operation takes a matrix and computes the
-		//unique values (or rows in case of multiple column inputs)
-		
+		return getUniqueValues(blkIn, dir, 1);
```
gaturchenko, line 54: The degree of parallelism is hard-coded here, so `AggregateUnaryCPInstruction` will always call the single-threaded implementation.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+		if( config == null )
+			return getUniqueValuesSequential(blkIn, dir);
+
+		ExecutorService pool = CommonThreadPool.get(config._numThreads);
+		try {
+			HashSet<Double> hashSet = new HashSet<>();
+			for( int pos = 0; pos < config._len; ) {
+				ArrayList<UniqueValueTask> tasks = new ArrayList<>();
+				for( int i = 0; i < config._numThreads && pos < config._len; i++ ) {
+					int end = Math.min(pos + config._taskLen, config._len);
```
gaturchenko: This will overflow when `pos` is close to `Integer.MAX_VALUE`, i.e., when the number of rows in the matrix is very large

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+	/**
+	 * Batched variant of getMaxUniqueValues.
+	 */
+	private static int getMaxUniqueValuesBatched(ExecutorService pool, MatrixBlock blkIn, Types.Direction dir,
+		BatchConfig config) throws Exception {
+		int ret = 0;
+		for( int pos = 0; pos < config._len; ) {
+			ArrayList<UniqueCountTask> tasks = new ArrayList<>();
+			for( int i = 0; i < config._numThreads && pos < config._len; i++ ) {
+				int end = Math.min(pos + config._taskLen, config._len);
```
gaturchenko: Same overflow issue

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+
+	/**
+	 * Batched variant of fillUniqueValues.
+	 */
+	private static void fillUniqueValuesBatched(ExecutorService pool, MatrixBlock blkIn, MatrixBlock blkOut,
+		Types.Direction dir, BatchConfig config) throws Exception {
+		for( int pos = 0; pos < config._len; ) {
+			ArrayList<UniqueOutputTask> tasks = new ArrayList<>();
+			for( int i = 0; i < config._numThreads && pos < config._len; i++ ) {
+				int end = Math.min(pos + config._taskLen, config._len);
```
gaturchenko: Same overflow issue

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+	 * @return one-column matrix block containing the unique values
+	 */
+	private static MatrixBlock getUniqueValuesRowColBatchedParallel(MatrixBlock blkIn, Types.Direction dir, int k) {
+		BatchConfig config = getBatchConfig(blkIn, dir, k);
+		if( config == null )
+			return getUniqueValuesSequential(blkIn, dir);
+
+		ExecutorService pool = CommonThreadPool.get(config._numThreads);
+		try {
+			HashSet<Double> hashSet = new HashSet<>();
```
gaturchenko, line 298: This `HashSet` is unbounded. When almost all values are unique, the set grows to `rlen x clen`, so it in principle matches the sequential case.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+		int len = getPartitionLength(blkIn, dir);
+		long maxBatchIndexes = getMaxLocalDedupIndexes(blkIn, dir);
+		if( maxBatchIndexes < 2 )
+			return null;
+
+		int numThreads = getNumThreads(k, (int) Math.min(len, maxBatchIndexes));
+		if( numThreads <= 1 )
+			return null;
+
+		int taskLen = Math.max(1, (int) Math.min(Integer.MAX_VALUE, maxBatchIndexes / numThreads));
```
gaturchenko, line 535: Then `numThreads x taskLen` consumes the entire 1/4 of the heap in the worst case

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+	/**
+	 * Conservative memory guard for full local deduplication. The estimate includes
+	 * a small overhead factor for set objects.
+	 *
+	 * @param blkIn input matrix block
+	 * @param dir unique direction
+	 * @return true if local deduplication is small enough for the parallel path
+	 */
+	private static boolean isLocalDedupMemoryBudgetSafe(MatrixBlock blkIn, Types.Direction dir) {
+		return getMaxLocalDedupIndexes(blkIn, dir) >= getPartitionLength(blkIn, dir);
```
gaturchenko: If I understand correctly, you check here whether you can hold in memory each dedup set for every row at the same time. This is correct for `RowCol`, but due to `hashSet.clear()` it is unnecessarily conservative for `Row` and `Col`, because the peak memory is then `numThreads × clen × 64 bytes` which doesn't depend on `rlen`.

src/test/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketchUniqueParallelTest.java
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sysds.runtime.matrix.data;
```
gaturchenko: Please, move the tests to respective files in `src/test/java/org/apache/sysds/test/functions/unique`, feel free to extend for the multi-threaded case if necessary

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,695 @@
...
+	 * @return maximum number of rows or columns per batch
+	 */
+	private static long getMaxLocalDedupIndexes(MatrixBlock blkIn, Types.Direction dir) {
+		long cellsPerIndex = dir == Types.Direction.Col ? blkIn.getNumRows() : blkIn.getNumColumns();
+		if( cellsPerIndex <= 0 ||
+			cellsPerIndex > Long.MAX_VALUE / Double.BYTES / PAR_UNIQUE_LOCAL_BYTES_OVERHEAD )
+			return 0;
+
+		long bytesPerIndex = cellsPerIndex * Double.BYTES * PAR_UNIQUE_LOCAL_BYTES_OVERHEAD;
+		long maxLocalBytes = Runtime.getRuntime().maxMemory() / PAR_UNIQUE_MAX_LOCAL_BYTES_FRACTION;
```
gaturchenko: Read from `InfrastructureAnalyzer.getLocalMaxMemory()` instead

gaturchenko (changes requested):
Thank you for the contribution @shieru1214. It looks good overall, but there are some issues I would like you to go over.

First, the most critical problem is that the `AggregateUnary` instruction will always call the single-threaded implementation. With multi-threading in place, it should be only used as a fallback.

Second, the batched multi-threading path has a number of problems related to memory estimation and guards against excessive usage. Below is the summary of suggested fixes:
1. Int overflow: compute as `int end = (int) Math.min((long) pos + config._taskLen, config._len);`
2. The safety condition for `Row` and `Col` is `numThreads × cellsPerIndex × 64 ≤ budget` (one live row/column set per thread). Otherwise, you force batched execution where it can be safely avoided
3. You have to keep in mind that for batched `RowCol` the merged `HashSet` grows toward the distinct count. So you should size `taskLen` against a fraction of the budget with the remainder reserved for the merged set. If the worst-case simply doesn't fit, fall back to sequential execution.
4. I'm pretty sure batched tests do not test the respective path in our CI since the JVM heap is never sufficiently low (the worst-case requirement is ~77 MiB). You can, e.g., add an overload like `maxLocalBytes`, or account for it in any other way you prefer

Also, the formatting job is failing, you need to run the `dev/format-changed.sh` script. Your fork doesn't have it, so please rebase on the current main so that it's present in your repository as well.

Isso-W:
## Update: review feedback addressed

Thanks for the detailed review. All points are addressed:

**1. `AggregateUnary` always called the single-threaded implementation.**
`AggregateUnaryCPInstruction` now passes `op.getNumThreads()` into `LibMatrixSketch.getUniqueValues(...)`. The thread count was already plumbed into `UnarySketchOperator` by `InstructionUtils`, it just was not forwarded, so the multi-threaded path was unreachable from DML and the 2-argument overload is now only the `k = 1` fallback.

**2. Int overflow in the batched task ranges.**
All three occurrences now compute `int end = (int) Math.min((long) pos + config._taskLen, config._len);`.

**3. Safety condition for `Row` and `Col`.**
The row-wise and column-wise workers reuse a single set that is cleared per row/column, so only one live set per thread has to be charged. The guard is now `numThreads * cellsPerIndex * 64 <= budget` instead of charging every index, which no longer forces batched execution where it can safely be avoided.

**4. Batched `RowCol` and the merged set.**
The `RowCol` workers accumulate over their whole range and are merged into a set that grows toward the total distinct count. The batched configuration now reserves the all-distinct worst case for the merged set up front and sizes `taskLen` against the remaining budget only. If not even the merged set fits, batching cannot help and the code falls back to sequential execution.

**5. The batched paths were not covered in CI.**
There is now a `getUniqueValues(blkIn, dir, k, maxLocalBytes)` overload, so tests can inject a small budget instead of relying on a low JVM heap. The test also moved from `org.apache.sysds.runtime.matrix.data` to `org.apache.sysds.test.functions.unique`, because the CI test selection only scans `org.apache.sysds.test.*` and the test was therefore never executed. Added cases now exercise all three batched paths and their sequential fallbacks deterministically.

Additionally, the memory budget is now derived from `InfrastructureAnalyzer.getLocalMaxMemory()` rather than `Runtime.getRuntime().maxMemory()`, so it reflects the local JVM memory that parfor adjusts per worker, matching the convention used in the neighbouring matrix libraries.

The branch is rebased on current `main`, and `dev/format-changed.sh` has been applied.

## Testing

`src/test/java/org/apache/sysds/test/functions/unique/LibMatrixSketchUniqueParallelTest.java`

```
mvn -Dtest=LibMatrixSketchUniqueParallelTest test   ->  10 tests, 0 failures
mvn -Dtest=UniqueRow,UniqueRowCol test              ->   9 tests, 0 failures
```

19 tests pass in total. Checkstyle reports 0 violations and the format check on the PR-edited lines is clean.

## Runtime experiments

The benchmark is included as `src/test/java/org/apache/sysds/performance/matrix/UniquePerf.java`, so the numbers below can be reproduced:

```
java -Xmx3g --add-modules jdk.incubator.vector \
  -cp target/classes:target/test-classes:<deps> \
  org.apache.sysds.performance.matrix.UniquePerf 4 5
```

The experiments were run locally on an Intel Core i7-13700H (20 logical cores) using Java 17, with the JVM heap fixed at 3 GB and dense `MatrixBlock` inputs throughout.

For each configuration I used three warm-up iterations followed by five measured iterations. The order of the `k=1` and `k=4` runs is alternated between iterations to reduce ordering effects, and the median runtime is reported. Input generation is not included in the measurements.

The experiments cover the directions `RowCol`, `Row` and `Col`, a small and a large input, unique shares of 1%, 10%, 50% and 100%, and thread counts `k=1` and `k=4`. `Unique Values` is the number of distinct scalar values in the whole input for `RowCol`, per row for `Row` and per column for `Col`. Speedup is `k=1` time divided by `k=4` time, so a value above 1 means the multi-threaded path was faster.

These are local measurements and are intended as an initial comparison rather than a full benchmark.

### RowCol

| Size | Input | Unique Share | Unique Values | k=1 (ms) | k=4 (ms) | Speedup |
|---|---|---|---|---|---|---|
| small | 100000x1 | 1% | 1000 | 3.186 | 21.061 | 0.15 |
| small | 100000x1 | 10% | 10000 | 3.164 | 7.690 | 0.41 |
| small | 100000x1 | 50% | 50000 | 2.407 | 6.757 | 0.36 |
| small | 100000x1 | 100% | 100000 | 3.228 | 5.975 | 0.54 |
| large | 2000000x1 | 1% | 20000 | 23.134 | 23.172 | 1.00 |
| large | 2000000x1 | 10% | 200000 | 31.379 | 58.022 | 0.54 |
| large | 2000000x1 | 50% | 1000000 | 180.886 | 248.605 | 0.73 |
| large | 2000000x1 | 100% | 2000000 | 248.350 | 325.830 | 0.76 |

### Row

| Size | Input | Unique Share | Unique Values per Row | k=1 (ms) | k=4 (ms) | Speedup |
|---|---|---|---|---|---|---|
| small | 10000x64 | 1% | 1 | 41.429 | 13.344 | 3.10 |
| small | 10000x64 | 10% | 6 | 44.689 | 15.397 | 2.90 |
| small | 10000x64 | 50% | 32 | 64.297 | 32.751 | 1.96 |
| small | 10000x64 | 100% | 64 | 88.566 | 55.402 | 1.60 |
| large | 200000x64 | 1% | 1 | 821.775 | 220.942 | 3.72 |
| large | 200000x64 | 10% | 6 | 921.537 | 260.261 | 3.54 |
| large | 200000x64 | 50% | 32 | 1290.565 | 568.493 | 2.27 |
| large | 200000x64 | 100% | 64 | 1837.356 | 1175.834 | 1.56 |

### Col

| Size | Input | Unique Share | Unique Values per Column | k=1 (ms) | k=4 (ms) | Speedup |
|---|---|---|---|---|---|---|
| small | 64x10000 | 1% | 1 | 46.585 | 15.993 | 2.91 |
| small | 64x10000 | 10% | 6 | 52.774 | 16.794 | 3.14 |
| small | 64x10000 | 50% | 32 | 71.740 | 34.172 | 2.10 |
| small | 64x10000 | 100% | 64 | 105.375 | 58.113 | 1.81 |
| large | 64x200000 | 1% | 1 | 995.851 | 289.124 | 3.44 |
| large | 64x200000 | 10% | 6 | 1117.002 | 335.624 | 3.33 |
| large | 64x200000 | 50% | 32 | 1553.361 | 798.953 | 1.94 |
| large | 64x200000 | 100% | 64 | 2253.123 | 1512.326 | 1.49 |

`Row` and `Col` show consistent speedups, decreasing as the share of unique values grows: at a low share the output is narrow and the runtime is dominated by the parallel scan, while at 100% the second pass writes a full-size output and becomes memory-bound.

`RowCol` still does not benefit. Its thread-local sets must be merged into one global set, and the merge cost grows with the total distinct count, so at a 100% share the merge re-inserts every distinct value and effectively repeats the sequential work on top of the parallel pass. At low shares the merge is cheap, but the concurrent construction of the per-thread `HashSet<Double>` instances costs more than the saved scan time. Making `RowCol` scale would need a different strategy — a shared concurrent set, a tree-shaped merge, or a primitive-keyed hash map to avoid boxing — which seems out of scope for this PR.

src/test/java/org/apache/sysds/test/functions/unique/LibMatrixSketchUniqueParallelTest.java
gaturchenko, line 1: This test file is rather inconsistent with the general structure of function tests. The biggest issue is that you don't actually invoke a DML script but rather call the instruction directly. So essentially instead of end-to-end you have unit tests, and we generally prefer the former. Therefore, please have another look at how existing unique tests are implemented. If necessary, you can extend the `UniqueBase` class. Regarding the tests that constrain the memory budget to trigger the batched path, you can in principle leave them as they are, but I'd still appreciate if you could find a way to make them DML-based as well.

gaturchenko (changes requested):
Thank you for the fixes @shieru1214 @Isso-W, we appreciate your effort. I added a comment regarding the test file, please address when you can.

Isso-W:
Update: tests are now script-based

Thanks, that was a fair point. The parallel test was indeed a unit test in disguise. It is removed and replaced with end-to-end tests that follow the existing unique tests.

**What changed**

- Deleted `LibMatrixSketchUniqueParallelTest` (344 lines of direct `LibMatrixSketch` calls).
- Added `uniqueCol.dml` and `UniqueCol`. The column-wise direction had no script-level coverage at all before.
- Extended `UniqueBase` with two variants:
  - `uniqueTestOrdered(...)` compares row by row instead of ignoring row order. Usable where every row or column of the result holds a single value, so the hash set iteration order cannot affect the outcome.
  - `uniqueTestConstrainedMemory(...)` temporarily lowers `InfrastructureAnalyzer.setLocalMaxMemory(...)` around the script run and restores it in a `finally` block, following the pattern already used in `ParForNaiveBayesTest` and `CompilerTestBase`.
- Added multi-threaded and sparse cases for all three directions. The sparse inputs populate only every eighth row, every eighth column or every tenth cell, which mirrors the dense/sparse split of the countDistinct tests.

24 cases in total, 22 of them run the DML script.

**Batched paths**

`RowCol` batching is now reachable end-to-end: `UniqueRowCol.testBatchedCP` runs the script under a reduced budget and selects the batched path.

For `Row` and `Col` I did not manage to do the same. Those select batching only when `numThreads * cellsPerIndex * 64` exceeds the budget, and reaching at least two batches additionally requires `budget >= 2 * cellsPerIndex * 64`. Both together put the budget into a `[2, numThreads)` window that depends on the thread count of the executing machine, and it is empty for `numThreads = 2`. That is a consequence of the corrected guard: after it, `Row` and `Col` only need batching when a single row or column is very large. So those two paths are covered by a small unit test, `UniqueBatchedPathTest`, in the same spirit as `CountDistinctRowOrColBase.testCPSparseLarge`, which also calls the library directly to reach a state that DML cannot express.

Happy to drop that test if you would rather have the two paths uncovered.

**Verification**

I instrumented the path selection temporarily to confirm the tests reach what they claim: the script runs use `k = 20`, the `RowCol` budget case selects the batched path, and the sparse inputs are read in sparse format. All 24 cases pass, Checkstyle reports 0 violations, and the format check is clean.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
```diff
@@ -19,89 +19,807 @@
...
+	 */
+	public static MatrixBlock getUniqueValues(MatrixBlock blkIn, Types.Direction dir, int k, long maxLocalBytes) {
+		// Similar to R's unique, this operation computes unique values according
+		// to the requested direction.
+		if(!satisfiesMultiThreadingConstraints(blkIn, dir, k))
+			return getUniqueValuesSequential(blkIn, dir);
+
+		boolean localDedupMemorySafe = isLocalDedupMemoryBudgetSafe(blkIn, dir, k, maxLocalBytes);
+		switch(dir) {
+			case RowCol:
```
gaturchenko, line 88: Sorry, I missed this one in my prior review. Based on your benchmarking results, it seems that multi-threading yields a slowdown for `RowCol`. Can we then just fall back to single-threaded here?

src/test/java/org/apache/sysds/test/functions/unique/UniqueBase.java
```diff
@@ -40,8 +41,39 @@ public void setUp() {
...
 							Types.ExecType instType, double epsilon) {
+		uniqueTest(inputMatrix, expectedMatrix, instType, epsilon, -1, false);
+	}
+
+	/**
+	 * Runs the unique script and compares the result row by row, in order. Use this where the expected output is
+	 * unambiguous, i.e. where every row or column of the result holds a single value and the iteration order of the
+	 * internal hash sets cannot affect the outcome.
+	 */
+	protected void uniqueTestOrdered(double[][] inputMatrix, double[][] expectedMatrix, Types.ExecType instType,
```
gaturchenko: Why do you need this separate function? The existing `uniqueTest` function compares the results out-of-order that works both when the output is deterministic or random. Am I missing something here?

src/test/java/org/apache/sysds/test/functions/unique/UniqueBase.java
```diff
@@ -52,9 +84,13 @@ protected void uniqueTest(double[][] inputMatrix, double[][] expectedMatrix,
 			runTest(true, false, null, -1);
 			writeExpectedMatrix("A", expectedMatrix);
 
-			compareResultsRowsOutOfOrder(epsilon);
+			if(orderedComparison)
```
gaturchenko: Again as per my comment above, I don't really understand the purpose of `orderedComparison`

src/test/java/org/apache/sysds/test/functions/unique/UniqueBatchedPathTest.java
gaturchenko, line 1: Why is this still a unit test? You have limited the memory with `InfrastructureAnalyzer` but you can limit concurrency as well by calling `InfrastructureAnalyzer.setLocalPar()`, so I don't quite see what holds you back from exercising the batched path for all 3 cases end-to-end

src/test/java/org/apache/sysds/test/functions/unique/UniqueCol.java
```diff
@@ -0,0 +1,127 @@
...
+		double[][] expectedMatrix = {{1, 2, 3}, {4, 5, 6}, {7, 8, 9}};
+		uniqueTest(inputMatrix, expectedMatrix, Types.ExecType.CP, 0.0);
+	}
+
+	/**
+	 * Large enough to take the multi-threaded path. Every column holds a single distinct value, so the expected result
+	 * is one row and independent of any hash set iteration order.
+	 */
+	@Test
+	public void testMultiThreadedCP() {
```
gaturchenko, line 80: Thank you for covering the `Col` case as well. However, what do you mean by `large enough`? The multi-threaded execution should be picked up by default unless the constraints are violated. Why are the 4 existing tests insufficient?

src/test/java/org/apache/sysds/test/functions/unique/UniqueCol.java
```diff
@@ -0,0 +1,127 @@
...
+		uniqueTestOrdered(inputMatrix, expectedMatrix, Types.ExecType.CP, 0.0);
+	}
+
+	/**
+	 * Same input under a heavily reduced local memory budget. Column-wise workers reuse a single set that is cleared
+	 * per column, so only one live set per thread is charged and the parallel path stays applicable; this guards
+	 * against needlessly falling back to batched or sequential execution.
+	 */
+	@Test
+	public void testReducedMemoryBudgetCP() {
```
gaturchenko: This should rather be in the batched test, although in principle you compare the results here, but do not actually check that the batched path was selected. So maybe if there is no way to verify it we don't need this test at all

src/test/java/org/apache/sysds/test/functions/unique/UniqueRow.java
```diff
@@ -76,4 +76,58 @@ public void testNoDuplicatesCP() {
 		double[][] expectedMatrix = {{1, 2, 3}, {4, 5, 6}, {7, 8, 9}};
 		uniqueTest(inputMatrix, expectedMatrix, Types.ExecType.CP, 0.0);
 	}
+
+	/**
+	 * Large enough to take the multi-threaded path. Every row holds a single distinct value, so the expected result is
+	 * one column and independent of any hash set iteration order.
+	 */
+	@Test
+	public void testMultiThreadedCP() {
```
gaturchenko, line 85: Same comment as for `Col`

src/test/java/org/apache/sysds/test/functions/unique/UniqueRow.java
```diff
@@ -76,4 +76,58 @@ public void testNoDuplicatesCP() {
...
+		uniqueTestOrdered(constantRows(400, 64), expectedConstantRows(400), Types.ExecType.CP, 0.0);
+	}
+
+	/**
+	 * Same input under a heavily reduced local memory budget. Row-wise workers reuse a single set that is cleared per
+	 * row, so only one live set per thread is charged and the parallel path stays applicable; this guards against
+	 * needlessly falling back to batched or sequential execution.
+	 */
+	@Test
+	public void testReducedMemoryBudgetCP() {
```
gaturchenko: Same comment as for `Col`

src/test/java/org/apache/sysds/test/functions/unique/UniqueRow.java
```diff
@@ -76,4 +76,58 @@ public void testNoDuplicatesCP() {
...
+			16 * 1024 * 1024);
+	}
+
+	/**
+	 * Sparse counterpart of the multi-threaded case: only every eighth row is populated, so the input is read in sparse
+	 * format. Every row still holds a single distinct value, either its filler or zero, so the expected result stays
+	 * one column.
+	 */
+	@Test
+	public void testSparseMultiThreadedCP() {
```
gaturchenko, line 105: What do you verify with this test? Why is this case different from the other tests?

src/test/java/org/apache/sysds/test/functions/unique/UniqueRowCol.java
gaturchenko, line 1: If multi-threading is not beneficial for `RowCol`, we don't need any new tests here

gaturchenko (changes requested):
Hey @shieru1214 @Isso-W, some more comments from my side. NB: since we are already past the project submission deadline, please feel free to take as much time as you need to make the changes. Also, if some of my comments are unclear or you disagree with them, feel absolutely free to respond, we can certainly have a discussion about this.

src/test/java/org/apache/sysds/test/functions/unique/UniqueCol.java
Isso-W: Sorry, that comment was vague. The multi-threaded path is guarded by `PAR_UNIQUE_NUMCELL_THRESHOLD`, which is 16,384 cells, and below it the implementation always goes sequential regardless of the thread count. The existing cases are at most 12 cells, so they never leave the sequential path.

src/test/java/org/apache/sysds/test/functions/unique/UniqueBase.java
Isso-W: It was unnecessary, Removed in next push

src/test/java/org/apache/sysds/test/functions/unique/UniqueBase.java
Isso-W: Also removed

src/test/java/org/apache/sysds/test/functions/unique/UniqueCol.java
Isso-W: Agreed, that test only compared results under a smaller budget without establishing which path ran. It is replaced by `testBatchedCP`

src/test/java/org/apache/sysds/test/functions/unique/UniqueRow.java
Isso-W: It differs only in that the input is read in sparse format, if you consider that too thin a reason I am happy to drop it

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
Isso-W: I prototyped an alternative before giving up on it.

Today each worker builds a local set and the caller merges them serially, and that merge grows with the total distinct count. At a 100% share it re-inserts everything, which caps the speedup at `k/(k+1)`. A tree-shaped merge does not help, as the last merge is still `O(d)` on one thread.

The prototype partitions by value hash, so each partition can be merged independently and in parallel, and uses `DoubleCountHashMap` to avoid boxing. Same machine, k=8, results verified against sequential:

input | share | distinct | sequential [ms] | current [ms] | partitioned [ms] | current | partitioned
-- | -- | -- | -- | -- | -- | -- | --
2000000x1 | 1% | 20000 | 77.0 | 15.5 | 18.0 | 4.97 | 4.28
2000000x1 | 10% | 200000 | 145.6 | 153.2 | 52.4 | 0.95 | 2.78
2000000x1 | 50% | 1000000 | 128.4 | 204.1 | 118.2 | 0.63 | 1.09
2000000x1 | 100% | 2000000 | 170.0 | 241.7 | 129.7 | 0.70 | 1.31
200000x64 | 1% | 128000 | 667.5 | 238.5 | 161.6 | 2.80 | 4.13
200000x64 | 10% | 1280000 | 803.8 | 1127.3 | 734.4 | 0.71 | 1.09
200000x64 | 50% | 6400000 | 1078.8 | 1572.8 | 1000.9 | 0.69 | 1.08
200000x64 | 100% | 12800000 | 1437.7 | 1928.6 | 1148.2 | 0.75 | 1.25

So the partitioned variant is never slower than sequential, while the current one drops to 0.63. But it only gains about 1.1 at high distinct counts, where the work becomes memory bound.

Does that result worth a multi-threading? Or I can also return to single-threaded.

src/test/java/org/apache/sysds/test/functions/unique/UniqueBatchedPathTest.java
Isso-W: You are right, `uniqueTestConstrainedMemory` now takes a `localPar` argument and restores both values in a `finally` block. All three directions reach the batched path end-to-end at k=8.

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
gaturchenko: What about smaller inputs, does this approach yield a speed-up as well? If so, we could opt for it but that would require further review, see for yourselves if you want to commit to it.

src/test/java/org/apache/sysds/test/functions/unique/UniqueCol.java
gaturchenko: I see, but where does the threshold come from? It sounds like for small inputs multi-threading doesn't make much sense, which likely comes from associated fixed overheads, but have you determined `PAR_UNIQUE_NUMCELL_THRESHOLD = 16,384` empirically? What's the rationale behind this heuristic?

src/test/java/org/apache/sysds/test/functions/unique/UniqueRow.java
gaturchenko: I see, agreed, let's keep it

src/main/java/org/apache/sysds/runtime/matrix/data/LibMatrixSketch.java
Isso-W: Before answering the two questions, a correction. `UniquePerf` warmed up each configuration individually but not the whole run.

The benchmark now warms up all three directions before measuring anything. Corrected results below.

RowCol
Size | Input | Unique Share | Unique Values | k=1 (ms) | k=4 (ms) | Speedup
-- | -- | -- | -- | -- | -- | --
small | 100000x1 | 1% | 1000 | 1.966 | 3.402 | 0.58
small | 100000x1 | 10% | 10000 | 1.713 | 5.572 | 0.31
small | 100000x1 | 50% | 50000 | 2.258 | 5.867 | 0.38
small | 100000x1 | 100% | 100000 | 3.349 | 6.589 | 0.51
large | 2000000x1 | 1% | 20000 | 34.466 | 30.974 | 1.11
large | 2000000x1 | 10% | 200000 | 38.477 | 94.247 | 0.41
large | 2000000x1 | 50% | 1000000 | 203.514 | 296.856 | 0.69
large | 2000000x1 | 100% | 2000000 | 252.824 | 370.762 | 0.68

Row
Size | Input | Unique Share | Unique Values per Row | k=1 (ms) | k=4 (ms) | Speedup
-- | -- | -- | -- | -- | -- | --
small | 10000x64 | 1% | 1 | 45.488 | 10.565 | 4.31
small | 10000x64 | 10% | 6 | 50.046 | 13.272 | 3.77
small | 10000x64 | 50% | 32 | 60.186 | 29.949 | 2.01
small | 10000x64 | 100% | 64 | 96.148 | 56.697 | 1.70
large | 200000x64 | 1% | 1 | 905.786 | 215.934 | 4.19
large | 200000x64 | 10% | 6 | 975.045 | 250.145 | 3.90
large | 200000x64 | 50% | 32 | 1274.481 | 637.637 | 2.00
large | 200000x64 | 100% | 64 | 1910.039 | 1166.702 | 1.64

Col
Size | Input | Unique Share | Unique Values per Column | k=1 (ms) | k=4 (ms) | Speedup
-- | -- | -- | -- | -- | -- | --
small | 64x10000 | 1% | 1 | 48.583 | 12.632 | 3.85
small | 64x10000 | 10% | 6 | 56.231 | 14.099 | 3.99
small | 64x10000 | 50% | 32 | 71.206 | 30.036 | 2.37
small | 64x10000 | 100% | 64 | 108.219 | 51.479 | 2.10
large | 64x200000 | 1% | 1 | 1105.272 | 254.552 | 4.34
large | 64x200000 | 10% | 6 | 1169.894 | 306.159 | 3.82
large | 64x200000 | 50% | 32 | 1563.054 | 627.662 | 2.49
large | 64x200000 | 100% | 64 | 2206.368 | 1147.500 | 1.92


While for smaller inputs, it doesn't work well.
| Input     | Cells | 1%   | 10%  | 50%  | 100% |
| --------- | ----- | ---- | ---- | ---- | ---- |
| 4096x1    | 4K    | 0.10 | 0.06 | 0.13 | 0.25 |
| 16384x1   | 16K   | 0.36 | 0.42 | 0.48 | 0.54 |
| 65536x1   | 64K   | 1.30 | 0.72 | 0.95 | 0.82 |

But there are some interesting findings, The row-wise and column-wise sequential implementation runs two passes, one to determine the output width and one to fill it, and both hash every value.

cells | input | distinct per row | 2-pass (current) | 1-pass | 1-pass + dense read
-- | -- | -- | -- | -- | --
1048576 | 8 MB | 4 | 22.9 ms | 12.4 ms (1.85x) | 10.1 ms (2.27x)
16777216 | 128 MB | 4 | 1041.4 ms | 539.8 ms (1.93x) | 482.0 ms (2.16x)
1048576 | 8 MB | 64 | 94.9 ms | 58.1 ms (1.63x) | 52.7 ms (1.80x)
16777216 | 128 MB | 64 | 1535.2 ms | 963.6 ms (1.59x) | 851.9 ms (1.80x)

The gain does not decay with scale, since what is saved is one hash per cell rather than a fixed cost. The buffer is at most as large as the input, reached only when every value in a row is distinct.

Is this still in the scope of this PR? We are glad to follow or starting with an new issue. @gaturchenko

PR #2557, opened 2026-07-19
[MINOR] off-by-one error in CSV header writing 

t99-i:
fixed incorrect column name indexing (off-by-one error) in CSV header writing


t99-i:
@janniklinde

janniklinde:
Thanks for identifying and fixing the issue @t99-i. It would be great if you could include a test case which would fail with the old implementation.

Baunsgaard:
Thanks @t99-i  for the bug fix, while merging i just fixed the code style formatting.

PR #2569, opened 2026-07-29
[SYSTEMDS-3955] Add signed SSL certificates support

gaturchenko:
This PR adds support for reading in a signed certificate and makes the coordinator verify it in place of the existing self-signed certificate generation. To that end, the XML config is extended with the following:

| Key                                      | Side        | Comment                                                           |
|------------------------------------------|-------------|--------------------------------------------------------------------|
| sysds.federated.ssl.cert                 | worker      | X.509 certificate chain, PEM, leaf first                           |
| sysds.federated.ssl.key                  | worker      | matching private key, PKCS#8 PEM                                   |
| sysds.federated.ssl.keyPassword          | worker      | read from `SYSTEMDS_FEDERATED_SSL_KEY_PASSWORD` env var |
| sysds.federated.ssl.trust                | coordinator | certificates trusted to sign worker certificates, typically the CA |

Both certificate and key are read by each worker from its own config, so they are local per worker. The coordinator has a single trust file, where one CA certificate covers any number of workers, and multiple CAs can be concatenated into one PEM. With hostname verification on, a worker's certificate must be issued for the address the script connects to (`san=dns:...`, or `san=ip:...` for IP literals), so a certificate valid for one worker cannot be replayed for another.

NB: private keys and certificates are added as a part of this PR for SSL-enabled tests to work. They are generated with a dedicated shell script and are irrelevant for anything except the tests.


src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
```diff
@@ -309,20 +311,35 @@ public synchronized static void createWorkGroup() {
...
 	private static class DataRequestHandler extends ChannelInboundHandlerAdapter {
 		private Promise<FederatedResponse> _prom;
+		// A failure observed before the promise was assigned, e.g., a rejected SSL handshake
+		private Throwable _failure;
 
 		public DataRequestHandler() {
 		}
 
-		public void setPromise(Promise<FederatedResponse> prom) {
+		public synchronized void setPromise(Promise<FederatedResponse> prom) {
```
ywcb00: Use `synchronized{ }` blocks to mark specific critical sections of the code instead of declaring the whole method as 'synchronized'.

ywcb00 (changes requested):
Thank you very much for this PR @gaturchenko. :)
Here a few general comments / questions for my understanding:
- Please do not include the key password in the XML config file. Instead, use an environment variable for the key password to avoid leaking this secret when publishing the configuration.
- Regarding the configurable hostname verification: In which case do we not want to check whether the hostname matches the certificate? Should we remove this configuration to make the hostname verification mandatory whenever ssl is enabled?

All the best,
David

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
gaturchenko: Done

gaturchenko:
> Thank you very much for this PR @gaturchenko. :) Here a few general comments / questions for my understanding:
> 
> * Please do not include the key password in the XML config file. Instead, use an environment variable for the key password to avoid leaking this secret when publishing the configuration.
> * Regarding the configurable hostname verification: In which case do we not want to check whether the hostname matches the certificate? Should we remove this configuration to make the hostname verification mandatory whenever ssl is enabled?
> 
> All the best, David

Hey @ywcb00, thanks for the review. Regarding the first point, implemented the change as suggested. About the second, I think the only case is SSH port forwarding, but I think we can remove this option, which I did. Let me know if you have any further comments.

scripts/tutorials/federated/conf/ssl.xml
```diff
@@ -18,4 +18,13 @@
...
 <root>
     <sysds.federated.ssl>true</sysds.federated.ssl>
+
+    <!-- Required with SSL: the workers need a certificate issued for the host the coordinator contacts them on
+         and signed by an authority the coordinator trusts, fill in the paths below before using this
+         configuration. An encrypted private key takes its password from the environment variable
+         SYSTEMDS_FEDERATED_SSL_KEY_PASSWORD.
+    <sysds.federated.ssl.cert>/path/to/worker-cert.pem</sysds.federated.ssl.cert>
+    <sysds.federated.ssl.key>/path/to/worker-key.pem</sysds.federated.ssl.key>
+    <sysds.federated.ssl.trust>/path/to/ca-cert.pem</sysds.federated.ssl.trust>
```
ywcb00, line 28: Since these configurations are required when ssl is enabled, I would include the configurations in this xml file. I see the point that we do not have any certificates for the corresponding tutorial yet, leaving us with three options to resolve this issue:
1. Generate respective certificates for the tutorial and include them in the repository.
2. Include the configurations with the current placeholder value (path/to/...) and ensure that, when executing the tutorial with ssl, it is clear from the error messages that these certificates are missing.
3. Remove the 'ssl' option from `scripts/tutorials/federated/parameters.sh`.
I think the first option would be the best, and the second option fits the scope of this PR best.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
```diff
@@ -308,18 +310,43 @@ public synchronized static void createWorkGroup() {
 	}
 
 	private static class DataRequestHandler extends ChannelInboundHandlerAdapter {
+		// The promise is assigned by the requesting thread, while the channel events below are handled on the
+		// event loop, and the two orders are not guaranteed: a rejected SSL handshake already fails the channel
+		// while the requesting thread is still connecting.
+		private final Object _promLock = new Object();
```
ywcb00: Is there a specific reason for creating an object for locking the member variable `_prom`? I think that we can just use the intrinsic lock from the member variable `_prom` directly, and thus synchronize on `_prom` instead of `_promLock`.

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedSSLUtil.java
```diff
@@ -40,24 +50,93 @@ private FederatedSSLUtil(){
...
+			final DMLConfig conf = ConfigurationManager.getDMLConfig();
+			final String trustPath = conf.getTextValue(DMLConfig.FEDERATED_SSL_TRUST);
+
+			if(!isSet(trustPath))
+				throw new DMLRuntimeException("Federated SSL requires the certificates that are trusted to sign "
+					+ "worker certificates, configure them in " + DMLConfig.FEDERATED_SSL_TRUST + ".");
+
 			try {
-				context = SslContextBuilder.forClient().trustManager(InsecureTrustManagerFactory.INSTANCE).build();
+				LOG.info("Federated SSL trusting certificates in: " + trustPath);
```
ywcb00: This LOG should be at debug level

src/test/java/org/apache/sysds/test/functions/federated/io/FederatedSSLTest.java
```diff
@@ -18,132 +18,249 @@
...
 
-	@Parameterized.Parameters
-	public static Collection<Object[]> data() {
-		// number of rows or cols has to be >= number of federated locations.
-		return Arrays.asList(new Object[][] {{10, 13, true, 2}});
+	@Before
+	public void clearSSLState() {
+		// The SSL context of the coordinator is cached for the JVM, so it should be cleared in between the tests
+		FederatedSSLUtil.resetClientContext();
+		FederatedData.resetFederatedSites();
```
ywcb00, line 81: Can we move this code to the `setUp` method to only have one method that is executed before the tests?

ywcb00 (changes requested):
Thank you very much @gaturchenko. :)
I left again some minor comments in the code, could you please have a look? Other than that, the PR looks good to me.
All the best, David

PS: Please do not overwrite the past commits. Instead, just append a new commit to the PR with the changes. This allows for reviewers to only review the new commits instead of reviewing all changes again. Thanks.

scripts/tutorials/federated/conf/ssl.xml
gaturchenko: Done (option 2)

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedSSLUtil.java
gaturchenko: Done

src/test/java/org/apache/sysds/test/functions/federated/io/FederatedSSLTest.java
gaturchenko: Done

src/main/java/org/apache/sysds/runtime/controlprogram/federated/FederatedData.java
gaturchenko: Done (can't lock on `_prom` but can lock on the class itself via `this`)

PR #2580, opened 2026-08-05
[SYSTEMDS-3949] Extend Delta Spark round trip tests

Jakob-al28:
The Spark round trip tests cover multi-file layouts, deletion vectors,
partitioning, column mapping and schema evolution. This adds three that were
missing: a log that has been checkpointed, a column type SystemDS cannot map,
and a table more than one engine has written to.

| test | covers |
| --- | --- |
| `checkpointBackedHistoryRead` | a log that rolled into a parquet checkpoint, with a delete after it, so the snapshot comes from a checkpoint plus trailing commits |
| `unsupportedColumnTypeRead` | DATE / TIMESTAMP / DECIMAL columns, which have no SystemDS value type and must be rejected by name before decoding |
| `systemdsCreateSparkAppendRead` | SystemDS creates the table, Spark appends to it, so a mixed engine commit history still reads back as one table |


Baunsgaard (approved):
LGTM, if it passes the CI tests we merge it.

PR #2586, opened 2026-08-11
[SYSTEMDS-3959] Report an error when a federated worker cannot bind its port

gaturchenko:
The command `systemds WORKER 80505` terminated silently with exit code 0. There are 2 causes:

1. No range check for `-w <port>`, neither in `bin/systemds` nor when parsing the option
2. `FederatedWorker.run()` captured every exception in a catch-all and logged only `e.getMessage()`, which is empty or ambigous for a bind failure. The same held for occupied and reserved ports, and for the `-fedMonitoring` backend

The following changes were implemented:

- **`PortUtils`** (new): the shared range constants and validation, as well as explicit port and cause naming in case of errors;
- **`DMLOptions`**: `-w` and `-fedMonitoring` reject non-integers and ports outside `[1, 65535]` at parse time, naming the incorrect option. A missing (optional) `-w` argument now falls back to the default port instead of throwing `NumberFormatException`.
- **`FederatedWorker`** / **`FederatedMonitoringServer`**: a failed bind is logged through `explainBindFailure` and re-thrown as `DMLRuntimeException`. The monitoring backend now separates `InterruptedException` (a normal shutdown) from a real failure.
- **`bin/systemds`**: adds a range check, a warning for reserved ports, and `exit 1`. Previously, an error was printed and the JVM was launched regardless.
- **`FederatedWorkerPortTest`** (new): 14 tests covering the range predicates, option parsing, and the two worker paths (out-of-range and occupied port).

Resulting messages:
```
$ systemds WORKER 80505
error: Port 80505 for the federated worker is out of range, expected a port in [1, 65535]

$ systemds WORKER 80
warning: Port 80 for the federated worker is a reserved system port, binding it requires elevated privileges
ERROR federated.FederatedWorker: Federated worker stopped: no permission to bind port 80, ports below 1024 are reserved and require elevated privileges (Permission denied)

$ systemds WORKER 8001   # port taken
ERROR federated.FederatedWorker: Federated worker stopped: port 8001 is already in use (Address already in use)
```

### DIFF
--- a/bin/systemds
+++ b/bin/systemds
@@ -173,6 +173,32 @@ script.
 EOF
 }
 
+# verify that $1 is a port that a server can be bound to, otherwise abort with an error.
+# $2 names the server the port is meant for.
+function checkPort {
+  local port=$1
+  local target=$2
+  local re='^[0-9]+$'
+  if ! [[ $port =~ $re ]] ; then
+    echo "error: Port '$port' for the $target is not a number"
+    printUsage
+    exit 1
+  fi
+  # drop leading zeros, so that the length check below is not fooled by e.g. 0000080505
+  local num=$port
+  while [[ ${#num} -gt 1 && $num == 0* ]] ; do num=${num#0} ; done
+  # more than 5 digits is out of range by definition, and comparing it would silently
+  # overflow the 64 bit integers of the shell for very long inputs
+  if [ ${#num} -gt 5 ] || [ "$num" -lt 1 ] || [ "$num" -gt 65535 ] ; then
+    echo "error: Port $port for the $target is out of range, expected a port in [1, 65535]"
+    printUsage
+    exit 1
+  fi
+  if [ "$num" -lt 1024 ] ; then
+    echo "warning: Port $port for the $target is a reserved system port, binding it requires elevated privileges"
+  fi
+}
+
 # print an error if no argument is supplied.
 if [ -z "$1" ] ; then
     echo "Wrong Usage. Add -help for additional parameters.";
@@ -242,11 +268,7 @@ elif echo "$1" | grep -q "WORKER"; then
     shift
   fi
   PORT=$1
-  re='^[0-9]+$'
-  if ! [[ $PORT =~ $re ]] ; then
-    echo "error: Port is not a number"
-    printUsage
-  fi
+  checkPort "$PORT" "federated worker"
   shift
 elif echo "$1" | grep -q "FEDMONITORING"; then
   FEDMONITORING=1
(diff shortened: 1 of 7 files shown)


gaturchenko:
Waiting for the review by @ywcb00

bin/systemds
```diff
@@ -173,6 +173,32 @@ script.
...
+  while [[ ${#num} -gt 1 && $num == 0* ]] ; do num=${num#0} ; done
+  # more than 5 digits is out of range by definition, and comparing it would silently
+  # overflow the 64 bit integers of the shell for very long inputs
+  if [ ${#num} -gt 5 ] || [ "$num" -lt 1 ] || [ "$num" -gt 65535 ] ; then
+    echo "error: Port $port for the $target is out of range, expected a port in [1, 65535]"
+    printUsage
+    exit 1
+  fi
+  if [ "$num" -lt 1024 ] ; then
+    echo "warning: Port $port for the $target is a reserved system port, binding it requires elevated privileges"
```
ywcb00: After reading this error message, I would expect it to work when runing the same command as superuser (root; sudo) to acquire the 'elevated privileges'.

bin/systemds
```diff
@@ -173,6 +173,32 @@ script.
...
+    echo "error: Port '$port' for the $target is not a number"
+    printUsage
+    exit 1
+  fi
+  # drop leading zeros, so that the length check below is not fooled by e.g. 0000080505
+  local num=$port
+  while [[ ${#num} -gt 1 && $num == 0* ]] ; do num=${num#0} ; done
+  # more than 5 digits is out of range by definition, and comparing it would silently
+  # overflow the 64 bit integers of the shell for very long inputs
+  if [ ${#num} -gt 5 ] || [ "$num" -lt 1 ] || [ "$num" -gt 65535 ] ; then
```
ywcb00: The first condition is already covered by the second and third condition.

bin/systemds
```diff
@@ -173,6 +173,32 @@ script.
...
+    echo "error: Port $port for the $target is out of range, expected a port in [1, 65535]"
+    printUsage
+    exit 1
+  fi
+  if [ "$num" -lt 1024 ] ; then
+    echo "warning: Port $port for the $target is a reserved system port, binding it requires elevated privileges"
+  fi
+}
+
```
Baunsgaard, line 201: Only criticism. I am strongly against adding more logic to the systemds file. Can we not just have the error message and verification inside Java. This gives a single point of failure, rather than splitting the logic in two.

While there was some logic already verifying basic number compatibility, i think that should also just be cleaned up and moved into the java PortUtils you have defined.

Baunsgaard (approved):

PR #2587, opened 2026-08-11
[SYSTEMDS-3960] Fix federated worker readiness probe on delayed connections

gaturchenko:
With any delay on the worker's interface, currently every federated test fails before it starts, while the worker is listening the whole time. `FederatedWorkerUtils.tryConnect` allocated 25ms per connect attempt, but a TCP handshake needs two traversals of the link which are ~100ms under the delay above. Every attempt expired with a `SocketTimeoutException`, which is an `IOException` and therefore indistinguishable from a closed port, so the probe never succeeded.

The following changes were implemented:

- Per-attempt connect budget increased from 25ms to 2s. A closed port is refused by the kernel immediately, so it only applies once a handshake is in progress. Additionally, 2s covers one lost SYN, retransmitted after ~1s
- Each attempt is capped by the time left until the overall deadline (`attemptTimeout`), so a slow connect cannot exceed it. The returned value of 0 implies the attempt was never made
- The `waitForWorkers` now rechecks the deadline per port, which can now cost up to 2s each
- Six new tests are added verifying the change


gaturchenko:
Waiting for the review by @ywcb00

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
```diff
@@ -25,20 +25,24 @@
...
 
 /**
- * Test helpers that block until a federated worker is accepting TCP connections on its port.
- *
- * <p>The federated worker opens its TCP port after Netty's {@code bind().sync()} returns; a successful
- * TCP connect to that port therefore indicates that the worker is ready to accept requests. The methods
- * here poll for that signal and throw {@link RuntimeException} on timeout or if the underlying
- * {@code Process}/{@code Thread} exits before the port becomes ready.
+ * Test helpers that block until a federated worker is accepting TCP connections on its port. The federated worker opens
+ * its TCP port after Netty's {@code bind().sync()} returns; a successful TCP connect to that port therefore indicates
```
ywcb00: While removing the html paragraph tag is good, please preserve the line break.

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
```diff
@@ -174,16 +180,33 @@ public static void waitForWorkers(int[] ports, int timeoutMs, java.util.function
...
 
+	/**
+	 * Budget for a single connect attempt, capped by the time left until the overall deadline so that one slow attempt
+	 * cannot substantially exceed the limit.
+	 *
+	 * @param remainingMs time left until the deadline, in ms
+	 * @return the timeout to pass to {@link Socket#connect}, or 0 if no attempt should be made. Never returns 0 while
+	 *         time is left, because {@code connect} reads a timeout of 0 as 'infinite'.
+	 */
+	public static int attemptTimeout(long remainingMs) {
```
ywcb00: This name here is ambiguous. While this method returns the timeout for a connection attempt, the name could also mean that it attempts the connection with a timeout.

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
```diff
@@ -174,16 +180,33 @@ public static void waitForWorkers(int[] ports, int timeoutMs, java.util.function
 		}
 	}
 
-	private static boolean tryConnect(int port) {
+	private static boolean tryConnect(int port, long deadline) {
+		final int timeout = attemptTimeout(deadline - System.currentTimeMillis());
+		if(timeout == 0) // out of time, do not start another attempt
```
ywcb00: This condition here can be rewritten to a "less than or equal" condition, thereby removing the necessity of the separate condition in `attemptTimeout()` and allowing for replacing the call to `attemptTimeout()` by the single remaining code line.

src/test/java/org/apache/sysds/test/component/federated/FederatedWorkerUtilsTest.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Tests for the readiness probe that blocks until a federated worker accepts connections.
+ */
+public class FederatedWorkerUtilsTest {
+
+	/** Round-trip time the probe should tolerate without giving up, in ms. */
+	private static final int TOLERATED_HANDSHAKE_MS = 1000;
+
+	@Test
+	public void attemptBudgetCoversADelayedHandshake() {
```
ywcb00: I appreciate verifying the code through these unit tests.
However, I think these tests are too fine-grained for testing the test framework. Please remove the tests that evaluate `attemptTimeout()` directly (aligning with removing of this method, as commented below), and move the remaining tests to the `FederatedUrlParserTest`.

ywcb00 (changes requested):
Thank you very much for this PR @gaturchenko. :)
I left some comments in the code. Also, could you please try to shorten the method headers and in-line documentation if possible?
Thanks and all the best, David

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
gaturchenko: Done, please see [a4b2adc](https://github.com/apache/systemds/pull/2587/commits/a4b2adc2c09f20bd732cec975c9846707cff5324)

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
gaturchenko: Done, please see [a4b2adc](https://github.com/apache/systemds/pull/2587/commits/a4b2adc2c09f20bd732cec975c9846707cff5324)

src/test/java/org/apache/sysds/test/FederatedWorkerUtils.java
gaturchenko: Done, please see [a4b2adc](https://github.com/apache/systemds/pull/2587/commits/a4b2adc2c09f20bd732cec975c9846707cff5324)

src/test/java/org/apache/sysds/test/component/federated/FederatedWorkerUtilsTest.java
gaturchenko: Done, please see [c58a3b1](https://github.com/apache/systemds/pull/2587/commits/c58a3b125deea6a003b5fc0a6e6093dfe36eb3d2)

PR #2596, opened 2026-08-25
[SYSTEMDS-3961] Add Sparsity-based Recompiler Class to Recompilation Routine

ywcb00:
This PR introduces the sparsity-based recompiler which replicates the sparsity-based rewrites at the moment. It is planned to extend this recompiler to retrieve additional data properties and to restructure of the entire computation graph.
The new sparsity-based recompiler is integrated in the recompilation procedure with a configuration to enable/disable it (disabled by default).
NOTE: Currently, the sparsity-based recompiler performs identical rewrites as `RewriteMatrixMultChainOptimizationSparse.java`. It will be extended in the future.

All the best,
David

### DIFF
--- a/src/main/java/org/apache/sysds/conf/DMLConfig.java
+++ b/src/main/java/org/apache/sysds/conf/DMLConfig.java
@@ -98,8 +98,9 @@ public class DMLConfig
 	public static final String NATIVE_BLAS          = "sysds.native.blas";
 	public static final String NATIVE_BLAS_DIR      = "sysds.native.blas.directory";
 	public static final String DAG_LINEARIZATION    = "sysds.compile.linearization";
-	public static final String SPARSITY_REWRITES    = "sysds.rewrites.sparsity.enabled"; // boolean
-	public static final String SPARSITY_ESTIMATOR   = "sysds.rewrites.sparsity.estimator"; // see EstiamtionUtils.EstimatorType
+	public static final String SPARSITY_REWRITES    = "sysds.sparsity.rewrites.enabled"; // boolean
+	public static final String SPARSITY_RECOMPILE   = "sysds.sparsity.recompile.enabled"; // boolean
+	public static final String SPARSITY_ESTIMATOR   = "sysds.sparsity.estimator"; // see EstiamtionUtils.EstimatorType
 	public static final String CODEGEN              = "sysds.codegen.enabled"; //boolean
 	public static final String CODEGEN_API          = "sysds.codegen.api"; // see SpoofCompiler.API
 	public static final String CODEGEN_COMPILER     = "sysds.codegen.compiler"; //see SpoofCompiler.CompilerType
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/sysds/conf/DMLConfig.java
```diff
@@ -98,8 +98,9 @@ public class DMLConfig
 	public static final String NATIVE_BLAS          = "sysds.native.blas";
 	public static final String NATIVE_BLAS_DIR      = "sysds.native.blas.directory";
 	public static final String DAG_LINEARIZATION    = "sysds.compile.linearization";
-	public static final String SPARSITY_REWRITES    = "sysds.rewrites.sparsity.enabled"; // boolean
-	public static final String SPARSITY_ESTIMATOR   = "sysds.rewrites.sparsity.estimator"; // see EstiamtionUtils.EstimatorType
+	public static final String SPARSITY_REWRITES    = "sysds.sparsity.rewrites.enabled"; // boolean
+	public static final String SPARSITY_RECOMPILE   = "sysds.sparsity.recompile.enabled"; // boolean
```
Baunsgaard, line 102: would it not be better not to rename these environment variables? 
Or are they currently not used in the system?

src/main/java/org/apache/sysds/hops/recompile/Recompiler.java
```diff
@@ -387,6 +388,11 @@ else if( !codegen ) {
 			memo.extract(hops, status);
 		}
 		
+		// sparsity-based DAG recompilation if enabled
+		if(ConfigurationManager.getDMLConfig().getBooleanValue(DMLConfig.SPARSITY_RECOMPILE)) {
+			hops = SparsityDAGRecompiler.optimize(hops, ec);
+		}
+
```
Baunsgaard, line 395: would it not be better to have it as part of the above statistics optimizations. Since if you change the graph based on the sparsity estimatings the statistics would also change ?

src/main/java/org/apache/sysds/hops/recompile/Recompiler.java
Baunsgaard, line 419: Also be aware of this one. since the rewrites you make above could be effected by the low level rewrites.

src/main/java/org/apache/sysds/hops/recompile/SparsityDAGRecompiler.java
Baunsgaard, line 1: So the idea is to start by targeting mm chains ? and do these based on the sparsity estimations?

Baunsgaard (commented):
only nits

Baunsgaard:
Maybe change the description of the PR, and name, to highlight it is for Rewriting MMChains to start with?

src/main/java/org/apache/sysds/conf/DMLConfig.java
ywcb00: These variables are unused so far. I have introduced these variables two weeks ago.

src/main/java/org/apache/sysds/hops/recompile/Recompiler.java
ywcb00: As it is right now, yes absolutely. The current version replicates the `RewriteMatrixMultChainOptimizationSparse`, which is still a part of the statistics optimizations.
In the future, the sparsity-based recompiler will perform modifications that are comparable to the code generation. Considering the planned future progress, I believe this will be the correct location.

src/main/java/org/apache/sysds/hops/recompile/SparsityDAGRecompiler.java
ywcb00: The sparsity-based recompiler currently replicates the `RewriteMatrixMultChainOptimizationSparse` which does only work for matrix multiplication chains. That is why it targets matrix multiplication chains as a first step.
The idea is to support modifications to additional operations by retrieving their sparsity information through pre-fetching of input data and partial computations.

src/main/java/org/apache/sysds/hops/recompile/Recompiler.java
ywcb00: I stand corrected. I misunderstood something about the current Recompiler. You are absolutely right! Consequently, I am thinking about not merging this PR, since this separation is not needed. Thank you for pointing this out. :)

